PR #113, opened 2014-06-10
DNN-5455 Extension method for binding PetaPoco dynamic result set 

weggetor:
PetaPoco is mostly used together with strong typed Info-classes. Thats nice and working well but sometimes we need a more flexible way to retrieve data. Think of a simple module where the user enters an arbitrary select statement to any table of the database (yes, I wont do that! Its just for explanation!). With PetaPoco this could be done with ExecuteQuery<dynamic> and the result is of type IEnumerable<dynamic>. This is nice and working "out of the box". But you can not databind this IEnumerable<dynamic> to a datagrid : even if the data is retrieved correctly, nothing is shown in the grid. So I implemented an extension method for IEnumerable<dynamic> named "ToDataTable"

### DIFF
--- a/DNN Platform/Library/Collections/EnumerableExtensions.cs
+++ b/DNN Platform/Library/Collections/EnumerableExtensions.cs
@@ -1,0 +1,66 @@
+﻿#region Copyright
+// 
+// DotNetNuke® - http://www.dotnetnuke.com
+// Copyright (c) 2002-2014
+// by DotNetNuke Corporation
+// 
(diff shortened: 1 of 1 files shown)


jbrinkman:
Given the level of effort needed to convert to NPoco, we have decided to remain on PetaPoco and merge this change.

cnurse:
Pull request was merged - had to add new file to project and cleaned up file to remove unused using directives.

PR #201, opened 2014-11-19
DNN-6102 Better exception management in event log

donker:
This change allows us to provide better analysis services in case of problems in DNN.
- Added table Exceptions that stores message/stacktrace/etc of an exception. It is identified through an MD5 hash. The EventLog optionaly includes the hash as a field to refer to this.
- Added table ExceptionEvents that stores DNN specific exception fields we find in the BasePortalException class such as PortalId, TabId, Filename, etc. It references the EventLog table through the LogEventId.
- A view is added to draw out the totality of the EventLog including the above two tables.
- I've rewired everything from the business layer so the outside doesn't see the difference (for now)

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.00.SqlDataProvider
@@ -952,14 +952,7 @@ VALUES
 	@StackTrace,
 	@InnerMessage,
 	@InnerStackTrace,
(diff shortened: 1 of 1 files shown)


ashishpd:
Do we delete records from the new Exceptions or ExceptionEvents tables after a while?

donker:
Not in this revision yet, AFAIK. I'm all in favor of doing something like that in the future. But I'm taking this one step at a time.

PR #232, opened 2014-12-11
DNN-6233 AutoCompleteControl for Profile properties

donker:


jbrinkman:
This was closes and resubmitted as #298

PR #236, opened 2014-12-18
performance updates in purge schedule history

ohine:
It seems the old script doesn't work too well. This change reduces the time to clear the history table with 465k rows by 97%. However, because it limited to 50k rows per execution, it will take multiple days to fully clean up the scheduled history table. 

Total Rows: 465k
Default Query: 3992 seconds (1hr 6m 32s)
Optimized (all at once): 96 seconds
100k batches: 143 seconds total, 5 runs (14, 34, 35, 31, 29)
50k batches: 94 seconds total, 10 runs (6, 6, 11, 7, 8, 8, 10, 14, 20, 5)


evidnn:
With this change Retain in Schedule History is ignored and all records are deleted:
DNN-6877

PR #238, opened 2014-12-20
DNN-6282: Fixed URLRewriter to work with VS2013 browser link feature

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-6282


cathalconnolly:
merging into 7.4.1

PR #241, opened 2014-12-20
DNN-6285: Some string (performance) optimizations in DefaultPage (and base classes)

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-6285

### DIFF
--- a/DNN Platform/Library/Framework/CDefault.cs
+++ b/DNN Platform/Library/Framework/CDefault.cs
@@ -49,13 +49,13 @@ namespace DotNetNuke.Framework
     /// -----------------------------------------------------------------------------
     public class CDefault : PageBase
     {
-        public string Author = "";
-        public string Comment = "";
-        public string Copyright = "";
-        public string Description = "";
-        public string Generator = "";
-        public string KeyWords = "";
-        public new string Title = "";
+        public string Author = string.Empty;
+        public string Comment = string.Empty;
+        public string Copyright = string.Empty;
+        public string Description = string.Empty;
+        public string Generator = string.Empty;
+        public string KeyWords = string.Empty;
+        public new string Title = string.Empty;
 
         protected override void RegisterAjaxScript()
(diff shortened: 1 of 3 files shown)


KenGrierson:
Hi Jan;
I see a LOT of code changed here what is testable from the UI?

DNN Platform/Library/Framework/CDefault.cs
```diff
@@ -121,7 +121,7 @@ protected string AdvancedSettingsPageUrl
                 if (modules.Count > 0)
                 {
                     var pmb = new PortalModuleBase();
-                    result = pmb.EditUrl(tab.TabID, "", false, "mid=" + modules.ElementAt(0).ModuleID, "popUp=true", "ReturnUrl=" + Server.UrlEncode(TestableGlobals.Instance.NavigateURL()));
+                    result = pmb.EditUrl(tab.TabID, "", false, string.Concat("mid=", modules.ElementAt(0).ModuleID), "popUp=true", string.Concat("ReturnUrl=", Server.UrlEncode(TestableGlobals.Instance.NavigateURL())));
```
bdukes: The `+` operator compiles to `string.Concat`, so there is no different between the two (see http://stackoverflow.com/q/1787696/2688).  At some point it make sense to switch from concatenating to using a `StringBuilder`, but changing from `+` to `Concat` has no difference in the underlying code that is generated (except in the case where you have two constant expression, in which case the compiler will pre-concatenate them when using `+`, before transforming it into a `Concat` call)

janjonas: @bdukes: Yes, you are right. I thought, that there is a difference, but that was a false conclusion. Maybe there a minimal differences between string.empty and "", but I'm not sure if this has any impact on the total performance of the system.

@KenGrierson: Since this pull request changes some base classes there is no explicit UI test. Everything should work as before :)

PR #245, opened 2014-12-21
DNN-6286: Performance issue in UriExtensions.IsLocalUri with DNS checks

bogdan-litescu:


cnurse:
I am not 100% convinced that the behavior is the same, so I am leaving this for now.

cathalconnolly:
agreed - it removes the functionality added in https://github.com/dnnsoftware/Dnn.Platform/commit/7834816fd4daeef2e76270482ff4f7c18ea994b9 which was added in response to https://dnntracker.atlassian.net/browse/DNN-3472

bogdan-litescu:
Not only it removes, but it eliminates the need for it since there's no DNS
check to slow things down.
I don't understand why the DNS check is there. Do you know of some use
cases? Checking DNS is also very unreliable.

jbrinkman:
So, maybe, rather than speculating it would be helpful to add some Unit Tests so that we can validate that behavior didn't change.  That will make it obvious whether we could safely accept this change.

cathalconnolly:
that code was there from quite some time ago (I only added caching code in dnn-3472). Out of interest I checked the CDF codebase, and it appears that umbraco/cdf changed that check a while back so it no longer does a DNS check https://github.com/Shazwazza/ClientDependency/blob/5ed7ac24f37061c12f2db4140219a4eb1f8f8856/ClientDependency.Core/UriExtensions.cs

I'm going to assign this to myself and do some further testing

jbrinkman:
This should be an easy win for Unit Tests.  That would document the behavior we expect.  We need to get in the habit of doing Unit Tests wherever possible. This is definitely one of the places where it is possible.

cathalconnolly:
We feel that the upgrade to the latest CDF is better kept until 7.5.0 (to allow for additional testing), so I will be merging this change into 7.4.1

PR #250, opened 2014-12-28
DNN-6265 Update OAuthClientBase to support optional parameters

skydnn:
Updated OAuthClientBase to support the optional parameter "resource" of the OAuth V2 standard, as this field is required to support the Microsoft Azure Active Directory OAuth implementation.
Also added UrlEncode logic to the APISecret, as apparently secret keys generated by Azure are not url friendly. 
Tested implementation on dev box and an Azure site for LiveConnect, Google, Facebook and Azure Active Directory.

### DIFF
--- a/DNN Platform/Library/Services/Authentication/OAuth/OAuthClientBase.cs
+++ b/DNN Platform/Library/Services/Authentication/OAuth/OAuthClientBase.cs
@@ -28,6 +28,10 @@
 //
 // Some modifications by Shannon Whitley
 // Author Url: http://voiceoftech.com/
+//
+// Additional modifications by Evan Smith (DNN-4143 & DNN-6265)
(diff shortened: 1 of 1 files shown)


skydnn:
Looks like I left off the bug number in the title.  It is DNN-6265

cathalconnolly:
thanks -we're in code-freeze for 7.4.0, so this will be processed for the 7.4.1 release

tatigo:
I'm having issues integrating UseOpenIdConnectAuthentication to Azure (from Owin startup class). 
I'm on DNN  7.04. Do I need to pull the most current Dnn version?

bdukes:
@tatigo it looks like this change was included in 7.4.1, so if you're planning to use DNN's `OAuthClientBase` to create your authentication provider, than you will need to get that version of DNN.

PR #251, opened 2014-12-31
Fixed DotNetNuke.UI.Skins.ControlsLanguageTokenReplace.NewUrl() to not remove query string parameters

janjonas:
Please see https://dnntracker.atlassian.net/browse/DNN-6310 for more information.

This pull request removes a logic that was introduced with this pull request https://github.com/sachatrauwaen/Dnn.Platform/commit/f98b4a3e5e05bc67200d1a6572c43f0c5bfd00dd created by @sachatrauwaen 

if (DotNetNuke.Entities.Host.Host.UseFriendlyUrls && !islocalized ) {
  [...]
}

I do not see a reason for not adding the query string when creating the "change language"-links, because query parameters that are "encoded" in the path (like the "Id" parameter in http://.../Home/Id/1?category=2) are not removed when creating links to the same page in other languages.

In other words: I would either remove all parameters (from query string and paths) or none of them when creating "change language links". IMHO, the current logic is confusing, because only query string parameters are removed.

### DIFF
--- a/DNN Platform/Library/UI/Skins/Controls/LanguageTokenReplace.cs
+++ b/DNN Platform/Library/UI/Skins/Controls/LanguageTokenReplace.cs
@@ -31,6 +31,7 @@ using DotNetNuke.Common.Utilities;
 using DotNetNuke.Entities.Modules;
 using DotNetNuke.Entities.Portals;
 using DotNetNuke.Entities.Tabs;
+using DotNetNuke.Entities.Urls;
 using DotNetNuke.Entities.Users;
 using DotNetNuke.Security;
 using DotNetNuke.Services.Localization;
@@ -233,24 +234,24 @@ namespace DotNetNuke.UI.Skins.Controls
         private string NewUrl(string newLanguage)
         {
             var objSecurity = new PortalSecurity();
-            Locale newLocale = LocaleController.Instance.GetLocale(newLanguage);
+            var newLocale = LocaleController.Instance.GetLocale(newLanguage);
 
             //Ensure that the current ActiveTab is the culture of the new language
-            int tabId = objPortal.ActiveTab.TabID;
-            bool islocalized = false;
+            var tabId = objPortal.ActiveTab.TabID;
+            var islocalized = false;
 
-            TabInfo localizedTab = TabController.Instance.GetTabByCulture(tabId, objPortal.PortalId, newLocale);
+            var localizedTab = TabController.Instance.GetTabByCulture(tabId, objPortal.PortalId, newLocale);
             if (localizedTab != null)
             {
                 islocalized = true;
                 if (localizedTab.IsDeleted || !TabPermissionController.CanViewPage(localizedTab))
                 {
-                    PortalInfo localizedPortal = PortalController.Instance.GetPortal(objPortal.PortalId, newLocale.Code);
+                    var localizedPortal = PortalController.Instance.GetPortal(objPortal.PortalId, newLocale.Code);
                     tabId = localizedPortal.HomeTabId;
                 }
                 else
                 {
-                    string fullurl = "";
+                    var fullurl = string.Empty;
                     switch (localizedTab.TabType)
                     {
                         case TabType.Normal:
@@ -277,14 +278,7 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
             }
 
-            // on localised pages most of the querystring parameters have no sense and generate duplicate urls for the same content
-            // because we are on a other tab with other modules (example : ?returntab=/en-US/about)
-            string rawQueryString = "";
-            if (DotNetNuke.Entities.Host.Host.UseFriendlyUrls && !islocalized )
-            {
-                rawQueryString = new Uri(HttpContext.Current.Request.Url.Scheme + "://" + HttpContext.Current.Request.Url.Authority + HttpContext.Current.Request.RawUrl).Query;
-            }
-
+            var rawQueryString = new Uri(string.Concat(HttpContext.Current.Request.Url.GetLeftPart(UriPartial.Authority), HttpContext.Current.Request.RawUrl)).Query;
             return
                 objSecurity.InputFilter(
                     TestableGlobals.Instance.NavigateURL(tabId, objPortal.ActiveTab.IsSuperTab, objPortal, HttpContext.Current.Request.QueryString["ctl"], newLanguage, GetQsParams(newLocale.Code, islocalized)) +


janjonas:
One more thought: Is the DotNetNuke.Entities.Host.Host.UseFriendlyUrls flag still in use, i.e. is it possible to change the value to false using the UI? I do not see any user interface to configure the UseFriendlyUrls flag?

If I set UseFriendlyUrls to "N" in the database the change language links seems to be broken, if a a "?category=someValue" query string is present.
Each time the language are switched, a new category parameter is added to the URLs:
 * http://localhost/dnn_platform/de-de/?category=someValue
 * http://localhost/dnn_platform/en-us/?category=someValue?category=someValue
 * http://localhost/dnn_platform/de-de/?category=someValue%3fcategory%3dsomeValue?category=someValue?category=someValue
 * [...]

IMHO, the URL settings should not change the parameters that are added when creating links. In other words: The URL settings should be transparent to the (query) parameters that are passed to the pages.

sachatrauwaen:
1) One more thought: Is the DotNetNuke.Entities.Host.Host.UseFriendlyUrls flag still in use, i.e. is it possible to change the value to false using the UI? I do not see any user interface to configure the UseFriendlyUrls flag?

> Yes this is possible. But only if you are not in advanced mode (in the web.config).
> Setting this flag to false while in advanced mode, is not supported.

> UseFriendlyUrls to "N", means also using query string parameters for the page and language (tabid=123&languague=de-DE)

2) If I set UseFriendlyUrls to "N" in the database the change language links seems to be broken, if a a "?category=someValue" query string is present.
Each time the language are switched, a new category parameter is added to the URLs:

http://localhost/dnn_platform/de-de/?category=someValue
http://localhost/dnn_platform/en-us/?category=someValue?category=someValue
http://localhost/dnn_platform/de-de/?category=someValue%3fcategory%3dsomeValue?category=someValue?category=someValue
[...]

> It seams that your changes  "Removing if (DotNetNuke.Entities.Host.Host.UseFriendlyUrls ..." add a second time the the query parameters when UseFriendlyUrls is false 
> That was on part of my fix

3) IMHO, the URL settings should not change the parameters that are added when creating links. In other words: The URL settings should be transparent to the (query) parameters that are passed to the pages.

> In reality to solve the whole problem it is really complicate (not possible without big architecture change)
> Let me explain 2 use cases for multi language web sites with content localization enabled :

a) 2 localized version of the same page contain a article module that not manage content localization. So each page had a independent module to manage the content in each language
-> on the first local page the Url of the first article = www.mysite.com/en/mypage/articleid/1
-> on the second local page the Url of the first article = www.mysite.com/de/mypage/articleid/2

DNN can not know that the url on the language selector need to be www.mysite.com/de/mypage//articleid/2 when www.mysite.com/en/mypage/articleid/1 is shown.

Actually the url in the language selector is www.mysite.com/de/mypage/articleid/1. 
Very bad because this generate a error or shown the list of articles or something else depending on the used module
 because articleid/1 correspond to the module on page en/mypage and not de/mypage.
 Secondary this generates duplicates content in Google (bad for SEO and users).
 
 b) When you click on login (on a site with popups disabled) the url is www.mysite.com/en/login?returnurl=/en/home then you click on the language selector.
 In the past (and after removing my fix), the url becaume www.mysite.com/de/login?returnurl=/en/home
 Then after login , you are redirected to /en/home, this is not the right page because you just login in DE and you are redirected to EN page.

The idea of my fix was, when you click on the language selector, it's better to be redirected to a "good page" but less corresponding to the content on what you are, 
then sometimes be redirected to a better page and sometimes to "bad page".

The fix for a) is not accepted by the coorporation because it brokes social groups modules tests cases, because the tests uses social groups as modules 
which manage content localization, which is not true. Or thinks the groups don't need to be translated.
And the core don't include really article modules.
And it's true that my fix make the language selector less accurate for modules which manage language.
That the raison i don't remove query string for neutral pages.

The code change ai madethat  you want to remove, was to fix b)

(waw, Not easy to explain)

janjonas:
Thanks @sachatrauwaen for the detailed answer and explanations.

After changing the urlFormat from "Advanced" to "Standard" in the Web.config, I was able to toggle the "UseFriendlyUrls" flag (on Host > Host Settings > Advanced Settings). And you are right, the problem that the query parameter is added again and again is caused by my pull request. So please do NOT merge this pull request.

I also understand your localized article scenario. Nevertheless, in my opinion it is better to pass all parameters, irrespectively of the fact that they might make no sense in other languages. IMHO, this problem should be handled by sone custom logic in the modules (for example by displaying a message like "This item is not available in the current language, please click here to see change the language"). If you just remove the parameter the module has no information about the content the user has requested and there is no way to provide the user some "fallback".

But: If all path parameters are passed to the localized pages, why not also pass all query parameters? I do not see any reason why the "category" parameter is removed from the language links.

I also see the point with the "returnurl" parameter. But I would handle the returnurl parameter as exception and either
 * remove it (but keep all other query parameters)
 * Change the languages in the returnurl parameter (so that the language in this parameter always match the current language)

Regarding your SEO and duplicate content problem: IMHO this problem is not directly related to the language links. In a default installation you can access the home page by an endless number of URLs:
 * http://localhost/dnn_platform/de-de/
 * http://localhost/dnn_platform/de-de/Home/Some
 * http://localhost/dnn_platform/de-de/Home/Some/New
 * http://localhost/dnn_platform/de-de/Home/Some/New/Paths
 * [...]
Of course, all there are no such links by default and one could argue, that the search bot will never "find" those URLs, but there is no guarantee that some other sites use them to link your page.

If you want to fix this problem, you need to implement some custom redirect rules, that redirects the user to the "real" URL or you could use canonical links.

Summarized:
 * My pull request introduces some new problems, please do NOT merge it
 * I recommend to fix the problem that the query parameters (like "category") are removed from the language links
 * I would add some special logic for the returnurl parameter

@sachatrauwaen: What do you think?

sachatrauwaen:
I also understand your localized article scenario. Nevertheless, in my opinion it is better to pass all parameters, irrespectively of the fact that they might make no sense in other languages. IMHO, this problem should be handled by sone custom logic in the modules (for example by displaying a message like "This item is not available in the current language, please click here to see change the language"). If you just remove the parameter the module has no information about the content the user has requested and there is no way to provide the user some "fallback".

> My personal opinion is, asking a module which are not made for multi-language senarios to display a message, is a littlebit optimistic (i don't think many module developpers will do it). I tink it's common, when the same content not exist on the other language, to redirect to a page which is not exactelly the same. Norma users change only 1 time in a website.

> But for modules which manage multi language content, i agree that finding a solutions which don't remove the module parameters (like category) is best. The problem is to find a solution to know when to remove them and when to leave them. We need something like a flag on the page.

> The recomendation is that the module raise a 404 page not found when the url is bad, not to display a message.

But: If all path parameters are passed to the localized pages, why not also pass all query parameters? I do not see any reason why the "category" parameter is removed from the language links.
> Same raison

I also see the point with the "returnurl" parameter. But I would handle the returnurl parameter as exception and either
remove it (but keep all other query parameters)
Change the languages in the returnurl parameter (so that the language in this parameter always match the current language)
> the page name need to be switched also, and sometime it can include more then that to switch (module path/query parameters)

Regarding your SEO and duplicate content problem: IMHO this problem is not directly related to the language links. In a default installation you can access the home page by an endless number of URLs:
http://localhost/dnn_platform/de-de/
http://localhost/dnn_platform/de-de/Home/Some
http://localhost/dnn_platform/de-de/Home/Some/New
http://localhost/dnn_platform/de-de/Home/Some/New/Paths
[...] Of course, all there are no such links by default and one could argue, that the search bot will never "find" those URLs, but there is no guarantee that some other sites use them to link your page.
If you want to fix this problem, you need to implement some custom redirect rules, that redirects the user to the "real" URL or you could use canonical links.

> Yes that right. 
> But the first think, is not create bad links, because they are surrely integrated by google.
> The actual Dnn core rewriter, cannot handle this kind of redirection.
> The recomendation is that the module raise a 404 page not found

Summary 
I don't know how to fix this simply.
My first approch was to make it less bad.
An other solution that solve maybe your use case, is to make the chose to remove or not (path and query parameters ) optional by parameters in the skin object.

janjonas:
If created a new pull request (https://github.com/dnnsoftware/Dnn.Platform/pull/256), that replaces this one.

@sachatrauwaen: Thanks again for your answer. Image a scenario where you have a "article detail module" (i.e. a module that does not show an article overview, but only one article) that requires an ID (passed as paths or in the query string). Removing the ID information from the language links also results in a broken link which needs to be handled in any way (404, redirect or show a "Sorry that item does not exist" message). I think, if modules on a multi-language page need to support this and there is no easy solution for all possible problems.

In my opinion, if there was a decision that paths parameters are not removed, query parameters should also be not removed for consistency. I've created a new pull request (https://github.com/dnnsoftware/Dnn.Platform/pull/256) which keeps all query parameters except the "returnurl" parameter.

sachatrauwaen:
Image a scenario where you have a "article detail module" (i.e. a module that does not show an article overview, but only one article) that requires an ID (passed as paths or in the query string). Removing the ID information from the language links also results in a broken link which needs to be handled in any way (404, redirect or show a "Sorry that item does not exist" message). 

> I don't agree with this, because in DNN a page allways have to exist. The page url is always must be valid. I think a module have always to shows something in that case. Something like a article list or message because you can not make this page not exist and not accessible otherwise you can not  manage the module on it.

I think, if modules on a multi-language page need to support this and there is no easy solution for all possible problems.

> My opinion, it is beter not support some use cases, then generating errors in some use cases.

In my opinion, if there was a decision that paths parameters are not removed, query parameters should also be not removed for consistency. I've created a new pull request (#256) which keeps all query parameters except the "returnurl" parameter.
> I agree

PR #253, opened 2015-01-03
DNN-6323 Clean list name before using as resource file name

bdukes:
#128 and #148 added localization to list values, by looking up a resource file with the same name as the list (plus a prefix). However, if the list name contains characters that are invalid for a file, an exception is thrown (see the stack trace on https://github.com/EngageSoftware/Engage-Employment/issues/6).

See [DNN-6323](https://dnntracker.atlassian.net/browse/DNN-6323)

### DIFF
--- a/DNN Platform/Library/Common/Lists/ListEntryInfo.cs
+++ b/DNN Platform/Library/Common/Lists/ListEntryInfo.cs
@@ -21,6 +21,8 @@
 #region Usings
 
 using System;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


cathalconnolly:
merging into 7.4.1

Bing-Wu:
Hi Brian, I am QA and testing for DNN-6323. If possible, could you please provide me a sample module in JIRA, by which I can repro this problem and validate the fixes. Thanks in advance.

PR #254, opened 2015-01-03
DNN-6324 Don't HTML decode the Razor engine's output

bdukes:
The `RazorModuleBase` class (used by the Razor Host module) calls `Server.HtmlDecode` on the output from the `RazorEngine`. This is probably to combat Razor's default HTML encoding, but also means that anyone using the Razor Host module will not get the normal behavior of Razor, and will not be able to easily output HTML encoded content.

I think it is important for the behavior of the Razor Host module to match Razor itself, however this will be a breaking change for existing Razor Host scripts that need the extra HTML decoding.

See [DNN-6324](https://dnntracker.atlassian.net/browse/DNN-6324)


cnurse:
I am going to merge this because I agree with Brian's argument, and while it does break binary compatability - Joe has scheduled the fix.

PR #255, opened 2015-01-03
Fix broken SQL script

bdukes:
The INFORMATION_SCHEMA views don't provide access to the schema, so
they can't be used in places where the {databaseOwner} token needs to be used
(which is to say, they can't be used pretty much at all in DNN).  I've replaced
the usages of INFORMATION_SCHEMA.COLUMNS with sys.columns

This fixes issues introduced in 643d88 (labeled for DNN-5943)

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
@@ -299,9 +299,9 @@ BEGIN
 END
 GO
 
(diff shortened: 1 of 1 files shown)


cathalconnolly:
I discussed this with Joe and we're going to pull the change forward to 7.4.0. I will close this out and assign it to the 7.4.0 milestone so you get credit, but I will do the change manually in the 7.4.0.sqldataprovider file.

cathalconnolly:
manually merged into release/7.4.0

PR #257, opened 2015-01-04
DNN-23079: Fixed DnnLanguageComboBox width in FF

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-23079 for more details


janjonas:
Using CssClass= "dnnFormItem" is maybe not the best way to set 100% width, but I tried various settings of the DnnComboBox to defined the width. I used the class dnnFormItem because it contains a "width: 100%" definition.

Another solution would be to add a CSS rule to some default css fule:

.dnnLanguageCombo > div.RadComboBox {
  width: 100%
}

Maybe someone of the platform developers could decide which way the problem should be fixed.

janjonas:
Please see for a css solution https://github.com/dnnsoftware/Dnn.Platform/pull/258

cathalconnolly:
Jan, thanks for this - I'm going to merge #258 though as I prefer the css approach (it also allows us to later tweak if any other browser layout items affect the field). I'll close this and mark it as merged so you get credit

PR #265, opened 2015-01-09
DNN-5834 - SI: Page Refresh Interval - Page does not refresh

kp-xcess:
Enabled refresh metatag to be rendered only when in VIEW-mode.


jbrinkman:
Thanks for the quick turn around on updating the pull request.

PR #268, opened 2015-01-13
DNN-6415 Added support for Http-Equiv on Meta skin object

bdukes:
The Meta skin object has an `HttpEquiv` property, but it should also support `Http-Equiv`, since may skinners will see it work for `Name` and `Content` and assume that it works with `http-equiv`.

In a skin file, with the `meta` tag registered:

    <%@ Register TagPrefix="dnn" TagName="META" Src="~/admin/Skins/Meta.ascx" %>

the following skin object reference should work:

    <dnn:Meta http-equiv="X-UA-Compatible" content="IE=edge" InsertFirst="true" runat="server"/>

See [DNN-6415](https://dnntracker.atlassian.net/browse/DNN-6415)


cathalconnolly:
merging into 7.4.1

PR #269, opened 2015-01-13
DNN-6417 Allow skin and container images to be png

bdukes:
See [DNN-6417](https://dnntracker.atlassian.net/browse/DNN-6417)


cathalconnolly:
merging into 7.4.1

PR #272, opened 2015-01-15
DNN-6437 Restore deprecated GetModule methods

jbrinkman:
These are simple overloads of a more complex method call.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -1382,9 +1382,34 @@ namespace DotNetNuke.Entities.Modules
         /// get a Module object
         /// </summary>
         /// <param name="moduleID">ID of the module</param>
+        /// <returns>ModuleInfo object</returns>
+        /// <remarks>This overload ignores any cached values and always retrieves the latest data
+        /// from the database.</remarks>
+        public ModuleInfo GetModule(int moduleID)
+        {
+            return GetModule(moduleID, Null.NullInteger, true);
+        }
+
+        /// <summary>
+        /// get a Module object
+        /// </summary>
+        /// <param name="moduleID">ID of the module</param>
+        /// <param name="tabID">ID of the page</param>
+        /// <returns>ModuleInfo object</returns>
+        /// <remarks>This overload ignores any cached values and always retrieves the latest data
+        /// from the database.</remarks>
+        public ModuleInfo GetModule(int moduleID, int tabID)
+        {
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -1382,9 +1382,34 @@ public ArrayList GetAllTabsModulesByModuleID(int moduleID)
...
+        /// get a Module object
+        /// </summary>
+        /// <param name="moduleID">ID of the module</param>
+        /// <param name="tabID">ID of the page</param>
+        /// <returns>ModuleInfo object</returns>
+        /// <remarks>This overload ignores any cached values and always retrieves the latest data
+        /// from the database.</remarks>
+        public ModuleInfo GetModule(int moduleID, int tabID)
+        {
+            return GetModule(moduleID, tabID, false);
```
bdukes: This passes `false` for `ignoreCache`, but the documentation says that it _does_ ignore the cache.

DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -1382,9 +1382,34 @@ public ArrayList GetAllTabsModulesByModuleID(int moduleID)
         /// get a Module object
         /// </summary>
         /// <param name="moduleID">ID of the module</param>
+        /// <returns>ModuleInfo object</returns>
+        /// <remarks>This overload ignores any cached values and always retrieves the latest data
+        /// from the database.</remarks>
+        public ModuleInfo GetModule(int moduleID)
+        {
+            return GetModule(moduleID, Null.NullInteger, true);
```
bdukes: I understand that this is just reinstating the obsolete method, but is there a good reason to have this API ignore the cache by default?  Shouldn't you have to make an explicit request to go around the cache?

bdukes:
See JIRA issue at https://dnntracker.atlassian.net/browse/DNN-6437

jbrinkman:
I did not alter either one of the defaults in these methods because it was unclear what the legacy behavior of these methods was prior to 7.3.  I presume that they had different semantics which is why they use different cache settings.

I'll investigate the history and make changes if appropriate.

jbrinkman:
After further research, the caching behavior is correct. DNN caches moduleinfo objects using TabID as part of the cachekey.  When you don't have a TabID, then you don't have the info needed to pull the object from the cache, thus GetModule(int moduleId) defaults to ignoring the cache since it won't find anything in the cache anyway.

bdukes:
rebased and merged

PR #273, opened 2015-01-15
DNN-6439 Update jQuery UI JavaScript from 1.10.3 to 1.11.2

bdukes:
I ran dnn.js through uglify, not sure if there's a standard minifier that is used
This commit doesn't attempt to update CSS, but I didn't see any immediate red
flags when looking through the DNN form pattern (dialogs, tabs, section expand/collapse)


cathalconnolly:
reviewed looks good (good to see the temporary code removed also) - merging into 7.4.1

cathalconnolly:
im going to revert part of this change as it doesn't work - modal dialogs are no longer centered.

cathalconnolly:
tested this again with jquery-ui 11.1.2 and jquery 11.1.3 and both still cause the modal windows to render in the bottom left of the screen. I tried debugging and couldn't see a quick resolution to this, so I am going to revert this pull request fully.

w8tcha:
I created a new pull request with a correct fix for the modal dialog

https://github.com/dnnsoftware/Dnn.Platform/pull/330

PR #275, opened 2015-01-19
DNN-6404 There should be more meaningful word 

EPTamminga:
in the help text. SImple change, but somebody has to do it. 
Good excercise for working with GitHub as well


cathalconnolly:
code reviewed and MarkAllPagesTranslated does not use placeholders, so I will merge this into 7.4.1

PR #276, opened 2015-01-19
Missing Resource Keys - Site Settings

EPTamminga:
It is a simple and dirty job, but somebody has to do it :-)


cathalconnolly:
thanks, every bit helps :)

cathalconnolly:
merging into 7.4.1

PR #277, opened 2015-01-20
DNN-6464 Upgrade to 7.4.0 - Windows Live Auth Hangs 

skydnn:
Updated urls used for live oauth service to be consistent with current Microsoft documentation for OAuth v2 authorization.


cathalconnolly:
as this is a recently introduced error and is localized to a small (and rarely used) function, I asked and have been given permission to merge this into release/7.4.0 so it will be part of 7.4.0

KenGrierson:
Hello;
I'm testing this fix in Platform 7.4.0 today and I need to know which which live domain should I be using to register an Application? 
http://dev.live.com
OR
https://account.live.com/developers/applications/index

Thank you!!

PR #278, opened 2015-01-20
(doc) Clarify use of portalId param

roman-yagodin:
It's not obvious how to use GetTab() method in the case we don't know portalId, but only tabId (may happen in some cross-portal scenario), without looking on GetPortalId() source.


jbrinkman:
Thanks for the commit.  This is a common pattern throughout DNN and I think that sometimes we forget this when creating documentation which causes problems for new platform developers.

PR #279, opened 2015-01-20
Dnn 6096: Member Directory SortBy error

svdv71:
See: https://dnntracker.atlassian.net/browse/DNN-6096
- Instead of changing the available SortFields, we'll use "UserId" as SortField if the TabModuleSetting SortField was CreatedOnDate. This is because the GetUsersBasicSearch and GetUsersAdvancedSearch do not allow sorting on CreatedOnDate. Sorting on UserId however has the same effect as sorting on CreatedOnDate.
- Also invalid T-SQL statements are filtered from the inputstring.


cnurse:
While I will merge this - as is - the regex change for invalid T-SQL should really be a separate JIRA.

PR #282, opened 2015-01-22
DNN-6493 Don't treat profanity filter values as regex

bdukes:
See [DNN-6493](https://dnntracker.atlassian.net/browse/DNN-6493)


cathalconnolly:
im going to check up about this as this is already extensively used in evoq, so this would be a "breaking" change for them (and anyone else who consumes the API method)

EPTamminga:
I do not understand why this would be a breaking change? Or am I missing a point?

cathalconnolly:
today a term such as "fuc*" would catch a common 4 letter swear word and a number of variants - changing to a string compare would mean that would no longer work. As this is a change in expected behaviour it qualifies as a breaking change

EPTamminga:
OK, now clear to me. Tnx.

cathalconnolly:
I've sent out a few mails about this and no-one has objected, so I'm merging into 7.4.1

PR #284, opened 2015-01-23
dnn-6499 Path too long exception when deleting a portal.

JoshuaBradley:
Moving recursive folder and file delete methods to FileSystemUtil and using long path dlls to prevent path too long exceptions.


cathalconnolly:
thanks, interesting approach. I notice that the license for LongPath is GPL -we used not to accept GPL projects, I will check and see if that is still the case

jbrinkman:
LongPath uses the LGPL v3 License (https://github.com/peteraritchie/LongPath/blob/master/LICENSE) so it is compatible with our MIT license.

More problematic is that this library requires .Net 4.5 so it **cannot be considered prior to DNN 7.5.0.**

JoshuaBradley:
I'm okay with this waiting till 7.5.0.

ohine:
This item as issues when running on a unc based installation. Josh, any plans to revisit this for unc support? Or should we close it till a later date?

PR #285, opened 2015-01-27
DNN-6506 Integration of a flexible imagehandler

weggetor:
To enable imagehandler functionality add following in web.config to "system.webServers/Handlers":

<pre>&lt;add name="DnnImageHandler" path="DnnImageHandler.ashx" verb="*" type="DotNetNuke.Services.GeneratedImage.DnnImageHandler, DotNetNuke" preCondition="integratedMode" /&gt;</pre>

Further documentation follows


jbrinkman:
Thanks Torsten.  Sorry for the delayed response.  We are a bit heads down trying to finish 7.4.0 testing.

weggetor:
Just added the documentation to the jira case: https://dnntracker.atlassian.net/browse/DNN-6506

bdukes:
Curious what folks think about the license on http://imageresizing.net/licenses/freedom.  It'd be nice to have this built upon its solid foundation, but I'm not exactly sure how its licensing would apply to our scenario.

weggetor:
Brian, its not only about the image resizing. Its about generating images too. And I made it extensible so you can add your own dll with a class derived from ImageTransform baseclass - really easy to implement. So why use another product when its all here ? About solid foundation: Its based on Microsoft original GeneratedImage project on codeplex: https://aspnet.codeplex.com/releases/view/16449

jbrinkman:
The Imageresizing.net solution is essentially a commercial solution. It offers a crippled version as Open Source.  The Open Source version does not have a caching feature (which is in their commercial product).  Because of the license, we could not build our own caching feature and distribute it with DNN.  Thus, I don’t see a way to use this with DNN since we would not want to ship a feature which would kill scalability for DNN.

roman-yagodin:
About another GeneratedImage- derived projects - please look at: http://bbimagehandler.codeplex.com/ and also my humble https://github.com/roman-yagodin/R7.ImageHandler, which is based on BBImageHandler code.

But really, using `System.Drawing` to resize images will not give you great results...

ohine:
51 Degrees has also included an image optimizer in pull request #621 

https://github.com/51Degrees/Dnn.Platform/commit/5ce23639be355d9b34cba351e156243f983b0e6b#diff-a128e381d81a7adf1019c7f5bf38c6b5R22

weggetor:
@roman-yagodin , this pull request is based on bbimagehandler (which is my project). The DNNImageHandler is the successor. If the code does not make it into the core, I'll update bbimagehandler
@ohine Oliver this pull request is not only for "optimizing" images, it is also for generating images (bar graphs etc.) with an implemented extension point.

roman-yagodin:
I'll hope it does make into the DNN core, @weggetor.
And thanks for your great job with BBImageHandler anyway :+1:

jbrinkman:
After reviewing ImageResizer.net I found that it caused performance degradation, even on pages where it is not "used".  We have spent a lot of effort identifying performance bottlenecks in DNN and have worked with Mitchel to implement a number of enhancements which will be included in DNN 8 (a few are included in 7.4.2).  One big performance boost came from ensuring that we don't run images, CSS or JavaScript through any HttpModules.  On my local machine this is a 10 second run on an unauthenticated page:
742speed-clean

I then did another test running with imageresizer which resulted in losing about 400 requests per second in performance.
742speed-imageresizer

In addition, the current imageresizer.net does not include a caching solution in the base package. In order to get disc caching, you have to purchase a performance license.  This would result in even worse performance for sites that actually used the resizer since the image resizing would have to be performed on each page load.  For a single user, this might not be too bad, but once you start adding load to the site, then it would become very noticable.

As a result, we will use Torsten's solution and make the needed modifications to apply our anti-DOS logic from the current profilepic handler.

PR #287, opened 2015-01-31
correct fix for DNN-6518

sleupold:
use portal default language, if no culture code is specified

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -2350,37 +2350,41 @@ namespace DotNetNuke.Entities.Portals
         }
 
         /// <summary>
-        /// Gets the portal.
+        /// Gets the portal (in portal default language).
         /// </summary>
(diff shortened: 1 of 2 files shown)


jbrinkman:
How extensively was this change tested?  There are many places in core platform and in 3rd party systems which rely on GetPortals returning data correctly in all scenarios.

sleupold:
stored procedure has been tested on live installs.
Code changes cannot result in a different result as before the first fix.
The previous version of the fix would return all sites with any portal language, if the requested language is not used for the site, i.e. requesting GetPortals('zn-CH') would return all sites, even if they don't have this language assigned.

jbrinkman:
I have tested this based on both upgrades and new installs and have included it with some modification to ensure that we retained behavioral compatibility with prior versions.  The new version is still undergoing internal testing but should be merged either today or tomorrow.  Thanks for catching the issue Sebastian.

cathalconnolly:
tested and merged into release/7.4.0

PR #288, opened 2015-02-03
DNN-6539 Only English ASCII nicknames are searchable!

cathalconnolly:


jbrinkman:
:shipit:

cathalconnolly:
im going to close this, the clone project had an issue part way through and it appears it missed a few files - whilst it doesn't seem to have hurt in this case, I would prefer to get it right, so i'll copy the change, close this and reclone the repo.

PR #289, opened 2015-02-04
DNN-6539 Only English ASCII nicknames are searchable!

cathalconnolly:
Fixed my repo - here is a clean checkin which can be automerged


donker:
Are we removing diacritics in search or is this too difficult to do? It's one of those things I do for my non-english projects. Maybe this is a matter of taste, though. To make sure we're on the same page: what I'm referring to is that "mel" will draw out "Méline" for instance.

cathalconnolly:
there's no intent on doing that in this pull request, it simply ensures that unicode text matches occur (i.e. not ascii)

jbrinkman:
Rebased and merged

PR #290, opened 2015-02-04
Update ControlPanelBase.cs

liffeyd:
It seems to me that PortalController.Instance.GetCurrentPortalSettings().ControlPanelSecurity == PortalSettings.ControlPanelPermission.ModuleEditor, at line 184, is preventing Admin > Site Settings > Usability Settings > Control Panel Settings from being set to anything other than Module Editors. If this setting is set to Page Editors the module cannot be edited.

### DIFF
--- a/DNN Platform/Library/UI/ControlPanels/ControlPanelBase.cs
+++ b/DNN Platform/Library/UI/ControlPanels/ControlPanelBase.cs
@@ -1,6 +1,6 @@
 #region Copyright
 // 
-// DotNetNuke® - http://www.dotnetnuke.com
+// DotNetNukeÂ® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2014
 // by DotNetNuke Corporation
 // 
@@ -181,7 +181,11 @@ namespace DotNetNuke.UI.ControlPanels
                     }
                 }
             }
-            return PortalController.Instance.GetCurrentPortalSettings().ControlPanelSecurity == PortalSettings.ControlPanelPermission.ModuleEditor && _IsModuleAdmin;
+            // Unless ControlPanelSecurity (Admin > Site Settings > Usability Settings > Control Panel Settings) 
+            //			== ControlPanelPermission.ModuleEditor
+            // this will always return false.
+            //return PortalController.Instance.GetCurrentPortalSettings().ControlPanelSecurity == PortalSettings.ControlPanelPermission.ModuleEditor && _IsModuleAdmin;
+            return _IsModuleAdmin;
         }
 
         internal static bool IsPageAdminInternal()


vmasanas:
Declan, can you provide a Jira ticket for this where you explain the use case? I'm not 100% sure on what this is trying to fix. A sample on how to reproduce would be helpfull.

cathalconnolly:
@cnurse and I discussed this last night - here's the relevant text:

[20/03/2015 21:20:05] cathalconnolly: the current check is if the site allows module editors to see the control panel and they are an editor on 1 or more modules on a page
[20/03/2015 21:20:15] cathalconnolly: changing it to Declans changes that logic
[20/03/2015 21:20:37] Charles Nurse: I agree
[20/03/2015 21:20:52] Charles Nurse: The point is that admins MUST allow Module Editors to access the controlpanel
[20/03/2015 21:21:07] Charles Nurse: otherwise that setting is meaningless
[20/03/2015 21:22:07] Charles Nurse: Now - this is only a property that returns IsModuelAdmin
[20/03/2015 21:22:25] Charles Nurse: the visibility may be enforced later in conjunction with the ControlPanelMode
[20/03/2015 21:22:52] Charles Nurse: But if this is all that controls the visibility then we can't change it
[20/03/2015 21:23:00] cathalconnolly: yup
[20/03/2015 21:23:22] Charles Nurse: So you should check the usage of the property
[20/03/2015 21:23:45] cathalconnolly: i figured it should be ControlPanelPermission.PageEditor or (ControlPanelPermission.ModuleEditor && _IsModuleAdmin;)
[20/03/2015 21:25:00] Charles Nurse: I guess what I am saying is that the property name is IsModuleAdminInternal
[20/03/2015 21:25:16] Charles Nurse: to satisfy the name of the property the change is actually valid
[20/03/2015 21:26:01] Charles Nurse: we need to check how the property is used
[20/03/2015 21:26:19] Charles Nurse: ie - is the extra check in this property redundant
[20/03/2015 21:26:31] Charles Nurse: because the security setting is checked later
[20/03/2015 21:26:56] cathalconnolly: sure -i'll follow it through but its only refernced in controlbar.ascx

liffeyd:
Vicenç,
It is detailed here https://dnntracker.atlassian.net/plugins/servlet/mobile#issue/DNN-6087

vmasanas:
Cathal I'll leave that up to you since you seem more familiar with that

cnurse:
This is an incorrect fix for the issue referenced.  This is demonstrated by the comment in the JIRA ticket that other modules work OK, but the HTML module does not work.

This fix basically ignores/overrides the Site Setting for Control Panel Visibility.

The correct fix is in the module that fails to work properly.

PR #291, opened 2015-02-06
Dnn-3522: If-tokens in DDRmenu templates fail

ohine:
The token template processor would only check for true/false values (0, 1). Added a super ugly check if they're not 0 or 1 and also not an empty string. It seems to solve the issue with checking for icons or any other string node via the ddr token text template.

### DIFF
--- a/DNN Platform/Modules/DDRMenu/MenuBase.cs
+++ b/DNN Platform/Modules/DDRMenu/MenuBase.cs
@@ -9,6 +9,7 @@ using System.Web.UI;
 using System.Xml;
 using System.Xml.Serialization;
 using DotNetNuke.Common.Utilities;
+using DotNetNuke.Entities.Host;
 using DotNetNuke.Entities.Portals;
 using DotNetNuke.Entities.Tabs;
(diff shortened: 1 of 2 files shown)


KenGrierson:
Hi Oliver
Can you provide an example use case for this fix?
I've tried adding the page icon definition in the JIRA but I don't get any change after I've done that and recycled the AppPool from the Toolbar

Thx,
Ken

ohine:
The ddr menu template needs to be modified to include the page icon token. Then need to modify the page settings and select icons for the page. Let me know if you are still having issues, and I can try and package up my test skin to the jira ticket.

cathalconnolly:
QA still reports issues with missing images in the JIRA - https://dnntracker.atlassian.net/browse/DNN-3522 . As we are closing the release today, I am going to revert this back, and we can then resolve it in 7.4.2 i.e. see if the issue is in our setup, or perhaps a bug in the pull request.

cathalconnolly:
QA now reports they have this working, so I will leave everything as is.

PR #292, opened 2015-02-06
DNN-6562 Add cache buster to logo uri so that site management may add ca...

RichardHowells:
...ching headers to logos


RichardHowells:
I'm withdrawing this request because I think I have an improved solution.

PR #295, opened 2015-02-10
DNN-6580 - 07.04.00.sqldataprovider is not rerunnable and other issues

sleupold:
smaller fixes, Github might be unable to display due to tab/space issues

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.00.SqlDataProvider
@@ -650,6 +650,8 @@ BEGIN
 	DELETE FROM {databaseOwner}{objectQualifier}CoreMessaging_NotificationTypeActions WHERE NotificationTypeId = @NotificationTypeId AND NameResourceKey = 'Review'
 
 	/* Add Approve notification action */
+  IF NOT EXISTS(SELECT * FROM {databaseOwner}{objectQualifier}CoreMessaging_NotificationTypeActions 
+                        WHERE NotificationTypeId = @NotificationTypeId AND ([Order] = 1 OR [NameResourceKey] = 'Approve'))
     INSERT INTO {databaseOwner}{objectQualifier}CoreMessaging_NotificationTypeActions(
         [NotificationTypeID],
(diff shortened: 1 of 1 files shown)


Website/Providers/DataProviders/SqlDataProvider/07.04.00.SqlDataProvider
```diff
@@ -691,7 +695,7 @@ BEGIN
         'Reject',
         'Reject',
         NULL,
-        1,
+        2,
```
bdukes: Why are you changing the order from 1 to 2 for this and the discard action later?

sleupold: Order should be unique per notification type (see all other items). 
I'll submit a pull request for 07.04.01 to modify these values and add a unique key

sleupold: PS: and it broke update for users of my Turbo scripts, which established an unique index on these columns

jbrinkman: Breaking Turbo scripts is not a valid reason for altering the behavior of DNN. If there is an index that you feel should be added to the core then submit the change and it can be reviewed and approved. Until that index is an accepted part of the standard DNN platform it should not be part of deliberation process for accepting or rejecting additional changes.

Just based on looking at other NotificationTypeActions it does appear that the correct value should be a 3 (not a 1 or a 2).  I'll check with the original developers of the notification feature to determine the correct behavior.

sleupold: thanks for the feedback, jbrinkman. 
Is there anywere a documentation of workflow, how it is constructed and should be used (it will be difficult to create a UI for it without further background. thank you.

sleupold: added DNN-6603: Add unique key on table CoreMessaging_NotificationTypeActions

jbrinkman:
Manually merged for 7.4.1

PR #296, opened 2015-02-10
Dnn 6560 string performance

mitchelsellers:
Optimizations across the platform focused on improving the management of strings.  Use of StringBuilder, the proper instantiation of strings, as well as avoiding unnecessary allocations.


donker:
I ran all tests and the code looks good to me. Merging.

PR #297, opened 2015-02-10
DNN-6585: Increased maxUrlLength

stephen-lim:


cathalconnolly:
I'm not sure about accepting this pull request as it changes the relaxedUrlToFileSystemMapping value. I got this from an asp.net 4.0 document written by Stefan Schackow of the asp.net team

"By default, as part of the configuration lookup, ASP.NET assumes the path portion URL is a valid NTFS file path—that is, this attribute is set to false by default. You can
disable this constraint by setting relaxedUrlToFileSystemMapping to true.


If you disable this constraint, you should validate request URLs and constrain them to work only with URL patterns that the application expects. In other words, by disabling this check, you disable the implicit protection that is provided by ASP.NET where the configuration system enforced the constraint that request URLs are valid NTFS directory paths."

stephen-lim:
I did some quick tests to see if we can do without the relaxedUrlToFileSystemMapping. It appears the maxUrlLength attribute is sufficient to extend the URL length. So I'm going to remove the relaxedUrlToFileSystemMapping setting. Please see the pull request for the files changed have now been updated.

Thanks,

cathalconnolly:
thanks - we'll go ahead and process this in 7.4.1

cathalconnolly:
i'm going to merge this today - please note, the submitting pull request only updates our default installation settings, if you want this to apply during an upgrade we'll need a relevant xml merge config

PR #300, opened 2015-02-12
DNN-6562 Browsercaching

RichardHowells:
DNN-6562 - Add cache buster to uris for images and other resources accessed from the browser via a FileUrl

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -626,6 +626,19 @@ namespace DotNetNuke.Entities.Portals
             }
         }
 
+        /*
+         * add a cachebuster parameter to generated file URI's
+         * 
+         * of the form ver=[file timestame] ie ver=2015-02-17-162255-735
+         * 
+         */
+        public bool AddCachebusterToResourceUris
+        {
+            get
+            {
+                return PortalController.GetPortalSettingAsBoolean("AddCachebusterToResourceUris", PortalId, true);
+            }
+        }
 
 		#endregion
 
--- a/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
+++ b/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
@@ -208,7 +208,7 @@ namespace DotNetNuke.Services.FileSystem
             {
                 return Globals.LinkClick(String.Format("fileid={0}", file.FileId), Null.NullInteger, Null.NullInteger);
             }
-            return TestableGlobals.Instance.ResolveUrl(rootFolder + file.Folder + file.FileName);
+            return TestableGlobals.Instance.ResolveUrl(rootFolder + file.Folder + file.FileName + "?ver=" + file.LastModificationTime.ToString("yyyy-MM-dd-HHmmss-fff"));
         }
 
         public override string GetFolderProviderIconPath()
--- a/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
+++ b/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
@@ -37,7 +37,7 @@ namespace DotNetNuke.Services.FileSystem
 {
     public class StandardFolderProvider : FolderProvider
     {
-    	private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof (StandardFolderProvider));
+        private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(StandardFolderProvider));
 
         #region Public Properties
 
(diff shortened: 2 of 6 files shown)


janjonas:
@RichardHowells: Thanks for the detailed feedback. I understand all of your arguments.

IMHO, your HTML module scenario shows the big problem: Even if the module is clever enough to update the image URLs if the file changes you have the same problem with external sources that refer to your resources (websites or maybe search engines) because they can't know that a resource (and it's URL) has been updated.

I would suggest the following: Instead of (automatically) adding a version token to the URLs, DNN provides functionalities (for administrator) to define (client-side) cache times (for files and/or folders). At the moment in a default DNN 7.4.0 installation there is a default cache time of 365 days for static resources defined in the web.config:
<clientCache cacheControlCustom="public" cacheControlMode="UseMaxAge" cacheControlMaxAge="365.00:00:00" />
Depending on the frequency a resource changes, cache times could be defined (for logos a cache time of 7 days could be reasonable). Of course, suggestion does not fix the problem that there is some delay before all clients get the new resources, but this feature will not introduce any other drawbacks. If a developer wants to make sure that clients always gets the most recent version, he could either use some custom logic or the abstract FolderProvider provides a new method GetFileUrl(IFileInfo file, bool disableCache) which is implemented by all file providers.

Btw: Maybe a default caching of 365 day is a bit too long and one should decrease the default caching (for example to one day) and override the default value with a higher one for files/folders that most likely won't change. But this question should not be discussed in the pull request.

jbrinkman:
This change was originally introduced in 7.3.0 (673d5fd91ae6e3c2ca59531197d051bd1a931fd8).  I agree that the MaxAge value is overly aggressive and should be reduced to a 1 day cache value.  I think users will derive more performance benefit from moving the resource to a CDN than creating an extremely long caching duration.  I think this enhancement would be better than adding cache busting logic to the platform.

RichardHowells:
@janjonas Do you have access to the MVP forum? There is a thread there relevant to this subject.  I'd value your views. http://www.dnnsoftware.com/forums/forumid/228/threadid/515268/scope/posts

jbrinkman:
@RichardHowells Jan doesn't currently have access to MVP forums. This is the better place to have the discussion since it is open to anyone in the community who is interested in this feature.

RichardHowells:
Hi Joe,

May I present a different point of view?

- moving content to a CDN is a choice that is already in the user's hands. (S)he can run the tests/simulations/whatever and decide.  They don't need us.

- cache headers are in the users' hands.  They have to be tech savvy enough to add the right web config lines or click the right pieces in the IIS manger GUI but they don't need us.  Now we even do it for them.  By doing it we prevent them properly controlling their own content.

- cache busting is something the user has no chance of doing alone.  Ok they could create a custom fork of the DNN source code, but that is a pretty major thing to do.  Only we can do cache busting.

Cache busting is an essential corollary to caching.  It's the only way to keep the browser cache under control.  We already recognize the fact by adding the magic "Current host version" to most resources injected into the page.  That is a massive contribution.  The argument about "how long should we cache for" is pointless.  Any chosen duration is wrong.  Worse it puts the developers and the site managers in an impossible position.  We cannot know what near random collection of scripts, images, CSS files and other resources any particular user's browser may have cached.  The immense number of combinations makes it impractical to test what we release.  It causes support calls and encourages call handlers to insist on 'clearing the cache' before they do anything useful.  Nobody likes to lose all their saved passwords and cookies, but real call handlers cannot make the distinction.

There is only one proper duration for caching - forever.  In practice the relevant RFC says one year max and that's close enough to forever.  I won't repeat the arguments.  This article [1] is great, does it better than I can, and has diagrams.

Even with a CDN you still require caching, cache headers, and proper cache busting.  The CDN may be 'closer' but nothing will beat the zero cost of serving a resource from the local cache.  Inadequate cache busting will result in the CDN serving the wrong resource.  Having the wrong resource faster is no comfort.

A CDN may give better performance than cache and cache busting.  CDN's are expensive and many, even quite large sites, will not have the budget.  Adding cache headers is free, we've already done it.  We should complete the job by doing cache busting as well as we can.  We've covered part of the job with the "Current host version".  Pull request #300 (https://github.com/dnnsoftware/Dnn.Platform/pull/300) is another part of the job.


[1]   http://blog.httpwatch.com/2007/12/10/two-simple-rules-for-http-caching/

janjonas:
@RichardHowells: What about adding your logic in the a new FolderProvider (that may extends the StandardFolderProvider)? This special FolderProvider could be part of a separate extension. Of course, to benefit from these FolderProvider the administrator has install the extension and use configure the the folders to use the FolderProvider.
This would keep the StandardFolderProvider as "clean" as possible. IMHO, this special cache busting is nothing that should be added to a StandardFolderProvider. It's just a standard folder provider, as the name implies, so there should be no magic :).

RichardHowells:
@JanJonas: Thank you for the suggestions.  IMHO cache busting is not special.  Caching and cache control measures are essential to the operation of the internet.

Cache busting has been a standard part of DNN through the "Current host version" setting for a significant time.  Adding cache headers is standard in new DNN7.3 installs.  These things are built in standard features.

I think it introduces a significant extra barrier if an administrator has to install an extension to get control over something we introduce automatically.

PR #301, opened 2015-02-12
DNN-3796 Update RewriterUtils.cs

mortezakermani:


KenGrierson:
Hello Morteza,
I re-opened the JIRA item associated to this Pull Request.
In IE 11 I get an incorrect value /key/%D8%B3%DB%8C%D8%A8
In FF, Chrome and Safari I get the correct value /key/سیب

PR #307, opened 2015-02-17
DNN-5881 Enable Module Effect? feature does not work.

cathalconnolly:
removed enable module effect related code as it hasn't applied for a few versions now (the two edit states shared the same opacity)


EPTamminga:
Tnx! I always hated the opacity effect.

PR #310, opened 2015-02-18
DNN-6557 - added web method to delete a tab module.

fjlucas:
Added new web method to allow deleting tab modules.

Note: it has been introduced for integration testing purpuses.


francescorivola:
reviewed and merged.

PR #315, opened 2015-02-20
DNN-6609 HTTPS 301 redirect

cathalconnolly:
When SSL-enforced is enabled, and the page is secure, it will now do a http 301 redirect when using advanced url's (basic already does a 301 redirect)


janjonas:
@cathalconnolly: I would not change the redirect to a permanent one. IMHO, this makes deactivating SSL (to switch back to http://...) very difficult or even impossible for most users. The 301 redirects are cached on clients and the site become inaccessible and the user manually has to delete the cache. There should at least exist an option (which is false by default) to activate permanent redirecting.

cathalconnolly:
@janjonas that seems a valid point (though that is the way the basic provider has worked since 3.0.0 and it hasn't been an issue) - i'll log it to the jira and see the response

jbrinkman:
We already have too many "settings" for the application.  We should be optimizing for the 90% use case as long as there is a workaround for the 10%.  You could make the case using the same logic, that we should never use 301's for anything since it is always possible that the admin might change it back someday. The change to a 301 is correct.

PR #321, opened 2015-02-25
DNN-6649 Now the ModuleRemoved in IModuleEventHandler is raised when a m...

pberna:
Now the ModuleRemoved in IModuleEventHandler is raised when a module is removed from a page.
Also a complete ModuleInfo object is used when raising the event of removing a module from all pages instead of one with only the ModuleId.
Could you please review the code and merge it?


jbrinkman:
Just a reminder that when merging a commit, please update the milestone so we know which release includes this change. This really simplifies our reporting.

PR #325, opened 2015-02-26
DNN-6662 Load full hierarchy of JS library dependencies

bdukes:
[DNN-6662](https://dnntracker.atlassian.net/browse/DNN-6662)

>When requesting a JavaScript Library to be included on the page, its direct dependencies are also included on the page. However, any dependencies of those direct dependencies are not included.


cathalconnolly:
@cnurse Can you please review this - I believe this was the plan for JSL, but we ran out of time and it seems Brian's done the work for us, but if you could confirm that this matches your planned architecture it would be appreciated.

PR #327, opened 2015-02-27
DNN-6660 ClientDependancy breaks dataUris in css

bradbamford:
Fix for the CDF striping out the data: attribute from CSS embed images. Images can actually render now.


cathalconnolly:
quick update: we're considering updating CDF to the latest release - if we do that I will verify if it fixes this issue, if not I'll merge the check as part of 7.4.1. If we do not have time for a full CDF update, I'll merge this issue in 7.4.1

cathalconnolly:
We feel that the upgrade to the latest CDF is better kept until 7.5.0 (to allow for additional testing), so I will be merging this change into 7.4.1

PR #328, opened 2015-02-27
DNN-6622- Hide Enable Language Parameter in URLs if using AUM.

ludikduran:
Updated Local Resource for Language Enabler.
Added check on urlFormat value in web.config. If using advanced, checkbox is disabled.


cathalconnolly:
merging into 7.4.1

PR #333, opened 2015-03-02
DNN-6746: improving returned load size from /Install/installstat.log....

galatrash-at-dnn:
...resources.txt endpoint

I added two minor fixes as well
- formatting installer log date as yyyyMMdd instead of yyyyMd
- made installer log entries date/time stamp local (same as regular logs) instead of UTC.

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
+++ b/DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
@@ -67,7 +67,7 @@ namespace DotNetNuke.Services.Upgrade
 			var logfilePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"Portals\_default\Logs\" + logFile);
 			using (var writer = new StreamWriter(logfilePath, true))
 			{
-				writer.WriteLine(DateTime.Now.ToString(CultureInfo.InvariantCulture) + " " + logType + " " + CallingType + " " + message);
+				writer.WriteLine(DateTime.UtcNow.ToString(CultureInfo.InvariantCulture) + " " + logType + " " + CallingType + " " + message);
 			}
 
 		}
--- a/DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
+++ b/DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
@@ -1,13 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
-using System.Linq;
-using System.Text;
 using System.Web.Compilation;
-
-using DotNetNuke.Instrumentation;
+using DotNetNuke.Services.Upgrade.Internals;
 
 namespace DotNetNuke.Services.Upgrade
 {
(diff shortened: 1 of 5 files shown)


DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
```diff
@@ -67,11 +63,11 @@ public static void InstallLogInfo(object message)
...
 		private static void LogInstaller(string logType, string message)
 		{
-			string logFile = "InstallerLog" + DateTime.Now.Year.ToString(CultureInfo.InvariantCulture) + DateTime.Now.Month.ToString(CultureInfo.InvariantCulture) + DateTime.Now.Day.ToString(CultureInfo.InvariantCulture) + ".resources";
-			string logfilePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"Portals\_default\Logs\" + logFile);
+		    var logFile = InstallController.Instance.InstallerLogName;
+			var logfilePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"Portals\_default\Logs\" + logFile);
 			using (var writer = new StreamWriter(logfilePath, true))
 			{
-				writer.WriteLine(DateTime.UtcNow.ToString(CultureInfo.InvariantCulture) + " " + logType + " " + CallingType + " " + message);
+				writer.WriteLine(DateTime.Now.ToString(CultureInfo.InvariantCulture) + " " + logType + " " + CallingType + " " + message);
```
jbrinkman: We should keep the install at UTC time.  Installs occur on the "server" which is not guaranteed to be in the same timezone as the administrator doing the install (this is especially true in cloud environments). Using UTC ensures that users can easily figure out the correct time for their location.

cathalconnolly:
can you please clone the social ticket as a platform ticket so we have a valid public jira to associate this with

DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
galatrash-at-dnn: I did this to be the same as the regular log files. I heard it from QA that they had problem relating bugs to actual lines between the two log files (regular and installation). IMO, both should have the same format; either both are UTC or local time but not one UTC and the other local.

ashishpd:
Here is the Platform Jira for this issue: https://dnntracker.atlassian.net/browse/DNN-6746

cathalconnolly:
thanks, i'll review and merge tomorrow

DNN Platform/Library/Services/Upgrade/DnnInstallLogger.cs
jbrinkman: I think all dates should be in the UTC time, and translated for display.  If QA has a problem with the way dates are stored then they need to create a proper Jira so that it can be discussed by the PMs. Please back out that single change so we can merge the rest of this fix.  I am happy to deal with the date issue on a separate Jira item.

galatrash-at-dnn: I've reverted that date back to UTC and included it in the same pull request.

cathalconnolly:
merging into development and cherrypicking into release/7.4.1

PR #334, opened 2015-03-03
DNN-6170 Switching Language does not clear portalsettings cache

evidnn:
Since PortalCacheKey now has a culture variation, each culture, including neutral one would have its own portalsettings dictionary. 
When site settings are loaded both dictionaries would be used, depending on the current culture and setting type (culture dependent or not). When setting gets update we need to ensure both cachekeys are cleared (when clear cache is set to true) because we don't know if setting is culture specific or not.

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -2265,7 +2265,8 @@ namespace DotNetNuke.Entities.Portals
                 }
                 else
                 {
-                    list.Add(new PortalTemplateInfo(templateFilePath, ""));
+                    //DNN-6544 portal creation requires valid culture, if template has no culture defined, then use current 
+                    list.Add(new PortalTemplateInfo(templateFilePath, GetCurrentPortalSettingsInternal().CultureCode ?? Thread.CurrentThread.CurrentCulture.Name));
                 }
             }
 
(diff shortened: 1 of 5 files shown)


jbrinkman:
During testing, I found that this does not actually resolve 6170.

See this video http://1drv.ms/1ENAByr

evidnn:
Here is a summary of 4 commits that are all required to resolve this issue.
These two commits resolve first part of the problem reported:
"English texts are loaded, but not Skin Settings":
https://github.com/dnnsoftware/Dnn.Platform/commit/2f8bf713c140ff35fc788e91e2a3f7a9ae28636b
https://github.com/dnnsoftware/Dnn.Platform/commit/54d854923060e8bc7336c6720c5aa2e316df7be5
This commit resolves the second issue of skin loading "English Skin is applied, once you clear Portal Cache"
https://github.com/dnnsoftware/Dnn.Platform/commit/87c8c6735543ae85ad224b8a5099176b0bfee908
Now after switching language culture specific skin (according to the the site settings) is loaded:
This commit resolves the issue reported by Ben (it is PortalSettings related so I resolved it within the scope of this JIRA as it affects not only host->extensions, but other site settings)
https://github.com/dnnsoftware/Dnn.Platform/commit/200206c4fc79c9eaa6f3b9d0bb2f7a3689b188a4

Test scenarios that were successfully completed:
Static Localization with 2 enabled languages
Page Skin load verification
Site settings: switching between languages
Pages: switching between languages on pages with no skin defined, with skin defined
Site Settings verification
Site settings: saving settings based on language: basic, advanced
AdminSkin load verification
Pages->editing pages: switching between languages on pages with no skin defined, with skin defined
Content Localization with 2 enabled languages (Content has been localized)
Page Skin load verification
Site settings: switching between languages
Pages: switching between languages on pages with no skin defined, with skin defined
Site Settings verification
Site settings: saving settings based on language: basic, advanced
AdminSkin load verification
Pages->editing pages: switching between languages on pages with no skin defined, with skin defined

PR #335, opened 2015-03-03
DNN-6673  try to authenticate request, if IIS is set to Windows Authentication

BarryWaluszko:
This patch fixing the issue described here: https://dnntracker.atlassian.net/browse/DNN-6673


cathalconnolly:
closed per user's request (will be handled as #440 )

PR #338, opened 2015-03-04
DNN-6672: Do not include GhostDoc solution files in commits

EPTamminga:
Needed for people working with Ghostdoc for auto code documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.user
 *.sln.docstates
 *.local.sln
+*.local.sln.GhostDoc.xml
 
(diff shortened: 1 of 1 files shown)


cathalconnolly:
I'm not sure we should be updating the .gitignore for 3rd party products - I'd suggest that it would be better for any user of a conflicting product to use "git update-index --assume-unchanged <file>" - i'll raise this with @cnurse to get his thoughts (as architect he "owns" the gitignore file)

cnurse:
This is fine.  We can accept this.

PR #348, opened 2015-03-09
DNN-6594: update CDN path to remove protocol.

zyhfish:


bdukes:
In my experience, the Client Resource Manager doesn't properly handle extension-less URLs when combining scripts (i.e. it doesn't see it as a remote URL and tries to combine it).  Has that been fixed?

PR #354, opened 2015-03-10
DNN-6105: if register failed, keep password field's value; if register c...

zyhfish:
...omplete but show message, then hide register form.


cathalconnolly:
this is not a valid checkin as it violates security policy by adding the password into the rendered output of the page i.e. password.Attributes.Add("value", Convert.ToString(Value)); - this would mean that in theory the cached page could be retrieved from a users temporary internet files and their password (or attempted password -which would likely be close to the actual password) would be revealed.

bdukes:
This looks like it should address the bug fine if you remove the change to `DnnFormPasswordItem.cs`

PR #357, opened 2015-03-10
DNN-6702: Issue with Shared Modules in Site Groups

sleupold:
DNN-6702: fix an unique index issue with shared modules in Site Groups

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
@@ -139,11 +139,6 @@ IF EXISTS(SELECT * FROM sys.indexes WHERE name = 'IX_{objectQualifier}ModulePerm
 	DROP INDEX [IX_{objectQualifier}ModulePermission_Roles] ON {databaseOwner}[{objectQualifier}ModulePermission]
 GO
 
-CREATE UNIQUE NONCLUSTERED INDEX IX_{objectQualifier}ModulePermission_Portals ON {databaseOwner}[{objectQualifier}ModulePermission]
-	(PortalID, ModuleID, PermissionID, RoleID, UserID) 
-	INCLUDE (AllowAccess)
-GO
-
 CREATE UNIQUE NONCLUSTERED INDEX IX_{objectQualifier}ModulePermission_Modules ON {databaseOwner}[{objectQualifier}ModulePermission]
 	(ModuleID, PermissionID, PortalID, RoleID, UserID) 
 	INCLUDE (AllowAccess)
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
@@ -117,6 +117,49 @@ AS
 
 GO
 
+-- DNN-6702: fix an issue with shared modules in Site Groups
+-- ---------------------------------------------------------
+IF EXISTS(SELECT * FROM sys.indexes WHERE name = 'IX_{objectQualifier}ModulePermission' AND object_id = OBJECT_ID('{objectQualifier}ModulePermission'))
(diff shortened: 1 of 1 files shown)


vmasanas:
I see you are adding an extra index IX_ModulePermission_Portals. This is the same as IX_ModulePermission_Modules just with a change in the order of columns. Is this really necessary?

sleupold:
Vicenç, IX_ModulePermission_Portals is not necessary, it has just been added for performance, especially of Groups module.

Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
```diff
@@ -117,6 +117,49 @@ AS
 
 GO
 
+-- DNN-6702: fix an issue with shared modules in Site Groups
+-- ---------------------------------------------------------
+IF EXISTS(SELECT * FROM sys.indexes WHERE name = 'IX_{objectQualifier}ModulePermission' AND object_id = OBJECT_ID('{objectQualifier}ModulePermission'))
+	ALTER TABLE	{databaseOwner}[{objectQualifier}ModulePermission] DROP CONSTRAINT [IX_{objectQualifier}ModulePermission] 
+GO
```
jbrinkman: There is no need to drop this constraint since it was removed in DNN 7.2.2 and hasn't been re-added to the platform.

Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
```diff
@@ -117,6 +117,49 @@ AS
...
+
+IF EXISTS(SELECT * FROM sys.indexes WHERE name = 'IX_{objectQualifier}ModulePermission_Roles' AND object_id = OBJECT_ID('{objectQualifier}ModulePermission'))
+	DROP INDEX [IX_{objectQualifier}ModulePermission_Roles] ON {databaseOwner}[{objectQualifier}ModulePermission]
+GO
+
+CREATE UNIQUE NONCLUSTERED INDEX IX_{objectQualifier}ModulePermission_Portals ON {databaseOwner}[{objectQualifier}ModulePermission]
+	(PortalID, ModuleID, PermissionID, RoleID, UserID) 
+	INCLUDE (AllowAccess)
+GO
```
jbrinkman: The need to add PortalID to the other indexes is clear enough from the issue history.  Why are we adding the ModulePermission_Portals index?  Which specific queries or SPROCS will use this Index versus the others?

Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
```diff
@@ -117,6 +117,49 @@ AS
...
+
```
jbrinkman: I understand the need for updating the other indexes, but why are we adding a new index (ModulePermission_Portals)? Is there a specific SPROC which will use this Index so we can look at the query plan?

Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
sleupold: that's right, but I've ssen too many sites, where Upgrades had issues and it doesn't cost any effort. But it is up to you.

Website/Providers/DataProviders/SqlDataProvider/07.04.01.SqlDataProvider
sleupold: There is GetModulePermissionsByPortal, but I don't have a site at hand with a large number of modules to validate. you may simply ignore this index.

jbrinkman: In looking through code, I can't find a single stored procedure which would use this query.  Even ones that look like they might are only called from deprecated methods and even then the actual query plan doesn't show this index getting used.  

As a result, please remove the ModulePermissions_Portals index as it appears unnecessary and will just add extra overhead to maintain the index.

PR #363, opened 2015-03-11
DNN-6594: update CDN path to remove protocol.

zyhfish:


bdukes:
Interesting, just saw a recommendation to HTTPS all the things, rather than protocol relative (http://twitter.com/Encosia/status/576074384487686144).

PR #365, opened 2015-03-11
DNN-6105: if register failed, keep password field's value; if register c...

zyhfish:
...omplete but show message, then hide register form.


zyhfish:
The change of password field will be ok, as confirm password field also output the password value.

cathalconnolly:
thanks for making the updates, merg9ing into 7.4.1

PR #383, opened 2015-03-17
DNN-6099: always remove user roles in current portal.

zyhfish:


cnurse:
Manually merged pull request

PR #388, opened 2015-03-19
Feature/dnn 6700

cathalconnolly:
First parts of DNN-6700
Contains:
1. update dotnetnuke.web to use WebAPI 2.0 (and dependant json.net 4.7.11)
2. update service framework consuming projects to also use WebAPI 2.0
3. fix a few namespace ambiguity issues (due to new requirement for json.net 4.7.11)
4. add assembly binding redirects to 07.05.00.config


cathalconnolly:
closing this as sourcetree made it against "development" when it should be against "feature/7.5.0"

PR #390, opened 2015-03-19
DNN-5373 SI: Email "From:" formatting

evidnn:
Added From and Sender to the mail. Both use combination of email + displayname


cathalconnolly:
reverting this due to https://dnntracker.atlassian.net/browse/DNN-6823 being logged (as per QA's request)

PR #395, opened 2015-03-20
DNN-6760 - Added some links to resources that help you to contribute in ...

robsiera:
...the correct way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,3 +33,10 @@ Quick Links
 * [Community Blogs](http://www.dnnsoftware.com/community-blog)
 * [Extensions Forge](http://www.dnnsoftware.com/Community/Download/DNN-Forge)
 * [DNN Store](http://store.dnnsoftware.com/)
+
+How to Contribute
+-----------------
+Some resources that might be helpful when you want to contribute
+* [Contributing to the DNN Platform using Git and Github](http://dnn-connect.org/community/blogs/contributing-to-the-dnn-platform-using-git-and-github)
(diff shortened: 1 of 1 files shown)


bdukes:
:+1: looks good to me

cathalconnolly:
thanks - though I would inclined to remove the first two links and use http://www.dnnsoftware.com/wiki/page/github instead as it collates all the useful github links

robsiera:
Ah, even better.

jbrinkman:
@robsiera If you just update it in your DNN-6760 branch, it will automatically update the Pull Request and we can go ahead and merge the change.

robsiera:
Don't hesitate to make the necessary linguistic improvements!

cathalconnolly:
it just occurred to me that we have all the relevant information on the contributing page at https://github.com/dnnsoftware/Dnn.Platform/blob/master/CONTRIBUTING.md - but that some people are clearly not seeing it as they pull the development repo - perhaps a change to point to the contributing.md page would be good?

cathalconnolly:
I've updated the wiki page to contain a link to the contributing.md, so I will merge this as is for 7.4.1 - thanks

cathalconnolly:
Note: we don't require JIRA tickets for documentation updates , so if you have any future pull requests to do documentation/xml documentation tags, don't worry about creating a jira. I will merge this and mark the jira as complete - however @jbrinkman  may want to delete the jira as it doesn't correspond with a change in platform.

PR #400, opened 2015-03-21
DNN-6636 Sitemap : Shows 404 Page in the Sitemap

vmasanas:
Sets the default priority for 404 pages to 0 so they are not included in the sitemaps.
Also fixes current 404 and 500 to priority 0.


cathalconnolly:
I believe part of this is already fixed as https://github.com/dnnsoftware/Dnn.Platform/pull/337

vmasanas:
No, this is independent. See my comment here https://dnntracker.atlassian.net/browse/DNN-6636

cathalconnolly:
makes sense, thanks for the explanation

PR #401, opened 2015-03-21
DNN-4130: only check check property on checkbox.

zyhfish:


cnurse:
This has been separately fixed.

PR #402, opened 2015-03-22
DNN-6758: Increase URL size

sleupold:
Increase column size for IconFile and IconFileLarge in Tabs table.
(I just added 4 the lines starting from "-- DNN-6758")


cnurse:
Manually merged to 7.4.2.SqlDataProvider

PR #405, opened 2015-03-23
DNN-6308: only reverse the save order for database folder type.

zyhfish:


cathalconnolly:
hmm, I noticed the comment needs updated after I merged this - I will fix that up when I next manually merge an issue (no point reopening an item for comment text)

PR #410, opened 2015-03-24
DNN-6772 - Added a default False value to the SendToast property on Work...

agrueso:
...flow Notifications


cathalconnolly:
"development" is now 7.4.2 - I have reassigned this to that milestone. If you need this in 7.4.1 please email Joe so he approve me cherry-picking this over.

PR #412, opened 2015-03-25
DNN-6777

robsiera:
My proposal to fix issue DNN-6777


cnurse:
Assigned to Cathal - this has Security implications (see discussion in JIRA)

cathalconnolly:
copying my comment from JIRA
"i'm afraid I'm not in favour of this change - having the ability to move a page in the hierarchy is entirely different from editing the page itself e.g. one example would be a site has "personal" pages under a "users" menu, the hierarchy clearly provides some "meta" information that the page is a personal one. Allowing the user to move the page to a different level breaks this admin decided taxonomy 

Typically I would recommend this be handled via an additional delegated permission (e.g. perhaps an extension to the permissions provider to allow for a new "pages editor" permission), or else via a custom module.

Assigning to Joe for his review"

cpaterra:
Closing for reasons noted by Cathal.

PR #413, opened 2015-03-26
DNN-6742: return proper settings for 'en-US'

sleupold:
DNN-6742: Return proper PortalSettings for cultureCode 'en-US'


jbrinkman:
Manually merged

sleupold:
thanks!

PR #415, opened 2015-03-26
DNN-6509 investigate and fix instances of hardcoded en-us/localzation.sy...

evidnn:
1. Here usage of SystemLocale is fine as we define new language in case if it doesn't exist.
ParseEnabledLocales
// if language does not exist in the installation, create it
locale = new Locale { Code = cultureCode, Fallback = Localization.SystemLocale, Text = CultureInfo.GetCultureInfo(cultureCode).NativeName };

2. Here usage of SystemLocale is fine as we define new language in case if it doesn't exist.
ParseTemplateInternal
defaultLocale = new Locale { Code = portalInfo.DefaultLanguage, Fallback = Localization.SystemLocale, Text = CultureInfo.GetCultureInfo(portalInfo.DefaultLanguage).NativeName };

3.GetPortal
if (string.IsNullOrEmpty(cultureCode)) cultureCode = GetPortalDefaultLanguage(portalId); //Localization.SystemLocale;
fallbackLanguage = LocaleController.Instance.GetDefaultLocale(portalId).Fallback; //Localization.SystemLocale);

4.GetActivePortalLanguage
This is avalid use case:
            // get Language
            string Language = Localization.SystemLocale;

5. LoadPortal
This is a valid use case
portalSettings.DefaultLanguage = Null.IsNull(portal.DefaultLanguage) ? Localization.SystemLocale : portal.DefaultLanguage;

6. Dashboard module uses following datareader:
        public static IDataReader GetPortals()
        {
            string cultureCode = Localization.SystemLocale;
            return provider.GetPortals(cultureCode);
        }

This will return portals with default language culture only  - this needs further investigation (new JIRA issue).

7. ProcessLegacyLanguages
                        //Save Language
                        Localization.Localization.SaveLanguage(language);
                        if (language.Code != Localization.Localization.SystemLocale)
This is a valid case

8. LanguagePackWriter
            _Language.Fallback = Localization.Localization.SystemLocale;
This is a valid use case

9. BestCultureCodeBasedOnBrowserLanguages
            return BestCultureCodeBasedOnBrowserLanguages(cultureCodes, Localization.SystemLocale);
This is a valid use case as second param is a fallback language value

10. GetCurrentLocale
        public Locale GetCurrentLocale(int PortalId)
        {
            Locale locale = null;

            if (HttpContext.Current != null && !string.IsNullOrEmpty(HttpContext.Current.Request.QueryString["language"]))
            {
                locale = GetLocale(HttpContext.Current.Request.QueryString["language"]);
            }
            return locale ?? ((PortalId == Null.NullInteger)
                                ? GetLocale(Localization.SystemLocale)
                                : GetDefaultLocale(PortalId));
        }
This is a valid use case as we'll use hardcoded value only of PortalId=-1 and there is no way to get active portal language

11.  GetPageLocale
           // 6. system default
            if (pageCulture == null)
                pageCulture = new CultureInfo(SystemLocale);
This is a valid use case 

12. TryGetStringInternal
string fallbackLanguage = Localization.SystemLocale;
This is a valid use case as value is redefined within the method.

13. GetInstallConfig
installConfig.InstallCulture = XmlUtils.GetNodeValue(rootNode.CreateNavigator(), "installCulture") ?? Localization.Localization.SystemLocale;
Valid use case

14.         internal static PortalController.PortalTemplateInfo FindBestTemplate(string templateFileName)
            if (String.IsNullOrEmpty(currentCulture))
            {
                currentCulture = Localization.Localization.SystemLocale;
            }
This is a valid use case

15. GetNavigationNodes
var hostTabs = TabController.GetTabsBySortOrder(Null.NullInteger, Localization.SystemLocale, true);
Even though here passed System.Locale last param (includeNeutral) is set to true so that way host pages are returned anyways 

16. GetLinkUrl
string cultureCode = Localization.SystemLocale;
Valid use case as value is redefined later on in the method

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -2329,7 +2329,7 @@ namespace DotNetNuke.Entities.Portals
                     //Get Fallback language
                     string fallbackLanguage = string.Empty;
 
-                    if (string.IsNullOrEmpty(cultureCode)) cultureCode = GetPortalDefaultLanguage(portalId) ?? Localization.SystemLocale;
+                    if (string.IsNullOrEmpty(cultureCode)) cultureCode = GetPortalDefaultLanguage(portalId);
 
                     Locale userLocale = LocaleController.Instance.GetLocale(cultureCode);
                     if (userLocale != null && !string.IsNullOrEmpty(userLocale.Fallback))
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -1035,8 +1035,7 @@ namespace DotNetNuke.Entities.Portals
             try
             {
                 var setting = Null.NullString;
-                //DNN-6509 - using empty string will ensure active portal language is used
-                GetPortalSettingsDictionary(portalId, Null.NullString).TryGetValue("EnableBrowserLanguage", out setting);
+                GetPortalSettingsDictionary(portalId, Localization.SystemLocale).TryGetValue("EnableBrowserLanguage", out setting);
                 if (string.IsNullOrEmpty(setting))
                 {
                     retValue = Host.Host.EnableBrowserLanguage;
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -2329,7 +2329,7 @@ namespace DotNetNuke.Entities.Portals
                     //Get Fallback language
                     string fallbackLanguage = string.Empty;
 
-                    if (string.IsNullOrEmpty(cultureCode)) cultureCode = GetPortalDefaultLanguage(portalId); //Localization.SystemLocale;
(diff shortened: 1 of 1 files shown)


evidnn:
So this is the sequence in which fallbacklanguage is defined in  
public PortalInfo GetPortal(int portalId, string cultureCode)

First we attempt to get portal using supplied cultureCode.
If there more than one active language and we couldn't get portal using supplied culture,
then we get portal default language:
if (string.IsNullOrEmpty(cultureCode)) cultureCode = GetPortalDefaultLanguage(portalId);

and get locale using new value and get fallback language: 
Locale userLocale = LocaleController.Instance.GetLocale(cultureCode);
                    if (userLocale != null && !string.IsNullOrEmpty(userLocale.Fallback))
                    {
                        fallbackLanguage = userLocale.Fallback;
                    }

If we are unable to retrieve fallback language from userLocale:
we use Localization.SystemLocale:
                    if (String.IsNullOrEmpty(fallbackLanguage))
                    {
                        fallbackLanguage = Localization.SystemLocale; 
                    }

PR #419, opened 2015-03-30
DNN-6814 Open the Manage Profile tab instead of the Communications tab w...

pberna:
Please review the changes I made.


cathalconnolly:
merging in to development and then cherry-picking over to release/7.5.0

PR #422, opened 2015-03-30
DNN-5373 SI: Email "From:" formatting

evidnn:
added defensive code in case when PortalSettings or GetUserByEmail return null


zyhfish:
This should still have issues in scheduler or for host users as the portal id is wrong and can't get correct display name.

evidnn:
I've tested it with host user - it displays correctly...if user can't be defined, then portal name would be used. Ben could you describe test scenario in details so that I can test it?

cathalconnolly:
reverting this due to https://dnntracker.atlassian.net/browse/DNN-6823 being logged (as per QA's request)

PR #426, opened 2015-03-31
DNN-6818 - Added logic to check is DisplaySearch setting has a valid boo...

agrueso:
...lean value or not


cathalconnolly:
merging into 7.4.2 and cherrypicking into release/7.4.1

PR #431, opened 2015-04-01
Support virtual relative path "~" symbol in URL link type for tabs.

DotNetNuclear:
Very often, administrators use the External URL Link type when creating internal links where they need to specify querystring parameters to an internal page or need to link to a child portal alias.  Supporting the virtual symbol "~" in the URL means that the full path and "http/https" designation does not need to be hardcoded in the URL and can therefore make it much easier to manage links between stage and production environments where the portal alias and SSL strategy is different.

### DIFF
--- a/DNN Platform/Library/Entities/Tabs/TabInfo.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabInfo.cs
@@ -543,8 +543,8 @@ namespace DotNetNuke.Entities.Tabs
         {
             get
             {
-                var key = string.Format("{0}_{1}", TestableGlobals.Instance.AddHTTP(PortalSettings.Current.PortalAlias.HTTPAlias),
-                                            Thread.CurrentThread.CurrentCulture);
+                string urlAlias = TestableGlobals.Instance.AddHTTP(PortalSettings.Current.PortalAlias.HTTPAlias);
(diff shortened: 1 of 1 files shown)


bdukes:
You're going to need to create a JIRA issue for this before we can accept the pull request.

I like this idea (we've run into the same thing).  The implementation seems pretty simple, though.  I've got to believe that there's somewhere else in the framework that's looking at the `Url` property of a tab.  Also, what about consumers outside of the framework?

Perhaps it would be better to add an "internal URL" link type, in addition to the external URL link type, and really treat this scenario as a fully supported option.  While adding `~` support helps you, no one else is ever going to know that it's an option, unless we change the UI to indicate that it's supported.

jbrinkman:
This can't be used as coded.  There are many references to Tab.Url in the framework - including in the URL rewriter.  At a minimum, those references would need to each be evaluated to determine if they should use the URL property or the FullURL property (in some cases it is just an IsNullOrEmpty check which is ok).

I think the better way to handle this is to create a new TabType.RelativeUrl.  This would allow everyone to detect this new format and then handle it appropriately.

cpaterra:
For reasons noted by Joe, closing this out. If interested, please resubmit with the necessary changes (and please start w/ 8 branch).

PR #433, opened 2015-04-01
DNN-6829 + CONTENT-2544 - Force folders deletion when they should be emp...

agrueso:
...ty


cathalconnolly:
merging into 7.4.2 and cherrypicking into release/7.4.1

PR #440, opened 2015-04-03
DNN-6673: fixed issue

BarryWaluszko:
Fixed issue where some of WebAPI requests was not authenticated, when IIS was working in "Windows Authentication" mode.
More info: https://dnntracker.atlassian.net/browse/DNN-6673

### DIFF
--- a/DNN Platform/HttpModules/Membership/MembershipModule.cs
+++ b/DNN Platform/HttpModules/Membership/MembershipModule.cs
@@ -144,7 +144,35 @@ namespace DotNetNuke.HttpModules.Membership
                 }
             }
 
-            if (request.IsAuthenticated && !isActiveDirectoryAuthHeaderPresent && portalSettings != null)
+            //DNN-6673 START
+            //check if it's Windows Authentication request, and try to authenticate it
+            if (request.IsAuthenticated
+                && context.User != null
+                && portalSettings != null
+                && (context.User is WindowsPrincipal || isActiveDirectoryAuthHeaderPresent))
+            {
+                string userName = string.Empty;
+                //get WinAuth username from context 
+                if (context != null && context.User != null && context.User.Identity != null)
+                {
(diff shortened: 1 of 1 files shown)


cathalconnolly:
is this not the same pull request as #335 ?

BarryWaluszko:
Yes, there is only one difference, how the "context.Items[UserInfo]" is stored. 
Sorry for the mess, I have not fully "git workflow" knowledge.

cathalconnolly:
no problem - I take it I can close #335 then? As for git workflow, the best way is typically to create a new branch for each change, that way if you have to alter/add to a fix you can update that branch

BarryWaluszko:
Please close the #335. 
I was confused because I didn't see my changes (DNN-6673) in any of the dnnsoftware/Dnn.Platform repo.

cathalconnolly:
thanks, i'll close #335 - typically we (the community committers and relevant DNN Corp employees) try to process pull requests quickly, but when they need some supporting environment setup to test (such as windows authentication) they can take longer

BarryWaluszko:
Hi Cathal,
If I want to add/change code in DNN-6673, what should I do? Branch DNN-6673 doesn't exist or I can't find it in "git upstream". Should I create new issue in DNN JIRA?

jbrinkman:
@BarryWaluszko It depends: Is this to fix some additional problem for WebAPI auth or is it truly the same bug?  Often people lump multiple related items under one bug which makes it hard to close the bug.  

If this is a behavior that can be tested and validated independent of the original bug then create a new Jira and a new pull request (PR).  In this case DNN-6673 might fail testing but the new bug could pass.  

If this is just a case where you found that the original code did not fully fix the original bug as described in DNN-6673, then just submit a new PR and reference the existing bug. In this case DNN-6673 would have likely failed testing which would have required a fix anyway and you are just providing the fix before the tests have been run.

BarryWaluszko:
Thanks @jbrinkman , your advice was very helpful. The original code from DNN-6673 didn't fully fix the issue, I submit new PR.

PR #442, opened 2015-04-07
DNN-3000: add enable/disable cache expiration api.

zyhfish:


cnurse:
This was reverted as the code change should have been made in the 8.0 branch.

I have manually done the correct merge.

PR #446, opened 2015-04-07
DNN-6845: Multiple uses of non-standard target="_new"

roman-yagodin:
This will replace all occurrences of target="_new" with target="new" to get better standards compatibility.

### DIFF
--- a/DNN Platform/Library/Services/Vendors/BannerController.cs
+++ b/DNN Platform/Library/Services/Vendors/BannerController.cs
@@ -1,6 +1,6 @@
 #region Copyright
 // 
-// DotNetNuke® - http://www.dotnetnuke.com
+// DotNetNukeÂ® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2014
 // by DotNetNuke Corporation
 // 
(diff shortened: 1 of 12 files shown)


cathalconnolly:
do you have a W3C reference for this change (e.g. http://www.w3.org/TR/WCAG20-TECHS/H83.html) - we need to verify that it is valid, and then to check that it works in all our supported browsers (IE8+, FF, Chrome)

vmasanas:
I think it would actually make more sense to change to _blank. See some of the discussions in this post http://stackoverflow.com/questions/4964130/target-blank-vs-target-new

roman-yagodin:
http://www.w3.org/html/wg/drafts/html/master/single-page.html#browsing-context-names describes browsing context names. According to it, "new" is valid browsing context name, though it's not a keyword (so, don't have any special meaning defined by standard).

AFAIK, target="_new" is also browser-specific. I think that most occurrences of "_new" could be safely replaced with "_blank" (e.g. in banners), except some really special cases.

cathalconnolly:
thanks - however based on Vicenc's link I think perhaps using _blank would be a better approach as it's valid and we know that it works in all supported browsers ("new" would be fine unless there was a window already called "new")

roman-yagodin:
Well, "_new" would be also fine unless there was a window already called "_new" :) 
I think I should close this pull request and create new one, thanks.

cathalconnolly:
FYI in future I recommend that you create a local branch for your repository and make the changes there. Then if there is feedback that requires change you can simply change your local branch.

roman-yagodin:
Sorry, I'll do better next time. I just get lazy to clone entire DNN.Platform repo locally and pushed separate commits via GitHub website. Luckily, there are `git clone -b <branch-name> --single-branch` thing for lazy peoples like me.

PR #454, opened 2015-04-08
DNN-6719 Create HMAC Authentication Attribute

cathalconnolly:
add a HMAC Authentication Attribute

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/HMACAuthenticationAttribute.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/HMACAuthenticationAttribute.cs
@@ -1,0 +1,195 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Security.Cryptography;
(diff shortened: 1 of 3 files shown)


DNN Platform/DotNetNuke.Web/Api/HMACAuthenticationAttribute.cs
```diff
@@ -0,0 +1,195 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web;
+using System.Web.Http;
+using System.Web.Http.Filters;
+using System.Web.Http.Results;
+
+namespace DotNetNuke.Web.Api
+{
+    public class HMACAuthenticationAttribute : Attribute, IAuthenticationFilter
```
bdukes: I'd prefer `HmacAuthenticationAttribute` (see where https://msdn.microsoft.com/en-us/library/ms229043 mentions acronyms)

cathalconnolly:
@bdukes 
I'll make that update when I return to the class (still a piece of work to do to integrate dnn's cache with the cached nonce values) - I checked this in a little early to make it easier to build my test library

PR #463, opened 2015-04-11
update captcha control

ohine:


sleupold:
great enhancement.

cathalconnolly:
is there a jira for this?

ohine:
I didn't create one, not sure if one exists in the backlog.

cathalconnolly:
cheers -we need to know what the problem is that's being fixed, so we can evaluate if the proposed change is appropriate, and so that QA can test it. Whilst I can look at the code and make a reasonable guess, it's better to get it from you - in this case it looks like you've changed from an encrypted value stored in a cookie to a querystring parameter , which whilst it adds some other permutations, I don't believe is adding additional security - hence I would like to see an explanation. 

thanks

ohine:
I'm not sure how the bots are cracking the system. I figured if they were somehow decrypting the querystring value, it wouldn't hurt to add some variation into that function since it's been the same since the catpcha control was introduced. 

I did bring up a few points in the MVP forum post on this subject. Maybe it would be better to continue the discussion over there. Since it's somewhat security related, I'm not sure if this is the best place for that conversation.

cathalconnolly:
sure - I suspect it's just a simple OCR crack of the image, but we can discuss elsewhere

ohine:
Ran into a few issues while testing, closing this pull request for now and will revisit for 7.4.2

cathalconnolly:
thanks for the update

PR #464, opened 2015-04-11
DNN-6897 add option to encrypt ids and a honeypot trap in the registration form

ohine:


sleupold:
great job!

RichardHowells:
Hi @ohine. Have you been able to test this site on any kind of a honeypot?  Ie a site known to get spam registrations?  I'm somewhat concerned that we may be attempting to block a route the spam registration scripts don't actually use.

ohine:
I have not, just thinking outside the box to attempt to make it more difficult to programmatically submit registration forms. 

I'd be happy to provide a compiled version if you'd like to run this on any site experiencing a registration issue. Most of my sites are not public, or if they are have registration disabled, which makes real world testing a bit more difficult. 

Plus, I'm pretty sure if you were to apply either of these into a site today, it would stop the bots. It won't be till it's in an actual release they notice a change and attempt to exploit it again. At least, I would think that's the case.

RichardHowells:
I do have a site (well several) that get registration issues.  I'd be happy to give it a go on one of them.  

BUT.  Of necessity this has to go on a live public facing site and I need an open upgrade path for the future.  It's currently running on 7.4.0 so I'd like this change merged into a 7.4.0 version.  I don't want to risk taking anything that has, or relies on, database changes.

> It won't be till it's in an actual release they notice a change and attempt to exploit it again.

I'm not sure they have much in the way of feedback loops.  I turned registration off and deleted all the dead accounts.  They keep trying to register more accounts and they keep coming back to (fail to) log into deleted ones.

ohine:
That's understandable. I'd be happy to backport these suggested fixes into 740. It's very minor and doesn't require any db changes. This specific pull request also has a switch so you could turn it on/off if you'd like.

RichardHowells:
Ok - go for it.  You already have my email address if you need it.  We have corresponded before.  If you need a reminder of it ask.

RichardHowells:
There was a mention in the DNN forums of implementing rate limiting.  So far my spammers have been very restrained registering six spam accounts in two hours.

It may be that this is going to ramp up or it may be a planned approach designed to evade defenses.  It may also be that different groups would have different policies.  (I am assuming that there is more than one spam group operating this approach.)

ohine:
I've updated the ID encryption slightly, and added a honeypot trap. I've done a bit of local testing to make sure it doesn't break a real user attempting to register. I'll try and get the v740 build generated and posted tomorrow for @RichardHowells to test in production. If anyone else would like a backported build, let me know the version and I'll do my best.

RichardHowells:
@ohine backported this to 7.4.0 for me.  It's run on a live site (www.wva.org.uk) since 14:20(UTC) yesterday - almost 18 hours.  In that period there have been no spam registrations.  Historically this site would see one or two spam registrations per hour.

I think it's a good change.

jbrinkman:
Oliver,
   Is there a Jira associated with this issue?

ohine:
Not yet, I was waiting for feedback on how effective it was before taking the time to create the jira item.

jbrinkman:
If there is any hope of getting this in 7.4.1 then all the bits have to be in place and that includes having a Jira which clearly identifies what the specific issue is that is being addressed and preferably how a QA could test the fix actually works.

ohine:
I can get that created, but I fear it will also be directions the spammers can use to quickly bypass this fix. It's your call. I'll create the jira, but being a security issue I'm hesitant to be descriptive with what was changed and how to verify it's working.

jbrinkman:
OK... Create a simple Jira and email me the QA steps and I'll add it as a private comment for our QA people

ohine:
Done. https://dnntracker.atlassian.net/browse/DNN-6897

jbrinkman:
This was reverted since the PR was submitted against the wrong branch.  PRs should NEVER be submitted against release branches.  I will work to get it moved over to the development branch.

ohine:
I can resubmit it against dev if that helps.

jbrinkman:
I have moved it and checked it into Dev... it is currently in 7.4.2 and we'll see if we can get it into 7.4.1.  Realistically though, the improvement that @RichardHowells sees is a temporary improvement.  This will have little to no impact long-term to the spam problem as it is easily circumvented by SPAMBOTs who will update their scripts once this update becomes widespread.

mitchelsellers:
@jbrinkman Although possibly a temporary fix, the spam registration issue is plaguing the platform.  To the tune of 4.75 MILLION spam users that I have had to personally remove from DNN installs since January 1st, of this year.  I have tested this pull request as well on an actively exploited site, and spam registrations have stopped entirely.  The solution here isn't necessarily "foolproof" but the impact of benefit, even if temporary is huge for the platform

nvisionative:
Very nice!

jbrinkman:
@mitchelsellers I have not discounted the short-term benefits, but it is an aspirin when what we really need is a cure.

mitchelsellers:
@jbrinkman I'm all for a cure....but we are almost a year into this being a horrendous issue and this is the closest thing to a cure that has been proposed.  Additionally we are only speculating on the true nature of the short term benefits, it could be longer than that.  Simple "What is 3 + 3" captcha's have worked for other platforms.  This solution is more complex than that.  My only comment here is that the longer that this fix is sitting here, publicly displayed and not implemented the more time its out there in the wild to be analysed etc while users are still getting overrun with uncontrollable spam.

BobKruger:
Team,

This is an interesting dialog. A long-term fix is needed. That said, we are proceeding with the existing fix in 7.4.2 and will try to get it into 7.4.1. Joe has already processed the pull request.

Just trying to save everyone’s fingers from writing more mail.  :-)

Thanx.

-Bob

BOB KRUGER
SVP Engineering
DNN Corp.
<http://www.dnnsoftware.com>www.dnnsoftware.com<http://www.dnnsoftware.com>

m +1 408-421-8153
o  +1 650-288-3687

WATCH 3 MINUTE VIDEO --><http://www.dnnsoftware.com/products/product-demos>
Publish / Measure / Optimize / Repeat



From: Mitchel Sellers
Reply-To: "dnnsoftware/Dnn.Platform"
Date: Thursday, April 23, 2015 at 10:38 AM
To: "dnnsoftware/Dnn.Platform"
Subject: Re: [Dnn.Platform] DNN-6897 add option to encrypt ids and a honeypot trap in the registration form (#464)


@jbrinkman<https://github.com/jbrinkman> I'm all for a cure....but we are almost a year into this being a horrendous issue and this is the closest thing to a cure that has been proposed. Additionally we are only speculating on the true nature of the short term benefits, it could be longer than that. Simple "What is 3 + 3" captcha's have worked for other platforms. This solution is more complex than that. My only comment here is that the longer that this fix is sitting here, publicly displayed and not implemented the more time its out there in the wild to be analysed etc while users are still getting overrun with uncontrollable spam.

nvisionative:
@mitchelsellers - I couldn't agree with you more.  

@jbrinkman - as an integrator that has been dealing with this on a daily basis for over a year, we have had to field more and more inquiries from our clients as to why they wouldn't switch platforms to something "more secure".  Please consider rolling this in ASAP to alleviate some of the pain (even if only temporary).  We have had to jump through way too many hoops and this has cost our clients way too much.  AND they are loosing faith in the platform.  We can only mitigate this so much, you know?

jbrinkman:
@nvisionative There are much better solutions already available in the Store.  InteractiveWebs registration does a great job of dealing with the problem using Google ReCaptcha.  The particular change was already merged earlier today into 7.4.2 and we are evaluating adding it to 7.4.1.

nvisionative:
@jbrinkman - thank you for merging - much appreciated.  I'll be glad to take this conversation offline, but in short, we are all too aware of the various 3rd-party solutions, configuration options, and even "hack" solutions to the problem (and have been implementing them all along).  That is my point though, we have been dealing with the "solution" outside of the platform rather than addressing it at a platform level.  This along with all the "clean up" effort associated with this issue (when not thwarted in advance) has cost us and our clients way too much!  It is important to understand this has sent a negative message about the platform.  We have always been able to say something along the lines of "DNN is very proactive in dealing with security related issues and that sets us apart from many other platforms".  However, our tune has honestly changed over the past year because we can no longer say this in good conscience.  I could go on and on about this, but will shut up now.  Great to hear this is still "on the radar" for addressing. 

David

ohine:
@nvisionative if you'd like a backported build, let me know. I already generated them for v740 and v703. I'm working on getting a v734 build going tonight.

nvisionative:
@ohine - thanks!

moorecreative:
I’ll be interested to see if it’s a self-contained honeypot or if it’s using an external service. I’ve always loved the idea of a honeypot and tarpit scenario… wouldn’t that change the face of spambots if it was just a standard part of all our security deployments.

 

Ryan Moore

 <mailto:ryan@moorecreative.com> ryan@moorecreative.com

704.896.6057

RichardHowells:
@moorecreative It's open source. Scroll up; click on commits; look at the code.  See whatever you want to see.

PR #468, opened 2015-04-13
DNN-6856:  Prevent Duplicate Entries

sleupold:
prevent duplicate entries in CoreMessaging_SubscriptionTypes


sleupold:
according to comments from DNN corp employees, this solution is too complex for them

sleupold:
meanwhile there will be a solution in TurboDNN (https://dnnscript.codeplex.com)

PR #469, opened 2015-04-14
DNN-4352: identify existing rows by PackageDependencyID

sleupold:
existing rows should be identified by passed ID, otherwise wie may simply skip passing the ID at all.


cathalconnolly:
manually merged into 7.4.2 (removed the set ansi_null's lines from 7.4.1 also)

cathalconnolly:
This change appears to work fine in platform but causes a major problem in evoq - I can prove this easily (took the 152 build of evoq saw the issue, then edited the 07.04.02.sqldataprovider file changes from this request and installed a different copy and everything worked fine). As such it seems that this change results is data changes that impact some scenarios. At this point I don't have the time to debug, so I'm going to roll it back.

PR #470, opened 2015-04-14
DNN-6868: Fix action button visibility in DAM module

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-6868


sleupold:
nice one :)

PR #471, opened 2015-04-15
Bug/dnn 4122

evidnn:
Sitemap is generated for the current language (based on the culture returned by GetPageLocale)
Sitemap would have entries if language is published


cathalconnolly:
FYI this pull request seems to have 2 commits in it (the dnn-6840 one is already processed), you may need to take care when creating branches for PR's

PR #472, opened 2015-04-15
DNN-6865 changing internal username forces users to login with email bet...

jcgomeztg:
DNN-6865 theres no need to update internal username, it causes user to login with email in all portals where is registered.


cnurse:
See my comment in the original JIRA issue - I have a concern that this is not the correct solution and might introduce more problems than it solves.   I am not prepared to go ahead without more testing to validate the fix.

cnurse:
Joe - can you make a decision on this one.

cpaterra:
The code changes are made in Login.ascx.cs. Not in UserController. It seems that we are handling the problem at the wrong place and providing a band-aid solution. We need to think about a better solution

PR #473, opened 2015-04-15
DNN-6856: change to prevent duplicate entries in 'CoreMessaging_Subscrip...

galatrash-at-dnn:
...tionTypes' table.


cathalconnolly:
merged directly into release/7.4.1 (so no cherrypick required)

PR #483, opened 2015-04-21
DNN-6554 PortalSettings not getting parsed correctly

evidnn:
Added defensive code in the `GetValueOrDefault` to catch exception when string value is invalid and can't be converted

### DIFF
--- a/DNN Platform/Library/Collections/CollectionExtensions.cs
+++ b/DNN Platform/Library/Collections/CollectionExtensions.cs
@@ -755,8 +755,16 @@ namespace DotNetNuke.Collections
         {
             Requires.NotNull("dictionary", dictionary);
             Requires.NotNull("converter", converter);
-
-            return dictionary.Contains(key) ? converter(dictionary[key]) : defaultValue;
+            var value = defaultValue;
+            try
+            {
+                value = dictionary.Contains(key) ? converter(dictionary[key]) : defaultValue;
+            }
+            catch (Exception ex)
+            {
+                // ignored
+            }
+            return value;
         }
 
         #endregion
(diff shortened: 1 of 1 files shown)


bdukes:
I'm really not comfortable with silently ignoring _any_ exception that occurs during converting.  If we're going to change the behavior of this method to ignore conversion errors (which can be argued for or against, but I'd be okay with), I think it should happen within the default converter, so that someone can specify a more strict converter if they want.  And, again, I would only want to catch conversion exceptions, not all.

jbrinkman:
We should catch the specific conversion error and not a generic exception.  The whole point of the GetValueOrDefault is that it should return default if a conversion fails, so don't "swallow" the error but actually return a default value. Also, the fix should resolve the root cause of the problem which is incorrect data getting stored in the system. Look at the location that Ben identifies and add code to ensure that if the value is deleted, that a valid integer is stored.

bdukes:
It may not be totally intuitive from the name, but it was a purposeful decision to let errors bubble from `GetValueOrDefault` (so that the default only comes through when a value isn't defined, but not for invalid data by default).  I agree that it's more friendly to also return the default for invalid data, but I think it should still support throwing.  And I definitely agree that the root cause needs to be fixed, in addition to how we read data that is invalid.

bdukes:
Also, I'm out of the country and won't be able to continue this conversation, so feel free to do whatever makes sense :smile:

evidnn:
Current fix don't simply swallow exception - it does return a default value in case of the exception, but doesn't log exception.
I would make changes to ensure exception is logged and check how invalid values are getting to the DB in the first place.

PR #489, opened 2015-04-25
DNN-6913: Removed unused paramter hideSync in function prepareForFilteredContent()

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-6913

### DIFF
--- a/DNN Platform/Modules/DigitalAssets/ClientScripts/dnn.DigitalAssets.js
+++ b/DNN Platform/Modules/DigitalAssets/ClientScripts/dnn.DigitalAssets.js
@@ -3263,16 +3263,12 @@ dnnModule.digitalAssets = function ($, $find, $telerik, dnnModal) {
         }
     }
     
-    function prepareForFilteredContent(hideSync) {
+    function prepareForFilteredContent() {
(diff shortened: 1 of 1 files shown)


cathalconnolly:
I've raised this with the original developer, as it may not be used in platform, but I suspect it is used elsewhere

cathalconnolly:
As I suspected this is used, so we cannot accept this change. I will close it and mark it as invalid. Here is the comment from the developer:
"Hi Cathal, you were right.

This method is also used in content, and it is used passing the “hideSync” parameter.
The method is invoked when Tags/Search Results are going to be shown or Subscribed items are shown.

The propose of that parameter is to show/hide the Sync buttons when a content is going to be listed. Obviously, that buttons make no sense when Search Results are shown or Subscribed items, therefore we have to hide them"

janjonas:
@cathalconnolly: Thanks for the information. IMHO, there should be some kind of documented "public" JavaScript APIs or "public" functions should be marked as such. Otherwise you cannot change any JavaScript code without potentially breaking some functionalities...

erikvb:
at the very least, there should be more documentation in code...

PR #491, opened 2015-04-27
Revert "DNN-3000: add enable/disable cache expiration api."

cnurse:
this needs to be reverted as the fix should have been done in the 8.0.0 branch


ohine:
Can this item be made public? I can't see any details on DNN-3000 in jira

PR #502, opened 2015-05-01
Bug/dnn 6135

evidnn:
I've tested sitemap on localized site without fix that is provided in DNN-4122, but couldn't replicate the exception.
I think issue could be related to the tab settings or modules on the localized pages.
To prevent failure of the SiteMap creation I've added try catch around the
GetPageUrl call:
                        try
                        {
                            pageUrl = GetPageUrl(tab, (ps.ContentLocalizationEnabled) ? tab.CultureCode : null);
                            urls.Add(pageUrl);
                        }
                        catch (Exception ex)
                        {
                            Logger.ErrorFormat("Error has occurred getting PageUrl for {0}", tab.TabName);
                        }
This way even if one page has failed process will continue and problematic tab would be reported in the log file.

### DIFF
--- a/DNN Platform/Library/Entities/Urls/TabIndexController.cs
+++ b/DNN Platform/Library/Entities/Urls/TabIndexController.cs
@@ -421,6 +421,16 @@ namespace DotNetNuke.Entities.Urls
                             false,
                             false);
             AddToTabDict(tabIndex,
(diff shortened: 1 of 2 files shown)


cnurse:
Merged PR manually

PR #505, opened 2015-05-03
DNN-6015: add missing view

sleupold:
[DNN 6015]:added missing view (even if not used for DNN platform atm, useful for all other use cases)


cnurse:
Manually merged

PR #516, opened 2015-05-12
DNN-6982 Error while exporting a DNN 7.4 bacpac because of an error o…

evidnn:
…n view vw_ContentWorkflowUsage

trim the white space from PortalID column name


sleupold:
this fix is not re-runnable (see PR #521 for improved version)

cnurse:
I will use Sebastians PR (#521)

PR #519, opened 2015-05-13
Added code which prevent adding context.Items["UserInfo"] equal to null.

BarryWaluszko:
Key "UserInfo" that exist but it's equal to null leads to exception: Item has already been added. Key in dictionary: 'UserInfo'  Key being added: 'UserInfo'


evidnn:
I've applied changes, proposed in this PR, with slight change:
https://github.com/dnnsoftware/Dnn.Platform/pull/535

This one can be closed.

cathalconnolly:
I've marked this as merged into 7.4.2 so Barry still get's credit for finding the issue and suggesting the solution

PR #521, opened 2015-05-14
DNN-6982: fix column Name

sleupold:
re-runnable Version of the fix provided by Eleonora


cathalconnolly:
thanks for the pull request, but the better approach is to add the suggested fix as a comment to #516 -that way Eleonora can update her branch and we ensure that good code is added. It also means that additional pull requests such as this don't need to be raised (and tested/merged)

sleupold:
thanks for the feedback, nice - and now??

cnurse:
Merged PR manually

PR #532, opened 2015-05-20
DNN-7014 Fix issue that did not allow a user to restore deleted pages…

pberna:
Please review my changes


francescorivola:
The fix is no longer needed as we fix the issue by cherry picking a missing commit from DNN-6734

PR #563, opened 2015-06-03
DNN-4734: Full URL as ReturnURL does not work anymore

stephen-lim:


Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -539,22 +334,29 @@ private void CreateUser()
...
+            {
 				User.UpdateDisplayName(PortalSettings.Registration.DisplayNameFormat);
-			}
-		}
+            }
+        }
 
 		private bool Validate()
 		{
-		    if (!string.IsNullOrEmpty(gotcha.Value))
```
bdukes: Why is this removed?

Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -518,14 +320,7 @@ private void CreateUser()
 
 					if ((string.IsNullOrEmpty(strMessage)))
 					{
-						Response.Redirect(GetRedirectUrl(), true);
-					}
-					else
```
bdukes: Why is the `else` removed?

Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -433,7 +236,6 @@ private void AddPasswordConfirmField(string dataField, string dataMember, bool r
 				Required = required,
 				TextMode = TextBoxMode.Password,
 				TextBoxCssClass = ConfirmPasswordTextBoxCssClass,
-				ClearContentInPasswordMode = true
```
bdukes: Why was this removed?

Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -726,45 +528,253 @@ private bool Validate()
...
+			}
+
+			//Verify that the current user has access to this page
+			if (PortalSettings.UserRegistration == (int)Globals.PortalRegistrationType.NoRegistration && Request.IsAuthenticated == false)
+			{
+				Response.Redirect(Globals.NavigateURL("Access Denied"), false);
+				Context.ApplicationInstance.CompleteRequest();
+			}
+
+			cancelButton.Click += cancelButton_Click;
```
bdukes: Why was this changed to a button that redirects instead of just a link?  The link that was used before seems more straightforward, is there some reason that the redirect needs to originate on the server?

Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -103,270 +104,72 @@ protected bool IsValid
...
-					AddField("LastName", String.Empty, true, String.Empty, TextBoxMode.SingleLine);
-				}
 
-				//Email
-				AddField("Email", String.Empty, true, PortalSettings.Registration.EmailValidator, TextBoxMode.SingleLine);
-
-				if (PortalSettings.Registration.RequireValidProfile)
-				{
-					foreach (ProfilePropertyDefinition property in User.Profile.ProfileProperties)
+					if (PortalSettings.Registration.RedirectAfterRegistration <= 0)
```
bdukes: this is inside of the `else` for `if (PortalSettings.Registration.RedirectAfterRegistration > 0)`, so it will always be `true`.  is there some particular case you're trying to address by adding this `if` check?

Website/admin/Security/SendPassword.ascx.cs
```diff
@@ -309,13 +309,8 @@ protected void OnSendPasswordClick(Object sender, EventArgs e)
                         {
                             canSend = false;
                         }
-                        else 
+                        else
                         {
-                            if (_user.Membership.Approved == false)
```
bdukes: Why was this removed?

stephen-lim:
Brian,

It looks like there was a conflict with some other code contributed by others that was added during the same moment when I made my pull request. I'm going to review the file changes here. They should be very straightforward a few line changes only, not as substantial as the lines changes shown in github.com

Website/admin/Security/SendPassword.ascx.cs
```diff
@@ -94,9 +95,8 @@ protected string RedirectURL
                         //return to the url passed to register
                         _RedirectURL = HttpUtility.UrlDecode(Request.QueryString["returnurl"]);
                         //redirect url should never contain a protocol ( if it does, it is likely a cross-site request forgery attempt )
-                        if (_RedirectURL.Contains("://") &&
-                            !_RedirectURL.StartsWith(Globals.AddHTTP(PortalSettings.PortalAlias.HTTPAlias),
-                                StringComparison.InvariantCultureIgnoreCase))
+                        if (_RedirectURL.Contains("://") &&
+							!Regex.Replace(_RedirectURL, @"^https*://", "", RegexOptions.IgnoreCase).StartsWith(PortalSettings.PortalAlias.HTTPAlias, StringComparison.InvariantCultureIgnoreCase))
```
bdukes: the regex should be `^https?://` (so that only zero or one `s` is accepted), otherwise this change (in all three files) looks good.  I'd probably separate it into two `if` statements, so that it's clearer what's going on:

    if (_RedirectURL.Contains("://")) {
        var redirectWithoutProtocol = Regex.Replace(_RedirectURL, @"^https?://", string.Empty, RegexOptions.IgnoreCase);
        if (redirectWIthoutProtocol.StartsWith(PortalSettings.PortalAlias.HTTPAlias, StringComparison.InvariantIgnoreCase)) {
            _RedirectURL = "";
        }
    }

stephen-lim:
Please close this pull request. I'm going to have to resubmit the PR. It should have been just a few regex line changes.

PR #565, opened 2015-06-04
DNN-7020

hkenuam:
DNN-7020 - The site settings control now provides an option to include attachments in outgoing messages. The default is off.


cnurse:
I have manually merged this

PR #566, opened 2015-06-04
DNN-7064 - Reset password form should allow hitting enter key to submit

ludikduran:
Added an ASP:Panel element. Set the DefaultButton value to cmdChangePassword.

Tested and enter key will submit form.

### DIFF
--- a/Website/admin/Security/PasswordReset.ascx
+++ b/Website/admin/Security/PasswordReset.ascx
@@ -6,27 +6,31 @@
         <span><asp:Label ID="lblInfo" runat="Server" /></span>
         <span class="error"><br /><br /><asp:Label ID="lblHelp" runat="Server" /></span>
     </div>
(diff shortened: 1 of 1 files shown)


bdukes:
Perhaps this has been fixed since I last used a `LinkButton`, but my understanding was that `DefaultButton` only works for `Button`s, not `LinkButton`s.  [MSDN still says](https://msdn.microsoft.com/en-us/library/system.web.ui.webcontrols.panel.defaultbutton.aspx) 
>The `DefaultButton` can be set to the identifier for a `Button` control or any control that implements the `IButtonControl` interface except a `LinkButton` control.

Also, can `divPassword` just be converted into a `Panel`, instead of adding an additional element?

cnurse:
Given that it "appears" to work - I will accept this change.  When I merge it I will review if we can remove the need for extra div element.

PR #570, opened 2015-06-09
DNN-6833: fixing unclosed db connection bug.

galatrash-at-dnn:


jbrinkman:
Please remember to mark the issue with the Merged label when you process a PR.  We use this label for reporting purposes.

PR #578, opened 2015-06-13
DNN-7097: Fixed stored procedure GetSchedule to return correct NextStart if scheduler should run on more than one server

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7097


cnurse:
Manually merged

PR #579, opened 2015-06-14
DNN-4496: SProc GetUsersAdvancedSearch missing the {databaseOwner}

EPTamminga:
placeholder several times calling a database.

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.04.02.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.04.02.SqlDataProvider
@@ -96,6 +96,223 @@ BEGIN
 	SELECT @@ROWCOUNT
 END
 GO
+
+/*** Fix bug DNN-4496: SProc GetUsersAdvancedSearch missing the {databaseOwner} placeholder several times calling a database. ***/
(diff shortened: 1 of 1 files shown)


cnurse:
This sproc is such a hairy one that I am reluctant to merge it for 7.4.2.  We can consider it for 8.0 so I will leave it open.

EPTamminga:
@cnurse It is, but the changes I have made should not be very harmful. It only involves adding the missing DBO to views and tables. Since 99,9% or more of the DNN installs do not use DBO, the chance that something is not working, is very very small.

cathalconnolly:
@EPTamminga from memory a similar change was logged before and found to not be necessary - when a stored procedure is executing within DBO scope any dynamic sql that is generated and ran runs within that scope meaning that dbo is not necessary - have you tested this to see if this is the case as I have a feeling this change is not required.

mitchelsellers:
@cnurse the change that @EPTamminga has here would really impact those very limited situations where someone set a custom database owner.  If left at the defaults you are correct the use of dbo. isn't really needed inside the context of a stored procedure.

cnurse:
Manually merged

PR #580, opened 2015-06-14
DNN-7098: Added parameter sendNotification to UserController.UpdateUser()

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7098 for more details.


cnurse:
Unfortunately we cannot accept this Pull Request as it breaks binary compatibility.   Any 3rd party developer that had used the event's code would fail.

We would be able accept the PR if submitted as an additional event - rather than modifying the existing one.

PR #583, opened 2015-06-15
DNN-7108: Fixed substring check in DotNetNuke.Services.Scheduling.DNNScheduler.CanRunOnThisServer

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7108

### DIFF
--- a/DNN Platform/Library/Services/Scheduling/DNNScheduler.cs
+++ b/DNN Platform/Library/Services/Scheduling/DNNScheduler.cs
@@ -54,19 +54,13 @@ namespace DotNetNuke.Services.Scheduling
 
         #region Private Methods
 
-        private bool CanRunOnThisServer(string servers)
+        private static bool CanRunOnThisServer(string servers)
         {
-            string lwrServers = "";
-            if (servers != null)
-            {
-                lwrServers = servers.ToLower();
-            }
-            if (String.IsNullOrEmpty(lwrServers) || lwrServers.Contains(Globals.ServerName.ToLower()))
-            {
-                return true;
-            }
-
-            return false;
+            return
+                // Scheduler should run on every server
+                string.IsNullOrWhiteSpace(servers)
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Scheduling/DNNScheduler.cs
```diff
@@ -54,19 +54,13 @@ public class DNNScheduler : SchedulingProvider
...
-            {
-                return true;
-            }
-
-            return false;
+            return
+                // Scheduler should run on every server
+                string.IsNullOrWhiteSpace(servers)
+                // Check if scheduler should run on this server
+                || servers.Split(',').Contains(Globals.ServerName, StringComparer.InvariantCultureIgnoreCase);
```
bdukes: In general, `InvariantCulture` and `InvariantCultureIgnoreCase` aren't really useful.  For non-culture-specific comparisons, `Ordinal` and `OrdinalIgnoreCase` should be used (see https://msdn.microsoft.com/en-us/library/dd465121.aspx).

bdukes: Also, do the servers need to be trimmed before comparing (e.g. would "Server1, Server2" be possible, with a space at the front of the second item)?  Perhaps it would be better to split on comma and space, and exclude empty entries.

janjonas: Hi @bdukes , thanks for your comment. I've made both changes (see https://github.com/janjonas/Dnn.Platform/commit/6328ac0088b5d9f0b9a78bf151d9214e3adf8aa2). I was not aware of OrdinalIgnoreCase but this definitively makes sense and add trimming is also a good idea to make the method a bit more "robust".

bdukes:
:+1: :shipit:

ohine:
:+1:

PR #584, opened 2015-06-15
DNN-7109: Hide "run now" button if scheduler should not run on this server

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7109

Please note: I also moved CanRunOnThisServer() from DNNScheduler to SchedulingController. Maybe there is a better place for this method.

### DIFF
--- a/DNN Platform/Library/Services/Scheduling/DNNScheduler.cs
+++ b/DNN Platform/Library/Services/Scheduling/DNNScheduler.cs
@@ -52,25 +52,6 @@ namespace DotNetNuke.Services.Scheduling
 
         #endregion
 
-        #region Private Methods
-
-        private bool CanRunOnThisServer(string servers)
-        {
-            string lwrServers = "";
-            if (servers != null)
-            {
-                lwrServers = servers.ToLower();
-            }
-            if (String.IsNullOrEmpty(lwrServers) || lwrServers.Contains(Globals.ServerName.ToLower()))
-            {
-                return true;
-            }
-
-            return false;
-        }
-
-        #endregion
-
         #region Public Methods
 
         public override int AddSchedule(ScheduleItem scheduleItem)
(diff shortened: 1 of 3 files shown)


bdukes:
I'm not totally familiar with how this part of the framework works, but would it be possible, instead, to start the schedule item on a server that it _can_ run on?  Does queuing the schedule task for immediate execution involve writing a database row, and, if so, could we write that row using a valid server name?

mitchelsellers:
Although I can understand how this is a good idea.  BUT, there are times where you might want to run a job on a non-standard server in a farm environment.  As such, I'm not sure that stopping the action is the best course of action.

mitchelsellers:
Scratch the prior comment - I just realized now that the current code lets the user click a button and nothing happens.  (All good here.)

janjonas:
@bdukes: I agree that a dialog to choose the server the scheduler should run on would be a much better solution but unfortunately there is no public API to run a scheduler on a specific server.

IMHO writing a record to the database would not be enough: If you take a look into the source code (https://github.com/dnnsoftware/Dnn.Platform/blob/development/Website/DesktopModules/Admin/Scheduler/EditSchedule.ascx.cs#L264) you find:

```
if (SchedulingProvider.SchedulerMode == SchedulerMode.TIMER_METHOD)
{
  SchedulingProvider.Instance().ReStart("Change made to schedule.");
}
```

So, if the scheduler mode is set to timer and the scheduler should run on another server there need to be a way to restart the scheduling provider on the other server.

ohine:
While the current scheduled task api doesn't support choosing which server a task is run next on, there are a few ways we could solve that.

The easiest way would be to save the current run on servers setting elsewhere (either another column in the schedule tasks table, or maybe a host setting). Then if you select run on server 2, we'd set the runat field to server 2 and the next start time to a few seconds from now. We would then modify the task complete step to check for the original run at setting and restore it when it marks the task complete. 

This could also be the start of an improved way to make sure tasks run on only a single server if we expanded this to all scheduled tasks. On Start, the service would write it's name into the Run at setting, on complete, it would clear it out and release that task again for all servers. 

Other ways would be to have one server call the target server and tell it to start up the task, but then we need to add a few more settings so separate instances can community and sync their scheduled tasks.

cnurse:
I am merging this - this change does what it says.  The other comments - while valid would change the scope of this change, and should be considered separately.

PR #592, opened 2015-06-18
DNN-6796 Cannot add existing module from hidden page as non-admin user.

SCullman:
Hidden Pages are not security relevant - they are just not included in the menu. There can be tons of reasons for that. This change is strongly requested from my power users.


SCullman:
Thanks

PR #595, opened 2015-06-22
DNN-7116 & DNN-7058 [REPLACED WITH SEPARATED PULL REQUESTS, TO BE DELETED]

carlinesrallysport:
Fixing LinkedIn oauth2 problem:
AccessToken var added to oauthclientbase.cs so every client.cs returns
their own access_token path.
LinkedIn's AccessToken path is oauth2_access_token (they didn't follow
the standard)

Added another var called OAuthHeaderCode to oauthclientbase.cs so every
client.cs returns their own authorization header if needed.

Fiware Needs Basic Authorization to work properly:
https://github.com/ging/fi-ware-idm/wiki/Using-the-FIWARE-LAB-instance

Files to compile and .zip Fiware provider are added, in case of DNN
implementation.

Around 1700 lines added due to the addition of 2 whole folders:
DotNetNuke.Authentication.LinkedIn and DotNetNuke.Authentication.Fiware

Anyway, if you are not interested on their addition there, it would be great to have the oauthclientbase.cs modified.

This will allow people to add their own authentication providers in the future with an extra couple of features in case they are needed.

The example is our Fiware provider for basic authorization and LinkedIn needing a custom access_token path name.

Thanks in advance.


cnurse:
You should separate these changes and submit 2 Pull Requests.  One for the Linked In fix and one for the FiWare addition.

jbrinkman:
This pull request was resubmitted as two different pull request.

PR #598, opened 2015-06-23
DNN-7116

carlinesrallysport:
Added Fiware OAuth2 authentication provider.
Added variable OAuthHeaderCode so people can include their own
authentication providers using basic or bearer authorization (My Fiware
case).

On any client.cs, we can fill OAuthHeaderCode with "Basic", this will
turn APIkey:APIsecret into base64 to send it as header as example below:

Authorization: Basic x54xffaftOHJFg3GGHr78=

Fiware standard can be found here:
https://github.com/ging/fi-ware-idm/wiki/Using-the-FIWARE-LAB-instance


galatrash-at-dnn:
@carlinesrallysport this PR is out of scope; the first few changes (the ones in existing files) are already implemented in the code. The rest are specific provider which can be included in a stand-alone where it can be installed when needed and shouldn't be part of the default installation.

Based on this, I'm going to close this issue without merging.

galatrash-at-dnn:
@carlinesrallysport please work with the coordinator of https://github.com/DNNCommunity to open-source this module there if you are interested.

PR #601, opened 2015-06-23
 DNN-7098: Added parameter sendNotification to UserController.UpdateUser()

janjonas:
This is a modified pull request of https://github.com/dnnsoftware/Dnn.Platform/pull/580; see https://dnntracker.atlassian.net/browse/DNN-7098 for more detail.

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserEventHandlers.cs
+++ b/DNN Platform/Library/Entities/Users/UserEventHandlers.cs
@@ -51,6 +51,7 @@ namespace DotNetNuke.Entities.Users
 
         public void UserApproved(object sender, UserEventArgs args)
         {
+            if (!args.SendNotification) return;
             Mail.SendMail(args.User, MessageType.UserRegistrationPublic, PortalSettings.Current);
             DeleteAllNotifications(args.User.UserID);
         }
--- a/DNN Platform/Library/Entities/Users/UserEventHandlers.cs
+++ b/DNN Platform/Library/Entities/Users/UserEventHandlers.cs
@@ -42,7 +42,7 @@ namespace DotNetNuke.Entities.Users
 
         public void UserRemoved(object sender, UserEventArgs args)
         {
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/Entities/Users/UserEventHandlers.cs
```diff
@@ -51,6 +51,7 @@ public void UserDeleted(object sender, UserEventArgs args)
 
         public void UserApproved(object sender, UserEventArgs args)
         {
+            if (!args.SendNotification) return;
             Mail.SendMail(args.User, MessageType.UserRegistrationPublic, PortalSettings.Current);
             DeleteAllNotifications(args.User.UserID);
```
bdukes: Does `DeleteAllNotifications` not need to happen if we aren't sending a new notification?

bdukes:
outside of my one comment. this looks good to me

DNN Platform/Library/Entities/Users/UserEventHandlers.cs
janjonas: There is a `return` statement, so `DeleteAllNotifications` won't get called if `args.SendNotification` is `false`.

janjonas:
Hi @bdukes, please see my comment below yours.

DNN Platform/Library/Entities/Users/UserEventHandlers.cs
bdukes: Right, that's my question.  It doesn't immediately seem to me like it should be skipped (but it's definitely possible that it's correct to skip it).

janjonas: You're right. I've added a new commit (https://github.com/janjonas/Dnn.Platform/commit/aa28f0ca6346fa239a63959cc0a295f69d8eb008) to this pull request. I also changed the name of the method to make more clear what the method does.

PR #603, opened 2015-06-24
DNN-7169 Created 2 more helpful error messages when trying to reset the password

donker:
More accurate reporting of failures when password reset doesn't work

### DIFF
--- a/Website/App_GlobalResources/SharedResources.resx
+++ b/Website/App_GlobalResources/SharedResources.resx
@@ -999,7 +999,13 @@
   <data name="PasswordResetFailed.Text" xml:space="preserve">
     <value>Your new password was not accepted for security reasons. Please make sure to choose a password which does not match a previously used password and which is long and complex enough to meet the sites password complexity requirements.</value>
   </data>
-  <data name="PasswordResetFailed_UserUndefined.Text" xml:space="preserve">
+ <data name="PasswordInadequate.Text" xml:space="preserve">
+    <value>Your new password was not accepted for security reasons. Please make sure to choose a password which is long and complex enough to meet the sites password complexity requirements.</value>
+  </data>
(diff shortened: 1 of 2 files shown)


Website/App_GlobalResources/SharedResources.resx
```diff
@@ -999,7 +999,13 @@
   <data name="PasswordResetFailed.Text" xml:space="preserve">
     <value>Your new password was not accepted for security reasons. Please make sure to choose a password which does not match a previously used password and which is long and complex enough to meet the sites password complexity requirements.</value>
   </data>
-  <data name="PasswordResetFailed_UserUndefined.Text" xml:space="preserve">
+ <data name="PasswordInadequate.Text" xml:space="preserve">
+    <value>Your new password was not accepted for security reasons. Please make sure to choose a password which is long and complex enough to meet the sites password complexity requirements.</value>
```
bdukes: can you please change "sites" to "site's"?

bdukes:
After fixing typo, looks good to me :+1:

cathalconnolly:
I've removed the milestone as assigning PR's to milestones is the PM's decision (i.e. @jbrinkman )

donker:
OK, I thought it was related to risk.

cathalconnolly:
it often is i.e. a large PR that touches on elements that run on each request (such as CDF) will be assigned to a "large" release such as 8.0.0 to ensure that we have a longer test cycle, or a change in a popular API (such as tabcontroller) may not be taken too close to codefreeze, but we also use it to organise what goes in each release i.e. Joe assigns items to a release, and Charles and I process them. In this case 7.4.2 is effectively codefrozen so when Joe reviews this I would expect it to be assigned to 8.0.0, unless he deems it low-risk (and important enough to come after the codefreeze) - we expect we may take an occasional additional PR but that's up to PM to decide

PR #604, opened 2015-06-25
DNN-5288: uninstall DK mobile skin during upgrade.

zyhfish:


mitchelsellers:
If I recall correctly, this will be a breaking change for anyone that is upgrading as part of this release that is using the DarkKnight skin.  I assume that this is an "acceptable" behavior, but want to confirm.

zyhfish:
this is DK Mobile skin/container, not DK skin.

jbrinkman:
We should not be uninstalling a skin on upgrade.  We should remove it from packaging, but if it exists in a prior installation then it stays in the installation.

PR #608, opened 2015-06-28
DNN-7176: Fix data type

sleupold:
Make sure, compared value is of correct data type, use UniCode where appropriate


zyhfish:
why there have CREATE, DROP, CREATE? maybe the first CREATE is no need.

sleupold:
agreed

cpaterra:
@zyhfish Can you please remove the initial create and merge this? Thanks.

sleupold:
should the function stay inside DNN Core or be moved to SiteLog module repository (which is still empty)?

PR #609, opened 2015-06-29
DNN-7135: add cache on get toasts api.

zyhfish:


jbrinkman:
@zyhfish We are not doing pull requests right now for Performance enhancements which are part of our testing efforts under the 'Steroids' epic.

PR #614, opened 2015-06-30
DNN-5288: only uninstall package when the skin/container folder not exist.

zyhfish:


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3041,6 +3041,17 @@ private static void UpgradeToVersion740()
...
 
+		private static void UpgradeToVersion742()
+		{
+			var containerFolder = string.Format("{0}Containers\\DarkKnightMobile", Globals.HostMapPath);
+			var skinFolder = string.Format("{0}Skins\\DarkKnightMobile", Globals.HostMapPath);
+			if (!Directory.Exists(skinFolder) && !Directory.Exists(containerFolder))
+			{
+				UninstallPackage("DarkKnightMobile"); //Container
+				UninstallPackage("DarkKnightMobile"); //Skin
+			}
```
jbrinkman: These two checks should be done separately and the corresponding package removed.

If (!Directory.Exists(skinfolder)) { UninstallPackage("...") }
If (!Directory.Exists(containerFolder)) { UninstallPackage("...") }

cnurse:
I will do the merge and apply Joe's commented change at the same time

cnurse:
The code cannot reliably be refactored as Joe suggests. This is because the two UnInstall calls are identical as they only pass the package name which is the same and don't use the PackageType property which is different.  This means we cannot just refactor and reliably uninstall the skin or container.

cnurse:
I mistakenly closed - but it probably should be resubmitted anyway as the code will be significantly different.

PR #615, opened 2015-06-30
DNN-4122: check content localization enabled setting, correct the cache file name.

zyhfish:


cnurse:
Cherrypicked into 7.4.2 release branch

PR #617, opened 2015-06-30
DNN-7193 Add Dynamic Content Creator Manager to Admin menu

cnurse:
Aded IUpgradable support to the Dynamic Content Creator Manager module so it is automatically added to the Admin menu on Install/Upgrade

### DIFF
--- a/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/Dnn.Modules.DynamicContentManager.csproj
+++ b/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/Dnn.Modules.DynamicContentManager.csproj
@@ -45,11 +45,19 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\DotNetNuke.Web\bin\DotNetNuke.dll</HintPath>
     </Reference>
+    <Reference Include="DotNetNuke.Instrumentation, Version=8.0.0.0, Culture=neutral, processorArchitecture=MSIL">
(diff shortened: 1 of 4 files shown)


DNN Platform/Modules/Dnn.Modules.DynamicContentManager/Dnn.Modules.DynamicContentManager.csproj
```diff
@@ -45,11 +45,19 @@
...
+    </Reference>
     <Reference Include="DotNetNuke.Web, Version=8.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\DotNetNuke.Web\bin\DotNetNuke.Web.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Moq, Version=4.2.1506.2515, Culture=neutral, PublicKeyToken=69f491c39445e920, processorArchitecture=MSIL">
+      <HintPath>..\..\..\packages\Moq.4.2.1506.2515\lib\net40\Moq.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
```
jbrinkman: Why do we need Moq in this module? Shouldn't that be contained within the test code?

cnurse: Yup - that was an inadvertent inclusion - should have Resharper to clean up references before checking in

PR #620, opened 2015-07-02
DNN-6630 Remove SiteLog from platform

jbrinkman:
This is the first part of the DNN-6630 change.  There is more likely more work to be done on DNN-6630 (create new forge project for sitelog) so please hold off on marking that complete.


DNN Platform/Library/Entities/Portals/PortalInfo.cs
```diff
@@ -827,7 +800,6 @@ public void Fill(IDataReader dr)
             PaymentProcessor = Null.SetNullString(dr["PaymentProcessor"]);
             ProcessorUserId = Null.SetNullString(dr["ProcessorUserId"]);
             ProcessorPassword = Null.SetNullString(dr["ProcessorPassword"]);
-            SiteLogHistory = Null.SetNullInteger(dr["SiteLogHistory"]);
```
francescorivola: Ideally, we should review the SProcs that returns SiteLogHistory field to remove that field

PR #621, opened 2015-07-02
Updates and improvements to the Provider.FiftyOneClientCapabilities.

nerru86:
Upd: 21/07/2015
Today we have released the stable version of the DLL: 3.2.3.2. In addition to the previous message:
- We have performed load testing and published the results: https://51degrees.com/blog/ArtMID/1641/ArticleID/368/Memory-and-Performance-Impact-of-51Degrees
- We have added multiple unit tests: https://github.com/51Degrees/.NET-Device-Detection/tree/master/UnitTests
- We have been running the updated version of the DLL on our own website (also powered by DNN) for quite some time now.

The design focus of the release is to integrate DNN efficiently with 51Degrees. Improvements include reducing memory consumption when the data file used directly from the disk, and to make use of 51Degrees performance improvements.

List of changes:
- When device detection is disabled the Client Capabilities Provider no longer forces it to start.
- All possible properties and values are no longer stored in a SortedList. Only required properties and values are loaded when needed and cached.
- Embedded data has been removed from the assembly and now must be provided from the App_Data folder.
- .NET 3.5 is no longer supported in order to use memory mapped files and simplify overriding default browser capabilities.
- In stream mode entity data properties that can allocate large arrays only initialise these arrays when needed.
- Caches used with stream operation are now fixed memory size and serviced via the thread pool.
- Automatic update processes uses temporary files rather than main memory to verify integrity of updated files prior to using them.
- Temporary files are now created in the App_Data/51Degrees folder of the web application rather than a UNC path or the master data file folder.
- Values associated with Profiles are now retrieved using a more efficient algorithm.
- DataSet.Properties collection now has a string accesser to make retrieving properties by name simpler.
- Memory mode no longer initialises the entire dataset. Startup time is now significantly improved.
- Memory mode uses less memory than prior versions.
- Supports 51Degrees V3.2 data format using approximately 25% smaller file sizes.


jbrinkman:
In performance/load testing, @mitchelsellers was seeing "File Access Exceptions" tied to it not being able to delete files in App_Data due to "Another process accessing the file."  Is this resolved by the change in how/where temp files are being created?

mitchelsellers:
@nerru86 As @jbrinkman mentioned under heavy log Log4Net is showing multiple locking issues on files in the /app_data folder tied to 51degrees.  The exceptions are related to "another process is using _____ file".  I don't have much more detail at this point on it though.

nerru86:
Hi,

Thank you for your feedback. These messages appear because of the way the detector is using the temporary files. For example there are two worker processes each using a separate temporary file, then a third worker process kicks in and attempts to do a bit of "housekeeping" by trying to delete the temporary data files. Since both temporary data files are in active use it fails to delete them and moves on to performing its job. This is useful for situations where a website has crashed and had to be restarted, it helps to remove the old unused temporary data files.

An alternative would be to use the original file, but that raises the problem of updating the master file as it will be locked by the worker processes. In addition the detector can run in medium trust.

One of the changes introduced in 3.1.12.6 that was carried on to 3.2 was the improved management of the temporary files that should result in less temporary files being created and less error messages displayed, but these messages are still possible.

That being said we are happy to consider changing the way temp files are created and handled if there is a better way of doing this or if you have any suggestions.

Kind Regards,
Mike

mitchelsellers:
@nerru86 If there are changes to this process included, I think that the best course of action would be to review the improved version, rather than trying to make changes without full education.

cathalconnolly:
merged as https://dnntracker.atlassian.net/browse/DNN-7361 - Note: I believe this will be backported to 7.4.2 - we also need to create a jira to xmlmerge 51Degrees.mobi.config changes into upgrades

PR #622, opened 2015-07-03
DNN-5785: add answer field in reset password control.

zyhfish:


cnurse:
@Ben - we cannot change the signature of the UserController method as it will break binary compatability.  We ned to deprecate the existing method and create the new method.  Call the new method with an empty answer.  I realise the old method will still fail, but that if other registration/login modules are using the API they will fail catastrophically as written.

zyhfish:
Hi Charles, add the original method in obsoleted user controller class.

cnurse:
Thats good now.

PR #624, opened 2015-07-06
DNN-6146: show clarify exception message when update user function failed.

zyhfish:


cnurse:
Exception message neds to be localized (use Exceptions.resx in Global Resources)

cnurse:
Thanks Ben

PR #625, opened 2015-07-06
DNN-7213: Removed DotNetNuke.Web.Client.ClientResourceManagement.ClientResourceManager.UpdateVersion()

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7213


cnurse:
This would break binary compatability for modules that take a dependency on this assembly. PLease deprecate the method and point the developer to the correct API method to use.  Then refactor the method to call the "new" API method.

PR #627, opened 2015-07-07
DNN-7201: move the setting into host setting.

zyhfish:


cnurse:
Assigned to Joe - based on a comment in the JIRA

jbrinkman:
@ashishpd Why was this merged?

PR #652, opened 2015-07-21
DNN-7142: tweak for MEF for Event Model.

galatrash-at-dnn:
This tweak saved about 10 seconds in the start-up time in our testing cloud environment.

### DIFF
--- a/DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
+++ b/DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
@@ -36,60 +36,87 @@ namespace DotNetNuke.ExtensionPoints
         public SafeDirectoryCatalog(string directory)
         {
             _catalog = new AggregateCatalog();
+            var matchingAssemblyFiles = GetMatchingAssemblies(directory);
 
-            // ASP.NET will load all the DLL files under the bin folder; the first loop would be sufficient
-            var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies()
-                                            .Where(asm => asm.GetName().CodeBase.StartsWith("file://"));
-
-            var processedFiles = new List<string>();
-            foreach (var assembly in loadedAssemblies)
+            foreach (var assembly in matchingAssemblyFiles.Select(f => f.Assembly))
             {
-                var fileUri = new Uri(assembly.GetName().CodeBase);
-                var fi = new FileInfo(fileUri.AbsolutePath);
-                if (fi.DirectoryName != null &&
-                    fi.DirectoryName.StartsWith(directory, StringComparison.InvariantCultureIgnoreCase))
-                {
-                    try
-                    {
-                        processedFiles.Add(fi.FullName.ToUpper());
-                        var asmCatalog = new AssemblyCatalog(assembly);
-                        //Force MEF to load the plugin and figure out if there are any exports
-                        // good assemblies will not throw the RTLE exception and can be added to the catalog
-                        if (asmCatalog.Parts.ToList().Count > 0) _catalog.Catalogs.Add(asmCatalog);
-                    }
-                    catch (ReflectionTypeLoadException)
-                    {
-                    }
-                    catch (BadImageFormatException)
-                    {
-                    }
-                    catch (FileLoadException) //ignore when the assembly load failed.
-                    {
-                    }
-                }
+                TryAddCatalog(_catalog, assembly);
             }
 
-            // process whatever left in subfolders
-            var files = Directory.EnumerateFiles(directory, "*.dll", SearchOption.AllDirectories)
-                                 .Select(f => f.ToUpper()).Except(processedFiles);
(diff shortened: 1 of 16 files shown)


DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
```diff
@@ -34,19 +35,51 @@ public class SafeDirectoryCatalog : ComposablePartCatalog
...
+                    }
+                    catch (FileLoadException) //ignore when the assembly load failed.
+                    {
+                    }
+                }
+            }
+
+            // process whatever left in subfolders
+            var files = Directory.EnumerateFiles(directory, "*.dll", SearchOption.AllDirectories)
+                                 .Select(f => f.ToUpper()).Except(processedFiles);
```
bdukes: This should use the `Except` overload that takes a `IEqualityComparer<string>` (https://msdn.microsoft.com/en-us/library/vstudio/bb336390.aspx), passing `StringComparer.OrdinalIgnoreCase`, rather than converting each file path to upper case

DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
```diff
@@ -34,19 +35,51 @@ public class SafeDirectoryCatalog : ComposablePartCatalog
...
             foreach (var file in files)
             {
                 try
                 {
-	                var asmCat = new AssemblyCatalog(file);
-
-	                //Force MEF to load the plugin and figure out if there are any exports
-	                // good assemblies will not throw the RTLE exception and can be added to the catalog
-	                if (asmCat.Parts.ToList().Count > 0) _catalog.Catalogs.Add(asmCat);
+                    var asmCatalog = new AssemblyCatalog(file);
```
bdukes: If we're going to do this twice, let's pull it out into its own method

DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
```diff
@@ -34,19 +35,51 @@ public class SafeDirectoryCatalog : ComposablePartCatalog
...
+                    //Force MEF to load the plugin and figure out if there are any exports
+                    // good assemblies will not throw the RTLE exception and can be added to the catalog
+                    if (asmCatalog.Parts.ToList().Count > 0) _catalog.Catalogs.Add(asmCatalog);
```
bdukes: I realize this isn't your code, but it seems like the `ToList()` is unnecessary here, I would expect that `asmCatalog.Parts.Any()` would be more efficient

DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
bdukes: nevermind, I guess this isn't really the same, since you're calling a different constructor

DNN Platform/Library/ExtensionPoints/SafeDirectoryCatalog.cs
```diff
@@ -34,19 +35,51 @@ public class SafeDirectoryCatalog : ComposablePartCatalog
 
         public SafeDirectoryCatalog(string directory)
         {
-            var files = Directory.EnumerateFiles(directory, "*.dll", SearchOption.AllDirectories);
-
             _catalog = new AggregateCatalog();
 
+            // ASP.NET will load all the DLL files under the bin folder; the first loop would be sufficient
+            var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies()
+                                            .Where(asm => asm.GetName().CodeBase.StartsWith("file://"));
```
bdukes: This is somewhat personal preference, but I'd prefer to see more of the conditions moved into the LINQ statement, and out of the loop (see http://www.martinfowler.com/articles/refactoring-pipelines.html).  Something like this, perhaps:

```
var matchingAssemblyFiles = (from assembly in AppDomain.CurrentDomain.GetAssemblies()
                             let codeBase = assembly.GetName().CodeBase
                             where codeBase.StartsWith("file://")
                             let fileUri = new Uri(codeBase)
                             let fileInfo = new FileInfo(fileUri.AbsolutePath)
                             where fileInfo.DirectoryName != null
                             where fileInfo.DirectoryName.StartsWith(directory, StringComparison.OrdinalIgnoreCase)
                             select new {
                                 fileInfo.FullName,
                                 Assembly = assembly,
                             }).ToList();

foreach (var assembly in matchingAssemblyFiles.Select(f => f.Assembly)) {
    TryAddCatalog(_catalog, assembly);
}

var otherAssemblyFiles = Directory.EnumerateFiles(directory, "*.dll", SearchOption.AllDirectories)
                                  .Except(matchingAssemblyFiles.Select (f => f.FullName), StringComparer.OrdinalIgnoreCase);
                                  
foreach (var assemblyPath in otherAssemblyFiles) {
    TryAddCatalog(_catalog, assemblyPath);
}
```

PR #653, opened 2015-07-21
DNN-7238: reducing user and module indexers memory footprint.

galatrash-at-dnn:
This is a fixed pull request for issue DNN-7238 (the older pull request was incorrect).


cnurse:
This will be resubmitted against the correct branch

PR #654, opened 2015-07-22
DNN-7138: Menu/Navigation (DDR Menu) optimization.

galatrash-at-dnn:
I've added some optimizations which cache the active-tab data per request and some other minor optimizations. The effect of these changes in a default installation of a Platform site was faster on average by about 2 to 3% per request in a single request mode, and about 5 to 10% per request when executing 10 concurrent requests on the Home page.


cnurse:
I spoke to George about removing the conditional compile code - George will push the changes.

PR #659, opened 2015-07-22
Changes to Insert DateTime at start of upgrade

mitchelsellers:
Changes to support DNN-7280 for enhanced experience when upgrading.  Localization values used for UpgradeStarted.Text were obtained from Google Translate, and a few validated to be correct.  Would appreciate a review from native speakers

### DIFF
--- a/Website/Install/App_LocalResources/UpgradeWizard.aspx.de-DE.resx
+++ b/Website/Install/App_LocalResources/UpgradeWizard.aspx.de-DE.resx
@@ -1,110 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
-    Version 2.0
-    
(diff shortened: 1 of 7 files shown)


cathalconnolly:
not sure how to process this - typically for translations you just add any new keys into the existing en-us resx file and then when I generate the difference files for the translators they provide the translations -I don't know if accepting the translations directly will cause any issues -in this case as it's the installation files which we ship I think it should be fine, but I'll probably check with the translators just in case

mitchelsellers:
@cathalconnolly if I need to pull back the changes to the others by all means I'll modify my repro and update the pull request.  Wasn't 100% sure on this one for process language

vmasanas:
There's only 1 change in the resx file, although most of the localized files seem changed. My suggestion would be to NOT include the localized values in the pull request and let the translators provide the new key.

mitchelsellers:
@vmasanas Done!  I updated things to remove the other changes.

PR #660, opened 2015-07-23
DNN-6330 Create FolderName for Packages for types that were not being…

ashishpd:
This needs to be cherry-picked into 7.4.2 as it is affecting CONTENT-5218, which in turn affecting translation of packages that are of type Library or EvoqConnector


cathalconnolly:
merged and cherrypicked as 7.4.2.147

PR #662, opened 2015-07-23
DNN-7249 DCC Add a context menu to insert Content Type fields to template 

cnurse:
This also includes the code for DNN-7250 (Add a context menu to insert Razor Helper tags into template )

### DIFF
--- a/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/App_LocalResources/Manager.resx
+++ b/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/App_LocalResources/Manager.resx
@@ -294,4 +294,10 @@
   <data name="Actions.Text" xml:space="preserve">
     <value>Actions</value>
   </data>
+  <data name="InsertField.Text" xml:space="preserve">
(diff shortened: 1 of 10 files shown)


jbrinkman:
This code is not working correctly.  I am seeing very strange behavior in Chrome, FF and Edge browsers.  In no case does the context menu display (although I can see that the underlying html is present).  In FF, the default browser context menu is being displayed.

cathalconnolly:
tested this (from the PR) - different results per browser: chrome (works fine), FF (partly works, browser context menu appears as well but can be ignored), IE (doesnt work, browser context menu overlays template context menu so it is unreachable)

cnurse:
I have it working in Edge, Chrome and Opera, but no matter what I try I cannot get IE and FF to stop showing their menus.

cnurse:
If in doubt use jQuery - switched from using code mirror's own event handling to using jQuery's bind method on the CodeMirror object and hey presto it works in all browsers - Chrome, Edge, Opera, IE and Firefox.

PR #663, opened 2015-07-23
DNN-7281 & CONTENT-5195: added/modified unit tests for this issue.

galatrash-at-dnn:
This should be imported in both branches for completeness of the task.


cnurse:
This should also be cherry-picked as the new code it is testing was.

PR #666, opened 2015-07-23
DNN-7287: Enable selecting child-pages of disabled pages as 404 Error Page or 500 Error Page

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7287

### DIFF
--- a/DNN Platform/DotNetNuke.Web/InternalServices/FileUploadController.cs
+++ b/DNN Platform/DotNetNuke.Web/InternalServices/FileUploadController.cs
@@ -292,7 +292,8 @@ namespace DotNetNuke.Web.InternalServices
                     return savedFileDto;
                 }
 
-                var file = FileManager.Instance.AddFile(folderInfo, fileName, stream, true, false, FileManager.Instance.GetContentType(Path.GetExtension(fileName)), userInfo.UserID);
+	            var contentType = FileContentTypeManager.Instance.GetContentType(Path.GetExtension(fileName));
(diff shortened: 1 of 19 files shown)


antoniodgonzalez:
@janjonas, it seems that this PR contains changes that are not related to the solution of the fix. Can you please review the modified files?

janjonas:
Hi @antoniodgonzalez, I do not know, why other people have added commits to this PR.
IMHO, we only need my commit https://github.com/janjonas/Dnn.Platform/commit/02cdfc3954a6028871550f5285f6a161e7a209c6 for this PR.

antoniodgonzalez:
Thanks @janjonas. Please, can you create a new pull request just including the changes to be merged? this way we can ensure we don't miss something, or we don't add some regression bug.

Also, we detected some things that can be improved, so we could use this same new PR for the changes:
- In ItemListServiceController.GetPageDescendants, the added disabledSelectable parameter is not being used, and it could be removed
- If a child page is disabled (for example the Error404 from the Jira) then it can still be selected from the dropdown, even if DisabledNotSelectable is set to true. This seems to work fine for the top level pages, tough.

Thanks

janjonas:
Hi @antoniodgonzalez,
I've created a new and "clean" pull request https://github.com/dnnsoftware/Dnn.Platform/pull/1181 which should be merged easily into development. I've also added a second commit that fixes the two mentioned problems:
* Disabled sub-pages are now no longer selectable
* The disabledSelectable parameter was renamed to disabledNotSelectable and is now used (to disable selecting of sub-pages)

ashishpd:
Closing this PR as new one was submitted

PR #670, opened 2015-07-24
DNN-7290 revert addition of language specific portal templates

cathalconnolly:
Reverts dnnsoftware/Dnn.Platform#656


cathalconnolly:
merged my own commit as it's breaking installs when a source build is used

PR #671, opened 2015-07-24
DNN-6546 - workaround, it's not a permanent solution but it seems to work

raphael-m:


jbrinkman:
Thanks @raphael-m We'll look at it and see if we can't find the root cause.

cnurse:
This has been closed as there is a Pull Request which handles the root cause.

PR #680, opened 2015-07-28
DNN-7144: removing 51Degrees dependency.

galatrash-at-dnn:
I created a new "ClientDependencyProvider" that replaced 51D's which uses the ASP.NET "HttpBrowserCapabilities" class.
To use it, there is a need to replace the 51D references in the distribution web.config file (which I didn't do as this should go into EVOQ codebase only and not Platform).

Note: some features that EVOQ Analytic is depending on from 51D (e.g., operating system) are not available through this provider and there is a need for a decision about them.


cnurse:
I am confused about this issue - the JIRA (which is not public) mentions that this should only be included in Evoq, so why is it here as a PR for Platform?

Also the problem with the ASP.NET capabilities is that it requires the end-user to manage new devices through the use of .browser files.  In the past we have found this to be a major head-ache and was the reason for adoption of 51 Degrees.

Assigning to Joe as I don't think it should be merged into Platform at this time (without further discussion/review)

galatrash-at-dnn:
This feature was requested originally for EVOQ. But since it was under DNN issue in Jira, I added it here. I will be moving it to CONTENT instead.

PR #682, opened 2015-07-30
DNN-6970: remove the view online help button as it not used anymore in codebase.

zyhfish:


cnurse:
@Ben - I am not sure this is correct - the JIRA says ONLY to remove online help if not specified in the manifest - not to remove in all cases

PR #711, opened 2015-08-12
DNN-7377: call renderModuleList to re-calculate the list position even ajax request return no data.

zyhfish:


cnurse:
Joe - does this need cherry-picking

PR #721, opened 2015-08-13
DNN-7306 DCC Quick Settings view for DCC Viewer

cnurse:


DNN Platform/Library/DotNetNuke.Library.csproj
```diff
@@ -1706,6 +1718,7 @@
     </Compile>
     <Compile Include="WebControls\FeedBrowser.cs" />
     <Compile Include="WebControls\WebControlBase.cs" />
+    <None Include="app.config" />
```
jbrinkman: Although this is shown as a change in the library project, there is no corresponding app.config that was checked in.  As a result the DNN Library won't compile.

PR #723, opened 2015-08-14
DNN-7400 - Ensured that Module Action menus are only added into the ma…

agrueso:
…in DNN form element

### DIFF
--- a/Website/admin/Menus/ModuleActions/ModuleActions.js
+++ b/Website/admin/Menus/ModuleActions/ModuleActions.js
@@ -16,7 +16,7 @@
         var isShared = opts.IsShared;
         
         if (count > 0 || supportsMove) {
-            var $form = $("body form");
+            var $form = $("body#Body form#Form");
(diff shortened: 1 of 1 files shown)


Website/admin/Menus/ModuleActions/ModuleActions.js
```diff
@@ -16,7 +16,7 @@
         var isShared = opts.IsShared;
         
         if (count > 0 || supportsMove) {
-            var $form = $("body form");
+            var $form = $("body#Body form#Form");
```
jbrinkman: This should really just be var $form = $("#Form");  The #Form should be unique on the page and therefore the rest of that selector is slowing down the page rendering since the browser must do a bunch of extra work to process the unnecessary selectors.

cnurse:
Given that I have heavily modified this file in a separate PR - I will merge this (including Joe's suggestion) once the QuickSettings PRs are prcessed.

PR #724, opened 2015-08-14
DNN-7403 Remove outdated Navigation providers

jbrinkman:
This PR removes the outdated Navigation providers - including solpart menu from the platform.


jbrinkman:
After doing some testing, everything seems to be working correctly.  This is a breaking change and I do expect some older skins and containers to break.

PR #727, opened 2015-08-16
DNN-7410: Fix GetSchedule

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7410

Please note: I'm not sure, if 08.00.00.SqlDataProvider is the right place or if its better to create a new 08.00.00.XX.SqlDataProvider file to support "in-place" upgrades?!

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/08.00.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/08.00.00.SqlDataProvider
@@ -8,6 +8,35 @@
 /*****                                                  *****/
 /************************************************************/
 
+/* DNN-7410 Fix GetSchedule */
+/****************************/
+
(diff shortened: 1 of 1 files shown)


cnurse:
We will need to put it in an incremental script file but I will do that as part of the merge

janjonas:
@cnurse & @jbrinkman: Is there any information available how the incremental script work? Do they need to be registered in the manifest file or is there any logic that checks which 08.00.00.xx.SqlDataProvider files exists and have not yet been executed in the environment? If yes, is this "magic" also available for modules?

cathalconnolly:
It works by checking for the existence of "build" versions of the sqldataprovider (e.g. 08.00.00.01.sqldataprovider) and verifying if they have been applied yet i.e. it checks to see the last build applied against the version table, and if an incremental file exists that is higher than that it applies it. When applying that it also looks for .config files of the same name as well as cleanup files of the same name and applies those if they exist.

jbrinkman:
@janjonas I expect we'll be extending this feature to other extensions in an upcoming DNN 8 sprint.  I have a blog post started on the topic as part of our upcoming CTP 4 release.

PR #735, opened 2015-08-18
DNN-7153: optimizing siteurl.config rewrite rules handling

galatrash-at-dnn:


zyhfish:
there have a possible issue with change in RewriterRule.cs, the method GetRuleRegex used parameter applicationPath to generate regular expression, then store in a variable, so if the parameter value use a different value next time, it will return wrong value.

galatrash-at-dnn:
I based my changes on the assumption that application path remains constant for the whole life-cycle of the app. How could the ``applicationPath`` change after the application starts?

PR #761, opened 2015-08-28
DNN-7488: remove the community version of detector http module if evoq version exists.

zyhfish:


ohine:
I would think it would make more sense to put this into the iUpgradeable function when Evoq is installed or upgraded instead of in the platform.

cathalconnolly:
Unless there is something I am not seeing here thent this should be something that is handled by a config xmlmerge (i.e. in the dnn manifest file) and not in code -you should be able to use the remove/removeattribute actions and then add back in the necessary nodes

cnurse:
I am rejecting this - per Oliver and Cathal's comments the fix is in Evoq's provider - it should remove the platform detector on install so their is no possible conflict.

PR #766, opened 2015-08-31
DNN-6790 add pieces for initial implementation of oauth server support (for ctp, will later be refactored as a library)

cathalconnolly:


cnurse:
Merged manually - fixed a couple of build issues and missing OQs

PR #777, opened 2015-09-05
DNN-7520: return proper number of records

sleupold:
- bug fix: Return proper number of records, even if multiple duplicate emails found.
- optimization: Use view
- improvement: include superusers


zyhfish:
both the original sp and updated sp in this PR have problem in below scenario:
1. in portal 0, create 2 users with email(testuser01@dnn.com) and 1 user with email(testuser02@dnn.com);
2. create 1 super user with email(testuser02@dnn.com);
3. execute sp GetDuplicateEmailCount with parameter value to 0.

Expected Result:
in portal 0 it should return 1 as only testuser01@dnn.com duplicated.

Actually Result:
it return 2, as its checked include super users. there also have problems in the original SP, we need fix this bug before merge.

ohine:
Also, this needs to target the development branch and not the v7.4.2 release branch

sleupold:
I'll Review and resubmit

antoniodgonzalez:
@sleupold, this pull request is pending to be merged, but we are waiting for the commented changes before we can merge. Thanks.

sleupold:
please check again. I fixed the procedure and moved it to its own file.
Note: it might be helpful to exclude deleted users, but this would require adding a check in DNN platform code, whenever the user is re-activated.

sleupold:
@zyhfish 
Ben, of course, the check needs to include superusers, otherwise you cannot enable login using email address. Exec GetDuplicateEmailCount -1 should only be used for superusers, which need to make sure, there is NO existing user with same email address in ANY portal.

antoniodgonzalez:
@sleupold, Sorry this PR is against release/7.4.2, but must be against development branch (8.0.0). Could you please fix this? Thanks

sleupold:
@antoniodgonzalez 
is there an option to change the target branch or do I need to close and re-open an new PR?

antoniodgonzalez:
@sleupold, for I'm aware of, there is no option to change the target branch. You can create a new PR

sleupold:
needs to be resubmitted against a different branch, but I don't have the time atm.
Fix is included in TurboScripts and will be re-submitted next year.

PR #784, opened 2015-09-11
DNN-7332 DCC Create Editor Templates for Base Types

cnurse:
This also includes the fix for DNN-7426 DCC Viewer - Page needs to be refreshed if settings have been changed


cathalconnolly:
merged as https://dnntracker.atlassian.net/browse/DNN-7322 (title incorrectly says DNN-7332)

PR #787, opened 2015-09-14
DNN-7446

BehnamEmamian:
These fields are needed to fetch from Facebook.


BehnamEmamian:
Hi
did you check it? it solve my problem on this site
http://crm.persianmigrant.com/

facebook api in version 2.4 removed default parameters so we should tell api witch parameters we need to fetch.

cathalconnolly:
FYI https://dnntracker.atlassian.net/browse/DNN-7446 is planned for 8.0.0

cathalconnolly:
this has also been done as #880 -ive added a comment to that to ensure that this PR is processed and you get credit for your work.

antoniodgonzalez:
#880 is now merged

PR #788, opened 2015-09-14
DNN-7546 - Added new icons and fixed CSS to reflect changes wanted.

mgatmaytan:


cnurse:
I am not sure about this change - these are platform images that have been modified to fit the Evoq default skin.  These should be handled through the Skin install for Evoq - it is possible to include images for this folder in the skin zip - assigning to Joe.

jbrinkman:
Charles is correct.  We should not change platform due to a skin requirement from Evoq.  This could easily be fixed in CSS in Evoq without requiring any change to platform.

PR #790, opened 2015-09-15
Added a few more properties to the token replace logic

donker:
I needed access to some of these IDs so I can call services.


cnurse:
I have reverted this as I noticed after merging that it was done against the incorrect branch - I will manually move the code to development.

donker:
Can you add this to the end of it:

                default:
                    if (_moduleContext.Settings.ContainsKey(propertyName))
                    {
                        return (string) _moduleContext.Settings[propertyName];
                    }
                    break;

cnurse:
Should probably be a separate PR but to reduce ceremony I will add it.

PR #796, opened 2015-09-16
DNN-7574: Fixed the install issue for Engage when there's no langu…

kanm2015:
…age pack available and the user local language is not English


ashishpd:
This PR has gotten wrong commits - we are submitting a new one https://github.com/dnnsoftware/Dnn.Platform/pull/804

PR #813, opened 2015-09-22
DNN-7525 Remove field length limitation on Answer field

cnurse:


cathalconnolly:
Arguably the limit shouldnt have been removed, but instead set to match the database where it appears this is th PasswordAnswer	nvarchar(128)	column in aspnet_membership. However with it being unlikely any user would provide such a long answer, I've elected to merge this.

cnurse:
I don't disagree - I thought about that - but that would have meant more work as we would have needed to change all occurrences of the Answer field (which has no limit in many other places).  I think this should be reviewed when we make changes in 8.x for Security

PR #816, opened 2015-09-23
DNN-6637 - Separate Skins and SkinDesigner modules into a single module

hismightiness:
* Combines Skins and Skin Designer (Attributes) into a single module with two definitions
* Added to the Admin Modules area of the Solution
* Builds moves files and creates packages
* Installs and upgrades in DNN 8
* Changes its version to 01.01.00
* Removes Skin pages and adds Theme pages
* Adds instances of the new Themes modules to the Themes page
* Updates any references to previous skin modules that might be on other pages
* Specifies a cleanup file to delete the old files
* Logs any errors that happen during install/upgrade
* Existing functionality was not updated (should work exactly the same as before this update)


hismightiness:
Before upgrade

module-data-before-upgrade

After upgrade

module-data-after-upgrade

cnurse:
In merging this I noticed that the version was set to 1.1.0.
All the other extracted modules are being given the same version as the core - 8.0.0, so I have updated that as part of the merge.

hismightiness:
Please revert your change. That was done intentionally, and with approval from Joe - pending testing.  I spent hours making sure that would work.  I didn't spend all that time making sure that the version number would accurately convey the features in the module, only to have it removed arbitrarily like this.

hismightiness:
http://www.dnnsoftware.com/forums/forumid/228/threadid/519915/scope/posts

hismightiness:
Also, please look at the attached screen shots that accompany the PR above for evidence of the working version update.

cnurse:
As Joe is on PTO - I will hold this PR until he returns. Originally when we decided to do this extraction we established a process - on step of which was to set the version of all modules extracted to the same as the core.

hismightiness:
You're right, for simplicity sake.  Changing the version number is a lot more work.  :)

cnurse:
I have assigned this to Joe.  While he indicated in the forum he was ok with the change, we have used 8.0 for all other modules extracted.

francescorivola:
I have merged this pull request manually in a new branch (feature/DNN-6637). I will create a new pull request from this new branch and mark this a closed once I verify some requirements with @jbrinkman or @zyhfish.

francescorivola:
@hismightiness, even if I close this PR as closed, I will label it as merged so you get credits for the good work that you have done on this item.

francescorivola:
commits from this pull request has been merged as part of https://github.com/dnnsoftware/Dnn.Platform/pull/987.

PR #818, opened 2015-09-23
DNN-7534: Updated the SharedConstants to be non-static so that its pr…

kanm2015:
…operty values get update when user switching languages.


bdukes:
I don't know of any uses of this in the wild, but changing from a field to a property is going to be a breaking change.

kanm2015:
Closing this pull request, we will do it in a different way.

PR #822, opened 2015-09-24
DNN-6563 Enable Switch UI Language for all users, not just superusers

vmasanas:


cathalconnolly:
I'm going to replicate the check for UI culture selection when content enabled in the SwitchLanguage service framework method (so it matches the logic that determines if it's displayed) i.e. "if (PortalSettings.AllowUserUICulture && PortalSettings.ContentLocalizationEnabled)" - i'll manually merge the changes into development and cherrypick to release/7.4.2

cathalconnolly:
working on manually merging this into the 7.4.2 release

cathalconnolly:
manually merged as 7.4.2.216

cathalconnolly:
FYI we merged the change into 8.0.0 and the release/7.4.2 branch, however testing shows that this only extends access to the switch UI language to admin users (previously it was only available to superusers). I inspected the source code and this has been “broken” since at least aug 16 2013 (when the source moved to github)
 
We don’t want to make another change so late to the 7.4.2 release, and want to take time to ensure we don’t accidentally expose other control panel elements to non-admin/host users, so have elected to do the page editor part of the enhancement as part of 8.0.0 –you can track the jira @ https://dnntracker.atlassian.net/browse/DNN-7642

PR #823, opened 2015-09-24
DNN-7278: keep checkpoint data if exception throw.

zyhfish:


cnurse:
@zyhfish see the comment in the JIRA - rather than re-opening a closed/regressed issue we need a new issue to track this change

PR #828, opened 2015-09-25
DNN-7603 Fix layout issue for multi-field content types

jbrinkman:


cnurse:
There is still a minor layout issue with the help icon in the vertical scenario - but that can be handled separately.

PR #725, opened 2015-08-16
DNN-7408: Fix validation of txtRetryTimeLapse control

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7408

PR #829, opened 2015-09-27
DNN-7408 & DNN-7409: Make RetryTimeLapse optional

janjonas:
This is a follow-up pull request to https://github.com/dnnsoftware/Dnn.Platform/pull/725 and https://github.com/dnnsoftware/Dnn.Platform/pull/726. I think the "retry time lapse" property should be optional and not mandatory.

### DIFF
--- a/Website/DesktopModules/Admin/Scheduler/App_LocalResources/EditSchedule.ascx.resx
+++ b/Website/DesktopModules/Admin/Scheduler/App_LocalResources/EditSchedule.ascx.resx
@@ -240,9 +240,6 @@
   <data name="RetryTimeLapseValidator.ErrorMessage" xml:space="preserve">
     <value>Retry Frequency range is from 1 to 999999</value>
   </data>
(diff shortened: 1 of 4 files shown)


donker:
This makes sense as in the actual scheduler itself the value is not required either. I.e. the scheduled item is not retried if the value is null.nullinteger.

donker:
@janjonas : can you create a new Jira ticket for this with the rationale?

janjonas:
When I started working on this problem I thought the RetryTimeLapse porperty should be mandatory (see https://dnntracker.atlassian.net/browse/DNN-7408 and https://dnntracker.atlassian.net/browse/DNN-7409). Then I realized that I was wrong and that's why I created comments in both tickets and both pull requests but it seems that nobody has read them because the tickets have been closed and the pull requests have been merged. That's why I created this new pull request (both tickets are referenced in the pull request's title). I would recommend to reopen the existing JIRA tickets and assign this pull request to them.

PR #844, opened 2015-10-01
DNN-7657 DDR Menu with Client Resource Manager

bdukes:
[DNN-7657](https://dnntracker.atlassian.net/browse/DNN-7657)

>The DDR Menu allows a menu template to define style sheets and scripts that it loads, but it does all of the loading manually.  Both JS and CSS should be loaded via the client resource manager.  Additionally, it would be nice if the DDR menu supported requesting JS Libraries, in addition to loose scripts.

### DIFF
--- a/DNN Platform/Modules/DDRMenu/DNNAbstract.cs
+++ b/DNN Platform/Modules/DDRMenu/DNNAbstract.cs
@@ -85,11 +85,5 @@ namespace DotNetNuke.Web.DDRMenu
 		{
 			menuNode.LargeImage = dnnNode.LargeImage;
 		}
-
-		public static bool RequestJQuery()
-		{
-            JavaScript.RequestRegistration(CommonJs.jQuery);
-			return true;
-		}
 	}
 }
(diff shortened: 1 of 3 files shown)


bdukes:
The next part of this change would be to include the ability to register JS libraries through a menu template, but it renders in the `PreRender` of the menu control, after the JS library requests have been dispatched (so new requests don't have an effect).  It seems like the fix for that should be in the JS Library code (this isn't the first time I've run into it), but I'm unsure of the correct approach.  

If there's a way to get that code to run after the `PreRender` is done, that would be great, but I don't know how to achieve that off the top of my head.  Alternatively, maybe requesting a JS library after they've been dispatched should cause it to be immediately dispatched.  I may just have to try it out and see what I can get to work, unless anyone has any insights to offer.

bdukes:
I've implemented the JS script and JS library portions.  #872 (DNN-7701) needs to be fixed in order for the JS libraries to work.  Once that's done, this can be processed

PR #846, opened 2015-10-02
DNN-7617 Module deletion leaves orphaned Content Items

cnurse:


sleupold:
as this is a data consistency issue, I would prefer this to be handled in the database by the stored procedure or a foreign key constraint.

PR #849, opened 2015-10-07
DNN-7672: Added cshtml as allowed file extension

EPTamminga:


cathalconnolly:
i would be inclined to not allow this as we would not want cshtml as an automatic file extension available anywhere that the file extensions list is allowed e.g. it might open up a security issue where users could upload them to their profile. Instead any modules (such as DCC) that require cshtml should add it to the list of supported extensions within their module -I believe there is an API method added by @cnurse to support this.

jbrinkman:
I agree with Cathal.  This opens up too big a security risk to be a default.  Hosts have the option to set this if it makes sense for their environment, but we should be secure by default.

PR #851, opened 2015-10-07
DNN-7458 Optimized Type scanning

cnurse:
This code was cherry-picked from the speed-base-dnn-7458 branch (which was based on an old 7.4.2 build)


zyhfish:
Reviewed, looks good.

PR #853, opened 2015-10-07
DNN-7142 Refactor MEF for better performance

cnurse:


zyhfish:
Reviewed, looks good.

PR #864, opened 2015-10-08
DNN-7139 Review web.config for Platform

cnurse:


zyhfish:
This only handled new install, but not include upgrade action.

PR #867, opened 2015-10-10
DNN-7699: Fixed AdvancedFriendlyUrlProvider to not change the PortalAlias of the passed PortalSettings

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7699

### DIFF
--- a/DNN Platform/Library/Entities/Urls/AdvancedFriendlyUrlProvider.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedFriendlyUrlProvider.cs
@@ -212,19 +212,6 @@ namespace DotNetNuke.Entities.Urls
             return friendlyPath;
         }
 
-        private static void CheckAndUpdatePortalSettingsForNewAlias(PortalSettings portalSettings, bool cultureSpecificAlias, string portalAlias)
-        {
-            if (cultureSpecificAlias && portalAlias != portalSettings.PortalAlias.HTTPAlias)
-            {
-                //was a change in alias, need to update the portalSettings with a new portal alias
-                PortalAliasInfo pa = PortalAliasController.Instance.GetPortalAlias(portalAlias, portalSettings.PortalId);
-                if (pa != null)
-                {
-                    portalSettings.PortalAlias = pa;
-                }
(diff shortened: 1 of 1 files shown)


donker:
Anyone know what the rationale behind CheckAndUpdatePortalSettingsForNewAlias was?

bdukes:
cc @brucerchapman

janjonas:
@donker: IMHO the intention was/is to find the correct (localized) portal alias. Basically, my code does the same but does not change the portal alias of the PortalSettings object that was passed as argument to FriendlyUrlInternal.

janjonas:
@fjlucas: Why is the pull request labled with "Waiting for Response"? Do you need any feedback?

janjonas:
@jbrinkman: Maybe you can take a look at this pull request. Thanks :)

DNN Platform/Library/Entities/Urls/AdvancedFriendlyUrlProvider.cs
```diff
@@ -443,38 +425,61 @@ private string FriendlyUrlInternal(TabInfo tab, string path, string pageName, st
...
-                if (portalSettings == null)
+                PortalSettings portalSettigsForFriendlyUrl = null;
+                if (cultureSpecificAlias && tab != null && portalSettings != null && portalAlias != portalSettings.PortalAlias.HTTPAlias)
                 {
-                    PortalAliasInfo alias = PortalAliasController.Instance.GetPortalAlias(portalAlias, tab.PortalID);
+                    // Gew new (culture specific) PortalAlias
+                    var pa = PortalAliasController.Instance.GetPortalAlias(portalAlias, portalSettings.PortalId);
+                    if (pa != null)
+                    {
+                        portalSettigsForFriendlyUrl = new PortalSettings(tab.TabID, pa);
```
zyhfish: you create a new portal settings object but not use the value from parameters, this may cause unexpected result, such as if user modify some settings in portal settings object before call this method, those changes will lost.

zyhfish:
add inline comment about concern, will close this and submit another PR.

PR #869, opened 2015-10-12
DNN-7569: Remove the FeedBrowser.

zyhfish:


cnurse:
@Ben - can you tweak this to use the incremental version 8.0.0.12 - rather than 8.0.0.6 - you will need to add a new cleanup file (08.00.00.00.12.txt) and a new private method in Upgrade.cs.  Otherwise people who are using CTPs won't get the change correctly.

Thanks

zyhfish:
done.

cathalconnolly:
sorry folks, 8.0.0.12 is being used already, it'll need moved again

jbrinkman:
:shipit:

PR #871, opened 2015-10-12
DNN-7702: fix issue with page template, skin and container dropdown lists loosi…

ohine:
…ng their values after a partial postback

https://dnntracker.atlassian.net/browse/DNN-7702


cathalconnolly:
@cnurse will need to have a look at this first as I believe it was he who added the viewstatemode attributes to reduce the size of the rendered output (and the viewstate)

cnurse:
This is fine.  We were probably over-zealous in removing ViewState. It is needed in some WebForm scenarios

PR #873, opened 2015-10-13
DNN-6788: search for dynamic content.

zyhfish:


cnurse:
@Ben - I chatted with Ash about this.  We were a little concerned that the Search Document may not be "committed", especially if the Search Crawler is running on node 1 in a Web Farm and your code is executing on Node 2.

PR #874, opened 2015-10-13
DNN-7568: Remove WhatsNew Module.

zyhfish:


cnurse:
@Ben - same comment as for the FeedBrowser - can you make modify this to use 8.0.0.12 incremental upgrade.

Thx

zyhfish:
done.

PR #880, opened 2015-10-14
DNN-7446: update api to match v2.4 requirement.

zyhfish:


cathalconnolly:
part of this already seems to be done as #787 -please make sure that user's pull request is updated accordingly and they get credit for doing the work

ashishpd:
This PR is on hold for the time being. Team needs to understand the oAuth status prior to jumping here.

PR #881, opened 2015-10-15
DNN-7694: update admin page value in manifest file.

zyhfish:


zyhfish:
This PR no need as the change will include in DNN-7565.

francescorivola:
I am deleting this branch as this PR is no longer needed.

PR #884, opened 2015-10-15
DNN-4133

JoshuaBradley:
Changes for the prefixing of the service, allowing it to be overridden if it is not wanted.


DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -869,45 +869,9 @@ private void ValidateForDuplicateDisplayName(UserInfo user, ref UserCreateStatus
         public override UserCreateStatus CreateUser(ref UserInfo user)
         {
             UserCreateStatus createStatus = ValidateForProfanity(user);
+            EventLogController aLog = new EventLogController();
             string service = HttpContext.Current != null ? HttpContext.Current.Request.Params["state"] : string.Empty;
 
-            //DNN-4016
```
bdukes: are you intending to remove the OAuth login logic?  or was this a bad merge?

bdukes:
@JoshuaBradley The only substantial change that I notice here is the removal of OAuth login logic.  Can you explain what you're changing and why (or resubmit the pull request without the extraneous changes)?  Thanks!

bdukes:
The changes to `OAuthClientBase.cs` look like they're probably good, but the other changes need to be undone

francescorivola:
Reviewing the corresponding jira look like this issue has been fixed in PR https://github.com/dnnsoftware/Dnn.Platform/pull/35. Also the jira is closed. 
I have removed the 8.0.0 milestone from this PR. @JoshuaBradley could you please provide some info about this PR and let here know if this still need to be merged? If is the case please create new jira documenting properly the issue and link the PR to it. Thank you so much.

JoshuaBradley:
Sorry for the delay on this. This was actually fixing a bug that PR #35 introduced with the prefixing on OAUTH providers. I think that this still needs more work, per @bdukes, but I haven't had a chance to look into this further.

PR #891, opened 2015-10-16
DNN-7744 Add missing members to DnnHtmlHelper and DnnUrlHelper

bdukes:
[DNN-7744](https://dnntracker.atlassian.net/browse/DNN-7744)

>The `DnnWebViewPage<TModel>` base class that MVC views and DCC templates use overrides the standard MVC helpers, `Html` and `Url`. There are some methods and properties available through those helpers in MVC that are not accessible in DNN's versions.
>
>For example, using `@Html.Raw(Dnn.ViewData.Model.Content.Fields["ProposalTitle"].Value)` in a DCC template gives an error that "`DotNetNuke.Web.Mvc.Helpers.DnnHtmlHelper<dynamic>` does not contain a definition for `Raw`"

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Helpers/DnnHtmlHelper.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Helpers/DnnHtmlHelper.cs
@@ -2,13 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Web;
 using System.Web.Mvc;
 using System.Web.Routing;
 using DotNetNuke.Framework;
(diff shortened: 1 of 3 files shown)


ohine:
This appears to use c# 6.0 functionality, I reverted the change to fix build issues.

bdukes:
I thought we had agreed that C# 6 was okay to use.  Does TeamCity need to have VS 2015 installed?

antoniodgonzalez:
C# 6 is now officially supported, and TC has been upgraded to VS 2015. 
This change has been added back.

bdukes:
:+1: thanks @antoniodgonzalez!

PR #898, opened 2015-10-18
DNN-7749: Add new option "All roles" to add all roles to the permission grid

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-7749


cpaterra:
The permissions grid is complex enough already. I have major concerns about adding more functionality to it that would benefit very few. I am marking it "merged" so you get community credit but rejecting the PR.

PR #901, opened 2015-10-19
DNN-7761 Initial Telerik Separation

ohine:
This is a breaking change for any modules which reference the dnn/telerik wrappers. If you are using these controls, you will need to update your references from DotNetNuke.Web to DotNetNuke.Web.Deprecated


DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/DnnDatePicker.cs
```diff
@@ -33,11 +33,6 @@ public class DnnDatePicker : RadDatePicker
     {
         protected override void OnInit(EventArgs e)
         {
-			if (CultureInfo.CurrentCulture.Name == "ar-SA")
```
bdukes: why is this removed?

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/PropertyEditorControls/DateEditControl.cs
```diff
@@ -199,18 +199,8 @@ public override string ID
 				base.ID = value;
 			}
 		}
-
-	    public override string EditControlClientId
```
bdukes: is there a reason this was removed?  (same question for `DateTimeEditControl`)

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/TermsSelector.cs
```diff
@@ -117,18 +115,10 @@ protected override void OnLoad(EventArgs e)
 				}
 			}
 
+			Page.ClientScript.RegisterClientScriptResource(GetType(), "DotNetNuke.Web.UI.WebControls.Resources.TermsSelector.js");
 
-			if (!Page.IsPostBack)
```
bdukes: what about this one?

DNN Platform/Library/Services/Exceptions/ExceptionInfo.cs
```diff
@@ -47,7 +47,9 @@ public ExceptionInfo(Exception e)
 				InnerMessage = e.InnerException.Message;
 				InnerStackTrace = e.InnerException.StackTrace;
 			}
-			ExceptionHash = e.Hash();
+
+            //only calculate the hash based on messages and the strack trace
+            ExceptionHash = string.Concat(e.Message.ToLower(), e.StackTrace == null ? string.Empty : e.StackTrace.ToLower(), InnerMessage == null ? string.Empty : InnerMessage.ToLower(), InnerStackTrace == null ? string.Empty : InnerStackTrace.ToLower()).GenerateHash();
```
bdukes: Is this a Telerik-related change?

DNN Platform/Library/Services/Exceptions/Exceptions.cs
```diff
@@ -305,14 +306,23 @@ public static void ProcessModuleLoadException(string FriendlyMessage, Control ct
...
+                    if (Host.DebugMode && exc is HttpParseException)
+                    {
+                        var httpParseError = ((HttpParseException)exc);
+                        friendlyMessageOverride = string.Concat(exc.Message, " in ", httpParseError.VirtualPath, ":line ", httpParseError.Line);
+                    }
+
                     if (ctrlModule == null)
                     {
-                        lex = new ModuleLoadException(exc.Message, exc);
+                        lex = new ModuleLoadException(friendlyMessageOverride, exc);
```
bdukes: is this a Telerik-related change?

DNN Platform/Library/Services/UserProfile/UserProfilePicHandler.cs
```diff
@@ -163,16 +163,7 @@ public void ProcessRequest(HttpContext context)
...
-	        try
-	        {
-				context.Response.End();
-	        }
-			catch (ThreadAbortException)//if ThreadAbortException will shown, should catch it and do nothing.
-			{
-
-			}
-            
+            context.ApplicationInstance.CompleteRequest();
```
bdukes: Is this a Telerik-related change?

DNN Platform/Library/Services/Exceptions/Exceptions.cs
ohine: Yes, I wanted to improve the errors when we run into an HttpParseException (triggered when the namespace doesn't match the control registered in the ascx). Currently you need to attach the debugger to figure out which sub ascx has a parse exception. This change will update the friendly exception message and include the control and line number with the parse exception.

DNN Platform/Library/Services/UserProfile/UserProfilePicHandler.cs
ohine: Not really, it was throwing an exception on each request and annoyed me while debugging the telerik changes. Figured it could be bundled in since the change is pretty minor.

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/DnnDatePicker.cs
ohine: I'll add this back, it was removed without my knowledge. Time gap between the start of this and finalizing the commit.

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/PropertyEditorControls/DateEditControl.cs
ohine: I'll add this back, it was removed without my knowledge. Time gap between the start of this and finalizing the commit.

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/TermsSelector.cs
ohine: I'll add this back, it was removed without my knowledge. Time gap between the start of this and finalizing the commit.

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/DnnDatePicker.cs
ohine: This is fixed, but I'm not sure why git didn't track the change and remove your comment like the others.... I'm guessing it's because of the indentation?

Website/DesktopModules/Admin/Security/Roles.ascx.cs
```diff
@@ -245,7 +245,7 @@ void cmdAddRoleGroup_Click(object sender, EventArgs e)
 
         void cmdAddRole_Click(object sender, EventArgs e)
         {
-            Response.Redirect(EditUrl("RoleGroupID", Request.QueryString["RoleGroupID"]), true);
```
bdukes: is this an intentional change?

Website/DotNetNuke.Website.csproj
```diff
@@ -1114,13 +1114,6 @@
     <Compile Include="Install\Install.aspx.designer.cs">
       <DependentUpon>install.aspx</DependentUpon>
     </Compile>
-    <Compile Include="Install\InstallWizard.aspx.cs">
```
bdukes: is this an intentional change?

ohine: Yes, the install wizard was moved outside of the website project, but the file was still listed in the project and was flagged as missing.

Website/DesktopModules/Admin/Security/Roles.ascx.cs
ohine: yes, I removed the true flag because it's the default action.

DNN Platform/Library/Services/UserProfile/UserProfilePicHandler.cs
bdukes: I don't really understand the full repercussions of this change, but this is my only outstanding question on the full PR, so if anyone else can sign off on this, the rest gets my :+1:

bdukes:
Thanks for your hard work on this @ohine, I know you put a lot of time into this one!

DNN Platform/Library/Services/UserProfile/UserProfilePicHandler.cs
ohine: My understanding is it reduces a thrown exception. Exceptions are costly and we shouldn't throw them if we can help it.

ashishpd:
Please be advised that @zyhfish will be doing some testing of this PR and will be providing feedback. He will be testing in both the Platform and the Commercial builds. Tests will include both installs  and upgrades. Thanks - Ash

ohine:
Thanks @bdukes!

ohine:
Thanks for the update @ashishpd! I'm sure we'll run into issues (it is a breaking change after all). I did clean install and surface testing with platform only so far. Looking forward to the results of testing by @zyhfish.

zyhfish:
IMHO this change is acceptable, in this change all controls which inherit from telerik moved into a separate project(which means it doesn’t really remove telerik reference), all these controls are still used(so it can’t means Deprecated), maybe this is the first step, second step is remove these controls or update to inherit from some other component, but in this mean time we can’t accept this PR, as it need lots of change in both evoq and cloud projects(these changes are useless, its need change again in future when telerik really removed).

My suggestion is that we need keep these control classes in DotNetNuke.Web project but let them inherit from DotNetNuke.Web.Deprecated project, in this case we doesn't need any change in controls which used them. then we work on these controls one by one to remove the inherit and implement the control by itself or reference new 3rd party components. after all these controls updated, we can remove the DotNetNuke.Web.Deprecated project and telerik assembly, this will be the most smooth change.

jbrinkman:
@zyhfish This change is exactly what was agreed upon as the first step in the Telerik removal process.  Yes, it is a breaking change in 8.0 and that is expected.  What this means, is that we won't have a breaking change in a later DNN release where there is less expectation that things would break.  When we finish the removal of all Telerik dependencies in the platform and in Evoq, then we will make Telerik an optional install.

This approach was done as much for our 3rd party ecosystem as it was for the platform.  We want them to be intentional of their continued use of Telerik, but if they make the assembly reference change and recompile, then they will be able to deprecate Telerik from their own code at their convenience.

Our purpose in this review is not to determine whether we will "accept" this pull request, but rather to understand the scope of changes we need to make to platform and Evoq for them to continue working, and to ensure that there are no unintended side-effects.

DNN Platform/Library/Services/UserProfile/UserProfilePicHandler.cs
jbrinkman: Oliver is correct.  We should not be using response.end() as it is evil.  Instead we should be doing CompleteRequest() as it is more friendly and won't drag down server performance or fill the eventlog with useless error messages.
(http://stackoverflow.com/questions/1087777/is-response-end-considered-harmful)

ohine:
Any update @jbrinkman or @ashishpd on the status of this pull request and when it will be merged?

ashishpd:
@ohine we are going to process this PR this week. I will be working with @kanm2015 on this.

ohine:
Thanks for the update @ashishpd! Let me know if you guys run into any issues or have any questions!

PR #904, opened 2015-10-19
Support for better settings management in modules

donker:
To simplify module development we need better support for how we deal with settings. Too often developers use the bare bones hashtable that is in PortalModuleBase. This is "dirty" and can be done in a more concise way helping developers avoid errors in their code.

The code is the work of Stefan Cullman with review by Brian Dukes and myself.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
+++ b/DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
@@ -1,0 +1,99 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using DotNetNuke.Common;
+
+namespace DotNetNuke.Entities.Modules.Settings
+{
+    public class StringBasedSettings : ISettingsStore
+    {
+        Dictionary<string, string> _cache;
+        Dictionary<string, Action> _dirty;
+        Func<string, string> _get;
+        Action<string, string> _set;
+
+        /// <param name="get">Function to get a value out of your setting store</param>
+        /// <param name="set">Action to save a value back to the setting store</param>
+        public StringBasedSettings(Func<string, string> get, Action<string, string> set)
+        {
+            Requires.NotNull("Get", get);
+            Requires.NotNull("Set", set);
+
+            _get = get;
+            _set = set;
+            _cache = new Dictionary<string, string>();
+            _dirty = new Dictionary<string, Action>();
+        }
+
+        // All changes to the settings are recorded and get only executed on demand 
+        public void Save()
+        {
+            foreach (var save in _dirty.Values) save();
+            _dirty.Clear();
+        }
+
+        // CallerMemberName is used for the name of the setting. No more magic strings
+        public string Get(string @default = default(string), [CallerMemberName] string name = null)
+        {
+            Requires.NotNull("Name", name);
+
+            if (_cache.ContainsKey(name))
+            {
+                return _cache[name];
+            }
+            else
+            {
+                var value = _get(name) ?? @default;
+                _cache[name] = value;
+                return value;
+            }
+        }
+
+        public T Get<T>(T @default = default(T), [CallerMemberName] string name = null)
+        {
+            //required to behave Get and Get<string> the same
+            if (typeof(T) == typeof(string)) return (T)(object)Get((string)(object)@default, name);
+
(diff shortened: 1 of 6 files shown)


EPTamminga:
It would be great is there is a sample module that demonstrates this way of using settings.

donker:
There is here:

https://github.com/DNN-Connect/Map/blob/master/Common/ModuleSettings.cs

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
```diff
@@ -0,0 +1,99 @@
...
+            _set = set;
+            _cache = new Dictionary<string, string>();
+            _dirty = new Dictionary<string, Action>();
+        }
+
+        // All changes to the settings are recorded and get only executed on demand 
+        public void Save()
+        {
+            foreach (var save in _dirty.Values) save();
+            _dirty.Clear();
```
bdukes: I'm concerned about thread-safety with this…  Can we use something thread-safe (which would, I assume, involve removing each item as we're saving it)?

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
```diff
@@ -0,0 +1,99 @@
...
+                return value;
+            }
+        }
+
+        public T Get<T>(T @default = default(T), [CallerMemberName] string name = null)
+        {
+            //required to behave Get and Get<string> the same
+            if (typeof(T) == typeof(string)) return (T)(object)Get((string)(object)@default, name);
+
+            var converter = TypeDescriptor.GetConverter(typeof(T));
```
bdukes: it needs to be possible to override the converter

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
SCullman: Would it be enough to replace `Dictionary` with `ConcurrentDictionary`?

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
SCullman: What about passing in the convert methods instead of an converter, for example like:

        public T Get<T>(T @default = default(T), [CallerMemberName] string name = null)
        {
            //required to behave Get and Get<string> the same
            if (typeof(T) == typeof(string)) return (T)(object)Get((string)(object)@default, name);

            var converter = TypeDescriptor.GetConverter(typeof(T));
            Requires.NotNull("Converter for T", converter);

            Func<T, string> toString = d => converter.ConvertToInvariantString(d);
            Func<string, T> fromString = s => (T)converter.ConvertFromInvariantString(s);

            return Get<T>( toString, fromString,@default, name);
        }

        public T Get<T>(Func<T, string> toString, Func<string, T> fromString,T @default = default(T),  [CallerMemberName] string name = null)
        {
            //required to behave Get and Get<string> the same
            if (typeof(T) == typeof(string)) return (T)(object)Get((string)(object)@default, name);

            try
            {
                var defaultValueAsString = toString(@default);
                string value = Get(defaultValueAsString, name);
                return fromString(value);
            }
            catch
            {
                return @default;
            }
        }

Similar in `Set` Method

bdukes: Yes, that would be perfect.  @donker do you want this change as a PR to your branch, or can you integrate it directly?

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
bdukes: @SCullman I may not have been thinking correctly about how this is used.  If we don't expect it to be used in a `static` context, then that shouldn't matter.  We probably don't _want_ to encourage it to be used that way, so I'm fine with it staying as is.  Sorry for my confusion.

DNN Platform/Library/Entities/Modules/Settings/StringBasedSettings.cs
donker: I'll take care of the next steps. Someone's also preparing an alternative solution. I'll enhance this and then we can compare when the other one is ready.

jbrinkman:
This was superseded by #1024 which used an alternate approach for handling mapping settings. This pull request has been closed.

PR #906, opened 2015-10-19
DNN-7273 Added js function to handle error messages from a service an…

pberna:
…d show them in the UI. Fix checking of existing content types and content fields: now it is case insensitive. Also changed HTTP status code returned when the entity is not valid: from 500 (Internal Server Error) to 422 (Unprocessable entity)


francescorivola:
hi @ohine, correct. This is a pending change. We plan to change DCC soon and use some Exception Filter Attribute so any unhandled exception is properly logged in the Event Viewer (right now it does not) and return only a friendly message (for UI perspective) or just the http status code as part http response. As you said we should never expose server side detail to the client.

PR #907, opened 2015-10-19
DNN-6240: Fix Exception hashing

ohine:


jbrinkman:
FYI: The SqlDataProvider file is missing objectqualifier and databaseowner tags.  We have a separate PR that will fix this oversight.

ohine:
@jbrinkman sorry about that! would you like me to fix that and resubmit the fix?

PR #917, opened 2015-10-19
bug/DNN-7738 - Added CSS logic to prevent string from extending table…

mgatmaytan:
… body width when input is long string.


DNN Platform/Modules/Dnn.Modules.DynamicContentManager/ClientScripts/Manager.js
```diff
@@ -1,4 +1,18 @@
-﻿function Manager($, ko, codeEditor, settings, resx){
+﻿ko.bindingHandlers.truncatedText = {
```
francescorivola: Hi @mgatmaytan, I would try to avoid knockout binding handlers any time that we can obtain the same result with just css code. Have a look at this example https://jsfiddle.net/w3avhaon/. Let me know if you have any questions.

PR #920, opened 2015-10-20
bug/DNN-7765 - Added logic to update the selected option onclick of a…

mgatmaytan:
…dd & edit.


mgatmaytan:
Update: Found more bugs with this, will update with new PR at some point.

PR #924, opened 2015-10-21
bug/DNN-7760 - Added UI validation to set an alert if the field type …

mgatmaytan:
…selected is invalid. Tested in IE9+, FireFox & Chrome.

### DIFF
--- a/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/ClientScripts/contentTypes.js
+++ b/DNN Platform/Modules/Dnn.Modules.DynamicContentManager/ClientScripts/contentTypes.js
@@ -457,45 +457,74 @@ dcc.contentFieldViewModel = function(parentViewModel, config) {
         }
     });
 
-    self.fieldTypes = ko.computed(function () {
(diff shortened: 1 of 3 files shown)


DNN Platform/Modules/Dnn.Modules.DynamicContentManager/ClientScripts/contentTypes.js
```diff
@@ -577,6 +577,8 @@ dcc.contentFieldViewModel = function(parentViewModel, config) {
     self.saveContentField = function(data) {
         if(!validate()) {
             util.alert(resx.invalidContentFieldMessage, resx.ok);
+        }  else if (self.fieldTypeId() === "C0" || self.fieldTypeId() === "D0") {
```
francescorivola: Hi @mgatmaytan, I would try to avoid fix this at javascript level with a user warning. We should not allow the user to select the "Content Type" and "Data Type" field in the drop down list. We may can fix this problem using optgroup (see https://developer.mozilla.org/en/docs/Web/HTML/Element/optgroup)

PR #935, opened 2015-10-26
DNN-7565: Change how pages are defined in manifest.

zyhfish:


kanm2015:
Tested couple of Admin/Host modules and the new manifest worked properly as expected. Changes are good.

PR #936, opened 2015-10-26
DNN-7813: support moving files between portals

sleupold:


fjlucas:
@sleupold, there was an error in the script that breaks its execution, so I couldn't test it. Could you please fix and test it again? Once tested the filename should be renamed in order to avoid a conflict with the files in Platform, which already has a sqldataprovider file with that number. Thanks!

sleupold:
@fjlucas , I fixed the typo, please try again

sleupold:
renamed file to 08.00.00.23.sqldataprovider

fjlucas:
Hi @sleupold, the resolution of the conflict of your last commit deletes the file '08.00.00.17.sqldataprovider' in Platform/development. It must be kept since contains new store procedures that are used in other parts of Platform.

Could you please fix this and try a clean instalation to be sure that everything works? Thanks

sleupold:
fixed deleted file (adjusted formatting)
Both procedures have been tested locally.

PR #943, opened 2015-10-28
DNN-7817: Host Unable to uninstall the SQL from Host Extension.

zyhfish:


zyhfish:
DNN-7817: update manifest file in ConfigurationManager, Console, Dashboard, Sql, DCCManager modules.

PR #952, opened 2015-11-01
BUG 7858 - WebApi Doesn't work correctly with Child Portals

markmcavoy:
When using the WebApi since the request to the controller will also remove the child portal sub folder the code loads the wrong portal instance. 
This is a problem since if you have a user called UserA in the child portal www.test.com/child then unless that same username exists in a parent portal called www.test.com then the user will not be able to call the action.
When on a child portal of www.test.com/child the controller will be called on the url
ww.test.com/... hence why the wrong portal instance is then loaded in the DNN WebApi code.

This fix provides a fallback lookup based on the TabId supplied in the request header

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/DnnContextMessageHandler.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/DnnContextMessageHandler.cs
@@ -17,6 +17,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System.Globalization;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


bdukes:
You _should_ be making web API requests with the child portal folder in the URL.  Are you using the `getServiceRoot` method on `$.ServiceFramework(moduleId)` to generate the URL on the client?

markmcavoy:
@bdukes Brian, you called. I failed to take into account I needed to use the jQuery ServiceFramework extension. I think this is a non-issue and this pull request can cancelled.
Thanks for the pointer :)

jbrinkman:
This is being closed based on the comments.

PR #953, opened 2015-11-01
Fix Bug7857 - WebAPI TabIsInPortal always returns true

markmcavoy:
fix to bug 7857,  before this would always return true regardless of the
portalId.

The method GetTab will work out the correct portalId to locate the tab
if the tabId doesn't existing in the portal. Therefore we need to test
to see if the returned TabInfo object has the portalId as the expected
value.

If this pull request is accepted then its import to apply pull request #952 since without this additional pull then anyone with a child portal won't be able to call the webAPI. This is due to the code needing an additional fix to correct the fact that portalId is wrong when called via a child portal.


antoniodgonzalez:
Reviewed and merged

PR #957, opened 2015-11-02
DNN-7544: Change default robots.txt

nvisionative:
Change default robots.txt to allow client side resources

### DIFF
--- a/Website/Robots.txt
+++ b/Website/Robots.txt
@@ -113,21 +113,10 @@ Disallow: /Components/
 Disallow: /Config/
 Disallow: /contest/
 Disallow: /controls/
-Disallow: /DesktopModules/
 Disallow: /Documentation/
 Disallow: /HttpModules/
 Disallow: /Install/
 Disallow: /Providers/
-Disallow: /Resources/ContentRotator/
-Disallow: /Resources/ControlPanel/
-Disallow: /Resources/Dashboard/
-Disallow: /Resources/FeedBrowser/
-Disallow: /Resources/OpenForceAd/
-Disallow: /Resources/Search/
-Disallow: /Resources/Shared/
-Disallow: /Resources/SkinWidgets/
-Disallow: /Resources/TabStrip/
-Disallow: /Resources/Widgets/
 Disallow: /Activity-Feed/userId/	# Do not index user profiles
 
 User-agent: Yahoo Pipes 1.0


bdukes:
Go ahead and also remove `DesktopModules` and the `Resources` exclusions, and I'll merge this in.

Thanks!

nvisionative:
@bdukes - done, thanks!

EPTamminga:
I am not sure if I (users) want to open the portals folder for bots by default. 
Is that one of the objectives of ths PR?

nvisionative:
## Per JIRA (DNN-7544)

**Background**
Google has changed the way that it crawls pages, so that it can properly render and understand the content in a page.

Part of this is accessing the js, css and image files that go into many webpages, particularly if they use a delayed load.

As per the official guidelines from Google Webmaster tools:

> For optimal rendering and indexing, our new guideline specifies that you should allow Googlebot access to the JavaScript, CSS, and image files that your pages use. This provides you optimal rendering and indexing for your site. Disallowing crawling of Javascript or CSS files in your site’s robots.txt directly harms how well our algorithms render and index your content and can result in suboptimal rankings.

The existing standard robots.txt file by default blocks access to specific DNN/CSS paths.

**Change**

* **Remove** the following lines from the googlebot section of the robots.txt file:

```
Disallow: /js/
Disallow: /Portals/
Disallow: /images/
```

This is a simple change which will help with Googlebot crawling Dnn sites still using the default robots.txt file.

bdukes:
@EPTamminga this change only affects Googlebot, because of the above recommendation from Google Webmaster Tools.

EPTamminga:
@bdukes I understand and understood that (and I knew the changes in Googlebot), but... my idea is that I do NOT want the portal folders to be indexed/crawled by default. There can be a lot of files in de portal folders that are not to be indexed, e.g. old pdf files, out of date images, old word and powerpoint presentations, etc. Or will these file not be indexed if there is no link anywhere in the site?

bdukes:
Right, a crawler isn't going to go around guessing at URLs, so if a file isn't linked to, it won't be indexed (and certainly won't be indexed _highly_).  Disallowing indexing the `Portals` folder means that the popular, up-to-date PDF that everyone is linking to won't get included in Google, which seems like the bigger problem, to me.

nvisionative:
Well stated.  That's the way I understand it.

PR #959, opened 2015-11-02
Research/speed dnn 7131

pberna:


pberna:
This pull request changes have been already merged. Renaming the sql data provider file was needed to match the current numbers.

PR #980, opened 2015-11-09
DNN-7875: Remove Advanced Settings.

zyhfish:


DNN Platform/Website/Templates/Default Website.template
```diff
@@ -2814,93 +2814,6 @@
         </pane>
       </panes>
     </tab>
-    <tab sku="EE,PE,CE">
-      <containersrc />
-      <description>[Tab.AdvancedSettings.Description.Text]</description>
```
francescorivola: Should we may remove the Description and Title keys from the template resource file.

PR #995, opened 2015-11-13
DNN-7913: Fixed login redirect issue on portal with specified login page

kanm2015:


kanm2015:
Closed this PR since there's a redirect loop issue.

PR #996, opened 2015-11-14
DNN-7927 making htmlhelper available to the rest of us

donker:


donker:
Fkin git did a lot more than what I proposed ...

PR #1005, opened 2015-11-17
Complete the fix for DNN-6758

sleupold:
re-submitted against the proper branch


sleupold:
oops, I think, I need to recreate my branch

PR #764, opened 2015-08-31
complete fix for DNN-6758

sleupold:
update stored procedures, which are using the columns with modified size.
Using proper size for CultureCode as well.

PR #1006, opened 2015-11-17
complete fix for DNN-6758

sleupold:
re-submitting PR #764 against the correct branch now


sleupold:
Update stored procedures, which are using the columns with modified size (using proper size for CultureCode as well).

PR #1013, opened 2015-11-18
Simplify fix Issues in DNN-7410

sleupold:
Simplify and fix novice issues in updated proc version, avoiding unnecessary comparisons and type casts.

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/08.00.00.05.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/08.00.00.05.SqlDataProvider
@@ -6,24 +6,17 @@ IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{
 GO
 
 CREATE PROCEDURE {databaseOwner}{objectQualifier}GetSchedule
- @Server varchar(150)
+ @Server nVarChar(150)
 AS
 BEGIN
-SELECT
-  S.*
-  , (SELECT max(S1.NextStart)
-   FROM {databaseOwner}{objectQualifier}ScheduleHistory S1
-   WHERE S1.ScheduleID = S.ScheduleID) as NextStart
- FROM {databaseOwner}{objectQualifier}Schedule S
- WHERE
-	(
-		@Server IS NULL OR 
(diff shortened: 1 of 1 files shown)


Website/Providers/DataProviders/SqlDataProvider/08.00.00.05.SqlDataProvider
```diff
@@ -6,24 +6,17 @@ IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{
 GO
 
 CREATE PROCEDURE {databaseOwner}{objectQualifier}GetSchedule
- @Server varchar(150)
+ @Server nVarChar(150)
```
pberna: Good point the change from varchar to nvarchar

Website/Providers/DataProviders/SqlDataProvider/08.00.00.05.SqlDataProvider
```diff
@@ -6,24 +6,17 @@ IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{
...
-		S.Servers LIKE '%,' + @Server OR
-		S.Servers = @Server OR
-		S.Servers IS NULL)
-	ORDER BY FriendlyName ASC
+    SELECT S.*, 
+           (SELECT max(S1.NextStart)
+             FROM  {databaseOwner}{objectQualifier}ScheduleHistory S1
+             WHERE S1.ScheduleID = S.ScheduleID) AS NextStart
+     FROM  {databaseOwner}{objectQualifier}Schedule S
+     WHERE IsNull(S.Servers, N'') = N''
```
pberna: The previous Where condition deals with different situations: server name stored in Schedule.Servers could not have commas, a comma before the name, a comma after the name or both commas. With this change we only can handle the server name with no commas at all or with both commas before and after the name. The other solution is more robust.

sleupold: Pedro, this is not correct, as I am now adding a Komma before and at the end of the servers list, so each server is embraced by commas and will be found. there is no part-name match anymore and this option is as stable as the previous one.

PR #1015, opened 2015-11-19
DNN-6719: HMAC attribute enhancements (refactoring, optimization, logging)

galatrash-at-dnn:
I reviewed the “DnnModuleAuthorizeAttribute” and had to made few changes (refactoring, optimization, and logging). From my previous experience, working with any security scheme without any logging is a nightmare. Therefore, I added some logging at TRACE level to allow developers to see why their calls fail when these happen.

Also, I change the time-stamp timeout to be 5 minutes instead of not 5 seconds because the server time could have more time drift than the 5 seconds. From a security perspective, this is not an issue as we have the NONCE caching to take care of this. Moreover, I set the key to be “APPID + NONCE”. This is extremely necessary to allow any service to server more than one user at the same time if NONCE was repeated by different clients (the clients may reset the NONCE every time they restart).

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/HMACAuthenticationAttribute.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/HMACAuthenticationAttribute.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Net;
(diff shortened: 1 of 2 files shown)


agrueso:
Merged before removing HAMC from Platform. See issue DNN-8058 for more details.

PR #1024, opened 2015-11-24
DNN-7751 Improve Settings Management API

jbrinkman:
Currently, when saving/retrieving values for module settings, tab module settings or portal settings, a module will only have a Dictionary of string values for the settings.  Every module must include code for serializing or deserializing these values into the appropriate value or reference type for use in their code.  This change provides a simple method for using attributes to define how settings should be stored and retrieved.  This greatly simplifies usage of settings within a module.  For example this code:
```
if (ModuleSettings["HtmlText_ReplaceTokens"] != null)
{
    chkReplaceTokens.Checked = Convert.ToBoolean(ModuleSettings["HtmlText_ReplaceTokens"]);
}
```
can be replaced with this code:
```
chkReplaceTokens.Checked = ModuleSettings.ReplaceTokens;
```

**Note:** This code was originally contributed by Peter Donker as a replacement for #904 I have done the review and will merge this contribution even though it appears that I created the pull request.


svdv71:
Joe, Peter; Great job! Thanks :)

PR #1026, opened 2015-11-24
DNN-6809 - Added meaningful error messages when following user.

mgatmaytan:


DNN Platform/Modules/CoreMessaging/App_LocalResources/View.ascx.resx
```diff
@@ -459,4 +459,10 @@
   <data name="DNNCorp_Wiki_Page.Text" xml:space="preserve">
     <value>Wiki Subscriptions</value>
   </data>
+  <data name="AlreadyFollowingUser.Text" xml:space="preserve">
```
francescorivola: @mgatmaytan, are you using these two keys in the code? I cannot find the usage of these new two keys.

Website/App_GlobalResources/Exceptions.resx
```diff
@@ -211,7 +211,7 @@
     <value>Target User '{0}' has Blocked Relationship '{1}' from Target User '{2}'.</value>
   </data>
   <data name="UserRelationshipExistsError.Text" xml:space="preserve">
-    <value>Relationship already exists for Initiating User '{0}' Target User '{1}' RelationshipID '{2}'.</value>
+    <value>You are already following this user.</value>
```
francescorivola: @mgatmaytan , this resource key was using a string.format values. You have changed the string removing the {} but I did not see in your commit the change in where this key is used in the code to adapt it to this change. Can you please review it? thks

PR #1028, opened 2015-11-24
DNN-7928 Improvements to Client Resource Manager

donker:
This PR will add features to the Client Resource Manager so module and skin developers can keep their files from clashing in a much easier way. It will also allow skin developers to opt out of default.css.


DNN Platform/Components/ClientDependency/Source/DependencySorter.cs
```diff
@@ -10,6 +10,30 @@ namespace ClientDependency.Core
...
+
+        public static IList<IClientDependencyFile> FilterDependencies(IList<IClientDependencyFile> dependencies)
+        {
+            if (dependencies.Any(f => f.Name != ""))
+            {
+                var newList = dependencies.Where(f => f.Name == "").ToList();
+                var frameworks = dependencies.Where(f => f.Name != "").GroupBy(f => f.Name.ToLower());
+                foreach (var framework in frameworks)
+                {
+                    var topPriority = framework.FirstOrDefault(d => d.ForceVersion);
```
bdukes: What happens if there are multiple files marked as `ForceVersion`?  Should this sort by `ForceVersion` then by `Version`?

DNN Platform/Components/ClientDependency/Source/DependencySorter.cs
```diff
@@ -10,6 +10,30 @@ namespace ClientDependency.Core
     /// </summary>
     internal static class DependencySorter
     {
+
+        public static IList<IClientDependencyFile> FilterDependencies(IList<IClientDependencyFile> dependencies)
+        {
+            if (dependencies.Any(f => f.Name != ""))
+            {
+                var newList = dependencies.Where(f => f.Name == "").ToList();
+                var frameworks = dependencies.Where(f => f.Name != "").GroupBy(f => f.Name.ToLower());
```
bdukes: This should probably use `ToLowerInvariant()`

DNN Platform/DotNetNuke.Web.Client/ClientResourceManager.cs
```diff
@@ -305,7 +312,21 @@ public static void RegisterScript(Page page, string filePath, FileOrder.Js prior
...
+        /// <summary>
+        /// Requests that a JavaScript file be registered on the client browser
+        /// </summary>
+        /// <param name="page">The current page. Used to get a reference to the client resource loader.</param>
+        /// <param name="filePath">The relative file path to the JavaScript resource.</param>
+        /// <param name="priority">The relative priority in which the file should be loaded.</param>
+        /// <param name="provider">The name of the provider responsible for rendering the script output.</param>
+        /// <param name="name">Name of framework like Bootstrap, Angular, etc</param>
+        /// <param name="version">Version nr of framework</param>
+        public static void RegisterScript(Page page, string filePath, int priority, string provider, string name, string version)
```
bdukes: Is there a reason that `version` would be a `string` instead of a `System.Version` object?

DNN Platform/Components/ClientDependency/Source/DependencySorter.cs
donker: This property is not available through the Client Resource Manager methods, but only in the control. This effectively makes it only available to skin developers. So you're down to one person.

If someone decides to "hack" or abuse the feature, the result is unpredictable. But at the end of the day it's down to instruction: this is only meant for Skin designers, not the modules.

bdukes: ​I use the various JS & CSS controls in my module all the time.  I prefer
to keep my views as declarative as possible.​

bdukes: It's not that big of a deal, and we should definitely make sure that the
documentation says that it's only for skin developers.  However, making it
do the (as much as possible) right thing in that scenario isn't difficult,
so why not?

donker: I'm not sure I follow. But I'm not very particular about this point. I do feel strongly about the use case that a skin dev should have the possibility to say: "ok, we're using this version (and file) of this framework". For the simple reason that the skin designer is the boss in the UI.

PR #1029, opened 2015-11-25
DNN-4282 - Added SenderDisplayName to use as title and alt text.

mgatmaytan:


DNN Platform/Modules/CoreMessaging/Services/MessagingServiceController.cs
```diff
@@ -248,6 +248,10 @@ public HttpResponseMessage Notifications(int afterNotificationId, int numberOfRe
 
                 foreach (var notification in notificationsDomainModel)
                 {
+                    var userObj =
```
francescorivola: you could call this just "user"

DNN Platform/Modules/CoreMessaging/Services/MessagingServiceController.cs
```diff
@@ -248,6 +248,10 @@ public HttpResponseMessage Notifications(int afterNotificationId, int numberOfRe
...
+                        UserController.Instance.GetUser(PortalSettings.PortalId, notification.SenderUserID);
+                    var displayName = "";
```
francescorivola: you could do this instruction and the above in the same line.

DNN Platform/Modules/CoreMessaging/ViewModels/NotificationViewModel.cs
```diff
@@ -31,6 +31,7 @@ public class NotificationViewModel
         public string Body { get; set; }
         public string SenderAvatar { get; set; }
         public string SenderProfileUrl { get; set; }
+        public string SenderDisplayName { get; set; }
```
francescorivola: Did you check all the place where this ViewModel is used?

PR #1039, opened 2015-11-30
DNN-6506 Replace profilepic with DNNImageHandler

jbrinkman:
Update DNN to use the new DNNImageHandler for displaying the profile picture.


Website/development.config
```diff
@@ -102,6 +102,7 @@
       <add name="UserProfilePageHandler" path="User.aspx" verb="*" type="DotNetNuke.Services.UserProfile.UserProfilePageHandler, DotNetNuke" preCondition="integratedMode"/>
       <add name="RadProgressHandler" verb="*" path="Telerik.RadUploadProgressHandler.ashx" type="Telerik.Web.UI.Upload.RadUploadProgressHandler, Telerik.Web.UI" preCondition="integratedMode"/>
       <add name="UserProfilePicHandler" path="ProfilePic.ashx" verb="*" type="DotNetNuke.Services.UserProfile.UserProfilePicHandler, DotNetNuke" preCondition="integratedMode" />
```
francescorivola: Should we remove UserProfilePicHandler from development.config and release.config? It is currently removed in the upgrade scenario based on the 08.00.00.26.config.

DNN Platform/Library/Services/GeneratedImage/IPCount.cs
```diff
@@ -83,14 +88,13 @@ static IPCount()
...
 
-        internal IPCount()
-        {
-            if (CachePath != null && !Directory.Exists(CachePath))
-            {
-                Directory.CreateDirectory(CachePath);
-            }
-        }
-
+        //internal IPCount()
```
francescorivola: we should remove this commented code.

PR #1041, opened 2015-11-30
Feature/dnn 8026

pberna:


DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
```diff
@@ -63,23 +63,16 @@ public void SaveSettings(ModuleInfo moduleContext, T settings)
...
                         {
-                            if (settingValue is DateTime)
-                            {
-                                settingValueAsString = ((DateTime)settingValue).ToString("u");
-                            }
-                            else
-                            {
-                                settingValueAsString = settingValue.ToString();
-                            }
+                            settingValueAsString = GetSettingValueAsString(settingValue);
```
bdukes: Should this just call `Convert.ToString(settingValue, CultureInfo.InvariantCulture)`, to cover every type that implements `IConvertible` or `IFormattable`?  The only difference that I see is that the date would be formatted differently, but so long as it's consistently the same culture, it shouldn't matter, right?

DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
```diff
@@ -228,6 +252,25 @@ private void DeserializeProperty(T settings, PropertyInfo property, ParameterAtt
                             Exceptions.LogException(exception);
                         }
                     }
+                } else if (propertyType.FullName == "System.TimeSpan")
```
bdukes: Why is this a string comparison, instead of `propertyType is TimeSpan` or `propertyType == typeof(TimeSpan)`?

donker: Not my code. So I can't say. This was as it was sent over. Feel free to change.

DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
donker: Right ... it's all about culture. So as soon as the thread culture changes you're in big trouble. This can happen for many reasons. 

This code is mine. The only robust method serializing dates is to use the universal format. I was disappointed that the original code just relied on the core to serialize. This leads to very awkwardly formatted dates which will fail in deserialization when the locale changes. That is not good behavior as the (de)serialization needs to be culture agnostic.

bdukes: So your concern is with code trying to deserialize a date outside of the `SettingsRepository`?  Sounds reasonable, just wanting to make sure I've got the scope of your concern.  `Convert.ToString` with `InvariantCulture` would be fine if everything went through the `SettingsRepository` during retrieval, yes?

bdukes:
@pberna I'm looking at some changes to simplify the complexity of the serialization/deserialization code, as well as adding tests in different cultures to make sure that we're being properly invariant in our conversions.  I'll push it to another branch for you to look at, unless you want me to just push the changes here

bdukes:
See my changes in the (feature/DNN-8026-tweaks)[https://github.com/dnnsoftware/Dnn.Platform/compare/feature/DNN-8026...feature/DNN-8026-tweaks] branch

DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
donker: Nono. It's about the app starting up in a different locale. You can't imagine how unsure you can be of the thread locale in multilingual sites. It really depends on who's logged in, etc.

bdukes: But when I'm specifying `InvariantCulture` then it doesn't matter what locale the site is executing under, unless I'm missing something.

bdukes:
@pberna I just pushed 545b3f59ab3ebda4cb8eb2f49b6bd92adbbe786b to that branch, which fixes the serializer support

pberna:
Hi, @bdukes , I have reviewed your changes and they look very good! The simplification of the serialization/deserialization code is a good improvement and your new tests cover areas that were not tested. I have merged your changes into this pull request to process them all together,
Good job and thank you for your efforts!

ashishpd:
It's really awesome to see unit tests being added with the new APIs. Ideally we'd want unit tests to be submitted as part of PR itself ;)

PR #1045, opened 2015-12-01
DNN-7894: Fixed the query string language code issue on language editor

kanm2015:


kanm2015:
Will be implemented in a different way

PR #1047, opened 2015-12-01
Feature/dnn 8026 tweaks

pberna:


pberna:
Added improvement from bdukes into dnn-8026. Very good job!

PR #1051, opened 2015-12-01
DNN-8048 Increment CRM version on package upgrade

bdukes:


zyhfish:
This PR is no valid as it won't update CRM version when use repair install option, i will submit another one for this item.

PR #1063, opened 2015-12-04
DNN-7693: Made the downloaded zip file named same as the language pac…

kanm2015:
…k name


kanm2015:
It'll be implemented in a better way.

PR #1068, opened 2015-12-07
DNN-8091 portal settings don't support empty values

bdukes:


bdukes:
The root cause of this issue is https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Entities/Portals/PortalController.cs#L3019, which returns the default value is the stored `string` is `null` _or `string.Empty`.  Since `string.Empty` is a valid value, and the most reasonable value to store is `null` is desired for a nullable value type, this makes it impossible to retrieve that value through this API.  I'd guess it's too dangerous to change that API now to be consistent, but if consistency is more desired, that method should be fixed. /cc @ashishpd

ashishpd:
Thanks for the PR @bdukes, I've assigned to @pberna for reviewing and merging. He will process this on Wednesday (he's away for 2 days).

PR #1069, opened 2015-12-07
DNN-7900 Fixed issues for trailing and leading white spaces

amarjit-dhunna:
DNN-7900 Fixed issues for trailing and leading white spaces in security roles


amarjit-dhunna:
Closing because of conflict. Will commit in a different branch

PR #1079, opened 2015-12-08
DNN-8093 Fixed the bug for missing controller name

amarjit-dhunna:
Fixed the bug for missing controller name in case of Action(actionName) overload.

Now the system will pick the current controller name in case it is not provided.


amarjit-dhunna:
Will submit new one with naming conventions changes.

PR #1101, opened 2015-12-11
Bug/dnn 7693

kanm2015:


kanm2015:
Will improve the performance and submit another pull request later

PR #1107, opened 2015-12-14
DNN-8147 - Adjusted CSS to allow overflow x scrolling on longer /…

mgatmaytan:
… more rooted folders.


mgatmaytan:
Marked as won't fix.

PR #1114, opened 2015-12-15
DNN-7861 - Update Platform Template Content

francescorivola:


agrueso:
Changes have been done into the resources file. I have seen the final result and look ok.

PR #1131, opened 2015-12-17
DNN-8162 fix assembly metadata

bdukes:
This pulls in the shared `SolutionInfo.cs` file and updates the `AssemblyInfo` to match other projects.


bdukes:
I'm going to resubmit with a different ID

PR #1136, opened 2015-12-17
DNN-8166 Add package to use latest .NET compilers

bdukes:


bdukes:
I added this as a standalone package, I'm not sure if it would be better as an Upgrade script instead

PR #1147, opened 2015-12-21
DNN-8134 : Added custom ValidateAntiForgeryToken for validation

amarjit-dhunna:


amarjit-dhunna:
Going to submit new one with service framework support

PR #1159, opened 2015-12-22
DNN-8078: Restricted the max date assignment to endDatePicker.MaxDate

amarjit-dhunna:
because of raddatetimepicker max date restrictions.


amarjit-dhunna:
Will submit new one with better changes

PR #1161, opened 2015-12-23
DNN-8196: Fixed the duplicate context key issue in MembershipModule.A…

kanm2015:
…uthenticateRequest


ashishpd:
It's good to have these safety checks in the system, though I am not sure about the root cause of this issue.

PR #1162, opened 2015-12-23
DNN-8125 - Update Newtonsoft to the latest version - revert

fjlucas:


bdukes:
Why's this getting reverted?

ashishpd:
There is no need to revert this as all the Packages are working pretty good with the latest Newtonsoft. @bdukes we have multiple issues in merging with latest Newtonsoft. The purpose of this PR was to keep the option of reverting handy in case we needed to bail out. This is no longer needed.

PR #1170, opened 2015-12-29
Fix for DNN-7731

juvander:
Should fix this https://dnntracker.atlassian.net/browse/DNN-7731

Note that in 8.0 branch there is also similar issue waiting few lines up from this.


kanm2015:
@juvander, it appears your change won't work with the case that there're multiple login controls on a single page. Will not merge it. I think we can keep the code block but make it start with "<%#" instead of "<%=" so that the CLR won't complain.

kanm2015:
I'll submit a new PR to fix this issue.

PR #1178, opened 2015-12-30
Feature/moniker api

galatrash-at-dnn:


ashishpd:
Amarjit has reviewed the code already.

PR #1185, opened 2016-01-04
DNN-8050 The comments will be injected when debug is enabled

amarjit-dhunna:
either from web.config or from host settings.

### DIFF
--- a/DNN Platform/Components/ClientDependency/Source/Controls/ClientDependencyInclude.cs
+++ b/DNN Platform/Components/ClientDependency/Source/Controls/ClientDependencyInclude.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Web.UI;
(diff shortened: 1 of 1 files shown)


bdukes:
These comments are required for correct operation, this commit needs to be reverted.  See [`ModuleHost.RestoreCachedClientResourceRegistrations`](https://github.com/dnnsoftware/Dnn.Platform/blob/115ae75da6b152f77ad36312eb76327cdc55edd7/DNN%20Platform/Library/UI/Modules/ModuleHost.cs#L448)

ohine:
The feature @bdukes mentioned was added in pull request https://github.com/dnnsoftware/Dnn.Platform/pull/324 and jira [DNN-6661](https://dnntracker.atlassian.net/browse/DNN-6661)

bdukes:
well, it was added for JS libraries in #324, it had already worked for direct CDF requests

PR #1191, opened 2016-01-05
DNN-6973: if the authentication cookie set by cookie slide in membership, then use SignIn method instead if current portal is in portal group.

zyhfish:


fjlucas:
Code review done.

PR #1192, opened 2016-01-05
CONTENT-5617: fix for invalid start/end date in PageSettings popup.

galatrash-at-dnn:


ashishpd:
Please submit a different PR with a Jira id "DNN-xxx".

PR #1207, opened 2016-01-08
DNN-8061: add support for condition attribute of xml merge.

zyhfish:


galatrash-at-dnn:
Closing without merging based on comments on Jira [DNN-8061](https://dnntracker.atlassian.net/browse/DNN-8061)

PR #1223, opened 2016-01-11
DNN-8295 - Reviewed & updated modal positioning.

mgatmaytan:
Keeps track of initial placement of modal - to be able to revert back.
Maximizing positions modal top left.
Changed to be fixed positioning rather than absolute.


Website/js/Debug/dnn.modalpopup.js
```diff
@@ -201,29 +201,41 @@
...
                         var $window = $(window),
                             newHeight,
-                            newWidth;
+                            newWidth,
+                            newTop,
+                            newLeft;
 
                         if ($modal.data('isMaximized')) {
-                            newHeight = $modal.data('height');
+                            newHeight = $modal.data('height') + 100;
```
francescorivola: That's very weird. We should investigate a bit why we need to sum 100 px in this scenario. In case we cannot fix this using the jQuery UI API I would at least store the 100 pixel in a variable and put a comment explaining this value. i.e.: var JQUERY_UI_HEIGHT_SRINK_OFFSET. Leaving this value without a named variable make the code very hard to understand.

Website/js/Debug/dnn.modalpopup.js
```diff
@@ -201,29 +201,41 @@
...
+
+                            newTop = "2%";
+                            newLeft = "1%";
+
                             $modal.data('isMaximized', true);
                         }
 
                         $modal.dialog({ height: newHeight, width: newWidth });
-                        $modal.dialog({ position: { my: "center", at: "center" } });
+                        $modal.parent().css({ "top": newTop, "left": newLeft, "position": "fixed" });
```
francescorivola: I have some concern about this fix. We should stick to the jQuery UI API as much as possible. I did a quick test and changing the calls in this way works to me in IE and Chrome:
    $modal.dialog( "option", "height", newHeight);
    $modal.dialog( "option", "width", newWidth);
    $modal.dialog( "option", "position", { my: "center", at: "center", of: window });
see http://api.jqueryui.com/dialog/#option-position

With this solution we do not need store the initial top and left + change the position to fixed. I do not see why we want have fixed here, we can talk on this today. But IMO we should always have dialog fixed or absolute and try to do not change it only when maximize or minimize.

Website/js/Debug/dnn.modalpopup.js
```diff
@@ -201,29 +201,41 @@
...
                             newWidth = $modal.data('width');
                             $modal.data('isMaximized', false);
+
+                            newTop = initialTop;
+                            newLeft = initialLeft;
+
                         } else {
                             $modal.data('height', $modal.dialog("option", "minHeight"))
-                                .data('width', $modal.dialog("option", "minWidth"))
-                                .data('position', $modal.dialog("option", "position"));
```
francescorivola: Good catch on removing this dead code

PR #1224, opened 2016-01-11
DNN-8279: Fixed double notifications showing up issue when user click…

kanm2015:
…ing the notification tab


kanm2015:
Will fix this issue in a different way

PR #1232, opened 2016-01-13
DNN-8312: merge incremental config files if version is incremental version.

zyhfish:


francescorivola:
The current fix only covers the install/upgrade wizard but do not the auto-install. I am working in a new PR to fix the issue properly avoiding process twice the internal config files.

francescorivola:
Closed PR. A new one has been provided https://github.com/dnnsoftware/Dnn.Platform/pull/1235

PR #1235, opened 2016-01-13
DNN-8312 - Changed code done in DNN-6919 to ensure that incremental c…

francescorivola:
…leanup and config are executed in the right step. Currently the incrementals are executed inside the scripts step.

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/Upgrade.cs
+++ b/DNN Platform/Library/Services/Upgrade/Upgrade.cs
@@ -4062,16 +4062,22 @@ namespace DotNetNuke.Services.Upgrade
         /// -----------------------------------------------------------------------------
         public static string DeleteFiles(string providerPath, Version version, bool writeFeedback)
         {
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -4062,65 +4062,22 @@ public static void DeleteInstallerFiles()
...
-                    }
-                }
-                catch (Exception exc)
-                {
-                    Logger.Error(exc);
-                }
-            }
-
-            if (writeFeedback)
+            var stringVersion = GetStringVersion(version);
```
amarjit-dhunna: I think we should use new method or overloaded version of this method (or may be just optional parameters) so that there we don't have the check revision and add it separately at each individual usage. Moreover, someone doing some implementation in future might miss that revision check if it is to be done separately. It would be easy for him to compare the two overloads and use them accordingly.

PR #1242, opened 2016-01-14
DNN-8331 - DnnImageHandlers > ProfilePic mode does not return the noA…

francescorivola:
…vatar and Log an NullReferenceException for no existing user Id


antoniodgonzalez:
Reviewed and merged

PR #1243, opened 2016-01-14
DNN-8273 - DAM - file move failed from Amazon S3 to Database folder type

fjlucas:
the position of the stream is only used when the steam is seekable


pberna:
Code review done

PR #1248, opened 2016-01-15
DNN-8337 Fixed the pixel format.

amarjit-dhunna:


francescorivola:
Hi @amarjit-dhunna , it is not clear to me why in the CopyImage method we generate a Graphics from destImage to write into srcImage, then we return the destImage.

Finally, could we move the CopyImage method in an helper and use that everywhere is needed in the dnnImageHandler instead to have this only in the UserProfilePicture Transformation. What do you think? Let's talk today about this, I can show you where we could apply this method as well.

PR #1256, opened 2016-01-18
DNN-8355: Add a message to module that's referenced on other pages

bdukes:
[DNN-8355](https://dnntracker.atlassian.net/browse/DNN-8355)

> Like with the "Shared Module" indicator in module actions for cross-site sharing withing a portal group, site administrators need to know if the module they're going to edit is shared by other pages on the same site.
> ### Steps to Reproduce
> 1. Add a module to a page
> 2. Create a new page
> 3. Select _Add Existing Module_ from the _Modules_ menu
> 4. Choose the page from step 1
> 5. Ensure _Make a Copy_ is unchecked
> 6. Choose the module from step 1 and add it to a pane
> ### Expected Results
> - There is an indication that the module is shared with the other page
> ### Actual Results
> - There is no indication

### DIFF
--- a/DNN Platform/Library/UI/Containers/Container.cs
+++ b/DNN Platform/Library/UI/Containers/Container.cs
@@ -36,6 +36,7 @@ using DotNetNuke.Entities.Host;
 using DotNetNuke.Entities.Modules;
 using DotNetNuke.Entities.Modules.Actions;
 using DotNetNuke.Entities.Portals;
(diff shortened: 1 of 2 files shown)


galatrash-at-dnn:
We decided to go with PR #1257 and not this for issue DNN-8355. Therefore, I'm closing this without merging it.

galatrash-at-dnn:
Replace with PR #1257

PR #1257, opened 2016-01-18
DNN-8355: Show "Shared Module" indicator for cross-page module sharing

bdukes:
This is an alternative implementation of #1256.

[DNN-8355](https://dnntracker.atlassian.net/browse/DNN-8355)

> Like with the "Shared Module" indicator in module actions for cross-site sharing withing a portal group, site administrators need to know if the module they're going to edit is shared by other pages on the same site.
> ### Steps to Reproduce
> 1. Add a module to a page
> 2. Create a new page
> 3. Select _Add Existing Module_ from the _Modules_ menu
> 4. Choose the page from step 1
> 5. Ensure _Make a Copy_ is unchecked
> 6. Choose the module from step 1 and add it to a pane
> ### Expected Results
> - There is an indication that the module is shared with the other page
> ### Actual Results
> - There is no indication

### DIFF
--- a/Website/admin/Menus/ModuleActions/ModuleActions.css
+++ b/Website/admin/Menus/ModuleActions/ModuleActions.css
@@ -176,18 +176,13 @@ div.actionMenu ul.dnn_mact > li.actionQuickSettings div.qsFooter a.secondarybtn 
 }
 div.actionMenu ul.dnn_mact li.dnn_shared {
   display: inline-block;
-  width: 20px;
-  height: 100px;
-  background-color: #ffffff;
 }
 div.actionMenu ul.dnn_mact li.dnn_shared > div {
   display: inline-block;
-  width: 20px;
-  height: 20px;
-  writing-mode: tb-rl;
-  -webkit-transform: rotate(90deg);
-  -moz-transform: rotate(90deg);
-  -o-transform: rotate(90deg);
+  padding: 0.5em 0;
+  -webkit-writing-mode: vertical-rl;
+  -ms-writing-mode: tb-rl;
+  writing-mode: vertical-rl;
   white-space: nowrap;
   color: white;
   background-color: #000;
--- a/Website/App_GlobalResources/SharedResources.resx
+++ b/Website/App_GlobalResources/SharedResources.resx
@@ -1962,4 +1962,7 @@ If you expect this addition, then just ignore this email; otherwise, an immediat
   <data name="RestrictFileMail_Subject.Text" xml:space="preserve">
     <value>Potentially dangerous file added to your website</value>
   </data>
+  <data name="ModuleShared.Text" xml:space="preserve">
+    <value>Shared Module</value>
+  </data>
 </root>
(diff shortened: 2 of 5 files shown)


tpluscode:
Also, please fix the changes to `Website/App_GlobalResources/SharedResources.resx`. It looks like line ending and encoding broke the diff and entire file is marked as changed

bdukes:
@tpluscode, I've rebased and fixed the whitespace problem

@galatrash-at-dnn unfortunately, that's the state of the shared module indicator in DNN Platform right now (you'll notice that my change doesn't implement that piece, it just reuses it).  That's how it looks right now if you use cross-site module sharing (e.g. if you manually create a portal group in the database or via API). Evoq uses a different technique to display that (a message at the top of the module, actually similar to the other implementation of DNN-8355, #1256)

I believe it's supposed to display with vertical text, but the CSS doesn't appear to have survived the years since it was implemented.  I can try to spend to some getting it turned around again, assuming that's truly how we want it to display.

bdukes:
Looks like that style was added in ec0cf96e1474ec6bb227a68b80991ec8563dd40a, for [CONTENT-1766](https://dnntracker.atlassian.net/browse/CONTENT-1766), FYI

bdukes:
@galatrash-at-dnn I've pushed a fix to the CSS for that shared item in the module actions, so that the orientation doesn't appear broken.

I discussed this with a colleague and we thought it would be nice if it appeared more as a standard module action, perhaps with an icon like http://fontawesome.io/icon/share/, and it displayed a small dialog (like with quick settings) with a statement about the module being shared, a link to the "master module" and either a report of which pages/sites it's shared on or a link to that section in the module settings.

That's probably more than I can tackle this week, but maybe an enhancement to log in JIRA.

Website/admin/Menus/ModuleActions/ModuleActions.css
```diff
@@ -178,16 +178,13 @@ div.actionMenu ul.dnn_mact li.dnn_shared {
...
 }
 div.actionMenu ul.dnn_mact li.dnn_shared > div {
   display: inline-block;
   width: 20px;
-  height: 20px;
-  writing-mode: tb-rl;
   -webkit-transform: rotate(90deg);
-  -moz-transform: rotate(90deg);
-  -o-transform: rotate(90deg);
+  transform: rotate(90deg);
```
galatrash-at-dnn: The above change (the two lines) are still not producing the expected results. I suggest replacing both with
``text-orientation: sideways;``

galatrash-at-dnn:
@bdukes, not sure we need that big change now. the only thing I recommend now is changing the css as I wrote in my comment (inline inside the code).

Website/admin/Menus/ModuleActions/ModuleActions.css
bdukes: It's the removal of `writing-mode` that causes it to display correctly for me (in Firefox, Edge, IE, and Chrome).  `text-orientation: sideways;` doesn't have support in IE (http://caniuse.com/#feat=css-text-orientation), and doesn't have any effect in Firefox in my testing.  Let me know if I'm missing something.

galatrash-at-dnn:
@amarjit-dhunna will you please validate from your end. @bdukes's tests and mine are conflicting and we need a third opinion. Amarjit, I'll put this issue in your court now :)

amarjit-dhunna:
@bdukes , The changes look good. Only issue I see is that the Shared Module text is not fully visible. The height should probably auto expand because the Shared Module text might be smaller length or longer in some other language.
dnn-8355

bdukes:
@amarjit-dhunnam, I've pushed a change to the styles (I also rebased on `development`).  I'm now using `writing-mode` instead of `transform` to do the rotation, so that we don't need explicit sizes.  I also added some padding.  Let me know if you see anything else you'd like adjusted

kanm2015:
looks good now, PR merged

PR #1262, opened 2016-01-19
DNN-5626 Fixed bug introduce with the previous commit: when the user …

pberna:
…clicked on an already selected radio button it was unselected


antoniodgonzalez:
Reviewed and merged

PR #1267, opened 2016-01-20
DNN-8348 Fixed the issue by removing MVC cache and using own cache key

amarjit-dhunna:
The issue was happening due to MVC cache key. Since MVC is designed to work on single site architecture and does not expect multiple modules and views folders.
The cache key has been modified to include the namespace in it so that each module is treated differently.
Also the call render module has been moved from onload to oninit because in onload case the if the page has different modules and then the view locations are overwritten to the last module at the time of onload. So all the other modules are not found and throws exception


fjlucas:
Code review done and merged

PR #1272, opened 2016-01-21
Allow locking site/instance DNN-8374

bdukes:
[DNN-8374](https://dnntracker.atlassian.net/browse/DNN-8374)

> We've made some customizations for a client which we think would be a good fit for inclusion in the core. There's a "Lock Site" action added to the control bar, which allows a super-user to turn off modifications to the site (probably for entering some sort of maintenance mode). For our client, the "good enough" way this is implemented is to disable the module actions whenever the site is locked (in addition, custom modules are able to access the locked status via a property on `PortalSettings` and do any additional changes, if desired).

### DIFF
--- a/DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
+++ b/DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
@@ -23,6 +23,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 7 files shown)


bdukes:
I've been pushing some changes as I added a little bit more functionality, but I'm done now, so it's ready for review.

ohine:
This looks good to me, but I'm a bit biased since I started the implementation. I'll let @jbrinkman, @ashishpd or @cpaterra make the final decision.

sleupold:
Putting a site on hold would be great, but this needs also preventing user input like comments or forum posts.

bdukes:
@wmorgenweck is there interest in this feature from DNN Corp?

sleupold:
@bdukes 
at least there is interest from the community!

wmorgenweck:
@bdukes There is definitely some interest in this idea.  How would this change now that we have the new persona bar?

bdukes:
@wmorgenweck that's a good question.  since the persona bar doesn't have an equivalent to the "tools" section in the old control panel, we'd need to find a new place for this.  The most straightforward plan that I can think of is adding a new persona bar extension which would let you switch the lock on and off (and potentially show some details about what locking does and doesn't do, etc.).  Alternatively, it could be added/buried as an additional host-level setting on one of the settings pages.

mitchelsellers:
I think that this would be a huge benefit, and could be as simple as a "toggle" on the persona bar, in the lower left corner that would allow a host user to enable/disable.  

From my perspective it would be awesome if there was an indication of "Editing Disabled" to the user

wmorgenweck:
@bdukes would it make sense to handle this at the site level and host level?  If that's the case then I would think that put this under Site Settings > Behavior.  Then we have a toggle that manages the setting scope similar to how we handle SMTP settings.

bdukes:
@wmorgenweck our use case for this functionality is only for the host level, I'm not sure how solid the use case is for site-level locking (it may be there, I just haven't thought through it very much).

galatrash-at-dnn:
@bdukes since the ControlBar is already gone from the new releases, this PR is invalid and need to be modified to fit within the new PersonBar structure. Therefore, I will close this issue without merging it.

As others commented, this is a great feature. Hence, if you are willing to have it back, please work on it from the PB perspective and submit a totally new PR for this change. Note that the PB feature is in another repository ( https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions ) - so you need to have to change in two places and two PR's.

PR #1273, opened 2016-01-22
DNN-8333: changes to make JWT work when module page is under host.

galatrash-at-dnn:
Note: I merged the two SqlProvider files into one as there is no meaning in having two of them now before we had any 8.0.1 CTP release.


kanm2015:
Changes look good

PR #1274, opened 2016-01-22
DNN-7128: only reset updatepassword flag when password change successful.

zyhfish:


amarjit-dhunna:
The same change should be performed in ChangePasswordByToken(int portalid, string username, string newPassword, string resetToken) method also because this one also does not check the return value from ResetAndChangePassword call. Line 733 of UserController.cs

Please make changes in that method also and commit the code.

PR #1275, opened 2016-01-22
DNN-8377: Installation fails on non-English databases

sleupold:
Use ISO date format for comparisons


bdukes:
:+1:

ohine:
Looks good to me, but since it touches a sql file from a released build. I'm deferring the final decision to @jbrinkman, @ashishpd or @cpaterra.

sleupold:
Oliver, if the current version of the script fails during the upgrade, there is no option to fix it in a later one. This will not affect users upgrading from DNN 8.0.0, but all users (with non-english SQL Server language) upgrading from a previous version.

ohine:
I completely agree, but I'm still more comfortable leaving that decision up to one of the PMs.

ashishpd:
In rare situations, we've gone and updated the older SQL providers to deal with Upgrade situations. Merging this change. Thanks guys!

PR #1277, opened 2016-01-22
DNN-8382: fixing issue for file create/copy/move when destination …

galatrash-at-dnn:
… folder does not exist.


kanm2015:
Code review done

PR #1279, opened 2016-01-22
DNN-8257: Updated the wording for RecycleApp.ToolTip

kanm2015:


galatrash-at-dnn:
I'm rejecting this because the term "Select to recycle the application pool" is incorrect. This should be "Select to restart the application" as this functionality has nothing to do with the domain or the pool;just the application.

PR #1281, opened 2016-01-22
DNN-8360 Used copy image method at all required calls

amarjit-dhunna:
Moved copyimage method into ImageTransform abstract class as a virtual method.
Added usage of CopyImage where ever it was required.


francescorivola:
code review done. It looks good.

PR #1282, opened 2016-01-22
Allow sharing OAuth service config across sites DNN-8383

bdukes:
[DNN-8383](https://dnntracker.atlassian.net/browse/DNN-8383)

>DNN installations with many sites sometimes want to share the API secret & key between those sites, instead of having to redefine them for each site.


bdukes:
@ohine or @JoshuaBradley do you have any examples of making use of this API, that we can include in the JIRA ticket for QA?

fjlucas:
The changes done in this pull request has been reverted in this other:
https://github.com/dnnsoftware/Dnn.Platform/pull/1338

bdukes:
:+1:  thanks, we'll resubmit when a more complete solution is ready

PR #1284, opened 2016-01-22
DNN-8346 Fixed the transparency.

amarjit-dhunna:
Changes done to get correct content type in all cases and create correct image.


DNN Platform/Library/Services/GeneratedImage/DnnImageHandler.cs
```diff
@@ -168,11 +168,13 @@ public override ImageInfo GenerateImage(NameValueCollection parameters)
                         if (!uppTrans.TryGetPhotoFile(out photoFile))
                         {
                             var noAvatar = uppTrans.GetNoAvatarImage();
+                            ContentType = ImageFormat.Gif;
```
amarjit-dhunna: Forcing the call to be routed via quantizer rather than UserProfilePicTransform

PR #1285, opened 2016-01-25
DNN-8387: Add Windows Authentication type to make API request works when windows authentication enabled. udo changes made in DNN-6673.

zyhfish:


kanm2015:
changes look good

PR #1288, opened 2016-01-25
DNN-8385 Fixed the if to check the parameters

amarjit-dhunna:


DNN Platform/Library/Services/GeneratedImage/DnnImageHandler.cs
```diff
@@ -351,7 +351,7 @@ public override ImageInfo GenerateImage(NameValueCollection parameters)
                 if (mode == "profilepic")
                 {
                     resizeMode = ImageResizeMode.FitSquare;
-                    if (!string.IsNullOrEmpty(parameters["w"]) && !string.IsNullOrEmpty(parameters["h"]) && width != height)
+                    if ((!string.IsNullOrEmpty(parameters["w"]) && width>0) && (!string.IsNullOrEmpty(parameters["h"]) && height>0) && width != height)
```
francescorivola: @amarjit-dhunna , could we simplify this logic as: width > 0 && height > 0 && width != height?
The method TryParseDimension called in the same method above checks for the existance of the query string "w" and "h" and, if them are not set or invalid, it set the value to 0.

PR #1296, opened 2016-01-26
DNN-8396: added a fix for web request cache provider which throws …

galatrash-at-dnn:
…exceptions some times.


kanm2015:
Code review done

PR #1307, opened 2016-01-28
DNN-7454 - added logic to disable button on form submit.

mgatmaytan:
It doesn't matter if the button doesn't get-reactivated because the web form posts back and re-renders the button if its invalid.

### DIFF
--- a/DNN Platform/Modules/Groups/Create.ascx
+++ b/DNN Platform/Modules/Groups/Create.ascx
@@ -15,8 +15,8 @@
                         </label>
                     </div>
                     <asp:TextBox ID="txtGroupName" runat="server" MaxLength="50"/>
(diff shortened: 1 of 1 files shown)


DNN Platform/Modules/Groups/Create.ascx
```diff
@@ -100,6 +100,12 @@
             $('#trMem').hide();
 
         });
+        $('.wizardStepBody').find('.dnnPrimaryAction').click(function () {
+            var btn = $(this);
+            if ($('.wizardStepBody').find('input[type="text"]').val().length > 0) {
```
amarjit-dhunna: This will work if the group name is blank but it will disable even if the group name is invalid.

PR #1310, opened 2016-01-29
DNN-7066: use text not localized to get editor info.

zyhfish:


galatrash-at-dnn:
This is not the proper fix - the error is in creating the type in "Type.GetType(systemType);"

galatrash-at-dnn:
Added enhanced checking.

kanm2015:
Code review done

PR #1312, opened 2016-01-29
DNN-8411: Fixed culture insensitivity issue when retrieving role by name

aclarklestein:
Removing ToUpper on roleName/roleGroupName, since it is only used in an InvariantCultureIgnoreCase string comparison and causes issues in Turkish websites (i capitalizes to Turkish i using ToUpper)


amarjit-dhunna:
This issue has already been logged and resolved. Please refer to https://dnntracker.atlassian.net/browse/DNN-8404

PR #1317, opened 2016-02-01
DNN-6578: update content item when tab name changed or module title changed.

zyhfish:


DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -835,6 +835,20 @@ private void DeleteTabModuleInternal(ModuleInfo moduleInfo, bool softDelete, boo
             }
         }
 
+        /// <summary>
+        /// Update content item when the module title changed.
+        /// </summary>
+        /// <param name="module"></param>
+        public void UpdateContentItem(ModuleInfo module)
```
francescorivola: Hi @zyhfish , I would change the visibility of this method to private as not used outside the ModuleController class.

zyhfish: Done, thanks.

PR #1320, opened 2016-02-02
DNN-5919: show 404 message when profile use doesnt exist or deleted, prevent log thread abort exception.

zyhfish:


kanm2015:
Code review done

PR #1325, opened 2016-02-03
DNN-7960 Fixed the last update time for the folders

amarjit-dhunna:
DNN-7960 Fixed the last update time for the folders
A new SP (UpdateFolderModifiedOnToCurrentDate) has been added to update the folders updated date to current recursively. This will update the datetime only if the old timestamp is more than 59 secs old. This has been done to secure too many frequent updates in case of massive sync.
All the add/update/delete files and folders methods has been updated to call the UpdateFolderModifiedOnToCurrentDate. Changes in code has been done to clear the folders  cache in case of file addition/updation/deletion.

DNN-7960 Added the changes from Angel into same sql provider file
i.e. 08.00.01.01.SqlDataProvider.
Fixed the mistyped names SetHasBeenPublished to SetFileHasBeenPublished which was added from DNN-8378. Cleared the content of 08.00.01.SqlDataProvider


Website/Providers/DataProviders/SqlDataProvider/08.00.01.01.SqlDataProvider
```diff
@@ -445,3 +445,913 @@ AS
...
+
+/* DNN-8378 Start */
+/* Create HasBeenPublished field */
+IF NOT EXISTS
+(SELECT * FROM sys.all_columns c INNER JOIN sys.tables t
+	ON c.object_id = t.object_id AND t.name = '{objectQualifier}Files'
+	WHERE c.name = 'HasBeenPublished')
+BEGIN
+	ALTER TABLE {databaseOwner}[{objectQualifier}Files]
+		ADD HasBeenPublished bit DEFAULT 1 NOT NULL
```
sleupold: the Flag "HasBeenPublished" is not really self-explaning and not sufficient. 
 If you continue the life-cycle of a file on a website, there will be a time when a cleanup is planned and a cleanup workflow identifies files to be deleted, i.e. another flag would be needed "readyForDelete" or "Deleted".
 Instead of a boolean flag I suggest a "PublishState" (tinyInt), where the different states can be stored (and a master table FileStates, where states are explained).

sleupold:
the Flag "HasBeenPublished" is not really self-explaning and not sufficient. 
 If you continue the life-cycle of a file on a website, there will be a time when a cleanup is planned and a cleanup workflow identifies files to be deleted, i.e. another flag would be needed "readyForDelete" or "Deleted".
 Instead of a boolean flag I suggest a "PublishState" (tinyInt), where the different states can be stored (and a master table FileStates, where states are explained).

amarjit-dhunna:
Will need to submit new PR after changes from DNN-7333

PR #1326, opened 2016-02-03
Task/dnn 7297

galatrash-at-dnn:
Allowing MVC partial views.


ashishpd:
This has been reviewed by @galatrash-at-dnn  and @amarjit-dhunna.

PR #1331, opened 2016-02-04
DNN-6888: replace underscore with star char.

zyhfish:


amarjit-dhunna:
As per discussion with PM, we should implement the GUID solution in email urls. So, I am closing this PR and keeping this JIRA open for resolution.

PR #1337, opened 2016-02-05
DNN-8306: copy permissions from source module.

zyhfish:


amarjit-dhunna:
After adding your changes there is still issue with copy page. It works fine if I provide Edit permissions to a new role in module but if I provided only view permissions (Even If I say not to inherit view permissions from page) it doesn't copy them to the new page in fact it creates another issue. It removes the newly added View permission from the source module also. Please check screencast in JIRA for more details

amarjit-dhunna:
It solves the issue if "Inherit View Properties from Parent Page" is not selected but there is still an issue if this check box is selected. It was shown in earlier screen cast also. Basically the subscribers role added to source page module is not copied to new page module in fact the permission from original page module is removed. Please check screencast in JIRA for more details

PR #1341, opened 2016-02-15
DNN-6523: update module definitions of users and roles package to set default cache time as -1.

zyhfish:


amarjit-dhunna:
Code changes are good. But since PR has conflicts it is being closed. I will add the changes in new branch and submit PR.

PR #1345, opened 2016-02-17
DNN-7857: return true when the tab's portal id and alias' portal id are in same portal group.

zyhfish:


antoniodgonzalez:
Reviewed. change looks good

PR #1346, opened 2016-02-17
DNN-8450: Allow Override of prefixing Service to Username for OAuth providers.

JoshuaBradley:
[DNN-8450](https://dnntracker.atlassian.net/browse/DNN-8450)

> The fix for [DNN-4133](https://dnntracker.atlassian.net/browse/DNN-4133) was to add a prefix for the service to the beginning of the username. This broke our custom OAuth provider and I believe that we should have the ability to override whether we want to prefix the service to the Username in OAuth providers. In addition to allowing that overriding, this also implements the option to automatically associate logins to existing users, based on email.
> Testing of this API change will require a custom OAuth provider implementation which sets the new `PrefixServiceToUserName` and `AutoMatchExistingUsers` properties in that provider's implementation of `OAuthClientBase` to `false` and `true`, respectively. That custom provider may be an existing provider (e.g. Facebook), tweaked to override those two properties.
> When configured as mentioned above, the username of users who log in via that provider should be the user's email (as provided by the OAuth provider). Additionally, if a user with that email address as their username already exists, that OAuth login would be associated with the existing account, rather than trying to create a new account.

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -617,7 +617,7 @@ namespace DotNetNuke.Security.Membership
         }
 
        
-        private UserInfo GetUserByAuthToken(int portalId, string userToken, string authType)
+        public override UserInfo GetUserByAuthToken(int portalId, string userToken, string authType)
         {
             IDataReader dr = _dataProvider.GetUserByAuthToken(portalId, userToken, authType);
             UserInfo objUserInfo = FillUserInfo(portalId, dr, true);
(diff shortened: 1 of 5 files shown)


bdukes:
@jbrinkman @ashishpd is there interest in this PR?

wmorgenweck:
@cpaterra & @ashishpd, I think this could be useful.  Can we consider bringing this in?

bdukes:
Any feedback/status, @wmorgenweck, @cpaterra, or @ashishpd?

galatrash-at-dnn:
@JoshuaBradley this PR has conflicts. To be able to review, you need to go and recreate it again against the latest codebase. I will close this now without merging and feel free to submit another PR after resolving the conflicts.

bdukes:
@galatrash-at-dnn I've rebased and fixed conflicts

PR #1348, opened 2016-02-19
Use the existing File Id for the new upload file when replacement hap…

kanm2015:
…pens


francescorivola:
Close PR. The change does not look bad, but the fix in https://github.com/dnnsoftware/Dnn.Platform/pull/1350 is more closer to the old logic and improve the changes done in Workflow first version of unpublished file.

PR #1351, opened 2016-02-19
DNN-6547: fixing unit and integration tests

galatrash-at-dnn:
Added new overloads and fixed the failing tests.


kanm2015:
Code review done

PR #1361, opened 2016-02-24
DNN-4075: also copy module settings when add exist module.

zyhfish:


Website/admin/ControlPanel/AddModule.ascx.cs
```diff
@@ -495,6 +509,13 @@ private void DoAddExistingModule(int moduleId, int tabId, string paneName, int p
                 }
                 else
                 {
+                    //copy tab module settings
+                    newModule.TabModuleSettings.Clear();
```
amarjit-dhunna: I believe the settings should be copied only when copy checkbox is checked. So there is not need to copy settings in else part.

DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
```diff
@@ -637,6 +651,13 @@ private int DoAddExistingModule(int moduleId, int tabId, string paneName, int po
                 }
                 else
                 {
+                    //copy tab module settings
```
amarjit-dhunna: I believe the settings should be copied only when copy checkbox is checked. So there is not need to copy settings in else part.

sleupold: TabSettings need to be copied in both cases (but with different ModuleID, as they are reference specific.

PR #1362, opened 2016-02-24
DNN-6583: Updated ConvertCoreNotificationTypeActionsFor700 in DotNetN…

kanm2015:
…uke.Services.Upgrade to handle null values of notification types

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/Upgrade.cs
+++ b/DNN Platform/Library/Services/Upgrade/Upgrade.cs
@@ -3214,18 +3214,63 @@ namespace DotNetNuke.Services.Upgrade
             {
                 var nt = NotificationsController.Instance.GetNotificationType(name);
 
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3214,18 +3214,63 @@ private static void ConvertCoreNotificationTypeActionsFor700()
...
+                            NotificationsController.Instance.DeleteNotificationTypeAction(
+                                action.NotificationTypeActionId);
+                        }
+
+                        NotificationsController.Instance.SetNotificationTypeActions(actions, nt.NotificationTypeId);
+                    }
+                }
+                else
+                {
+                    switch (name)
```
amarjit-dhunna: We can wrap this section in a seprate method

PR #1363, opened 2016-02-24
DNN-8464: Remove cache from notifications

amarjit-dhunna:


ohine:
Why was caching removed? I think the correct way to solve this issue would be to invalidate the cache for that specific user when they click on the dismiss notification action. disabling the cache completely is going to affect performance.

jbrinkman:
Notification caching was just added in November (DNN-7135).  The correct fix should be to make sure invalidation happens correctly, not that we remove caching, that will only cause the performance problem noted in 7135 to re-emerge.

PR #1368, opened 2016-02-25
Add ReSharper settings

bdukes:
I wanted to get started creating some shared settings, so folks can avoid simple mistakes that ReSharper might insert (or not correct).

To start, this only has a setting to avoid adding nested namespaces.  I'm sure there are more settings that can be added to this.

Does this need a JIRA if it's only affecting developer experience?


galatrash-at-dnn:
@bdukes I merged this now. But in the future, even if it affects developers only, we should add Jira for all commits. Anyway, there must be a caution when modifying this by any developer as we don't need anyone to change the rules to suit their own preference. Any change must be in a PR and discussed with the team leads and devs in general.

bdukes:
👍 thanks @galatrash-at-dnn, will do

PR #1371, opened 2016-02-26
DNN-8470: Added defensive code to protect against case where 1 file h…

kanm2015:
…as multiple copy requests


DNN Platform/Providers/HtmlEditorProviders/RadEditorProvider/Components/EditorProvider.cs
```diff
@@ -253,7 +253,14 @@ private static void EnsureDefaultFileExists(string path)
 
                 //if defaultConfigFile is missing there is a big problem
                 //let the error propogate to the module level
-                File.Copy(defaultConfigFile, path);
+                try
+                {
+                    File.Copy(defaultConfigFile, path);
```
amarjit-dhunna: We should avoid using try catch and use the overwrite overload of the File.Copy

PR #1373, opened 2016-02-26
DNN-27129: Added new stored proc GetNonExpiredUsersByRoleName

kanm2015:


sleupold:
where is this new procedure used?

mitchelsellers:
I have been looking to see if there was another PR or JIRA issue associated with this change, but I cannot find any place that the API's were updated to call this new procedure.  Given the framework desire to require individuals to navigate through the API rather than directly to the database is this planned to be added?

PR #1375, opened 2016-02-29
DNN-7777: set search document's portal id to module's portal id if its not set yet.

zyhfish:


zyhfish:
will submit another PR to handle this problem.

PR #1376, opened 2016-02-29
DNN-8474: Rolled back changes from DNN-7906

amarjit-dhunna:
to fix DNN-8474. Made changes to fix DNN-7906 using manual delete after the desktop module is uninstalled.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -1584,6 +1584,26 @@ namespace DotNetNuke.Entities.Modules
         }
 
         /// <summary>
+        /// Gets the modules by DesktopModuleId.
+        /// </summary>
+        /// <param name="desktopModuleId">The Desktop Module Id.</param>
+        /// <returns>module collection</returns>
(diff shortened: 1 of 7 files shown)


DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -1584,6 +1584,26 @@ public ArrayList GetModulesByDefinition(int portalID, string definitionName)
...
+        public ArrayList GetModulesByDesktopModuleId(int desktopModuleId)
+        {
+            var moduleDefinitions = ModuleDefinitionController.GetModuleDefinitionsByDesktopModuleID(desktopModuleId);
+            var modules = new ArrayList();
+            foreach (var moduleDefinition in moduleDefinitions)
+            {
+                var portals = PortalController.Instance.GetPortals();
+                foreach (PortalInfo portal in portals)
+                {
+                    modules.AddRange(GetModulesByDefinition(portal.PortalID, moduleDefinition.Key));
```
zyhfish: the dictionary's key is module definition's friendly name but not definition name(they are same value mostly, but it could be different), need use the definition name here.

PR #1381, opened 2016-03-01
Would close DNN-8479

iJungleboy:
This would allow the installer to include pdb files in the installation, as described in detail in https://dnntracker.atlassian.net/browse/DNN-8479

Here an excerpt:

When users install our modules and run into an error, they have trouble reporting what exactly the error is, because the stack in the events-module is not really helpful.
BUT there is a solution: if we deploy the built PDB files from visual studio together with the .DLL files, then the error messages include exact line numbers. This is very helpful in fixing issues.
PDB files have no security implication, they are like JS-map files and only help the error-system inform the user about the code-line where something happened.
Because of this, it's important that we can also include the PDB files in the installation. Unfortunately as of now, the installer blocks this extension - telling the user it's unsafe.
We should add pdb-files to the list of allowed files in the assembly installer to aid in better user error-feedback.


bdukes:
The symbols packages that DNN generates don't show any errors (they use a `ResourceFile` component to place the pdb files in the `bin` folder).  Does that work for you, or does using the `Assembly` component get you some other benefit?

bdukes:
For the record, I have no problem with this change, just wanting to better understand the different options.

iJungleboy:
I'm now also using the resourcefile workaround. It works, but it's very hacky / dirty / indiscoverable. IMHO this should be straight-forward and not feel like you're abusing a back-door.

ohine:
I believe the `ResourceFile` workaround is only valid for only a single package with pdb's (for uninstallation that is). It would be great to support pdb files along side their dll components. Thanks for the contribution @iJungleboy!

iJungleboy:
Will drop this pull - and re-create as it became stale...

iJungleboy:
Note: this looks long, but it's only 1 simple file that changes!

DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
```diff
@@ -110,7 +110,7 @@ public override string AllowableFiles
         {
             get
             {
-                return "dll";
```
ashishpd: One my engineer recommended removing space after comma.

iJungleboy:
Will cancel this and send a new pull-request with just this micro-change...
https://github.com/dnnsoftware/Dnn.Platform/pull/1769

iJungleboy:
Second micro-change in this new pull request: https://github.com/dnnsoftware/Dnn.Platform/pull/1770

PR #1382, opened 2016-03-02
DNN-8481: check whether the receipient already add by sp CreateMessageRecipientsForRole first.

zyhfish:


sleupold:
it would be more performant, if we just improve stored procedure SaveMessageRecipient to add an address only, if it is not already in the list.

sleupold:
```
IF  EXISTS (SELECT * FROM sys.Procedures WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}CoreMessaging_SaveMessageRecipient]'))
    DROP PROCEDURE {databaseOwner}[{objectQualifier}CoreMessaging_SaveMessageRecipient]
GO

CREATE PROCEDURE {databaseOwner}[{objectQualifier}CoreMessaging_SaveMessageRecipient]
    @RecipientID        Int, -- ignored
    @MessageID          Int, -- must be ID of an existing message
    @UserID             Int, -- ID of the recipient
    @Read               Bit, -- not Null
    @Archived           Bit, -- not Null
    @CreateUpdateUserID Int  -- not Null, ID of the user performing the action
AS
BEGIN
    MERGE INTO {databaseOwner}[{objectQualifier}CoreMessaging_MessageRecipients] T
     USING (SELECT @MessageID MessgID, @UserID UsID, @Read Rd, @Archived Archvd, @CreateUpdateUserID CrUserID) S
     ON (T.MessageID = S.MessgID AND T.UserID = S.UsID)
     WHEN MATCHED THEN UPDATE SET 
        [Read]                 = [Rd],
        [Archived]             = [Archvd],
        [LastModifiedByUserID] = [CrUserID],
        [LastModifiedOnDate]   = GETDATE()
     WHEN NOT MATCHED THEN 
        INSERT  ([MessageID], [UserID], [Read], [Archived], [CreatedByUserID], [CreatedOnDate], [LastModifiedByUserID], [LastModifiedOnDate])
         VALUES (S.[MessgID], S.[UsID], S.[Rd], S.[Archvd],      S.[CrUserID],       GETDATE(),           S.[CrUserID],            GETDATE());

    SELECT  RecipientID 
     FROM  {databaseOwner}[{objectQualifier}CoreMessaging_MessageRecipients]
     WHERE MessageID = @MessageID AND UserID = @UserID;
END -- PROCEDURE
GO
```

PR #1383, opened 2016-03-03
DNN-7777: add portal id in search document.

zyhfish:


ashishpd:
Went back and forth with @zyhfish on this issue, change looks good.

PR #1385, opened 2016-03-04
DNN-8438 Fixed the order of viewengine assignment.

amarjit-dhunna:


ashishpd:
I went back and forth on this issue with Amarjit - it looks good.

PR #1396, opened 2016-03-09
DNN-4864: check anonymous except support modules and ValidateAntiForgeryToken attributes, also these 2 attributes need always authorize.

zyhfish:


DNN Platform/DotNetNuke.Web/Api/AuthorizeAttributeBase.cs
```diff
@@ -70,7 +70,9 @@ protected virtual bool SkipAuthorization(HttpActionContext actionContext)
         public static bool IsAnonymousAttributePresent(HttpActionContext actionContext)
         {
             return actionContext.ActionDescriptor.GetCustomAttributes<AllowAnonymousAttribute>().Any()
-                   || actionContext.ControllerContext.ControllerDescriptor.GetCustomAttributes<AllowAnonymousAttribute>().Any();
+                   || (actionContext.ControllerContext.ControllerDescriptor.GetCustomAttributes<AllowAnonymousAttribute>().Any()
+                         && !actionContext.ActionDescriptor.GetCustomAttributes<AuthorizeAttributeBase>()
+                                .Any(t => !(t is SupportedModulesAttribute) && !(t is ValidateAntiForgeryTokenAttribute)));
         }
```
amarjit-dhunna: Hi @zyhfish , This still checks for the presence of the ValidateAntiForgeryTokenAttribute on the method. This check should work as follows:
1. If AllowAnonymous is present at method level = SkipAuthorization
2. If AllowAnonymous is present at Class level and There is not AuthorizeAttributeBase type attribute at method level = SkipAuthorization
3. If AllowAnonymous is present at Class level and There is AuthorizeAttributeBase type attribute at method level = PerformAuthorization
3. If no AllowAnonymous is present at any level = PerformAuthorization

The the actual check should be like below:
```
return actionContext.ActionDescriptor.GetCustomAttributes<AllowAnonymousAttribute>().Any()
                   || (actionContext.ControllerContext.ControllerDescriptor.GetCustomAttributes<AllowAnonymousAttribute>().Any()
                         && !actionContext.ActionDescriptor.GetCustomAttributes<AuthorizeAttributeBase>()
                                .Any(t => !(t is SupportedModulesAttribute)));
```

PR #1397, opened 2016-03-09
Dnn 8503: Fixed vw_Modules  casting code

sawest:
[DNN-8503](https://dnntracker.atlassian.net/browse/DNN-8503)
The VW_MODULES view has the below CASE statement causing an error:

CASE WHEN tm.IconFile LIKE 'fileid=%' THEN
 (SELECT IsNull(Folder, '') + [FileName]
 FROM dbo.[vw_Files]
 WHERE fileid = CAST(SUBSTRING(tm.IconFile, 8, 10) AS int)) ELSE COALESCE (tm.IconFile, '') END AS IconFile

When this view is joined to another table as in the stored procedure 'getModulebyDefinition', the subquery in the above codes is fired for some unknown reason. The substring returns a partial string of a file path for iconfile. It cannot be cast to integer and, therefore, causes a pagebase error stated in the issue tracker.

This contribution changes the code to the following:

CASE WHEN tm.IconFile LIKE 'fileid=%' THEN
 (SELECT IsNull(Folder, '') + [FileName]
 FROM dbo.[vw_Files]
 WHERE CAST(fileid AS nchar(10)) = SUBSTRING(tm.IconFile, 8, 10)) ELSE COALESCE (tm.IconFile, '') END AS IconFile

The result is the same but no accidental casting of strings to integers will happen. 

There are 2 commits with this pull request.  Github some how attached me with the wrong account on my first commit so I committed again with the correct account.

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/08.00.01.02.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/08.00.01.02.SqlDataProvider
@@ -10,8 +10,7 @@ SELECT M.PortalID AS OwnerPortalID, T.PortalID, TM.TabID, TM.TabModuleID, M.Modu
              WHERE CAST(fileid AS nchar(10)) = SUBSTRING(tm.IconFile, 8, 10)) 
 	   ELSE COALESCE (tm.IconFile, '') 
 	   END AS IconFile, 
-	   M.AllTabs, TM.Visibility, TM.IsDeleted, 
-       TM.Header, TM.Footer, M.StartDate, M.EndDate, TM.ContainerSrc, TM.DisplayTitle, TM.DisplayPrint, TM.DisplaySyndicate, TM.IsWebSlice, TM.WebSliceTitle, 
+	   M.AllTabs, TM.Visibility, TM.IsDeleted, TM.Header, TM.Footer, M.StartDate, M.EndDate, TM.ContainerSrc, TM.DisplayTitle, TM.DisplayPrint, TM.DisplaySyndicate, TM.IsWebSlice, TM.WebSliceTitle, 
        TM.WebSliceExpiryDate, TM.WebSliceTTL, M.InheritViewPermissions, M.IsShareable, M.IsShareableViewOnly, MD.DesktopModuleID, MD.DefaultCacheTime, 
        MC.ModuleControlID, DM.BusinessControllerClass, DM.IsAdmin, DM.SupportedFeatures, CI.ContentItemID, CI.[Content], CI.ContentTypeID, CI.ContentKey, CI.Indexed, 
        CI.StateID, M.CreatedByUserID, M.CreatedOnDate, M.LastModifiedByUserID, M.LastModifiedOnDate, M.LastContentModifiedOnDate, TM.UniqueId, TM.VersionGuid, 
--- a/Website/Providers/DataProviders/SqlDataProvider/08.00.01.02.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/08.00.01.02.SqlDataProvider
@@ -1,0 +1,28 @@
+/* DNN-8503 */
+
+ALTER VIEW {databaseOwner}[{objectQualifier}vw_Modules]
+AS
+SELECT M.PortalID AS OwnerPortalID, T.PortalID, TM.TabID, TM.TabModuleID, M.ModuleID, M.ModuleDefID, TM.ModuleOrder, TM.PaneName, TM.ModuleTitle, TM.CacheTime, 
+       TM.CacheMethod, TM.Alignment, TM.Color, TM.Border, 
+	   CASE WHEN tm.IconFile LIKE 'fileid=%' THEN
+			(SELECT IsNull(Folder, '') + [FileName]
+             FROM dbo.[vw_Files]
+             WHERE CAST(fileid AS nchar(10)) = SUBSTRING(tm.IconFile, 8, 10)) 
+	   ELSE COALESCE (tm.IconFile, '') 
+	   END AS IconFile, 
+	   M.AllTabs, TM.Visibility, TM.IsDeleted, 
+       TM.Header, TM.Footer, M.StartDate, M.EndDate, TM.ContainerSrc, TM.DisplayTitle, TM.DisplayPrint, TM.DisplaySyndicate, TM.IsWebSlice, TM.WebSliceTitle, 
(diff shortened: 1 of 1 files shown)


sleupold:
I suggest simply reverting the changes made in DNN 8.0.0 and use the previous version, which used static function "FilePath". I did a couple of performance tests and it turned out to be the fastest option (assuming, that only a percentage of modules used IconFile)

zyhfish:
1. i tried set some module icon file for html module, then try to run "exec GetModuleByDefinition 0, 'Text/HTML'", it works without any error. do we have a clarify steps to reproduce the issue?
2. we have similar code in vw_tabmodules, vw_tabs as well, if this is valid issue, these SP should be update together. thx

galatrash-at-dnn:
@sawest , you need to sign and send the CLA document before we look into and process this PR. See this https://github.com/dnnsoftware/Dnn.Platform/blob/development/CONTRIBUTING.md

sleupold:
@zyhfish the issue occurs when there is a file path stored in IconFile column (instead of "fileid=..."

sawest:
CLA has been submitted.

galatrash-at-dnn:
@sawest please check @zyhfish earlier comment and reply/modify you code accordingly.

sawest:
The environment in which I see this bug is using a SQL Express 2008 server.  I noticed that the bug does not occur in environments with later SQL server versions.  It appears that the SQL version is likely a factor with this bug.  Perhaps SQL versions parse/transact that CASE statement differently?  I also noticed that current minimum requirements for SQL versions is now 2008 R2.  At this point the bug may be a non issue if the SQL version it occurs in is not supported.

sawest:
I feel like the select statement is run to retrieve a return value before the when statement is evaluated.  If the when statement is true, then the retrieved value is returned.  This procedure may have changed with later versions of SQL server.  I have not been able to find any real answers in my research on SQL.

sleupold:
@sawest what was the database compatibility level of your SQL 2008 database?

sawest:
I still have the database up and am able to cause/fix this bug at will if needed.

SQL Express 2008 10.0.1600.22 
Compatibility level 100

sawest:
After a little more research on MS SQL's evaluation procedure for CASE referenced here: [https://msdn.microsoft.com/en-us/library/ms181765(v=sql.110).aspx](url) 

You will find the below quote:
> You should only depend on order of evaluation of the WHEN conditions for scalar expressions (including non-correlated sub-queries that return scalars), not for aggregate expressions.

Notice the "non-correlated sub-query" part.  The reason for this is that the return expressions can be evaluated prior to the WHEN expression.  For example, you would not want an UPDATE statement inside a THEN because the update statement could be executed prior to the WHEN.  IE, CASE cannot be used for control-flow either.  This must be the problem: the SELECT is executing prior to the WHEN and the SELECT is correlated (contingent) on the WHEN.  CASE can't reliably do this in MS SQL as per the above article.

The above article is regarding SQL Server 2012 so not limited to SQL Server 2008.  Since DNN is using a correlated sub-query in this CASE statement, the bug is probably intermittent and only shows it's ugly head in random situations.  This pull-request will fix the problem.  But the best solution may be to rethink the entire CASE statement in light of this research.  @sleupold seems to have tested and verified an alternative to the CASE statement.

sleupold:
@sawest 
the order of evaluation in the documentation is referring to multiple "WHEN" conditions inside the same case, which is not used here. Conditions are always evaluated before the term returned.
The difference of this pull request is the type cast in the last line - if the string after "fileid=" does not contain a valid integer.

I still prefer using static function FilePath - which would not be optimal, if all rows contain "fileid=" references, but in my experience, only a small percentage of all records do have an icon assigned.

`CREATE FUNCTION {databaseOwner}[{objectQualifier}FilePath]
(    @StrMayContainFileId nVarChar(500)   )
    RETURNS              nVarChar(500)
AS
BEGIN
    DECLARE @Path AS nVarChar(500);
    IF ISNUll(@StrMayContainFileId, N'') = N''
        SET @Path = N''
     ELSE IF @StrMayContainFileId LIKE N'fileid=%'
        SELECT @Path = FilePath FROM {databaseOwner}[{objectQualifier}vw_FilePaths]
         WHERE Cast(fileid as nVarchar(10)) = SUBSTRING(@StrMayContainFileId, 8, 10)
     ELSE
        SET @Path = @StrMayContainFileId
    RETURN @Path -- never Null!
END 
`

sleupold:
PS: performancewise it would be best to add a calculated persisted column fileID int Null, which could be linked to column vw_filePath.FileId in a left join (or via an inline table function)

sawest:
I think I have provided everything @zyhfish and @galatrash-at-dnn  need for this issue.  If either of you need anything else from me, don't hesitate to ask.

@sleupold reading the above article will show it covers much more than multiple WHENs.  Yes, the return value will always be returned AFTER the WHEN.  However, it is possible (obviously) that in the case of a THEN SELECT, that the SELECT be executed (not RETURNED) prior to the WHEN.  This is why return value should not be a sub-query that correlates in any way to the WHEN expression nor should it be any thing that controls flow of execution such as an UPDATE.  (All covered in the article).

sleupold:
please see alternative fix in PR#1455

mitchelsellers:
Closing this PR due to staleness and procedural changes to the PR process.  If this issue/contribution is still desired please resubmit, we apologize for any delay as the finalize the transition of PR management to the DNN Community.

If re-submitted please make note of the re-submission with the new PR and we will expedite review following the current procedures.

sawest:
Well that was disappointing after waiting 2+ years.  If first submission = waste of time, then surely re-submission = more wasted time.  Think I will pass on the re-submission.  Let the bug stay I know the work around.

mitchelsellers:
@sawest Our apologies on this, working a backlog with an item that is more than 6 releases behind is something that cannot be processed as-is.  We are extremely sorry that this pull request had sat out here as long as it did without action, but we have only taken over control of the process in the past few months.

I do guarantee that if you re-submit aligned with the 9.3 release that we WILL review it promptly and we WILL approve the submission.

Our goal is to make this process easier in the end, sadly there are a few hiccups in making this transition

sleupold:
ok, again:

the issue was introduced by a "performance fix" in DNN 8.0.0. reverting to my previous version from DNN 7.2.0 will only work for DNN 8, as in DNN 9.0.1 the field list has changed.

But you may just use the original line for file path in the field list:

 `{databaseOwner}{objectQualifier}FilePath(TM.IconFile) AS IconFile,`

PR #1402, opened 2016-03-14
Dnn 7522

JoshuaBradley:
Swallowing HTTPException on WriteFeedback method to prevent this from continuing to run after a request has timed out.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/HtmlUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/HtmlUtils.cs
@@ -524,10 +524,8 @@ namespace DotNetNuke.Common.Utilities
             try
             {
                 bool showInstallationMessages;
-                string configSetting = Config.GetSetting("ShowInstallationMessages");
-                if (configSetting != null
-                   && bool.TryParse(configSetting, out showInstallationMessages)
-                   && showInstallationMessages)
+                string configSetting = Config.GetSetting("ShowInstallationMessages") ?? "true";
+                if (bool.TryParse(configSetting, out showInstallationMessages) && showInstallationMessages)
                 {
                     //Get the time of the feedback
                     TimeSpan timeElapsed = Upgrade.RunTime;
--- a/DNN Platform/Library/Common/Utilities/HtmlUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/HtmlUtils.cs
@@ -504,7 +504,7 @@ namespace DotNetNuke.Common.Utilities
         /// <param name="message">The feedback message</param>
         /// -----------------------------------------------------------------------------
         public static void WriteFeedback(HttpResponse response, Int32 indent, string message)
-        { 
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Common/Utilities/HtmlUtils.cs
```diff
@@ -521,28 +521,36 @@ public static void WriteFeedback(HttpResponse response, Int32 indent, string mes
         /// -----------------------------------------------------------------------------
         public static void WriteFeedback(HttpResponse response, Int32 indent, string message, bool showtime)
         {
-            bool showInstallationMessages = true;
-            string ConfigSetting = Config.GetSetting("ShowInstallationMessages");
-            if (ConfigSetting != null)
-            {
-                showInstallationMessages = bool.Parse(ConfigSetting);
-            }
-            if (showInstallationMessages)
```
IDisposable: How about simpler:

```
bool showInstallationMessages = true;
string ConfigSetting = Config.GetSetting("ShowInstallationMessages");		
if (ConfigSetting != null 
    && bool.TryParse(ConfigSettings, out showInstallationMessages) 
    && showInstallationMessages)
{
```

DNN Platform/Library/Common/Utilities/HtmlUtils.cs
```diff
@@ -521,28 +521,36 @@ public static void WriteFeedback(HttpResponse response, Int32 indent, string mes
...
+                    }
+                    strMessage += message;
+                    response.Write(strMessage);
+                    response.Flush();
                 }
-                strMessage += message;
-                response.Write(strMessage);
-                response.Flush();
+            }
+            catch (HttpException)
```
IDisposable: Better to avoid exceptions in the first place, see comment on left

DNN Platform/Library/Common/Utilities/HtmlUtils.cs
```diff
@@ -504,7 +504,7 @@ public static void WriteError(HttpResponse response, string file, string message
         /// <param name="message">The feedback message</param>
         /// -----------------------------------------------------------------------------
         public static void WriteFeedback(HttpResponse response, Int32 indent, string message)
-        {
+        {
```
IDisposable: Unneeded change.

bdukes:
@ashishpd @jbrinkman is there interest in this PR?

bdukes:
Any feedback/status, @wmorgenweck, @cpaterra, @ashishpd, or @jbrinkman?

DNN Platform/Library/Common/Utilities/HtmlUtils.cs
galatrash-at-dnn: The problem is no tin the parsing; it is in writing to a closed stream; therefore, we still need the try/catch blocks.

zyhfish:

PR #1408, opened 2016-03-21
DNN-8538 and DNN-8334 : Update the RegEx during install/upgrade to al…

amarjit-dhunna:
…low Image Handler


zyhfish:
also need consider the settings may saved in host settings table.

zyhfish:
Code changes looks good.

PR #1416, opened 2016-03-24
DNN-8553: cleanup user input data & double encoding.

galatrash-at-dnn:


DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
```diff
@@ -243,7 +243,7 @@ public static string GetRichValue(ProfilePropertyDefinition property, string for
                         break;
                     case "richtext":
                         var objSecurity = new PortalSecurity();
-                        result = PropertyAccess.FormatString(objSecurity.InputFilter(HttpUtility.HtmlDecode(property.PropertyValue), PortalSecurity.FilterFlag.NoScripting), formatString);
+                        result = PropertyAccess.FormatString(objSecurity.InputFilter(property.PropertyValue, PortalSecurity.FilterFlag.NoScripting), formatString);
```
zyhfish: this will cause the biohraphy show the html tags as content:

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -2785,8 +2785,6 @@ private static void UpgradeToVersion721()
 
                         if (userName != HttpUtility.HtmlDecode(userName))
                         {
-                            
-                            userName = HttpUtility.HtmlDecode(userName);
```
zyhfish: i don't think this line can be removed, such as if the username is "&lt;testuser&gt;", if you don't decode it, the input filter can't remove the tags.

DNN Platform/Library/Common/Utilities/HtmlUtils.cs
```diff
@@ -49,7 +49,7 @@ public class HtmlUtils
         private static readonly Regex HtmlDetectionRegex = new Regex("<(.*\\s*)>", RegexOptions.Compiled);
         private static readonly Regex StripWhiteSpaceRegex = new Regex("\\s+", RegexOptions.Compiled);
         private static readonly Regex StripNonWordRegex = new Regex("\\W*", RegexOptions.Compiled);
-        private static readonly Regex StripTagsRegex = new Regex("<[^<>]*>", RegexOptions.Compiled);
+        private static readonly Regex StripTagsRegex = new Regex("<.*?>", RegexOptions.Compiled | RegexOptions.Multiline);
```
zyhfish: this regex can't match the string which have line break:

DNN Platform/Library/Entities/Profile/ProfileController.cs
```diff
@@ -602,11 +602,11 @@ private static string RemoveUnsafeAttributes(string content)
                 needEncode = true;
             }
 
-            var tagMatches = Regex.Matches(cleanContent, @"(<[^>]*?) on.*?\=(['""]?)[\s\S]*?(\2 )([^>]*?>)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+            var tagMatches = Regex.Matches(cleanContent, @"(<[^>]*?) on.*?\=(['""]?)[\s\S]*?(\2 )([^>]*?>)", RegexOptions.IgnoreCase);
```
mitchelsellers: For performance purposes, these RegEx statements should be compiled to ensure that they are properly cached and not re-interpreted by the .NET framework.  As only the "most commonly used" items are cached if not explicitly marked as compiled.

galatrash-at-dnn: Compiling Regex components that are created on the fly is very slow and keeps adding to the temporary IL generated code. I did few tests and compiling these (non-static) ones is 3 to 6 times slower than non compiled ones.

DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
galatrash-at-dnn: Biography can be added bu any registered user. Therefore, allowing users to add these kinds of tags opens the site to XSS. IMO, this should not be allowed and we display what the users entered.

DNN Platform/Library/Common/Utilities/HtmlUtils.cs
galatrash-at-dnn: Reverted this one back.

DNN Platform/Library/Services/Upgrade/Upgrade.cs
galatrash-at-dnn: This applies to upgrading older sites and not for noral user registration.

DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
galatrash-at-dnn: Reverted this change.

PR #1419, opened 2016-03-25
DNN-8559 enhance UX for Usage Details report in Extensions module

bdukes:
[DNN-8559](https://dnntracker.atlassian.net/browse/DNN-8559)
>There are a couple of small changes that the Usage Details report in the extensions module could make to make its user experience smoother.

[DNN-8560](https://dnntracker.atlassian.net/browse/DNN-8560) (sub-task)
>The module generates URLs using the first portal alias for the site, instead of the primary alias. Also, it generates the URL as `/Default.aspx?tabId=123` instead of generating the canonical URL for the page.

[DNN-8561](https://dnntracker.atlassian.net/browse/DNN-8561) (sub-task)
>When accessing the usage details of a module from _Host -> Extensions_, it requires you to pick the portal/site, even if your installation only contains one site. It should default to the current site.


kanm2015:
Change looks good, PR merged.

PR #1423, opened 2016-03-29
DNN-8565: added support for bulk-insertion of DataTable object

galatrash-at-dnn:
Support for bulk-insertion of DataTable object (with multiple rows) into the database.


kanm2015:
Code review done

PR #1430, opened 2016-04-01
DNN-8083: copy dispatcher code into platform.

zyhfish:


Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs
```diff
@@ -128,6 +128,8 @@ protected override void OnLoad(EventArgs e)
 
                     DisplayQuickSettings = ModuleContext.Configuration.ModuleSettings.GetValueOrDefault("QS_FirstLoad", true);
                     ModuleController.Instance.UpdateModuleSetting(ModuleContext.ModuleId, "QS_FirstLoad", "False");
+
+                    ClientResourceManager.RegisterScript(Page, "~/admin/menus/ModuleActions/dnnQuickSettings.js");
```
amarjit-dhunna: Is there any specific reason to register this script in OnLoad instead of OnInit along with ModuleActions.js? Won't it make more sense to load them at same place?

PR #1434, opened 2016-04-04
DNN-8576: Improve scripts load in ajax mode to support load in sequence in FF.

zyhfish:


DNN Platform/DotNetNuke.Web.Client/Controls/ClientResourceLoader.cs
```diff
@@ -51,6 +51,28 @@ protected override void OnPreRender(System.EventArgs e)
...
+        var s = document.createElement( 'script' );
+        s.type = 'text/javascript';
+        s.src = window.dnnLoadScriptsInAjaxMode[i];
+        s.onload = window.dnnLoadScriptsInAjaxModeComplete;
+        s.onerror = window.dnnLoadScriptsInAjaxModeComplete;
+        document.body.appendChild(s);
+    }
+};
+
+var isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
```
bdukes: If this requires user agent testing rather than feature testing, can we at least get some comments about what scenario causes us to behave differently in Firefox, so we can revisit that decision in the future?

zyhfish:
Here is an article which describe the behavior in FF4: http://www.softwareishard.com/blog/firebug/script-execution-analysis-in-firefox-4/.

bdukes:
Thanks @zyhfish, can that link be added as a comment in the source?

PR #1436, opened 2016-04-04
DNN-8573 Suppress X-Frame-Options header by default.

amarjit-dhunna:


mitchelsellers:
This change is a bit against the standard recommendations that are being made with regards to cross-site scripting recommendations.  By doing this, we will go back to a situation where a user could embed any page of DNN in another site with an IFrame.

I realize that we did allow htis with prior versions, but I'm curious if this is really the change in behavior we desire?

ashishpd:
@mitchelsellers Due to the MVC implementation in 8.0 this header was set by default for ALL requests.  This is standard Asp.Net MVC behavior. This was ignoring any X-Frame-Options that you might have configured in web.config.

This PR fixed the problem where system will respect the web.config setting. However, we take your point that by default there will be no option. We are back to pre 8.0.

We are going to add this option on **new installs** with value SAMEORIGIN. Users can change to whatever they desire.

Thoughts?

Thanks
Ash

mitchelsellers:
@ashishpd My only thought here is that the default behavior could be used to assume proper security by those that have developed new MVC applications.  I'd almost be in favor of leaving this code change in, however, additionally ensuring that on upgrade "SAMEORIGIN" is set for installations that didn't already override the default behavior.

My concern would be in the following example.

I set up a DNN 8.0.0 site, and everything is secured.  I then upgrade to the latest version, now all of my stuff can be viewed without the proper context.

If we set the value for those that haven't already explicitly set it, we would prevent this change in behavior.

amarjit-dhunna:
@ashishpd I have modified the code to Suppress X-Frame-Options Header only if there is some configuration specified in the Web.Config for it. It might be remove or add header. In this way it will have same behavior as 8.0 for all the websites without any settings for this header in the Web.config but if someone wants to override this behavior intentionally, they would be able to do it from web.config.

Let me know in case of any questions.

/Amarjit

zyhfish:
Changes looks good, merged.

PR #1437, opened 2016-04-05
DNN-8578: Xcillion cause site can't build in VS after install.

zyhfish:


ohine:
Shouldn't these changes get checked into the main repository first? https://github.com/XCESS-NL/xcillion

ashishpd:
@ohine agreed, though we have had more customization in DNN 8.0 release, which were not sent back to the main repo. We'd have to do a mass merge later.

PR #1438, opened 2016-04-05
DNN-8532 - Messaging Fails for users with an Apostrophe in their name

jsmontesinos:
Fixed username encoding so now can contains apostrophe without errors on messaging tab. Also applied same fix in portal name variable.


francescorivola:
Code review done. It looks good.

PR #1442, opened 2016-04-07
DNN-8588 Rotate the images on upload if required

amarjit-dhunna:
We check if there is a metadata property present for orientation. If present then we apply the rotation and remove the property


DNN Platform/Library/Services/FileSystem/FileManager.cs
```diff
@@ -174,18 +175,107 @@ private void OnFileAdded(IFileInfo fileInfo, IFolderInfo folderInfo, int userId)
...
+                    image.SetPropertyItem(orientation);
+                    content = ToStream(image, GetImageFormat(image));
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex);
+            }
+        }
+        private static ImageFormat GetImageFormat(Image img)
```
zyhfish: this method should change to use switch-case statment.

zyhfish:

PR #1443, opened 2016-04-08
DNN-8556: update tab reference when delete tab and files.

zyhfish:


sleupold:
this solution will not restore previous situation, when the page gets undeleted.
IMO we should check retrieved URL in DNN (after loading the tabs), whether the target page is deleted

galatrash-at-dnn:
Outdated issue and PR. I'm closing this based on `won't fix` status in the Jira issue.

PR #1451, opened 2016-04-14
DNN-8604 Avoid touching files when XmlMerge doesn't change anything

bdukes:
[DNN-8604](https://dnntracker.atlassian.net/browse/DNN-8604)

> It is common for packages to include a [config component](http://www.dnnsoftware.com/wiki/manifest-config-component) in their manifest if they require any adjustments to the `web.config` file.  On initial install, this change will be made correctly.  However, this component will also run on upgrades (ensuring that the config is correctly in place), and the vast majority of the time, won't do anything, since it was done on install.  The problem is that the config component still updates the `web.config` file, even if no changes are being made, which causes the website to be restarted unnecessarily.
> 
> I've attached a modified version of the Xcillion theme package which adds support for the `woff` and `woff2` font file formats.  When installing this a second time, it should not require restarting the site, but currently DNN will restart the site anyway.

### DIFF
--- a/DNN Platform/Library/Services/Installer/XmlMerge.cs
+++ b/DNN Platform/Library/Services/Installer/XmlMerge.cs
@@ -544,8 +544,7 @@ namespace DotNetNuke.Services.Installer
                     {
                         case "overwrite":
                             var oldContent = rootNode.InnerXml;
-                            rootNode.RemoveChild(targetNode);
-                            rootNode.InnerXml = rootNode.InnerXml + child.OuterXml;
+                            rootNode.ReplaceChild(TargetConfig.ImportNode(child, true), targetNode);
                             var newContent = rootNode.InnerXml;
                             changedNode = !string.Equals(oldContent, newContent, StringComparison.Ordinal);
                             break;
--- a/DNN Platform/Library/Services/Installer/XmlMerge.cs
+++ b/DNN Platform/Library/Services/Installer/XmlMerge.cs
@@ -158,6 +158,14 @@ namespace DotNetNuke.Services.Installer
         /// -----------------------------------------------------------------------------
         public string Version { get; private set; }
 
+        /// -----------------------------------------------------------------------------
+        /// <summary>
+        /// Gets a value indicating whether the last update performed by this instance resulted in any changes
+        /// </summary>
+        /// <value><c>true</c> if there were changes, <c>false</c> if no changes were made to the target document</value>
+        /// -----------------------------------------------------------------------------
+        public bool ConfigUpdateChangedNodes { get; private set; }
+
         public IDictionary<string, XmlDocument> PendingDocuments
         {
             get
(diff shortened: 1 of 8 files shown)


bdukes:
@ashishpd any interest in this one?

bradbamford:
Similarly, could we also check if dll files are different during installation of extensions.
That's another place that could cause the website to be restarted unnecessarily.

galatrash-at-dnn:
@bdukes, this change doesn't take into consideration if there are two changes and the first created a change while the second didn't. The ``ConfigUpdateChangedNodes`` flag must be an accumulative for these changes. Therefore, another change in ``XmlMerge.cs`` file is needed for the line ```ConfigUpdateChangedNodes = changedAnyNodes;``` to become ```ConfigUpdateChangedNodes |= changedAnyNodes;```

Waiting for this change and afterwards I will process this PR.

bdukes:
@galatrash-at-dnn is there a scenario during installation where there would be two changes on one instance of `XmlMerge`, or is this in a theoretical case where some third party code is re-using an `XmlMerge` instance?  

If the latter, I'm having a hard time coming up with a use case where you'd want it to behave that way.  I documented the property to indicate that it only reflects that last update, and I _think_ that's the least surprising implementation (or, otherwise, someone is using the `XmlMerge` class incorrectly.

If you still think it needs to be changed, I can do that, but I wanted to have a discussion to see if I was missing some details, first.  Thanks!

galatrash-at-dnn:
I'n not talking about other installer or library running in parallel; I'm talking about having multiple attributes in one manifest file. We do this in many places where we add more than one attribute to merge and some could already be there from previous installation(s).

bdukes:
@galatrash-at-dnn, you can have an XML Merge file like [`Net35.config`](https://github.com/dnnsoftware/Dnn.Platform/blob/2691908bee3c383ec8eaaaf16c87c0e677f9a1b7/DNN%20Platform/Website/Install/Config/Net35.config) which performs many actions, but all of those actions will be triggered by one call to `UpdateConfig`, which in turn calls [`ProcessNodes`](https://github.com/dnnsoftware/Dnn.Platform/blob/848d7f96c2023de7a1d410df465c73b47633beeb/DNN%20Platform/Library/Services/Installer/XmlMerge.cs#L311-L374).  That call uses `||` to keep track of whether a change happened via _any_ node in the XML merge file, so the `bool` returned and used to set `ConfigUpdateChangedNodes` takes that into account, for each call to `UpdateConfig` (or `UpdateConfigs`).

I hope that clarifies what I'm thinking.  Again, let me know if I'm missing a scenario that needs to be covered.

galatrash-at-dnn:
@bdukes, the code you are referring to is long gone. Here is the new code for [ProcessNodes](https://github.com/dnnsoftware/Dnn.Platform/blob/968d97a16277c5fb3de2ebc97dfadcb011f52c3a/DNN%20Platform/Library/Services/Installer/XmlMerge.cs#L297) which doesn't use any || and doesn't return any bool result. Maybe this code should be reintroduced back!

bdukes:
@galatrash-at-dnn, I was linking to the change introduced by this PR.  The `bool` return is new from my changes, and takes into account multiple changes from one XML Merge document

bdukes:
FYI, I've rebased this against `development` and have pushed that change. 

Let me know if there's still an issue with missing updates. If it would be easier to chat, let me know, and we can figure out the right channel for that. 

Thanks!

galatrash-at-dnn:
@bdukes I've sent a request to connect with you through Skype.

DNN Platform/Library/Services/Installer/XmlMerge.cs
```diff
@@ -505,8 +543,11 @@ private void UpdateNode(XmlNode rootNode, XmlNode actionNode)
                     switch (collisionAction.ToLowerInvariant())
                     {
                         case "overwrite":
+                            var oldContent = rootNode.InnerXml;
                             rootNode.RemoveChild(targetNode);
                             rootNode.InnerXml = rootNode.InnerXml + child.OuterXml;
+                            var newContent = rootNode.InnerXml;
+                            changedNode = !string.Equals(oldContent, newContent, StringComparison.Ordinal);
                             break;
                         case "save":
```
galatrash-at-dnn: This case will always return the nodes as not equal and must be changed to:
```
                        case "overwrite":
                            var oldXml = targetNode.OuterXml;
                            var newXml = child.OuterXml;
                            changedNode = !string.Equals(oldXml, newXml, StringComparison.Ordinal);
                            if (changedNode)
                            {
                                rootNode.InnerXml = rootNode.InnerXml.Replace(oldXml, newXml);
                            }
                            break;
```

PR #1452, opened 2016-04-17
Dnn 8613

sleupold:
Fix ViewingUserID and make sure, UserID in GetUsers is used consistently.


zyhfish:
Hi @stephen-lim , i have been able to valid this problem and describe the reproduce steps on JIRA. but need you take some work here:
1. resolve the PR conflicts so that we possible to process it;
2. check my inline comments to see whether need change.

thx so much for your help!

DNN Platform/Modules/MemberDirectory/Services/MemberDirectoryController.cs
```diff
@@ -138,7 +138,7 @@ private IEnumerable<UserInfo> GetUsers(int userId, int groupId, string searchTer
                     }
                     if (CanViewGroupMembers(portalId, groupId))
                     {
-                        users = UserController.Instance.GetUsersAdvancedSearch(portalId, userId, -1,
+                        users = UserController.Instance.GetUsersAdvancedSearch(portalId, PortalSettings.UserId, userId,
```
zyhfish: The filterUserId parameter only used when search type is "Relationship", so IMO it can keep use -1 here.

DNN Platform/Modules/MemberDirectory/Services/MemberDirectoryController.cs
```diff
@@ -159,7 +159,7 @@ private IEnumerable<UserInfo> GetUsers(int userId, int groupId, string searchTer
                     var propertyValue = GetSetting(ActiveModule.ModuleSettings, "FilterPropertyValue", String.Empty);
                     AddSearchTerm(ref propertyNames, ref propertyValues, filterValue, propertyValue);
 
-                    users = UserController.Instance.GetUsersAdvancedSearch(portalId, userId, -1, -1,
+                    users = UserController.Instance.GetUsersAdvancedSearch(portalId, PortalSettings.UserId, userId, -1,
```
zyhfish: The filterUserId parameter only used when search type is "Relationship", so IMO it can keep use -1 here.

DNN Platform/Modules/MemberDirectory/Services/MemberDirectoryController.cs
```diff
@@ -169,7 +169,7 @@ private IEnumerable<UserInfo> GetUsers(int userId, int groupId, string searchTer
                                                                            sortField, (sortOrder == "ASC"),
                                                                            "DisplayName", searchTerm)
                                                                            :
-                                                                           UserController.Instance.GetUsersAdvancedSearch(portalId, PortalSettings.UserId, -1, -1,
+                                                                           UserController.Instance.GetUsersAdvancedSearch(portalId, PortalSettings.UserId, userId, -1,
```
zyhfish: The filterUserId parameter only used when search type is "Relationship", so IMO it can keep use -1 here.

zyhfish:
Hi @stephen-lim , i have been able to valid this problem and describe the reproduce steps on JIRA. but need you take some work here:
1. resolve the PR conflicts so that we possible to process it;
2. check my inline comments to see whether need change.

thx so much for your help!

stephen-lim:
@zyhfish Sorry I don't remember what this change is anymore. It looks like something that was done back in 2015? I don't think it's reasonable to merge a pull after this long and expect to succeed. Maybe you can back out my changes? :(

galatrash-at-dnn:
As this had conflict ans based on outdated source code, I picked the change locally in my machine and committed it directly (commit [a823b03](https://github.com/dnnsoftware/Dnn.Platform/commit/a823b03cc9391e477359ff8171c9bb8bbe58c1b4))

PR #1422, opened 2016-03-29
DNN-8564: Performing Hard Tab Deletion, messes up TabOrder of other portals

JoshuaBradley:
See Jira Ticket: [https://dnntracker.atlassian.net/browse/DNN-8564](https://dnntracker.atlassian.net/browse/DNN-8564)

PR #1453, opened 2016-04-17
DNN-8564: Performing Hard Tab Deletion, messes up TabOrder of other portals

sleupold:
fix updating page order for top level pages


galatrash-at-dnn:
There is already an earlier PR for this same issue here https://github.com/dnnsoftware/Dnn.Platform/pull/1422

bdukes:
This change doesn't handle a `NULL` value for `@PortalId`, whereas #1422 does (otherwise they're functionally the same)

zyhfish:
i merge the changes from #1422 into new PR #1893. close this one.

PR #1454, opened 2016-04-17
DNN-4350

sleupold:
Procedure DeletePortalAlias needs to update TabUrls


zyhfish:
Hi @sleupold , i confirm that DNN-4350 is valid and assign that issue into 9.2 scope, but for this PR its not able to merge as the change is not valid for below scenario:
1. if we create 2 alias A and B,  for page, i create 2 URL U1 set alias to A, U2 set alias to B.
2. i delete alias A, your changes will both update U1 and U2's PortalAliasId to NULL, but what we expect is only update U1, U2 should still reference to alias B.
3. also the update should limit in current portal but not affect other portals.

BTW, this PR changes in old sqldataprovider file which won't valid, can you please submit a new PR to change in 9.2.0 sqldataprovider file? per these, i will close this PR for now, thx so much!

PR #1456, opened 2016-04-17
DNN-7719: Improve indexes on Lists table to allow same list name in multiple portals

sleupold:
see Jira ticket for details

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
@@ -166,3 +166,54 @@ AS
 
 	EXEC {databaseOwner}{objectQualifier}EnsureLocalizationExists @PortalID = @PortalId, @CultureCode = @CultureCode
 GO
+
+
+-- DNN-7719: Improve indexes on Lists table to allow same list name in multiple portals
+IF EXISTS (SELECT * FROM sys.indexes WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}Lists]') AND name = N'IX_{objectQualifier}Lists_ListName_ParentID')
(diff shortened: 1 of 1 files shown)


Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
```diff
@@ -166,3 +166,54 @@ AS
...
+GO
+
+-- remove orphaned entries
+DELETE FROM {databaseOwner}[{objectQualifier}Lists] 
+WHERE (DefinitionID NOT IN (SELECT ModuleDefID FROM {databaseOwner}[{objectQualifier}ModuleDefinitions]) AND DefinitionID > 0)
+   OR (PortalID NOT IN (SELECT PortalID FROM {databaseOwner}[{objectQualifier}Portals]) AND PortalID >= 0)
+GO
+
+-- ensure that there are no duplicates before creating the index
+DELETE FROM {databaseOwner}[{objectQualifier}Lists]
```
IDisposable: Whoa, that could be a very surprising result of an upgrade attempt... total loss of an earlier (presumably used) list and retention of only the last one added?  I don't know the whole context, obviously... but would it be better to just rename the duplicate lists (ala, the count trick) and then let the end user prune them down after the upgrade?
```
SELECT EntryID, NewName=ListName + COUNT(*)
FROM  {databaseOwner}[{objectQualifier}Lists] 
GROUP BY PortalID, ParentID, ListName, [Value]
```

sleupold: no, those items were never been able to be retrieved by the API.

galatrash-at-dnn:
This PR needs to be rebased on the latest source code to be able to process.

sleupold:
I know, it's on my list..

sleupold:
I'll review and resubmit asap

mitchelsellers:

PR #1461, opened 2016-04-17
DNN-4608: use new virtual roles

sleupold:


galatrash-at-dnn:
This issue is already closed based on Will's comment here https://dnntracker.atlassian.net/browse/DNN-4608?focusedCommentId=116531&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-116531

PR #1463, opened 2016-04-17
DNN-6696: avoid unnecessary db compatibility level upgrade 

sleupold:
level should only upgraded, if less than desired value (not equal).
There have been a number of users reporting errors due to missing permission - this change will not solve this problem, but at least reduce it.


galatrash-at-dnn:
@sleupold, I'm wondering how much this makes a difference. If the value is 100 and you set it to 100 again, then what harm doe this do? I don't think this worth fixing. Besides, the issue reported in DNN-6696 (closed at the moment) is due to the db user having insufficient rights to perform ``ALTER DATABASE ... ``.

I'll close this without merging unless there is a strong case otherwise.

sleupold:
[~galatrash-at-dnn],
the issue occurs, if the database user account has not been granted dbowner permission and is unable to set the compatibility level - in this case he gets an unnecessary error, if the code tries to update the level (to same value)

sleupold:
I will resubmit asap

mitchelsellers:

PR #1464, opened 2016-04-17
DNN-8358: Improve Journal_Save to support updates

sleupold:


Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
```diff
@@ -166,3 +166,65 @@ AS
...
+    @ItemData           nVarChar(2000),
+    @JournalXML         xml,
+    @ObjectKey          nVarChar(255),
+    @AccessKey          UniqueIdentifier,
+    @SecuritySet        nVarChar(2000),
+    @CommentsDisabled   Bit,
+    @CommentsHidden     Bit
+AS
+BEGIN
+    IF IsNull(@JournalId, -1) = -1 BEGIN
```
IDisposable: New to the DNN way of doing things... but can't we wrap this in a `BEGIN TRAN` / `COMMIT TRAN` so we don't end up modifying the `Journal`, then not finishing the `JournalSecurity` rows?

Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
```diff
@@ -166,3 +166,65 @@ AS
...
+			   Summary          = @Summary,
+			   ObjectKey        = @ObjectKey,
+			   AccessKey        = @AccessKey,
+			   ItemData         = @ItemData,
+			   CommentsHidden   = @CommentsHidden,
+			   CommentsDisabled = @CommentsDisabled
+		 WHERE JournalId = @JournalId;
+
+    IF @JournalID IS NOT NULL BEGIN
+		DELETE
```
IDisposable: Wouldn't a MERGE be much safer here?

Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
sleupold: DNN never uses transactions, usually it uses single statements per procedure.
The Journal is one of the most Anti-Database pieces in the structure - storing relations compacted in a single text column is just against all concepts. There are no foreign keys defined and I don't really see a reason to create a dedicated table for a single xml column.
My fix here is just to solve this one issue - because Active Forums module needs to delete and recreate a journal item upon every update, which is not reliable.

sleupold:
I have to be careful using merge statements in DNN scripts; I got the feedback that they are too complex for the developers.

jbrinkman:
This will require significant review as it changes a fundamental behavior of the Journal.

sleupold:
there is no significant change, just avoiding the need to delete an journal item and recreate it, once the content gets updated (e.g. like Active Forums does, which doesn't seem to be reliable - there are a number of records deleted and not recreated). Joe, if you prefer, I can rewrite  the procedure just to use a merge instead of delete and create.

cpaterra:
We fixed a bug around this some time ago and there is the ability to update a journal item already. I don't think it is necessary to change the Journal API, instead change Active Forums implementation. Please see: https://dnntracker.atlassian.net/browse/DNN-4253

PR #1477, opened 2016-04-22
Properly handle SearchResult XSS issues.

IDisposable:
The `SearchResult.ascx` page is susceptible to XSS when the search string has single-quotes (breaks JSON at page bottom). This patch fixes the issues the right way; using `HtmlAttributeEncode` when emitting user-supplied values (when echoing back the `SearchTerm` or `TagsQuery` values in the input fields) and `JavaScriptStringEncode` when emitting the `SeachTerm` in the `dnn.searchResult.queryOptions` script.

This also simplifies the code behind to null-coalesce the query parameters and not do the encoding on the input values (instead encoding them on the rendering out, as it should be)


zyhfish:
changes looks good and merge, btw, for JavaScriptStringEncode, we want to use Localization.GetSafeJSString instead of it.

PR #1484, opened 2016-05-03
DNN-6221: update service api path format to API/{module name}/Action/{controller}/{action}.

zyhfish:


mitchelsellers:
I like this change as it makes the routes shorter, however, this is a breaking change for ANY module that is useing API calls.  Is there any thought as to how this will be communicated?

DNN Platform/Library/Entities/Urls/RewriterUtils.cs
```diff
@@ -35,7 +35,7 @@ namespace DotNetNuke.Entities.Urls
 {
     public class RewriterUtils
     {
-        internal static readonly Regex ServicesFrameworkRegex = new Regex("DesktopModules/.+/API/", RegexOptions.Compiled);
+        internal static readonly Regex ServicesFrameworkRegex = new Regex("desktopModules/.+/api/|api/.+/action/", RegexOptions.Compiled);
```
galatrash-at-dnn: This is a breaking change as the other products/module are using it. It is better to add RegexOptions.IgnoreCase option than using ToLowerInvariant() all over the place.

galatrash-at-dnn: Why this "/API/.+/Action" ? Why not "/API/Action/" ? The latter would be faster to match against.

DNN Platform/Tests/DotNetNuke.Tests.Integration/Tests/Jwt/JwtAuthTest.cs
```diff
@@ -51,13 +51,13 @@ public class JwtAuthTest : IntegrationTestBase
...
-        private const string TestPostQuery = "/DesktopModules/JwtAuth/API/mobile/testpost";
-        private const string GetMonikerQuery = "/DesktopModules/web/API/mobilehelper/monikers?moduleList=";
-        private const string GetModuleDetailsQuery = "/DesktopModules/web/API/mobilehelper/moduledetails?moduleList=";
+        private const string LoginQuery = "/API/JwtAuth/Action/mobile/login";
+        private const string LogoutQuery = "/API/JwtAuth/Action/mobile/logout";
+        public const string ExtendTokenQuery = "/API/JwtAuth/Action/mobile/extendtoken";
+        private const string TestGetQuery = "/API/JwtAuth/Action/mobile/testget";
+        private const string TestPostQuery = "/API/JwtAuth/Action/mobile/testpost";
+        private const string GetMonikerQuery = "/API/web/Action/mobilehelper/monikers?moduleList=";
+        private const string GetModuleDetailsQuery = "/API/web/Action/mobilehelper/moduledetails?moduleList=";
```
galatrash-at-dnn: You should keep both forms of the URL's in the tests so we validate that backward compatibility is not broken.

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/PortalAliasRouteManagerTests.cs
```diff
@@ -47,10 +47,10 @@ public void TearDown()
...
         [Test]
-        [TestCase("mfn", "url", 0, "DesktopModules/mfn/API/url")]
-        [TestCase("mfn", "url", 1, "{prefix0}/DesktopModules/mfn/API/url")]
-        [TestCase("mfn", "url", 2, "{prefix0}/{prefix1}/DesktopModules/mfn/API/url")]
-        [TestCase("fee/foo", "{contoller}/{action}/{id}", 4, "{prefix0}/{prefix1}/{prefix2}/{prefix3}/DesktopModules/fee/foo/API/{contoller}/{action}/{id}")]
+        [TestCase("mfn", "url", 0, "API/mfn/Action/url")]
+        [TestCase("mfn", "url", 1, "{prefix0}/API/mfn/Action/url")]
+        [TestCase("mfn", "url", 2, "{prefix0}/{prefix1}/API/mfn/Action/url")]
+        [TestCase("fee/foo", "{contoller}/{action}/{id}", 4, "{prefix0}/{prefix1}/{prefix2}/{prefix3}/API/fee/foo/Action/{contoller}/{action}/{id}")]
         public void GetRouteUrl(string moduleFolderName, string url, int count, string expected)
```
galatrash-at-dnn: Same thing here; keep URL formats (old and new) to test for backward compatibility.

galatrash-at-dnn:
Did you apply this to other products (EVOQ Content and Engage)? They are using the API extensively and need to be changed as well.

DNN Platform/Library/Entities/Urls/RewriterUtils.cs
zyhfish: Hi @galatrash-at-dnn , this change is try to resolve the issue which introduced in DNN-8347, the service module won't match the regex in any case. if RegexOptions.IgnoreCase doesn't affect performance, i will change it to RegexOptions.IgnoreCase, can you help me confirm? thanks,

DNN Platform/Tests/DotNetNuke.Tests.Integration/Tests/Jwt/JwtAuthTest.cs
zyhfish: this is for JWT project which use new format now, so it doesn't need backward check in this test.

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/PortalAliasRouteManagerTests.cs
zyhfish: this test is for GetRouteUrl, it always return new format, the backward compatibility implement in ServicesRoutingManager, add a new tes in ServicesRoutingManagerTest to check backward compability.

DNN Platform/Library/Entities/Urls/RewriterUtils.cs
galatrash-at-dnn: That's not true if you include the forward slashes in the Regex as  DNN-8347 will not be valid in that case.

As for ignoring the case, yes it has a slight overhead but if you are matching without the wild card (`.+`), it won't degrade the performance by any noticeable/measurable value.

DNN Platform/Tests/DotNetNuke.Tests.Integration/Tests/Jwt/JwtAuthTest.cs
galatrash-at-dnn: JWT is out there for a couple of releases (since the beginning of this year) and you can't say it is not being used until now. It is a must to guarantee backward compatibility since day 1 after releasing it to the public.

galatrash-at-dnn:
Closed this as it has conflicts with latest code-base (Platform 08.00.05). I will create a new branch in the most recent code.

PR #1487, opened 2016-05-04
Refix the script injection issue on Search

IDisposable:
As documented in the comment for commit 6216b09acc41b59ab9a4655a25737fe9ad5f0ab4 `Localization.GetSafeJSString` is not enough to prevent issues with XSS injection. We should really replace ALL calls with the safer `HttpUtility.JavaScriptStringEncode` method, but for now at least let's fix the reported site.


IDisposable:
If we merge in #1492 then this pull isn't required, however this is an URGENT XSS and really aught to be merged ASAP if not.

jbrinkman:
Since #1492 was accepted and merged I am closing this issue.

PR #1488, opened 2016-05-11
DNN-8664: Replace Localization.GetSafeJSString w/ HttpUtility.JavaScriptStringEncode

IDisposable:
Since we've established in Pull request 1487 that `Localization.GetSafeJSString` actually isn't... replacing all occurrences with `HttpUtility.JavaScriptStringEncode` (which is provably safer)


IDisposable:
Note that I haven't added a deprecation to the `Localization.GetSafeJSString`, but it really actually should be deprecated. If you want that to be part of this pull request, please advise.  Alternatively, we could just have it be implemented as a trivial shim call to the `HttpUtility.JavaScriptStringEncode`

bdukes:
Yeah, it seems like the simpler path forward would be to make `GetSafeJSString` actually be safe (by making it delegate to `JavaScriptStringEncode`), unless there's an expected behavioral difference between the two calls (there's none that I'm aware of).

IDisposable:
The biggest behavior difference (other than fixing the partial escaping problems) is that `GetSafeJSString` **also** morphs CR and LF characters to blanks (which seems odd, but given the way it works it has to) while `JavaScriptStringEncode` emits them as unicode escape sequences.

bdukes:
I'm fine with encoding line breaks, if anyone else has strong opinions, we can strip line breaks then call `JavaScriptStringEncode` from `GetSafeJSString`

Website/DesktopModules/Admin/Security/editroles.ascx
```diff
@@ -18,7 +18,7 @@
                     <asp:TextBox ID="txtRoleName" runat="server" MaxLength="50" />
                     <asp:Label ID="lblRoleName" Visible="False" runat="server" />
                     <asp:RequiredFieldValidator ID="valRoleName" CssClass="dnnFormMessage dnnFormError" runat="server" resourcekey="valRoleName" ControlToValidate="txtRoleName" Display="Dynamic" />
-                    <asp:RegularExpressionValidator ID="valRoleName2" CssClass="dnnFormMessage dnnFormError" runat="server" resourcekey="valRoleName2" ControlToValidate="txtRoleName" Display="Dynamic" ValidationExpression="[^&\$\+,/?~#<>����!\.:\*'\[\]]*" />
+                    <asp:RegularExpressionValidator ID="valRoleName2" CssClass="dnnFormMessage dnnFormError" runat="server" resourcekey="valRoleName2" ControlToValidate="txtRoleName" Display="Dynamic" ValidationExpression="[^&\$\+,/?~#<>����!\.:\*'\[\]]*" />
```
IDisposable: DANGER This is a broken encoding. Probably need to resubmit patch with this fixed. Sorry

Website/DesktopModules/Admin/ViewProfile/ViewProfile.ascx.cs
```diff
@@ -1,6 +1,6 @@
 #region Copyright
 // 
-// DotNetNuke� - http://www.dotnetnuke.com
+// DotNetNuke� - http://www.dotnetnuke.com
```
IDisposable: Encoding issue, again. Sorry

Website/admin/Skins/Search.ascx.cs
```diff
@@ -1,5 +1,5 @@
 // 
-// DotNetNuke� - http://www.dotnetnuke.com
+// DotNetNuke� - http://www.dotnetnuke.com
```
IDisposable: Encoding issue again. Sorry

IDisposable:
What about the `ClientAPI.GetSafeJSString` variant?

IDisposable:
Squashed all the commits.

DNN Platform/Admin Modules/Dnn.Modules.ConfigurationManager/XmlMerge.ascx
```diff
@@ -58,9 +58,9 @@
 	jQuery(function($) {
 		$('#<%=cmdSave.ClientID%>, #<%= cmdExecute.ClientID %>').dnnConfirm({
 			text: '<%= ConfirmText %>',
-			yesText: '<%= Localization.GetSafeJSString("Yes.Text", Localization.SharedResourceFile) %>',
-			noText: '<%= Localization.GetSafeJSString("No.Text", Localization.SharedResourceFile) %>',
-			title: '<%= Localization.GetSafeJSString("Confirm.Text", Localization.SharedResourceFile) %>'
+			yesText: '<%= HttpUtility.JavaScriptStringEncode("Yes.Text", Localization.SharedResourceFile) %>',
+			noText: '<%= HttpUtility.JavaScriptStringEncode("No.Text", Localization.SharedResourceFile) %>',
+			title: '<%= HttpUtility.JavaScriptStringEncode("Confirm.Text", Localization.SharedResourceFile) %>'
```
bdukes: When a resource file path is passed as the second parameter, `Localization.GetString` is called on the values before they're "encoded" (see https://github.com/dnnsoftware/Dnn.Platform/blob/f50786674d8e9e2566189f4cdf80e24c6f840e02/DNN%20Platform/Library/Services/Localization/Localization.cs#L1263)

IDisposable: Yep, this is an error, I'm recommending we scrap this whole pull request if people are okay with a wholesale swap anyway and use pull #1492

IDisposable:
If everyone is okay with just swapping out the implementation of Localization.GetSafeJSString, let's just scrap this pull request and merge in @bdukes pull #1492

PR #1490, opened 2016-05-13
DNN-8637: Updated folder provider to work w/ top level folder renaming

kanm2015:


zyhfish:
changes looks good, merged.

PR #1491, opened 2016-05-13
DNN-8663: add file download event.

zyhfish:


kanm2015:
Could you please rename all FileDownload to FileDownloaded so we follow the events naming convention? And why you fire the new event in FileServerHandler and DigitalAssetsController but none of other events are there?

zyhfish:
the name reanamed to FileDownloaded. the event talked with Chris, it only called in download action from DAM module, and called in linkclick url, the event will subscribed in evoq side.

kanm2015:
Code review done

PR #1492, opened 2016-05-13
DNN-8664 Update GetSafeJSString to use built-in encoder

bdukes:
The current implementations are woefully incomplete and can cause breakages and
security risks

This is an alternate implementation of #1488, thanks to @IDisposable for taking on the initial work for fixing this!  I'm suggesting this implementation so that :one: other users of the `GetSafeJSString` methods are protected, not just usages within DNN, :two: to simplify the change and not have to touch as many core files, and :three: to avoid having to distinguish between the two overloads of `Localization.GetSafeJSString`.

[DNN-8664](https://dnntracker.atlassian.net/browse/DNN-8664)
>Through testing and correction of an XSS vulnerability found in the SearchResults module, we've discovered that the library routine used to escape strings for inclusion as a javascript variable in a script block does NOT safely escape HTML tags. This means that submitting an input that contains `</script>` will end the surrounding `<script>` block early. This means that arbitrary HTML can be injected, including other `<script>` blocks by user input. Rather than piecemeal fix this, I'm suggesting a wholesale swap-out from the `Localization.GetSafeJSString` to the Microsoft provided `HttpUtility.JavaScriptStringEncode` which is provably much safer.

### DIFF
--- a/DNN Platform/Library/Services/Localization/Localization.cs
+++ b/DNN Platform/Library/Services/Localization/Localization.cs
@@ -122,8 +122,6 @@ namespace DotNetNuke.Services.Localization
         //private static readonly ILocaleController LocaleController.Instance = LocaleController.Instance;
         //private static readonly ILocalizationProvider _localizationProvider = LocalizationProvider.Instance;
         private static bool? _showMissingKeys;
-
-        private static readonly Regex UnsafeJsRegex = new Regex("(['\"\\\\])", RegexOptions.Compiled);
         #endregion
 
         #region Public Shared Properties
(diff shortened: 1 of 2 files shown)


IDisposable:
Much simpler than my more sweeping change, one caveat occurred to me.. the existing implementation will morph a null to a String.Empty, this one will throw.

DNN Platform/Library/Services/Localization/Localization.cs
```diff
@@ -1245,13 +1243,7 @@ public static string GetStringUrl(string key, string resourceFileRoot)
...
         public static string GetSafeJSString(string unsafeString)
         {
-            var safeString = string.IsNullOrEmpty(unsafeString) ? "" : UnsafeJsRegex.Replace(unsafeString, "\\$1");
-			if (safeString.Length > 0)
-	        {
-				safeString = safeString.Replace("\r", string.Empty).Replace("\n", string.Empty);
-	        }
-
-			return safeString;
+            return HttpUtility.JavaScriptStringEncode(unsafeString);
```
IDisposable: This will throw a `NullReferenceException` if passed a null, the existing code will return an empty string.

bdukes:
@IDisposable :+1: I've updated both methods to return `string.Empty` when passed a `null`

IDisposable:
:shipit:

bdukes:
@jbrinkman This is a security issue that needs to be merged into 8.1.0, can you help ensure that happens before code freeze?  Thanks!

zyhfish:
Changes looks good, merged and will do more test with it.

PR #1493, opened 2016-05-15
DNN-8659 Change HostSettings and PortalSettings SettingValue column size

sleupold:
Adjust SettingValue column size for HostSettings, PortalSettings and RoleSettings and use same procedure construct for all settings tables.

### DIFF
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -779,10 +779,6 @@ namespace DotNetNuke.Data
                                         tab.IsSystem);
         }
 
-		public virtual void AddTabSetting(int TabId, string SettingName, string SettingValue, int createdByUserID)
-		{
-			ExecuteNonQuery("AddTabSetting", TabId, SettingName, SettingValue, createdByUserID);
-		}
 
         public virtual int AddTabToEnd(TabInfo tab, int createdByUserID)
         {
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -451,12 +451,6 @@ namespace DotNetNuke.Data
 
 		#region Host Settings Methods
 
-		public virtual void AddHostSetting(string SettingName, string SettingValue, bool SettingIsSecure,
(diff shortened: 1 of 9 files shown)


Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
```diff
@@ -166,3 +162,389 @@ AS
...
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}UpdateHostSetting]
+	@SettingName	   nvarchar( 50),
+	@SettingValue	   nvarchar(Max),
+	@SettingIsSecure   bit,
+	@UserID	           int
+AS
+BEGIN
+	UPDATE {databaseOwner}[{objectQualifier}HostSettings]
```
bdukes: Doesn't the `MERGE` below replace this `UPDATE`?

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -451,12 +451,6 @@ public virtual string GetUnappliedIterations(string version)
 
 		#region Host Settings Methods
 
-		public virtual void AddHostSetting(string SettingName, string SettingValue, bool SettingIsSecure,
```
bdukes: Is there not a backwards compatibility concern with removing these methods (and their corresponding stored procedures)?

Website/Providers/DataProviders/SqlDataProvider/08.01.00.01.SqlDataProvider
sleupold: you are right, I used the update to update the correct column and forgot to remove the statement

DNN Platform/Library/Data/DataProvider.cs
sleupold: Brian, 
the stored procedures are not part of the API, nor the DataProvider methods.
Only the controllers and objects, which still use same signatures and functionality.

sleupold:
changes should be complete now. Using same functionality and patterns for all settings tables (Site Settings is a bit ahead, as it already supports multiple languages).

PR #1494, opened 2016-05-16
DNN-7483: reset app start elapse time after install site complete, set the write lock file exception logged in Debug level.

zyhfish:


ashishpd:
This change should remove unnecessary write.lock errors during startup / installation.

PR #1496, opened 2016-05-18
Limit menu action ul selector to first ul

dafergu2:
DNN-8678: When using a <ul> element in a quick settings view, the jquery
ul selector would apply to all ul elements.  Since showMenu() applied
styles by calling .css(), the menu placement adjustments being performed
would incorrectly be applied to <ul> elements internal to the quick
settings view. This limits the jquery object to the first <ul>.


mitchelsellers:
I have tested this fix on multiple installations.  No negative impacts.

@jbrinkman Can we merge this one in?

PR #1499, opened 2016-05-24
DNN-8694: add user updated event handlers.

zyhfish:


DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -1691,6 +1694,8 @@ public override void UpdateUser(UserInfo user)
                                      user.IsDeleted,
                                      UserController.Instance.GetCurrentUserInfo().UserID);
 
+            EventManager.Instance.OnUserUpdated(new UpdateUserEventArgs {User = user, OldUser = oldUser});
```
kanm2015: Should we fire this update event from UserController like other user events?

kanm2015:
Code review done

PR #1503, opened 2016-05-30
DNN-8668: Event Viewer: Allow for simplified Purge procedure

sleupold:
Purge based on "keep for days" added and also purge events for disabled log entries.


zyhfish:
I will close this PR and submite a new one, as this PR have 2 areas can't merge:
1. it changed some other area which not affect to this issue;
2. it will both delete logs by keep days and keep number condition if keep days value set, this doesn't match the requirement.

sleupold:
the other changes are more or less formatting issues, however the procedure fulfills the requirements while providing consistency with UI settings. please consider to include the procedure as provided, thanks.

PR #1512, opened 2016-06-16
DNN-8766: Move modules still doesn't work correctly after DNN-7099 fix

roman-yagodin:
This will fix the remaining issues with moving modules remaining after DNN-7099 bugfix.

The original DNN-7099 fix checks for all deleted modules on the page while calculating corrected module order value, resulting moving up or down inside a pane with huge steps - depending on total number of deleted modules on the page. On the other hand, if there are only one pane (or just one pane with deleted modules) it works OK.

The solution is to count only deleted modules on same (target) pane while calculating corrected module order value.

> **Edit 06/23/2016:**
> 
> To reproduce the issue:
> 1. create new page with a skin having 2 or more panes;
> 2. create 4 modules on content pane;
> 3. create at least one module on any other pane (e.g. side pane);
> 4. delete module from side pane;
> 5. try to move first module on content pane down.
> 
> Expected result on step 5: module move to second place on the pane.
> 
> Actual result on step 5: module move to third place on the pane.
> 
> Workaround (same for DNN-7099): empty deleted modules from recycle bin for the page.

### DIFF
--- a/DNN Platform/DotNetNuke.Web/InternalServices/ModuleServiceController.cs
+++ b/DNN Platform/DotNetNuke.Web/InternalServices/ModuleServiceController.cs
@@ -1,6 +1,6 @@
 #region Copyright
 
-// DotNetNuke® - http://www.dotnetnuke.com
+// DotNetNukeÂ® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2016
 // by DotNetNuke Corporation
 // 
@@ -20,6 +20,7 @@
 
 #endregion
 
+using System;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


jbrinkman:
This needs further review and discussion to ensure we properly deal with all use cases.  There has been a lot of churn in this behavior and we want to ensure that we deal with this once and for all rather than piecemeal.

jbrinkman:
This needs a bigger story and review.  We'll close this item and do a more complete fix and still credit this fix to Roman.

roman-yagodin:
Branch rebased - added missing `using System` directive.

roman-yagodin:
Closing this, as DNN-8766 should be resolved by merged PR #1546

PR #1520, opened 2016-07-17
Changing the display of the login screen! (#1)

mauriliofilho:
Change in position of the login buttons , for better positioning and navigation experience !


galatrash-at-dnn:
@mauriliofilho , is there a Jira issue for this change? What is the justification. When I applied your change and compared the login screen layout, I get the following (
before the change
after the change

This change will not be accepted by our designer; therefore, I'm closing this PR without merging it.

PR #1519, opened 2016-07-15
DNN-8817: Fixed System.NullReferenceException in UpdateRole() CreateRoleGroup(), UpdateRoleGroup() and CreateRole()

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-8817 for more details...

PR #1521, opened 2016-07-19
DNN 8817

DNNMonster:
Adding null checks for role description create and update methods.


DNNMonster:
Sorry.  I didn't notice.  https://github.com/dnnsoftware/Dnn.Platform/pull/1519/commits/b1bd888c6bb6ce35bc70700720b9c2633242a59c

Looks like a more elegant fix.

PR #1401, opened 2016-03-14
DNN-8513: Create a new extension point in the DNN core at the Database Connection Level.

JoshuaBradley:
Can use this package for testing:[https://github.com/JoshuaBradley/DnnAzureSqlConnectionProvider](https://github.com/JoshuaBradley/DnnAzureSqlConnectionProvider)

PR #1522, opened 2016-07-27
DNN-8675: Add Support for .woff2 font format

sleupold:
still open:  merge it into web.config

### DIFF
--- a/DNN Platform/Library/Services/FileSystem/FileContentTypeManager.cs
+++ b/DNN Platform/Library/Services/FileSystem/FileContentTypeManager.cs
@@ -103,7 +103,8 @@ namespace DotNetNuke.Services.FileSystem
 			contentTypes.Add("svg", "image/svg+xml");
 			contentTypes.Add("ttf", "font/ttf");
 			contentTypes.Add("eot", "application/vnd.ms-fontobject");
-			contentTypes.Add("woff", "application/font-woff");
+			contentTypes.Add("woff", "font/font-woff");
+			contentTypes.Add("woff2", "font/woff2");
 			contentTypes.Add("mp3", "audio/mpeg");
 			contentTypes.Add("wma", "audio/x-ms-wma");
 			contentTypes.Add("mpg", "video/mpeg");


vmasanas:
I think the correct mime type are:
application/font-woff
application/font-woff2

sleupold:
Vicenç,
please see http://www.w3.org/TR/WOFF2/

vmasanas:
Checked that but I don't see where it says the mime type is what you proposed.

sleupold:
see section http://www.w3.org/TR/WOFF2/#IMT:
Type name:         font
Subtype name:   woff2

galatrash-at-dnn:
There is a problem with this PR; the MIME type is not correct. Also, you need to update the web.config file with these static file contents for both new installations and upgraded ones.

If you can do these in quickly, I'll work on adding them to the upcoming release 9.2.0.

sleupold:
George, which mime type do you expect? according to W3C, the mime types should be font/woff and font/woff2 - or do you have a newer information?

galatrash-at-dnn:
Just looking at the draft but it is not clear this is final or just a suggestion to use ``font/*`` and whether the old ``application/font-...``` was still valid or deprecated. Do you have a link where this is explained in more detail?

sleupold:
As far as I can see, http://www.w3.org/TR/WOFF2 is the current W3C candidate recommendation, and using font/woff2 (http://www.w3.org/TR/WOFF2/#IMT)

sleupold:
for WOFF (1.0), see http://www.w3.org/TR/WOFF/#appendix-b, but noticing http://www.w3.org/Fonts/REC-WOFF-20121213-errata.html

sleupold:
Any other questions?

galatrash-at-dnn:
Okay, I agree with these changes. But, to make it complete, there is a need for having these in the web.config of both new/upgrade scenarios; we need migrations there.

sleupold:
I haven't done this before, will need to investigate

galatrash-at-dnn:
@sleupold look at this for an example on how this can be done
https://github.com/dnnsoftware/Dnn.Platform/pull/1401/files#diff-96aac40a9c17ff8f9671a92ad3615f09

mitchelsellers:
Closing this PR due to staleness and procedural changes to the PR process.  If this issue/contribution is still desired please resubmit, we apologize for any delay as the finalize the transition of PR management to the DNN Community.

If re-submitted please make note of the re-submission with the new PR and we will expedite review following the current procedures.

Additionally, for this PR the submission would need to include the changes both to file-system and web.config with a proper solution for upgrade/install.

PR #1523, opened 2016-08-15
Feature/Beacon on release 8.0.4 branch

galatrash-at-dnn:


ohine:
@ashishpd Endpoint is misspelled in quite a few places. Search for Epdpoint

DNN Platform/Library/Services/ImprovementsProgram/BeaconService.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        private string _beaconEndpoint;
+        private readonly SHA256 _sha256 = SHA256.Create();
+
+        public string GetBeaconEndpoint()
+        {
+            if (string.IsNullOrEmpty(_beaconEndpoint))
+            {
+                var ep = ConfigurationManager.AppSettings["ImprovementProgram.Epdpoint"];
```
ohine: Spelling issue. Epdpoint should be Endpoint.

DNN Platform/Website/Install/Config/08.00.04.config
```diff
@@ -0,0 +1,7 @@
+<configuration>
+  <nodes configfile="web.config">
+    <node path="/configuration/appSettings" action="update" key="key" collision="ignore">
+      <add key="ImprovementProgram.Epdpoint" value="https://bcn.dnnapi.com/beacon" />
```
ohine: Spelling issue. Epdpoint should be Endpoint.

Website/development.config
```diff
@@ -62,6 +62,7 @@
     <add key="PreserveLoginUrl" value="true" />
 	  <add key="loginUrl" value="~/Login.aspx" />
     <add key="ValidationSettings:UnobtrusiveValidationMode" value="None" />
+    <add key="ImprovementProgram.Epdpoint" value="https://dev-bcn.dnnapi.com/beacon" />
```
ohine: Spelling issue. Epdpoint should be Endpoint.

Website/release.config
```diff
@@ -62,6 +62,7 @@
     <add key="PreserveLoginUrl" value="true" />
 	  <add key="loginUrl" value="~/Login.aspx" />
     <add key="ValidationSettings:UnobtrusiveValidationMode" value="None" />
+    <add key="ImprovementProgram.Epdpoint" value="https://bcn.dnnapi.com/beacon" />
```
ohine: Spelling issue. Epdpoint should be Endpoint.

mitchelsellers:
I am a bit concerned about the performance implications of this feature implementation.  In a quick review of the code there are multiple places that perform various conditional checks to things such as IsInRole, and then the operation is repeated a second time if it is in fact actually enabled.  

Has there been any possible consideration to further streamline this operation to avoid the redundancy?  My primary concern is that edit mode is already an area of the system that is taxed quite heavily performance wise and I'm cautious on adding more work than necessary.

DNN Platform/Library/Services/ImprovementsProgram/BeaconService.cs
```diff
@@ -0,0 +1,117 @@
...
+            // p: Portal ID
+            // a: Portal Alias
+            // r: Role(s) - bitmask - see RolesEnum
+            // u: User ID - hashed
+            // f: filename/path (optional; unused here)
+
+            var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
+            var qparams = new Dictionary<string, string>
+            {
+                // Remember to URL ENCODE values that can be ambigious
```
bdukes: why not just URL encode all values to make sure they're always safe (e.g. in the `Select` on line 95)?

DNN Platform/Library/Services/ImprovementsProgram/BeaconService.cs
```diff
@@ -0,0 +1,117 @@
...
+                    (tabPath.StartsWith("//Admin") || tabPath.StartsWith("//Host"));
+            }
+
+            return enabled;
+        }
+
+        public string GetBeaconQuery(UserInfo user, string filePath = null)
+        {
+            var roles = 0;
+            if (user.UserID >= 0)
```
bdukes: rather than duplicating this between here and `GetUserRolesBitValues`, should this just be `var roles = (int)GetUserRolesBitValues(user)`?

DNN Platform/Website/Install/InstallWizard.aspx
```diff
@@ -666,7 +675,8 @@
                             databaseSecurity: "",
                             databaseUsername: "",
                             databasePassword: "",
-                            databaseRunAsOwner: null
+                            databaseRunAsOwner: null,
+                            dnnImprovementProgram: $('#<%= chkImprovementProgram.ClientID %>')[0].is(":checked") ? "Y" : "N"
```
bdukes: does this work?  `[0]` should return a DOM element (not jQuery-wrapped), which wouldn't have an `is` method…  if you take out the `[0]` it should be fine

ohine: This was fixed after the code review by a direct commit into the release branch. 
https://github.com/dnnsoftware/Dnn.Platform/commit/b0f64827d07286176d4e2bfc4c9d8a6c06fa1c6a

Website/admin/ControlPanel/ControlBar.ascx
```diff
@@ -435,6 +435,10 @@
         $('a#ControlBar_ViewInPreview').click(function() {
             <%=PreviewPopup() %>;
         });
+
+        <% if (IsBeaconEnabled) { %>
+        (new Image()).src = "<%= GetBeaconUrl() %>";
```
bdukes: this should be properly encoded (e.g. `(new Image()).src = "<%=Localization.GetSafeJSString(GetBeaconUrl())%>"`

PR #1528, opened 2016-08-19
CONTENT-7377: fixing DB connections leaks or possible leaks.

galatrash-at-dnn:


DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -2094,11 +2095,10 @@ public void UpdateTabModuleOrder(int tabId)
         /// <remarks>empty SettingValue will relove the setting</remarks>
         public void UpdateTabModuleSetting(int tabModuleId, string settingName, string settingValue)
         {
-            IDataReader dr = null;
+            IDataReader dr = dataProvider.GetTabModuleSetting(tabModuleId, settingName);
```
mitchelsellers: Why not use a single using statement here?

zyhfish:
Changes looks good and merged into development branch.

PR #1529, opened 2016-08-23
DNN-8866: Prevent NullReferenceException in RemoveUnsafeAttributes()

janjonas:
See https://dnntracker.atlassian.net/browse/DNN-8866


galatrash-at-dnn:
This issue is no longer valid as the changed method doesn't exist in the source code now.

PR #1532, opened 2016-08-25
DNN-8867: improving folder permission caching.

galatrash-at-dnn:
Improving folder permission caching.


sleupold:
George, please remove "with (NoLock)" hints. I haven't been able to notice minimal performance improvements (tested on SQL Server 2012) and the risk to affect proper results is way to high and in a platform product like DNN, you cannot test all use cases. There are way better options to improve database performance than using critical edge case options like this.

zyhfish:
changes looks good, and merge into development branch.

PR #1533, opened 2016-08-26
DNN-8876: support large files

sleupold:
switch to long/bigint (and migrate database column type image to varbinary)


sleupold:
sorry, wrong files

PR #1536, opened 2016-08-31
Jira/dnn 8874

galatrash-at-dnn:


DNN Platform/Library/Common/Utilities/DataCache.cs
```diff
@@ -211,6 +211,7 @@ public class DataCache
         public const string UserProfileCacheKey = "UserProfile|{0}|{1}";
         public const int UserProfileCacheTimeOut = UserCacheTimeOut;
 
+        public const string UserNotificationsConversationCountCacheKey = "UserNitifConversationCount|{0}|{1}";
```
zyhfish: this cache seems useless with 2 points:
1. it will cause user won't receive notification even refresh page in 30 seconds;
2. if user stay on page, the ajax request every 30 seconds, that means cache will always expired in each request.

DNN Platform/Library/Entities/Content/Workflow/Entities/Workflow.cs
```diff
@@ -22,6 +22,7 @@
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using DotNetNuke.Common.Utilities;
```
zyhfish: looks like this using doesn't do anything.

DNN Platform/Library/Entities/Content/Workflow/Entities/WorkflowState.cs
```diff
@@ -20,6 +20,7 @@
 #endregion
 
 using System.ComponentModel.DataAnnotations;
+using DotNetNuke.Common.Utilities;
```
zyhfish: looks like this using doesn't do anything.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -42,13 +42,14 @@ public TabVersion GetTabVersion(int tabVersionId, int tabId, bool ignoreCache =
         
         public IEnumerable<TabVersion> GetTabVersions(int tabId, bool ignoreCache = false)
         {
-            //if we are not using the cache
+            //if we are not using the cache, then remove from cacehh and re-add loaded items when eeded later
+            var tabCacehKey = GetTabVersionsCacheKey(tabId);
```
zyhfish: typo here, should be cache, also typo in comment.

DNN Platform/Library/Services/FileSystem/FileManager.cs
```diff
@@ -899,7 +899,7 @@ public virtual IFileInfo GetFile(int fileID, bool retrieveUnpublishedFiles)
             var file = DataCache.GetCache(strCacheKey);
             if (file == null)
             {
-                file = CBO.Instance.FillObject<FileInfo>(DataProvider.Instance().GetFileById(fileID, retrieveUnpublishedFiles));
```
zyhfish: this and below changes same with this shouldn't change, otherwise will break all related unit tests.

DNN Platform/Library/Services/Social/Messaging/Data/DataService.cs
```diff
@@ -116,7 +116,7 @@ public int CreateMessageReply(int conversationId, int portalId, string body, int
         /// <returns>The count of recipients</returns>
         public int CheckReplyHasRecipients(int conversationId, int userId)
         {
-            return userId <= 0 ? 0 : _provider.ExecuteScalar<int>("CoreMessaging_CheckReplyHasRecipients", conversationId, userId);
```
zyhfish: these changes seems commit in 6a374e8f for performance stuff, should it really need remove?

Website/Providers/DataProviders/SqlDataProvider/08.00.05.01.SqlDataProvider
```diff
@@ -137,7 +137,7 @@ BEGIN
 	  PublishedVersion,  
 	  ContentItemID,
       HasBeenPublished
-	FROM {databaseOwner}[{objectQualifier}vw_Files]
```
zyhfish: this change shouldn't related to this item, right?

sleupold: this change needs to be reverted

DNN Platform/Library/Common/Utilities/DataCache.cs
galatrash-at-dnn: This is intentional and any page transitions withing 30 seconds will save some additional SQL requests.
This doesn't need any synchronization as it is meant to live for short time without the need to synchronize with other servers' caches.

DNN Platform/Library/Entities/Content/Workflow/Entities/Workflow.cs
galatrash-at-dnn: Maybe left there after trying some stuff then removing it - this doesn't affect anything.

DNN Platform/Library/Entities/Content/Workflow/Entities/WorkflowState.cs
galatrash-at-dnn: Maybe left there after trying some stuff then removing it - this doesn't affect anything.

DNN Platform/Library/Services/Social/Messaging/Data/DataService.cs
galatrash-at-dnn: YES, data layer should not be responsible about these checks.

DNN Platform/Library/Services/FileSystem/FileManager.cs
galatrash-at-dnn: This is a bad example of wsing"CBO.Instance.XXX"; there were only 6 places in the whole Platform source code that use CBO as an instance and it is used as a static object everywhere else (over 420 places). When we use the instance, it must be either all over the place or none as there is no meaning to use it this way in one place and write unit tests for this single case and leaving git all around the place as static object.

zyhfish: its not about unit tests for CBO, but other unit tests depend on CBO, and they will set mockup CBO when needed, like used in file manager tests.

zyhfish:

PR #1542, opened 2016-09-13
DNN-6221: removed '/action' from the API endpoints; it became as: '/a…

galatrash-at-dnn:
…pi/[module]/[controller]/[action]'


kanm2015:
Code review done

PR #1543, opened 2016-09-20
DNN-8900 Update Microsoft CodeDOM extension

bdukes:
See [DNN-8900](https://dnntracker.atlassian.net/browse/DNN-8900)

> A new version of Microsoft.CodeDom.Providers.DotNetCompilerPlatform has been released: https://blogs.msdn.microsoft.com/webdev/2016/09/20/announcing-the-dotnetcompilerplatform-1-0-2-release/, which improves performance and fixes a bug.


galatrash-at-dnn:
@bdukes this is an old PR. The latest CodeDOM library in now 1.0.7. Will you be able to update this PR with the latest one for me to process it?

galatrash-at-dnn:
@bdukes  any update?

bdukes:
@galatrash-at-dnn I intend to update this today

bdukes:
@galatrash-at-dnn I've updated this for 1.0.7 of the package, and verified that `int.TryParse(value, out int result)` compiles correctly in an ascx file.

bdukes:
I was _not_, however, able to get it to let me use the new `default` literal from C# 7.1, but I also don't see any specific details on whether this package is supposed to support C# 7.1 or not. 🤷‍♂️

galatrash-at-dnn:
We are not going to support C# 7 syntax yet in DNN source code (all our compilation is still based on VS 2015).

PR #1549, opened 2016-10-04
DNN-8553: cleanup user input data for tags.

galatrash-at-dnn:


zyhfish:

PR #1553, opened 2016-10-11
DNN-8845: switching to a FIPS compliant HashAlgorithm (plus a small m…

galatrash-at-dnn:
…odification to prevent clashing hashes).


galatrash-at-dnn:
This PR is discarded because it breaks the compatibility.

PR #1554, opened 2016-10-11
DNN-8201: fixing racing condition that creates an error in folder/fil…

galatrash-at-dnn:
…e permission checks during installation/upgrading.


zyhfish:

PR #1556, opened 2016-10-12
DNN-8935: Fix leaky resources which are not disposed properly.

galatrash-at-dnn:


zyhfish:

PR #1557, opened 2016-10-14
DNN-8935: More fixes for leaky resources.

galatrash-at-dnn:


zyhfish:

PR #1562, opened 2016-11-02
DNN-8921: removed redirect when only editor or setting views are requ…

galatrash-at-dnn:
…ested.


zyhfish:

PR #1563, opened 2016-11-02
DNN-8921: removed some unwanted control elements for PB actions.

galatrash-at-dnn:


zyhfish:

PR #1564, opened 2016-11-04
DNN-8953: set compability level on database.

zyhfish:


galatrash-at-dnn:
Rejected - not the proper solution.

The proper solution requires adding a semicolon (;) to the statement before the MERGE keyword that caused the problem. I'll create another fix and commit.

galatrash-at-dnn:
Rejected - not the proper solution.

The proper solution requires adding a semicolon (;) to the statement before the MERGE keyword that caused the problem. I'll create another fix and commit.

PR #1566, opened 2016-11-08
PAE-19: a fix for when a older DLL is installed in the site while a n…

galatrash-at-dnn:
…ewer one is being referenced by other libraries in the bin folder.


zyhfish:

PR #1569, opened 2016-11-14
DNN-8959: move purge log under thread lock.

zyhfish:


galatrash-at-dnn:
This PR does not resolve the problem of abnormal application termination.

PR #1571, opened 2016-11-16
DNN-8959: preventing exceptions from terminating the application at application end.

galatrash-at-dnn:


kanm2015:
Code review done

PR #1576, opened 2016-11-25
DNN-10273 Unclosed P tag in the resource string

roman-yagodin:
This with fix unclosing P tag in the FileUpload.UnzipFileSuccessPromptBody.Text resource string.


ashishpd:
Hi Roman - I realize we are looking at this PR after a long time. However, in order to process this PR we'll need a Jira. Ideally Jira should have enough information that someone in QA can verify the changes, and that to ensure it has not broken anything. I know this is a one liner change in a resource file, it will be great if you could still create a jira for this and follow the process (we will document the process as well).

roman-yagodin:
OK, here it is: https://dnntracker.atlassian.net/browse/DNN-10273

kanm2015:
code review done

PR #1579, opened 2016-11-29
DNN-9023: fix user profile create thumbnail issue (was caused by DNN-…

galatrash-at-dnn:
…8935; added some changes to original changes to fix this).


kanm2015:
Code review done

PR #1584, opened 2016-12-08
DNN-9048 Optimize Compose Message Dialogue for mobile

DNNMonster:
I think that this would help render the compose message box better for mobile devices.


amarjit-dhunna:
The changes do not fix the issue. Therefore I will close this PR.
I will submit PR with working changes.

PR #1586, opened 2016-12-09
DNN-9060: fix for languages aliases when changed from default.

galatrash-at-dnn:


zyhfish:
This change will break DNN-7274, can't merge.

PR #1590, opened 2016-12-13
DNN-9137: fix for anti-forgery-token validation

galatrash-at-dnn:


DNN Platform/DotNetNuke.Web/InternalServices/FileUploadController.cs
```diff
@@ -635,6 +635,21 @@ public Task<HttpResponseMessage> UploadFromLocal()
         }
 
         [HttpPost]
+        [AllowAnonymous]
+        [ValidateAntiForgeryToken]
+        public HttpResponseMessage GetUserInfo()
```
kanm2015: Is this for testing purpose? Can we get rid of it?

PR #1592, opened 2016-12-15
Nuspec cleanup

bdukes:
This PR improves the NuGet packages by using consistent terminology and naming, and expanding descriptive text.  

It also uses dependencies between packages instead of duplicating files in each package (though I believe this can be improved on more than is in place via this PR).

Finally, the new DotNetNuke.Bundle package is renamed to just DotNetNuke, to indicate its place as the standard "meta" package.

Any feedback is welcome!  Thanks!

### DIFF
--- a/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec
+++ b/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
-    <id>DotNetNuke</id>
+    <id>DotNetNuke.Bundle</id>
     <version>9.0.1</version>
     <title>DNN Platform</title>
     <authors>DNN Corp</authors>
--- a/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec
+++ b/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec
@@ -4,11 +4,11 @@
     <id>DotNetNuke.Bundle</id>
     <version>9.0.1</version>
     <title>DotNetNuke Bundle</title>
-    <authors>DotNetNuke Corporation</authors>
-    <owners>DotNetNuke Corporation</owners>
+    <authors>DNN Corp</authors>
+    <owners>DNN Corp</owners>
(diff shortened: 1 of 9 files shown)


ashishpd:
@bdukes thanks for cleaning this. I agree with everything, but for DotNetNuke.Bundle, name change. We have several internal builds that depend on this. Changing the name will break the chain. Can you pls. remove that and re-submit pls.

bdukes:
sounds good, I've reverted that commit.  is DotNetNuke.Bundle going to be published, or is it just for internal use?

ashishpd:
We use that internally and had not considered publishing it as it's huge.

bdukes:
Alright, that makes sense.  

I think there would be value in a "meta" DotNetNuke package which included most of what you'd need to develop extensions.  

Perhaps we just need to add some more dependencies, so that when you pick your flavor of development (Web API, MVC, or just Web), you'll already have all of the pieces you need, instead of trying to get started only to realize you're missing required references.  Does it make sense to add dependencies between those three packages and DotNetNuke.Core?  Should DotNetNuke.Core have a dependency on DotNetNuke.Instrumentation?

ashishpd:
Internally we do not use any of these packages except the .Bundle - it's sort of 'a la carta' for us. The other packages are for community to use. With the renames, will it affect anyone using the older packages, as we won't be updating them?

bdukes:
I've kept all of the IDs the same, so the existing packages should continue getting upgraded

ashishpd:
I've just now processed this PR. Let's keep an eye on the builds here: [https://build.dnnsoftware.com/viewType.html?buildTypeId=DnnPlatform910Public_Packaging]() You need to log-in as Guest. Ideally all the .nupkg packages should be listed under Artifacts. It should be available in 9.0.1.13 build

ashishpd:
image

It made to our unofficial (and unstable) public repo here:
[https://www.myget.org/feed/dnn-software-public/package/nuget/DotNetNuke.Bundle](url)

The dependencies are looking great. License is empty though. Thoughts @bdukes?

ashishpd:
Interestingly the `DotNetNuke.WebApi` is not getting pushed to MyGet since this change. Everything is else went fine.

image

bdukes:
I've added the license URL (https://github.com/dnnsoftware/Dnn.Platform/blob/development/LICENSE) to all of the packages

PR #1594, opened 2016-12-16
DNN-9155

DNNMonster:
There will likely be some refactoring but these changes hopefully help to target the areas that need attention to get the profile tab more mobile friendly.


pterenin:
I did some testing. Looks good @DNNMonster .

PR #1598, opened 2016-12-19
DNN-9152: ignore merge process when target file not exist.

zyhfish:


francescorivola:
I am not really happy with this solution, I have merged as currently the build is failing, however we should review if we can avoid suppress this error in the XmlMerge class to allow just a specific scenario.

PR #1602, opened 2016-12-21
DNN-8950: improvement for 'GetModuleByDefinition' stored procedure …

galatrash-at-dnn:
…and 'vw_Modules' view performance.


ashishpd:
Original request came from @mitchelsellers. @galatrash-at-dnn improvised it further. Measurements were taken and changes looked good. It should make DNN run faster.

PR #1604, opened 2016-12-22
DNN-9038: Updated Localization.GetString to return empty string if th…

kanm2015:
…e key is not found


bdukes:
@donker do you think this change could break any existing localization projects?

PR #1608, opened 2016-12-23
DNN-9141: redirect to access-denied page with messages coming from …

galatrash-at-dnn:
…the database and not the query string.


kanm2015:
Code review done

PR #1614, opened 2017-01-04
DNN-9040: fix for indexing profile properties when they contain $ at …

galatrash-at-dnn:
…beginning or end.


zyhfish:

PR #1617, opened 2017-01-05
DNN-9206: password should trimmed before use it in password reset.

zyhfish:


galatrash-at-dnn:
This PR doesn't fix the issue at hand.

PR #1620, opened 2017-01-09
DNN-9047: added new SQL execution methods with timeout parameter.

galatrash-at-dnn:
I used this timeout in the PurgeScheduleHistory task because it is timing out in systems with large number of history records.


kanm2015:
Code review done

PR #1635, opened 2017-01-27
DNN-9317 Use HTTPS for Xcillion font requests

bdukes:
This is a copy of XCESS-NL/xcillion#4 (a slightly improved version of XCESS-NL/xcillion#2)


bdukes:
Related to XCESS-NL/xcillion#1

PR #1639, opened 2017-01-30
DNN-9323 Media query & CSS to better render Digital Asset Manger

DNNMonster:
Things currently start getting disjointed at around 573px.


DNNMonster:
I overlooked some javascript work that needs to accompany this.

PR #1641, opened 2017-01-31
DNN-9323 Mobile friendly CSS/JS for Digital Asset Manager

DNNMonster:
Should make the Digital Asset Manager mobile friendly to about 350px.  It does not address all the necessary js to make it entirely responsive for window resize events.


galatrash-at-dnn:
@DNNMonster, according to the PM, "We have stopped enhancing this module as of 9.0." Therefore, I will close this issue without merging it.

PR #1646, opened 2017-02-04
DNN-9347 XCillion js mods to render search results on small screen devices.

DNNMonster:


zyhfish:

PR #1647, opened 2017-02-04
DNN 9348 Recalculate Search Results page's input box on window resize

DNNMonster:
Recalculate the input field and close icon's CSS attributes when the window has been resized.


zyhfish:

PR #1658, opened 2017-02-15
DNN-9232: upgrade jquery to 3.1.1.

zyhfish:
Note for compatible with jquery UI and other jquery plugin, the jquery migration plugin is required when register jquery.


jbrinkman:
This should be updated to jQuery 3.2.1.

kanm2015:
code review done

PR #1660, opened 2017-02-16
DNN-9405: Call ResetPassword after user unlocked

kanm2015:


zyhfish:
We can't call this method directly as the web config may disable retrieve password, need find another way.

PR #1673, opened 2017-03-01
DNN-9494: Do not track module deletion for hard delete and do not upd…

kanm2015:
…ate module order if order not changed


zyhfish:

PR #1678, opened 2017-03-06
NOJIRA: add comment in internal controls, add back console module.

zyhfish:


ashishpd:
Adding Console module back to Platform as it's needed in Profile area. It's dependency on Dotnetnuke.Deprecated.Dll is also being replaced as this DLL will eventually be removed from everywhere.

PR #1684, opened 2017-03-13
DNN-9552: add dotnetnuke.website.deprecated project to include the unused classes only for upgrade site.

zyhfish:


ashishpd:
This is needed to handle control panel code behind files during upgrade

PR #1685, opened 2017-03-14
DNN-9461: save extension install package file during install process.

zyhfish:


zyhfish:
This is sub task of Import/Export, so merge the changes into epic branch, and no need merge here.

PR #1696, opened 2017-04-04
DNN-9651: drop duplicate index and unused column.

sleupold:
fixing issues from DNN 7.2.0 upgrade and fixing typos in old script to make sure, future upgrades from previous versions don't have issues.

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/07.02.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/07.02.00.SqlDataProvider
@@ -346,12 +346,12 @@ GO
 /** DNN-2995 - SI: Performance improvement in dbo.GetUsersByEmail **/
 /*******************************************************************/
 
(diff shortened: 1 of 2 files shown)


galatrash-at-dnn:
@sleupold, this column is needed for DNN-2995 and can't be dropped. I will reject this for now.

sleupold:
@galatrash-at-dnn,
I don't have access to DNN-2995, however, the calculated column is stupid (DNN always uses CI collation) and the index does not use this column. There is already an index on Email.

galatrash-at-dnn:
@sleupold any update on when you will submit against the latest code base? Also, there is a need to fix the newly introduced C# ExportUser class as follows to since this column is dropped:
```
namespace Dnn.ExportImport.Dto.Users
{
    public class ExportUser : BasicExportImportDto
    {
       // ... everything stays as is except the LowerEmail property

        [Obsolete("Deprecated in 9.2.0 - use Email property instead.")]
        public string LowerEmail
        {
            get { return string.IsNullOrEmpty(Email) ? null : Email.ToLower(); }
            set { }
        }
```

sleupold:
I will review and resubmit asap

mitchelsellers:

PR #1698, opened 2017-04-07
DNN-9481: Add scheduler client feature to allow scheduled task to run on single/all servers.

zyhfish:


galatrash-at-dnn:
Closing based on comments in the Jira issue.

PR #1699, opened 2017-04-10
DNN-8709: Remove DNN_PlatformNoTests.sln.

zyhfish:


valadas:
Which solution should be used currently to build Dnn from source and submit pull requests. Following the instructions in the wiki does not work anymore, maybe the wiki should be updated.

galatrash-at-dnn:
@valadas, use the main solution "DNN_Platform.sln".

valadas:
Great it now builds. However the site does not work. I know the project has been split into multiple repositories because of the persona bar. Any instructions on how to build and run the website from the github repositories ?

PR #1702, opened 2017-04-11
DNN-9675: Updated portal.template.xsd for tabType

kanm2015:


zyhfish:

PR #1708, opened 2017-04-15
[DNN-9695] Release/9.0.2 Dnn.PersonaBar.Users SQLDataProviders issues

aarsys:
[SQLDataProviders.zip](https://github.com/dnnsoftware/Dnn.Platform/files/923791/SQLDataProviders.zip)

Dnn.PersonaBar.Users all Versions have an issue in all SQLDataProviders.

sysobject have always be called by the dbo and not by {databaseOwner}.

So you have to replace all **{databaseOwner}sysobjects** Statements with
**dbo.sysobjects**.

Otherwise this module will not work if anyone installs with other {databaseOwner} as 'dbo'!


galatrash-at-dnn:
@aarsys, rejected due to the large number of conflicts. We think this should have been fixed in PB code. But, if you think this issue is still valid, then you are welcome to submit a new PR that will be reviewed again.

PR #1710, opened 2017-04-19
DNN-9542: Fixed font color issue for asset name editing

kanm2015:


zyhfish:

roman-yagodin:
Shouldn't it be applied to .ItemNameEdit only inside .rgSelectedRow?
```CSS
.rgSelectedRow .ItemNameEdit {
    color:#000
}
```

PR #1719, opened 2017-04-25
DNN-9721: Updated code block with <%# instead of <%=

kanm2015:


vlk74:
Hello, I have a question if this change is correct. With that change it won't generate correct loginButton's id and I can't login by hitting ENTER. This problem appears only if you place login control to some page so module id must be used.

Thanks for the answer
   Vladimir

PR #1723, opened 2017-04-26
DNN-9728: Updated password reset page to respect expiration date and …

kanm2015:
…force to reset expiration date while generating password reset email


zyhfish:
Hi @kanm2015, that changes is made for DNN-7071, so this fix is not acceptable, we need find another way.

PR #1726, opened 2017-05-08
DNN-9670: changes to match the new check states of the PB page picker…

galatrash-at-dnn:
… UI control.


kanm2015:
code review done

PR #1727, opened 2017-05-09
Password history should have minimum timeframe for password retention (DNN-9759)

bdukes:
[DNN-9759](https://dnntracker.atlassian.net/browse/DNN-9759)

>As part of a security audit, it came up that we couldn't set a minimum timeframe for which to retain passwords. So, while we can say that you can't reuse your last six (for example) passwords, there's nothing to stop you from changing you password six times in a row and reusing your original password.
>
>I'd like to propose adding a new setting with the minimum number of days to retain passwords (our client's requirements are six months, but I would plan to default to zero, for backwards compatibility, unless y'all think there should be a default).

### DIFF
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -3317,14 +3317,26 @@ namespace DotNetNuke.Data
 
         #region Password History
 
+        [Obsolete("Deprecated in Platform 9.2.0, please use the overload that takes passwordsRetained and daysRetained")]
         public virtual IDataReader GetPasswordHistory(int userId)
         {
-            return ExecuteReader("GetPasswordHistory", GetNull(userId));
+            return this.GetPasswordHistory(userId, int.MaxValue, int.MaxValue);
         }
 
+        public virtual IDataReader GetPasswordHistory(int userId, int passwordsRetained, int daysRetained)
+        {
+            return ExecuteReader("GetPasswordHistory", GetNull(userId), passwordsRetained, daysRetained);
+        }
+
+        [Obsolete("Deprecated in Platform 9.2.0, please use the overload that takes daysRetained")]
         public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int retained)
         {
-            ExecuteNonQuery("AddPasswordHistory", GetNull(userId), password, passwordHistory, retained, GetNull(userId));
+            this.AddPasswordHistory(userId, password, passwordHistory, retained, int.MaxValue);
+        }
+
+        public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int passwordsRetained, int daysRetained)
+        {
+            ExecuteNonQuery("AddPasswordHistory", GetNull(userId), password, passwordHistory, passwordsRetained, daysRetained, GetNull(userId));
         }
 
         #endregion
--- a/Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
@@ -558,19 +558,10 @@ CREATE PROCEDURE {databaseOwner}[{objectQualifier}GetPasswordHistory]
 	@PasswordsRetained  int,
 	@DaysRetained       int
 AS
-	DELETE FROM {databaseOwner}{objectQualifier}PasswordHistory
-	WHERE UserID=@UserId
-	  AND PasswordHistoryID NOT IN (
-		SELECT TOP (@PasswordsRetained) PasswordHistoryID
-		FROM {databaseOwner}{objectQualifier}PasswordHistory
-		WHERE UserID=@UserId
-		ORDER BY CreatedOnDate DESC
-		)
-	  AND DATEDIFF(day, CreatedOnDate, GETDATE()) > @DaysRetained;
-
-	SELECT *
+	SELECT TOP (@PasswordsRetained) *
 	FROM {databaseOwner}{objectQualifier}PasswordHistory
 	WHERE UserID=@UserID
+	  AND DATEDIFF(day, CreatedOnDate, GETDATE()) <= @DaysRetained;
 GO
 
 /************************************************************/
--- a/Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
@@ -502,6 +502,77 @@ BEGIN
 END
 GO
 
+IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{objectQualifier}AddPasswordHistory]') AND OBJECTPROPERTY(id, N'IsPROCEDURE') = 1)
+	DROP PROCEDURE {databaseOwner}{objectQualifier}AddPasswordHistory
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}AddPasswordHistory]
+	@UserId             int,
+	@Password           nvarchar(128),
+	@PasswordSalt       nvarchar(128),
+	@PasswordsRetained  int,
+	@DaysRetained       int,
+	@CreatedByUserID    int
+AS
+
+	BEGIN
+
+	INSERT INTO {databaseOwner}{objectQualifier}PasswordHistory (
+		UserId,
+		[Password],
+		PasswordSalt,
+		CreatedByUserID,
+		CreatedOnDate,
+		LastModifiedByUserID,
+		LastModifiedOnDate
+	)
+	VALUES (
+		@UserId,
+		@Password,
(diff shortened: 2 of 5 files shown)


DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3317,14 +3317,14 @@ public virtual IDataReader GetFolderMappingSettings(int folderMappingID)
 
         #region Password History
 
-        public virtual IDataReader GetPasswordHistory(int userId)
+        public virtual IDataReader GetPasswordHistory(int userId, int passwordsRetained, int daysRetained)
```
galatrash-at-dnn: This creates compatibility issues. This must be deprecated and a new method created.

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3317,14 +3317,14 @@ public virtual IDataReader GetFolderMappingSettings(int folderMappingID)
...
         {
-            return ExecuteReader("GetPasswordHistory", GetNull(userId));
+            return ExecuteReader("GetPasswordHistory", GetNull(userId), passwordsRetained, daysRetained);
         }
 
-        public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int retained)
+        public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int passwordsRetained, int daysRetained)
```
galatrash-at-dnn: This creates compatibility issues. This must be deprecated and a new method created.

DNN Platform/Library/Data/DataProvider.cs
bdukes: In the past I had been told that `DataProvider` was not intended to be backwards compatible, but I'm happy to keep compatibility, if that's desired.  However, I won't be able to turn any attention to this until at least next week.

galatrash-at-dnn: That's right and the ideal case, but in practice, some third party modules might have used these. We need to try keeping backward compatibility as much as possible.

Website/Providers/DataProviders/SqlDataProvider/09.01.01.01.SqlDataProvider
```diff
@@ -0,0 +1,85 @@
...
+	  AND DATEDIFF(day, CreatedOnDate, GETDATE()) > @DaysRetained;
+
+	SELECT * 
+	FROM {databaseOwner}{objectQualifier}PasswordHistory 
+	WHERE UserID=@UserID
+GO
+
+/************************************************************/
+/*****              SqlDataProvider                     *****/
+/************************************************************/
```
galatrash-at-dnn: As we are now working on 9.2.0, this needs to go into ``09.02.00.SqlDataProvider`` (and not ``09.02.00.01.SqlDataProvider``)

Website/Providers/DataProviders/SqlDataProvider/DotNetNuke.Schema.SqlDataProvider
```diff
@@ -17246,9 +17251,23 @@ GO
...
+		ORDER BY CreatedOnDate DESC
+		)
+	  AND DATEDIFF(day, CreatedOnDate, GETDATE()) > @DaysRetained;
+
+	SELECT * 
+	FROM {databaseOwner}{objectQualifier}PasswordHistory 
+	WHERE UserID=@UserID
 GO
 PRINT N'Creating {databaseOwner}[{objectQualifier}SetFileHasBeenPublished]'
 GO
```
galatrash-at-dnn: You shouldn't change anything in this file as the individual migrations will take care of this.

galatrash-at-dnn:
@bdukes, please resolve the conflicts and pay attention to my comments and I'll look into this once you commit your changes.

No problem if it comes next week, but the sooner the better since we are getting closer to release date and code freeze.

bdukes:
@galatrash-at-dnn I've fixed and squashed these changes, let me know if there's anything else

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3317,14 +3317,26 @@ public virtual IDataReader GetFolderMappingSettings(int folderMappingID)
...
+        [Obsolete("Deprecated in Platform 9.2.0, please use the overload that takes daysRetained")]
         public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int retained)
         {
-            ExecuteNonQuery("AddPasswordHistory", GetNull(userId), password, passwordHistory, retained, GetNull(userId));
+            this.AddPasswordHistory(userId, password, passwordHistory, retained, int.MaxValue);
+        }
+
+        public virtual void AddPasswordHistory(int userId, string password, string passwordHistory, int passwordsRetained, int daysRetained)
+        {
+            ExecuteNonQuery("AddPasswordHistory", GetNull(userId), password, passwordHistory, passwordsRetained, daysRetained, GetNull(userId));
```
galatrash-at-dnn: daysRetained is not used in the SPROC, so why add it?

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
```diff
@@ -492,16 +492,87 @@ BEGIN
...
+IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{objectQualifier}AddPasswordHistory]') AND OBJECTPROPERTY(id, N'IsPROCEDURE') = 1)
+	DROP PROCEDURE {databaseOwner}{objectQualifier}AddPasswordHistory
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}AddPasswordHistory]
+	@UserId             int,
+	@Password           nvarchar(128),
+	@PasswordSalt       nvarchar(128),
+	@PasswordsRetained  int,
+	@DaysRetained       int,
```
galatrash-at-dnn: daysRetained is not used in the SPROC, so why add it?

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
```diff
@@ -492,16 +492,87 @@ BEGIN
...
+AS
+	DELETE FROM {databaseOwner}{objectQualifier}PasswordHistory
+	WHERE UserID=@UserId
+	  AND PasswordHistoryID NOT IN (
+		SELECT TOP (@PasswordsRetained) PasswordHistoryID
+		FROM {databaseOwner}{objectQualifier}PasswordHistory
+		WHERE UserID=@UserId
+		ORDER BY CreatedOnDate DESC
+		)
+	  AND DATEDIFF(day, CreatedOnDate, GETDATE()) > @DaysRetained;
```
galatrash-at-dnn: There is no need for this in this SPROC. This is for returning the old password. This is already achieved when you are adding a new password. The select statement can return top (xx) passwords withing a specific period of time.

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
bdukes: it is used on line 547

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
bdukes: The purpose of this change is that you can specify a minimum number of days to retain passwords, in addition to a minimum number of passwords to retain, so that you can't just change your password many times in one day in order to get around the restrictions of password retention.  So, `AddPasswordHistory` may not remove your old password today, but that old password may no longer be relevant tomorrow, which is why this change ensures that the list of password history respects the retention rules before retrieving the passwords.  If it didn't do this, a password which could be reused today would still be in the list.  

Let me know if there's something I'm missing, or if there's a different solution you'd prefer to see.

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
bdukes: it is used on line 547

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
bdukes: The purpose of this change is that you can specify a minimum number of days to retain passwords, in addition to a minimum number of passwords to retain, so that you can't just change your password many times in one day in order to get around the restrictions of password retention. So, `AddPasswordHistory` may not remove your old password today, but that old password may no longer be relevant tomorrow, which is why this change ensures that the list of password history respects the retention rules before retrieving the passwords. If it didn't do this, a password which could be reused today would still be in the list.

Let me know if there's something I'm missing, or if there's a different solution you'd prefer to see.

(sorry if you're getting these comments twice, I've not gotten the hang of GitHub reviews)

bdukes:
@galatrash-at-dnn I've pushed a change which removes the whitespace changes, sorry those snuck in there

PR #1733, opened 2017-05-19
DNN-9627: Bind data when headers get localized

kanm2015:


zyhfish:
IMO the LocalizeDataGrid line can move to line 361, so that we doesn't need duplicate call to Grid DataBind method.

zyhfish:

PR #1737, opened 2017-05-23
DNN-9838: ResourceFileInstaller fails to update resource files

sergey-ostapenko:
Fixes https://dnntracker.atlassian.net/browse/DNN-9838

### Fix description
Now DNN will retry to write locked files during module update process.
If for some reason we cannot overwrite old version of file during module installation, Dnn will retry to copy file during one second. This approach help to avoid failures if files are being used by caching, indexing, antivirus or whatever.

### Problem description
It seems that in some cases during upgrade of resource files dnn9 tries to create new resource file while old version has not been deleted yet.
We see this error on our production environments where we have separate machines for web-site hosting (VM03) and web-site files storage (VM07). The problems occurs with upgrade of modules containing large number of resource files. Symptoms: upgrade of modules fail on random resource files. For example:

```
Info Creating backup of previous version - images\action.gif
Failure Installation Failed - ResourceFile
```

Additional information from dnn log shows 
```
"The process cannot access the file xxx because it is being used by another process.":
2017-05-15 11:57:03,252 [ProductionVM03][Thread:24][ERROR] DotNetNuke.Services.Installer.Installers.ResourceFileInstaller - System.IO.IOException: The process cannot access the file '\\ProductionVM07\dnn9_website\DesktopModules\OurModule\images\help.gif' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
   at System.IO.FileInfo.Create()
   at DotNetNuke.Services.Installer.Util.WriteStream(Stream sourceStream, String destFileName)
   at DotNetNuke.Services.Installer.Installers.ResourceFileInstaller.InstallFile(InstallFile insFile)
```

It is known problem in .NET that after deletion of the file it takes some time before the file actually disappears, and instead of try to relay on File.Exists function it's better to wait until file will be unlocked using FileSystemWatcher.
Same problem could occur because of antivirus, indexing, caching, etc. But the fix remains the same since such locks usually lasts just few moments.


sergey-ostapenko:
Finished, covered with tests.
Ready for review.

sergey-ostapenko:
@kanm2015 Could you please make code review?
Thanks in advance.

sergey-ostapenko:
Any reason why this pull request is blocked?

sergey-ostapenko:
Conflicts were resolved. 
Rebased on top of current development branch.

PR #1742, opened 2017-05-31
SQL GetAllTabsModules optimisation

SergentGarcia:
Optimisation when there is no module


galatrash-at-dnn:
@SergentGarcia , the change you introduced went into a very old migration script. To have this included in the latest release, you must add it to the latest migrations file:
``` Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider```

PR #1748, opened 2017-06-05
CONTENT-8174: Add soure Unique Id for tab module while import.

amarjit-dhunna:
Fix restore and delete of modules on target.


DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -518,7 +518,7 @@ private int ImportTabModulesAndRelatedItems(TabInfo localTab, ExportTab otherTab
             var localExportModules = isNew ? new List<ExportModule>()
                 : EntitiesController.Instance.GetModules(localTab.TabID, true, Constants.MaxDbTime, null).ToList();
             var localTabModules = isNew ? new List<ModuleInfo>() : _moduleController.GetTabModules(localTab.TabID).Values.ToList();
-
+            var allTabUniqueIds = CBO.FillCollection<ModuleInfo>(DataProvider.Instance().GetAllTabModules(localTab.PortalID)).Select(x => x.UniqueId).ToList();
```
galatrash-at-dnn: Please change this to take the unique ID as a parameter from the SPROC to minimize the number of returned data from the DB instead of filtering through LINQ.

PR #1754, opened 2017-06-08
DNN-9835: added includeAllTypes=true to ItemListService/GetPages api …

kanm2015:
…call


zyhfish:

PR #1756, opened 2017-06-08
DNN-9542: Fixed font color issue for asset name editing in icons view

kanm2015:


zyhfish:

PR #1759, opened 2017-06-12
DNN-9903: replace fips compilance assembly in fips enabled env.

zyhfish:


galatrash-at-dnn:
See my comment in Jira for the reason of rejecting this PR.

kanm2015:
code review done

PR #1775, opened 2017-06-23
DNN-9947: Fix the Export_GetActiveTabModules UDF to return all module…

amarjit-dhunna:
…s in case of no versioning of the tab


kanm2015:
code review done

PR #1780, opened 2017-06-28
DNN-9970: do not delete module files during module uninstalling, file…

kanm2015:
…s will be handled by cleanup process separately


kanm2015:
will submit new PR, close this one for now

PR #1789, opened 2017-07-03
DNN-9428: if there already have higher version installed, remove current one from list.

zyhfish:


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -4963,9 +4963,20 @@ private static void ParsePackagesFromApplicationPath(IEnumerable<string> package
 	                {
 		                //check whether have version conflict and remove old version.
 		                var package = packages[file];
-		                if (packages.Values.Count(p => p.FriendlyName == package.FriendlyName) > 1)
+
+                        var installedPackage = PackageController.Instance.GetExtensionPackage(Null.NullInteger, 
+                            p => p.Name == package.Name && p.PackageType == package.PackageType);
```
galatrash-at-dnn: These comparison should be case insensitive.

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -4963,9 +4963,20 @@ private static void ParsePackagesFromApplicationPath(IEnumerable<string> package
 	                {
 		                //check whether have version conflict and remove old version.
 		                var package = packages[file];
-		                if (packages.Values.Count(p => p.FriendlyName == package.FriendlyName) > 1)
+
+                        var installedPackage = PackageController.Instance.GetExtensionPackage(Null.NullInteger, 
+                            p => p.Name == package.Name && p.PackageType == package.PackageType);
+
+                        if (packages.Values.Count(p => p.FriendlyName == package.FriendlyName) > 1 || installedPackage != null)
```
galatrash-at-dnn: Also make this comparison case insensitive

PR #1797, opened 2017-07-07
DNN-10016 Install - SQL Build Versions Corrected

AnonDevelopment:
DNN-10016 - https://dnntracker.atlassian.net/browse/DNN-10016
Causing invalid minimum requirement of SQL Server 2012 during install.


galatrash-at-dnn:
This whole ``switch`` block should be changed to this:
```
isValidVersion = versionNumber >= 10;  // SQL Server 2008 and above
```
This will eliminate the need to change this code every time a new SQL server version is released.

PR #1804, opened 2017-07-18
AUTO-606 / CONTENT-8411: added logic to suppress alerts during site i…

galatrash-at-dnn:
…nstall/upgrade process.


galatrash-at-dnn:
Note: the some commits contained AUTO-606 by mistake; the issue should have been AUTO-608.

PR #1805, opened 2017-07-18
DNN-10082 Fix issue with JSON when sending message

bdukes:
The JSON sent to MessagingServiceController from ComposeMessage.js is invalid,
but fails in an acceptable way with Newtonsoft.Json v7. However, in v8 (through
the current version, v10), this causes a larger failure, which breaks the API
(resulting in a null reference exception at the start of the call).  This
change adjusts the JSON to always send an array (serialized as a string),
rather than sending an empty object when the array would be empty.  This also
requires the calling code to expect empty lists, rather than null lists.

[DNN-10082](https://dnntracker.atlassian.net/browse/DNN-10082)
>If the Newtonsoft.Json dependency is updated (you could install [newtonsoft.json.8.0.3.zip](https://dnntracker.atlassian.net/secure/attachment/55312/55312_newtonsoft.json.8.0.3.zip) or [newtonsoft.json.10.0.3.zip]()https://dnntracker.atlassian.net/secure/attachment/55311/55311_newtonsoft.json.10.0.3.zip, or a 3rd party module which requires a newer version), there's an error when trying to send a message:
>
><a href="https://dnntracker.atlassian.net/secure/attachment/55310/55310_2017-07-18_15-37-29.png">An error occurred while creating the message:Object reference not set to an instance of an object.</a>
>
>This issue stems from the JavaScript for the Compose Message component sending an unexpected JSON structure to the MessagingController web API (specifically, the API is expecting strings, the component sends empty objects).  In Newtonsoft.Json v7, this results in those fields being `null`, while after version 7 (versions 8-10), this results in a failure to complete the model binding, so the whole DTO is `null`.

### DIFF
--- a/DNN Platform/Library/Services/Social/Messaging/MessagingController.cs
+++ b/DNN Platform/Library/Services/Social/Messaging/MessagingController.cs
@@ -1,21 +1,21 @@
 ﻿#region Copyright
-// 
+//
 // DotNetNuke® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2017
 // by DotNetNuke Corporation
-// 
-// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated 
-// documentation files (the "Software"), to deal in the Software without restriction, including without limitation 
-// the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and 
+//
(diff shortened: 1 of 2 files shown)


zyhfish:
Hi @bdukes , thanks for your hard work on this problem, i have included your changes in PR #1837, so that the reviewer can check these changes in same time. for this i will close this PR. thx again!

PR #1807, opened 2017-07-21
DNN-10091: Update DotNetNuke.install.config.resources to support html5 video files

valadas:


zyhfish:
Hi @valadas , this only change in the refresh install, can you please also handle the upgrade scenario? thx

valadas:
Hi @zyhfish I would be happy to, I will do some digging tomorrow, I am not sure where to implement that on upgrades. I will try to find that out tomorrow unless you already know and can point me in the right direction. Thanks.

Also, do you think it is a good idea to change allowed file types on upgrades. If for some reason I have a running website and I specifically set only .jpg are allowed, I might be surprised to see other extensions allowed after an upgrade without my knowledge and no warning? Let me know what you think.

zyhfish:
Hi @valadas , you can check Website\Providers\DataProviders\SqlDataProvider\05.01.00.SqlDataProvider line 2852 to see how add extension during upgrade, the code change should be done in 9..2.0 data provider file.

we always have release note for what have been changed in every release, use should be know we add this in 9.2.0. if we don't catch upgrade scenario, then user may see the new toolbar in ck editor, but he can't upload video files, this will cause function broken, so we should take care of install/upgrade process in most of time. thx!

valadas:
@zyhfish Makes a lot of sense, I will do that tomorrow. Thanks for the hint.

zyhfish:
Thanks so much for your hard work, im reviewing the changes in CK editor now, I may add comment on that PR if thing need to be changed as well, thx again!

valadas:
@zyhfish Ok, I got a little hickup with github, but I think the commits should be fine now. Let me know if there is any issue.

zyhfish:
Hi @valadas , thx for the changes, i think it should be good enough. please resolve the conflicts, then i can merge into into main branch, thx!

valadas:
It is done, sorry, I saw it before but I tought I needed write access to your repo to resolve it. I managed to do it, learned something new in the process. Let me know if there is anything else. And it is good to see pull requests getting processed, thanks for your help.

zyhfish:
Changes looks good and merged, great thx for your contribution!

PR #1810, opened 2017-07-24
DNN-8305: updated UrlUtils.InPopUp() to do ignore case comparison

kanm2015:


zyhfish:
DNN Platform\Library\Common\Globals.cs line 2852 need to update as well, thx.

kanm2015:
updated, thx for reviewing

zyhfish:

PR #1811, opened 2017-07-25
DNN-10101: updated to show user friendly error msg

kanm2015:


zyhfish:

PR #1812, opened 2017-07-26
DNN-8014: update user logout flag and check it via membership module.

zyhfish:


galatrash-at-dnn:
The design is not acceptable as it logs out the user from all other logged in devices.

PR #1815, opened 2017-07-28
DNN-10110: updated to catch StorageException for both StorageClientEx…

kanm2015:
…ception and StorageServerException and prevent from sending empty account name or key to Azure


zyhfish:

PR #1816, opened 2017-07-28
DNN-10012: Optimizing PortalSecurity object creation

galatrash-at-dnn:
(using static instance instead).


kanm2015:
code review done

PR #1842, opened 2017-09-05
Update ModuleController.cs

meowthing:
Surely when trying to make a deep copy, this line is to supposed to add a newModule, not sourceModule?


bdukes:
This change looks right to me.  

Is there a place in the standard UI where you're encountering this bug, or is it only in custom code?  I don't see a place (after a quick glance) where `AddModules` is called with `asReference` being `false`.

Also, this will probably get processed faster if you create an issue at https://dnntracker.atlassian.net/ and reference the issue number in the PR title here.

Thanks!

meowthing:
I was trying to figure out how to clone a module, since ModuleController.Instance.CopyModule seems to only allow a 'shallow' copy (same ModuleID). I'll make an issue in the tracker, thanks.

zyhfish:
I created JIRA https://dnntracker.atlassian.net/browse/DNN-10363 to tracking this PR. also will try to create a custom module to verify this API.

zyhfish:

PR #1844, opened 2017-09-06
Fixes DNN-10071

vano-maisuradze:
Fixes DNN-10071 issue by adding check for null.
When _relativePath_ paramter was null, it was throwing an exception.


fjlucas:
Code review done. It looks good.

PR #1846, opened 2017-09-07
DNN-10166 add IP Filter validation

tpluscode:
Added a check to validate the IP address and subnet mask which throws ArgumentException if either of them are not in correct format.

### DIFF
--- a/DNN Platform/Library/Entities/IPFilter/IPFilterController.cs
+++ b/DNN Platform/Library/Entities/IPFilter/IPFilterController.cs
@@ -36,6 +36,7 @@ using DotNetNuke.ComponentModel;
 using DotNetNuke.Data;
 using DotNetNuke.Entities.Users;
 using DotNetNuke.Instrumentation;
+using DotNetNuke.Services.Localization;
 using DotNetNuke.Services.Log.EventLog;
 
 #endregion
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Entities/IPFilter/IPFilterController.cs
```diff
@@ -256,5 +261,20 @@ public bool IsAllowableDeny(string currentIP, IPFilterInfo ipFilter)
         }
 
         #endregion
+
+        private static void AssertValidIPFilter(IPFilterInfo ipFilter)
+        {
+            IPAddress parsed;
+            if (IPAddress.TryParse(ipFilter.IPAddress, out parsed) == false)
+            {
+                throw new ArgumentException("IP address is not in correct format");
```
tpluscode: What is the best practice for adding localization? Maybe we don't want to localize this message at all?

DNN Platform/Library/Entities/IPFilter/IPFilterController.cs
```diff
@@ -256,5 +261,20 @@ public bool IsAllowableDeny(string currentIP, IPFilterInfo ipFilter)
...
+            }
+
+            bool isIPRange = string.IsNullOrEmpty(ipFilter.SubnetMask);
+            if (isIPRange && IPAddress.TryParse(ipFilter.SubnetMask, out parsed) == false)
+            {
+                throw new ArgumentException("Subnet mask is not in correct format");
```
tpluscode: Ditto

zyhfish: Hi @tpluscode , for the localization, you can check DNN Platform\Library\Common\Requires.cs line 102 for example. the localization strings can saved in Website\App_GlobalResources\Exceptions.resx.

zyhfish:

PR #1850, opened 2017-09-08
DNN-9410: Change the JWT logout to retun valid JSON instead of empty …

aliazizoglu:
…string.


galatrash-at-dnn:
The change is good, but I need you to modify the related integration test to take the new change into consideration.

PR #1862, opened 2017-09-15
 DNN-10243: KO templates relying on Biography field insert one if not

Mohtshum:


zyhfish:
Pending this PR as i don't think hard code stuff is a good choice, let me think about whether we can find out a better way. close PR for this mean time.

PR #1863, opened 2017-09-18
DNN-10247: added checks against null request and headers objects.

galatrash-at-dnn:


zyhfish:

PR #1867, opened 2017-09-20
DNN-10166 fix reversed check for subnet mask

tpluscode:


zyhfish:

PR #1871, opened 2017-09-21
DNN-9019: New Eventhandler subscriber added to handle profile updates

aliazizoglu:


aliazizoglu:
portalsecuritytest code should not be in PR

amarjit-dhunna:
I will submit PR for this JIRA

PR #1874, opened 2017-09-25
Release/9.1.1 - Open HTML Editor Manager causes 404 error

JaySMa:
Settings -> Site Settings -> Site Behavior -> More

HTML Editor Manager - Clicking the Open HTML Editor Manager causes 404 error in the window which appears.

Used Fiddler and it shows the URL trying to be called is:
     /Host/HTMLEditorManager?portalid=undefined&portpopUp=true

Installation was upgraded from DNN 7.4.1 to 8.0.4 to 9.1.1.


zyhfish:
Hi @JaySMa , This branch is not able to process as it target to a release branch, PR should only target to development branch. also please create an JIRA item first to describe what's the issue and how to reproduce it. thx so much!

zyhfish:
As i do more research, i believe what you are talking about is related to https://dnntracker.atlassian.net/browse/DNN-10063, and its already handled in 9.2.0, thx!

cxdone:
After disable a page and save will auto redirect to the page which already disabled and report 404 error.

PR #1877, opened 2017-09-28
DNN-10252: only allow attach files in current user folder.

zyhfish:


galatrash-at-dnn:
@zyhfish , you should add some tests both positive and negative) for your code changes.

DNN Platform/Modules/Journal/ServicesController.cs
```diff
@@ -119,14 +119,31 @@ public HttpResponseMessage Create(CreateDTO postData)
                     postData.ProfileId = UserInfo.UserID;
                 }
 
+                if (postData.ProfileId != UserInfo.UserID)
+                {
+                    var profileUser = UserController.Instance.GetUser(PortalSettings.PortalId, postData.ProfileId);
+                    if (profileUser == null || (!UserInfo.IsInRole(PortalSettings.AdministratorRoleName) && !Utilities.AreFriends(profileUser, UserInfo)))
+                    {
+                        throw new ArgumentException("you have no permission to post journal on current profile page");
```
galatrash-at-dnn: Use ``Localization.GetString(...)`` please

DNN Platform/Modules/Journal/ServicesController.cs
```diff
@@ -119,14 +119,31 @@ public HttpResponseMessage Create(CreateDTO postData)
...
                     postData.ProfileId = -1;
 
                     RoleInfo roleInfo = RoleController.Instance.GetRoleById(ActiveModule.OwnerPortalID, postData.GroupId);
-                    if (roleInfo != null && !roleInfo.IsPublic)
+                    if (roleInfo != null)
                     {
-                        postData.SecuritySet = "R";
+                        if (!UserInfo.IsInRole(PortalSettings.AdministratorRoleName) && !UserInfo.IsInRole(roleInfo.RoleName))
+                        {
+                            throw new ArgumentException("you have no permission to post journal on current group.");
```
galatrash-at-dnn: Use ``Localization.GetString(...)`` please

DNN Platform/Modules/Journal/ServicesController.cs
```diff
@@ -119,14 +119,31 @@ public HttpResponseMessage Create(CreateDTO postData)
                     postData.ProfileId = UserInfo.UserID;
```
galatrash-at-dnn: You need to remove the ``if`` condition of this block to become basically ``postData.ProfileId = UserInfo.UserID;`` instead of ``if(..) { postData.ProfileId = UserInfo.UserID; }

DNN Platform/Modules/Journal/ServicesController.cs
```diff
@@ -119,14 +119,31 @@ public HttpResponseMessage Create(CreateDTO postData)
...
+                    var profileUser = UserController.Instance.GetUser(PortalSettings.PortalId, postData.ProfileId);
+                    if (profileUser == null || (!UserInfo.IsInRole(PortalSettings.AdministratorRoleName) && !Utilities.AreFriends(profileUser, UserInfo)))
+                    {
+                        throw new ArgumentException("you have no permission to post journal on current profile page");
+                    }
+                }
+
                 if (postData.GroupId > 0)
                 {
                     postData.ProfileId = -1;
```
galatrash-at-dnn: I know this is an old line of code, but why is it is needed? don't we keep the UserID when posting to a group?

DNN Platform/Modules/Journal/ServicesController.cs
zyhfish: The if condition can't be removed, as have scenario user can post journal on friend's profile page.

PR #1878, opened 2017-09-28
DNN-9192: fixed open link in new window issue for Xcillion skin

kanm2015:


zyhfish:

PR #1880, opened 2017-10-04
Bug/dnn 8939

Mohtshum:
DNN-8939: SI: IP Address Tracking when using a Load Balancer
- Did the change for getting IP against any Header from HostSetttings meanwhile Default is X-Forwarded-For
- Change the existing methods to use new controller and obselete them
- Not returning DNS name as security con mentioned in CONTENT-8523

Put this controller in 
dnn platform\library\services\userrequest

Unittest added in 
DNN Platform\Tests\DotNetNuke.Tests.Core\Services\UserRequest\
With test cases for 
- Valid IP address
- Header/ServerVariable present to obtain IP against

Can be extended per customer requirement

### DIFF
--- a/DNN Platform/Library/Services/GeneratedImage/IPCount.cs
+++ b/DNN Platform/Library/Services/GeneratedImage/IPCount.cs
@@ -1,4 +1,5 @@
 ﻿using DotNetNuke.Common.Utils;
+using DotNetNuke.Services.UserRequest;
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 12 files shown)


DNN Platform/Library/Services/GeneratedImage/IPCount.cs
```diff
@@ -165,32 +166,10 @@ private static string BuildFilePath(string ipAddress)
...
-
-            if (string.IsNullOrEmpty(visitorIPAddress) || visitorIPAddress.Trim() == "::1")
-            {
-                visitorIPAddress = string.Empty;
-            }
-
-            if (string.IsNullOrEmpty(visitorIPAddress))
-            {
-                //This is for Local(LAN) Connected ID Address
-                var stringHostName = Dns.GetHostName();
```
Mohtshum: This logic is also not used for IP address for security concern mentioned in CONTENT-8523

DNN Platform/Library/Services/UserRequest/UserRequestIPAddressController.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        public string GetUserRequestIPAddress(HttpRequestBase request)
+        {            
+            var userRequestIPHeader = HostController.Instance.GetString("UserRequestIPHeader", "X-Forwarded-For");
+            var userIPAddress = string.Empty;
+
+            if (request.Headers.AllKeys.Contains(userRequestIPHeader))
+            {
+                userIPAddress = request.Headers[userRequestIPHeader];
+                userIPAddress = userIPAddress.Split(',')[0];
```
galatrash-at-dnn: You need to take the last IP as follows ``userIPAddress = userIPAddress.Split(',').Last().Trim();``

DNN Platform/Library/Services/GeneratedImage/IPCount.cs
galatrash-at-dnn: @Mohtshum please have a look at this 
```
        public static string GetIPAddress(this HttpRequestMessage request)
        {
            return GetHttpContext(request).Request.UserHostAddress;
        }
```
In DotNetNuke.Web.Api HttpRequestMessageExtensions clasas. This one should be fixed as well.

Mohtshum: ok, thnx

PR #1890, opened 2017-10-13
DNN-9394: remove invalid antiforgery cookie generated by old mvc assembly during upgrade.

zyhfish:


kanm2015:
looks good, merged

PR #1893, opened 2017-10-17
DNN-8564: Update DeleteTab so that TabOrder of other Portals is not affected on Delete

zyhfish:


zyhfish:
This PR created base on #1422 , thanks @JoshuaBradley for post the original PR.

sleupold:
@zyhfish,
I am using a more stable version to handle all these tab management in my turbo scripts, see https://dnnscript.codeplex.com/releases/view/615958 for an earlier version. If you are interested, I would submit an updated version after DNN 9.2.0 release (due to time restrictions)

galatrash-at-dnn:
@sleupold please submit yours for 9.2.0 as the release date has been pushed further at the moment; so we have time for this and some more ones from you. The team here will review these in a timely manner.

PR #1897, opened 2017-10-18
DNN-8450 Allow override of prefixing the service name for OAuth providers

bdukes:
This is a continuation of #1346 (after rebasing and fixing conflicts)

[DNN-8450](https://dnntracker.atlassian.net/browse/DNN-8450)

> The fix for [DNN-4133](https://dnntracker.atlassian.net/browse/DNN-4133) was to add a prefix for the service to the beginning of the username. This broke our custom OAuth provider and I believe that we should have the ability to override whether we want to prefix the service to the Username in OAuth providers. In addition to allowing that overriding, this also implements the option to automatically associate logins to existing users, based on email.
> Testing of this API change will require a custom OAuth provider implementation which sets the new `PrefixServiceToUserName` and `AutoMatchExistingUsers` properties in that provider's implementation of `OAuthClientBase` to `false` and `true`, respectively. That custom provider may be an existing provider (e.g. Facebook), tweaked to override those two properties.
> When configured as mentioned above, the username of users who log in via that provider should be the user's email (as provided by the OAuth provider). Additionally, if a user with that email address as their username already exists, that OAuth login would be associated with the existing account, rather than trying to create a new account.

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -614,7 +614,7 @@ namespace DotNetNuke.Security.Membership
         }
 
        
-        private UserInfo GetUserByAuthToken(int portalId, string userToken, string authType)
+        public override UserInfo GetUserByAuthToken(int portalId, string userToken, string authType)
         {
             IDataReader dr = _dataProvider.GetUserByAuthToken(portalId, userToken, authType);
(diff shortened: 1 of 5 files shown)


bdukes:
@zyhfish you're right, I'm not sure why those changes were made.  I've reverted them.  Let me know if you see anything else

ohine:
I'm bypassing the CLA requirement since @bdukes already has one signed and this PR was created before the CLA bot was added.

PR #1898, opened 2017-10-18
DNN-8374 adding locking functionality to the instance and individual sites

bdukes:
This is a continuation of #1272, after rebasing and fixing conflicts

[DNN-8374](https://dnntracker.atlassian.net/browse/DNN-8374)

> We've made some customizations for a client which we think would be a good fit for inclusion in the core. There's a "Lock Site" action added to the control bar, which allows a super-user to turn off modifications to the site (probably for entering some sort of maintenance mode). For our client, the "good enough" way this is implemented is to disable the module actions whenever the site is locked (in addition, custom modules are able to access the locked status via a property on `PortalSettings` and do any additional changes, if desired).

### DIFF
--- a/DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
+++ b/DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
@@ -23,6 +23,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -30,6 +31,7 @@ using System.Net.Http.Headers;
 using System.Web.Http;
 using DotNetNuke.Common;
 using DotNetNuke.Common.Utilities;
+using DotNetNuke.Entities.Controllers;
 using DotNetNuke.Entities.Portals;
 using DotNetNuke.Entities.Modules;
 using DotNetNuke.Entities.Modules.Definitions;
@@ -488,6 +490,31 @@ namespace DotNetNuke.Web.InternalServices
             Controller.SaveBookMark(PortalSettings.PortalId, UserInfo.UserID, bookmark.Title, bookmark.Bookmark);
 
             return Request.CreateResponse(HttpStatusCode.OK, new { Success = true });
+        }
+
+        public class LockingDTO
+        {
+            public bool Lock { get; set; }
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        [RequireHost]
+        public HttpResponseMessage LockInstance(LockingDTO lockingRequest)
+        {
+            HostController.Instance.Update("IsLocked", lockingRequest.Lock.ToString(), true);
+            HostController.Instance.Update("LockedByUserId", this.UserInfo.UserID.ToString(CultureInfo.InvariantCulture), true);
+            return Request.CreateResponse(HttpStatusCode.OK);
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        [RequireHost]
+        public HttpResponseMessage LockSite(LockingDTO lockingRequest)
+        {
+            PortalController.UpdatePortalSetting(PortalSettings.PortalId, "IsLocked", lockingRequest.Lock.ToString(), true);
+            PortalController.UpdatePortalSetting(PortalSettings.PortalId, "LockedByUserId", this.UserInfo.UserID.ToString(CultureInfo.InvariantCulture), true);
+            return Request.CreateResponse(HttpStatusCode.OK);
         }
 
         private void ToggleUserMode(string mode)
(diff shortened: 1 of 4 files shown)


bdukes:
This implements the functionality in the core, but does not have a UI.  I'll need some guidance on what y'all want to do regarding a Persona Bar UI (where it goes, what it says, etc.)

galatrash-at-dnn:
To develop for Admin Experience, you need first to checkout the repository. Then use Visual Studio as usual to compile the C# code and copy the relusting DLL's to your test site. Before that, the developers need to point to DNN's MyGet packages which are here: [DNN MyGet Public Packages](https://www.myget.org/F/dnn-software-public/api/v3/index.json)

As for React (Javascript code), each project has its own set of files under a folder inside the project's directory. and can be built using ``npm`` or ``yarn``. First, you need to prepare the machine as follows (must perform with elevated credentials):
- Download and Install ``npm``
- Build the different react components by changing to each directory and running: ``yarn run build`` or ``npm run build``
This will take a very long time during the first time you run it. If you want faster build cycles, use ``yarn``:
- Install ``yarn`` using:  ``npm install yarn``
- Repeat the build step by running: ``yarn run build``

If you see any errors during any of these preparatory steps, then check that you have elevated rights and that you have registered the various paths and npm feeds in your machine. Search the web "working with npm" for more information.

Here is the ``DNN public npm feed`` https://www.myget.org/F/dnn-software-public/npm/

Enjoy!

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -538,6 +538,30 @@ public Mode UserMode
...
+            get { return PortalController.GetPortalSettingAsBoolean("IsLocked", PortalId, false); }
+        }
+
+        /// <summary>
+        /// Get the ID of the user who locked this portal
+        /// </summary>
+        public int LockedByUserId
+        {
+            get { return PortalController.GetPortalSettingAsInteger("LockedByUserId", PortalId, Null.NullInteger); }
+        }
```
galatrash-at-dnn: There is no need for this field, as it can be obtained from the HostSettings table column [LastModifiedByUserID]
 or [CreatedByUserID] if the latter is null.

DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
```diff
@@ -488,6 +490,31 @@ public HttpResponseMessage SaveBookmark(BookmarkDTO bookmark)
...
+            public bool Lock { get; set; }
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        [RequireHost]
+        public HttpResponseMessage LockInstance(LockingDTO lockingRequest)
+        {
+            HostController.Instance.Update("IsLocked", lockingRequest.Lock.ToString(), true);
+            HostController.Instance.Update("LockedByUserId", this.UserInfo.UserID.ToString(CultureInfo.InvariantCulture), true);
```
galatrash-at-dnn: There is no need for this field (see other comments in this review).

DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
```diff
@@ -488,6 +490,31 @@ public HttpResponseMessage SaveBookmark(BookmarkDTO bookmark)
...
+            return Request.CreateResponse(HttpStatusCode.OK);
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        [RequireHost]
+        public HttpResponseMessage LockSite(LockingDTO lockingRequest)
+        {
+            PortalController.UpdatePortalSetting(PortalSettings.PortalId, "IsLocked", lockingRequest.Lock.ToString(), true);
+            PortalController.UpdatePortalSetting(PortalSettings.PortalId, "LockedByUserId", this.UserInfo.UserID.ToString(CultureInfo.InvariantCulture), true);
```
galatrash-at-dnn: There is no need for this field (see other comments in this review).

bdukes:
@galatrash-at-dnn I've removed the `LockedByUserId` settings

galatrash-at-dnn:
@bdukes, we are still missing unit/integration tests in your commits.

Also, the commits you supplied just add/modify the settings, but what measures/checks are taken to prevent site updates (whether in the Platform source code or third party extensions? There will be hundreds of places that need to take this into consideration for this to work. What do you think?

bdukes:
@galatrash-at-dnn The "simple" solution that this PR implements is to disable module actions and edit mode (see the changes to `ModuleInstanceContext`), which disables the vast majority of administrative actions to the site.

galatrash-at-dnn:
@bdukes m that's not enough as the PB code is not utilizing this context value and will be able to do changes in PB pages for modules regardless of this flag.

mitchelsellers:
@bdukes do you have plans for PB changes as well to bring this live?  I would prefer if we can approve this PR when we have a true implementation.

bdukes:
@mitchelsellers I won't have time in the near future to implement the PB component of this

PR #1902, opened 2017-10-20
DNN-8513: Create a new extension point in the DNN core at the Database Connection Level.

galatrash:
Refer to this PR #1401 for details.
Jira [DNN-8513](https://dnntracker.atlassian.net/browse/DNN-8513)
>I would like the ability to break away from using the standard SqlConnection, and switch to using ReliableSqlConnection for Azure sites. I would also like the ability to set up retry policies when performing database operations.
>
>I have/will be including a code fix that adds a new extension point at the database connection layer. We can then use a custom database connection that will allow us to use ReliableSqlConnection and the RetryManager functionality that Azure provides, without having to bring those things into the DNN Core.
>
>I have created an AzureSqlConnection Provider to test this functionality, which can be found here: https://github.com/JoshuaBradley/DnnAzureSqlConnectionProvider.

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
+++ b/DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
@@ -1,6 +1,6 @@
 #region Copyright
 // 
-// DotNetNuke® - http://www.dotnetnuke.com
+// DotNetNukeÂ® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2018
 // by DotNetNuke Corporation
 // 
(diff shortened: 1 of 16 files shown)


DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
```diff
@@ -82,7 +82,9 @@ private void Application_Start(object sender, EventArgs eventArgs)
 
             ComponentFactory.Container = new SimpleContainer();
 
+            ComponentFactory.InstallComponents(new ProviderInstaller("databaseConnection", typeof(DatabaseConnectionProvider), typeof(SqlDatabaseConnectionProvider)));
             ComponentFactory.InstallComponents(new ProviderInstaller("data", typeof(DataProvider), typeof(SqlDataProvider)));
+            ComponentFactory.InstallComponents(new ProviderInstaller("databaseConnection", typeof(DatabaseConnectionProvider), typeof(SqlDatabaseConnectionProvider)));
```
zyhfish: why we need add two same code here with line 85?

Website/development.config
```diff
@@ -158,7 +159,7 @@
     <globalization culture="en-US" uiCulture="en" requestEncoding="UTF-8" responseEncoding="UTF-8" fileEncoding="UTF-8"/>
     <!--<globalization culture="en-US" uiCulture="en"  fileEncoding="iso-8859-1" requestEncoding="iso-8859-1" responseEncoding="iso-8859-1"/>-->
     <!-- page level options -->
-    <pages validateRequest="false" enableViewStateMac="true" enableEventValidation="true" viewStateEncryptionMode="Always">
+    <pages validateRequest="false" enableViewStateMac="true" enableEventValidation="true"  viewStateEncryptionMode="Always">
```
zyhfish: no need duplicate space here.

galatrash:
@mitchelsellers these obsoleted public methods are already removed in 9.2.0 since they were deprecated pre v7.0. Therefore, this addition won't have a breaking change from 9.2 to 10.0.

As for fixing the conflict, someone has to take care of that as I'm no longer affiliated with DNN Corp.

bdukes:
@mitchelsellers, I've rebased this onto `development`, and moved the config change from 9.2.0 to 9.3.0 (I didn't see anywhere that referenced the 9.2.0 config file, does that get run automatically based on version number?).  I also reverted the `DataProvider.cs` changes which were unnecessary.

galatrash:
The config files are copied at compile time from `Dnn.Platform\DNN Platform\Website\Install\Config` folder to `Dnn.Platform\Website\Install\Config`

mitchelsellers:
@dnnsoftware/tag Input on this?

PR #1910, opened 2017-10-26
DNN-10389: fixing a potential platform crash issue.

galatrash-at-dnn:
Fixing a case where a disposed memory stream was passed as an argument to another method.


kanm2015:
merged

PR #1911, opened 2017-10-26
DNN-10390: fixing warning CA2217

galatrash-at-dnn:
Do not mark enums with FlagsAttribute in the Platform code.

See MSDN Documentation here http://msdn.microsoft.com/en-us/library/ms182335.aspx


kanm2015:
merged

PR #1923, opened 2017-11-06
DNN-10015 :  <source> tag added to filtered tags.

aliazizoglu:


DNN Platform/Tests/DotNetNuke.Tests.Core/Security/PortalSecurity/PortalSecurityTest.cs
```diff
@@ -0,0 +1,52 @@
...
+            //Arrange
+            var portalSecurity = new DotNetNuke.Security.PortalSecurity();
+            var htmlEntry = @"<source>
+                               Hi this is personal site
+                               Test for source tag  
+                              </source> ";
+            var markup = DotNetNuke.Security.PortalSecurity.FilterFlag.NoMarkup |
+                         DotNetNuke.Security.PortalSecurity.FilterFlag.NoScripting |
+                         DotNetNuke.Security.PortalSecurity.FilterFlag.NoAngleBrackets;
+            var expectedOutput = "  ";
```
galatrash-at-dnn: IMO, we need to have 3 or more separate tests for various combinations settings for these flags:
- none of these flags set; should return the same without filtering
- NoMarkup
- NoScripting
- NoAngleBrackets
- combination of 2 of these together (3 cases)
- all of them together (which is this test).

DNN Platform/Tests/DotNetNuke.Tests.Core/Security/PortalSecurity/PortalSecurityTest.cs
```diff
@@ -0,0 +1,52 @@
...
+        #region Tear Down
+        [TearDown]
+        public void TearDown()
+        {
+            //TODO:implement clean up
+        }
+        #endregion
+
+        #region Test Methods
+        [Test]
```
vano-maisuradze: We can use **TestCase** to easily test different scenarios. Like this:
```
[TestCase("<source></source>", "")]
[TestCase("<source>Hi this is personal Test for source tag</source>   ", "   ")]
[TestCase("<SOURCE>Upper case</SOURCE>", "")]
[TestCase("<source src="https://google.com/">Source with attribute</source>", "")]
[Add more test cases here]
public void Typing_Html_Source_Tag_Should_Not_Be_Allowed(string html, string expectedOutput)
{
        // Use html parameter as input and expectedOutput to assert the result
}
```

vano-maisuradze: OMHO, **Typing_** prefix is not necessary in the name. **Html_Source_Tag_Should_Not_Be_Allowed** is better.

vano-maisuradze:
All tests are passing. Merging to `development` branch

PR #1929, opened 2017-11-11
DNN-10465:The close button of popup get part hidden when change to ma…

Mohtshum:
…ximized mode

- Changes for close button clipped when  maximized and minmized


Website/js/Debug/dnn.modalpopup.js
```diff
@@ -222,14 +219,21 @@
...
+                            var closeButtonWidthCorrection = 0;
+                            var closeButtonHeightCorrection = 0;
+                            if ($('button.ui-dialog-titlebar-close').length) {
+                                closeButtonHeightCorrection = $('button.ui-dialog-titlebar-close').parent('.dnnModalCtrl').height();
+                                closeButtonWidthCorrection = $('button.ui-dialog-titlebar-close').parent('.dnnModalCtrl').width();
                             }
 
-                            newWidth = $window.width() - 40 - personaBarIFrameWidth;
-                            newHeight = $window.height() - 46;
+                            newWidth = $window.outerWidth() - personaBarIFrameWidth - (closeButtonWidthCorrection / 7.5) - 40;
```
zyhfish: Hi @Mohtshum , can we add comment to clarify whats the "closeButtonWidthCorrection / 7.5" mean? whats the "- 40" mean? same to below code about "closeButtonWidthCorrection / 5.5" and "closeButtonHeightCorrection / 11", without comment we can't know what that do and will be hard to maintenance.

Website/js/Debug/dnn.modalpopup.js
```diff
@@ -206,13 +204,12 @@
...
                         var $window = $(window),
                             newHeight,
-                            newWidth,
-                            JQUERY_UI_HEIGHT_SHRINK_OFFSET = 100,
+                            newWidth,                            
                             horizontalPosition = "center",
                             verticalPosition = "center";                            
 
                         if ($modal.data('isMaximized')) {
-                            newHeight = $modal.data('height') + JQUERY_UI_HEIGHT_SHRINK_OFFSET;
```
zyhfish: I'm not sure why this line get removed, its add in DNN-8295, if this line really need to removed, we should ask QA to verify DNN-8295 won't broken with the changes as well.

Website/js/Debug/dnn.modalpopup.js
```diff
@@ -191,9 +191,7 @@
                     closingUrl: closingUrl,
 					open: dialogOpened,
                     close: function() { window.dnnModal.closePopUp(refresh, closingUrl); }
-                })
```
Mohtshum: these lines where making impacts on loading of popup modal on first load and on subsequent minimized calls no impact so different UX experience for a users.

Website/js/Debug/dnn.modalpopup.js
Mohtshum: It was adding height to modal in minimized state in fact more height than maximized state need to check  not sure we need that. And as you mentioned why added in https://dnntracker.atlassian.net/browse/DNN-8295

zyhfish:

PR #1931, opened 2017-11-14
DNN-10507: MVC ActionFilter Now Supports Redirects

ahoefling:
Adds support for `ActionFilter` redirects inside the DNN MVC Framework.

* Redirects the action based on an `ActionFilter` by setting the `filterContext.Result` equal to an `ActionResult`.
* Handles Exception Filter Redirects
    * As long as you specify `filterContext.ExceptionHandled = true`
* Handles OnExecuted `ActionFilter` redirects
* Handles ONExecuting `ActionFilter` redirects

**Usage**
When you define your `ActionFilter` you need to set the `filterContext.Result` property just like you would in ASP.NET MVC development. The new code in this Pull Request then takes that `Result` object and sets it to the View where appropriate. 

**If you are setting the `filterContext.Result` from an exception it is critical that you flip the `filterContext.ExceptionHandled` bit or the view won't render correctly**

**My Use Case**
In my use case I want to display a special DNN MVC Module Error Page to the user when we have an exception thrown from the current controller where our controller has the filter `RedirectOnExceptionAttribute`. See code snippets below for usage of this use case

RedirectOnExceptionAttribute.cs:
```C#
[AttributeUsage(AttributeTargets.Class)]
public class RedirectOnExceptionAttribute : ActionFilterAttribute, IExceptionFilter
{
    public void OnException(ExceptionContext filterContext)
    {
        var controller = (IHasFilterActionMethods)filterContext.Controller;
        filterContext.Result = controller.RedirectToErrorPage();
    }
}
```
IHasFilterActionMethods.cs:
```C#
public interface IHasFilterActionMethods
{
    ActionResult RedirectToErrorPage();
}
```

FooController.cs:
```C#
[RedirectOnException]
public class FooController : DnnController, IHasFilterActionMethods
{
    public ActionResult RedirectToErrorPage()
    {
        return Redirect(Url.Action(nameof(ErrorPage)));
    }

    public ActionResult ErrorPage()
    {
        return View();
    }

    public ActionResult Index()
    {
        return View();
    }
}
```

**OnExecuted and OnExecuting Use Case**
The other supported redirects in this change handle setting the `filterContext.Result` for both overrides in `ActionFilterAttribute` 

* `OnActionExecuted(ActionExecutedContext filterContext)`
* `OnActionExecuting(ActionExecutingContext filterContext)`

The usage in this use case is the same as above. All you need to do is set the `filterContext.Result` in your approprate method in the `ActionFilter` and then the invoker will automatically handle displaying the view.

If you want to handle an exception in the `OnActionExecuted` you will still need to flip the `ExceptionHandled` bit as mentioned above for the exception use case.


ahoefling:
Any update on this?

DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules/ResultCapturingActionInvoker.cs
```diff
@@ -19,20 +19,33 @@
...
 
+using System;
+using System.Collections.Generic;
 using System.Web.Mvc;
 
 namespace DotNetNuke.Web.Mvc.Framework.Modules
 {
     public class ResultCapturingActionInvoker : ControllerActionInvoker
     {
         public ActionResult ResultOfLastInvoke { get; set; }
```
galatrash-at-dnn: Must change to private setter; i.e.
```
    public ActionResult ResultOfLastInvoke { get; private set; }
```

PR #1934, opened 2017-11-16
DNN-10289:Message/Email Tokens Don't Reflect SSL

Mohtshum:
- changes make in LocalizationProvider

### DIFF
--- a/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
+++ b/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
@@ -97,7 +97,19 @@ namespace DotNetNuke.Services.Localization
                 Logger.WarnFormat("Missing localization key. key:{0} resFileRoot:{1} threadCulture:{2} userlan:{3}", key, resourceFileRoot, Thread.CurrentThread.CurrentUICulture, language);
             }
 
-            return string.IsNullOrEmpty(resourceValue) ? string.Empty : resourceValue;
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Localization/LocalizationProvider.cs
```diff
@@ -97,7 +97,19 @@ public string GetString(string key, string resourceFileRoot, string language, Po
...
+            return string.IsNullOrEmpty(resourceValue) ? string.Empty : RemoveHttpUrlsIfSiteisSSLEnabled(resourceValue);
+        }
+
+        private string RemoveHttpUrlsIfSiteisSSLEnabled(string resourceValue)
+        {
+            var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
+
+            if (portalSettings.SSLEnabled)
+            {
+                resourceValue = resourceValue.Replace(@"http:", @"https:");
```
Mohtshum: I think resource entries should not be containing the external URLs. Either pointing to parent or child portals, which is taken care by GetCurrentPortalSettings

galatrash-at-dnn: I agree. But your check is still missing one check for the other setting
``if (portalSettings.SSLEnabled || portalSettings.SSLEnforced)``

Mohtshum: Ok thnx. Put that as well.

kanm2015:
code review done

PR #1938, opened 2017-11-22
DNN-10510: Cannot Request Password Reset Link

Mohtshum:
- Unlock the lock out user first before resetting the password.

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -1590,8 +1590,12 @@ namespace DotNetNuke.Security.Membership
 
 			//Get AspNet MembershipUser
 			MembershipUser aspnetUser = GetMembershipUser(user);
+            if (aspnetUser.IsLockedOut)
+            {
+                aspnetUser.UnlockUser();
+            }
 
-			string resetPassword = ResetPassword(user, answer);
+            string resetPassword = ResetPassword(user, answer);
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -1590,8 +1590,12 @@ public override bool ResetAndChangePassword(UserInfo user, string newPassword, s
 
 			//Get AspNet MembershipUser
 			MembershipUser aspnetUser = GetMembershipUser(user);
+            if (aspnetUser.IsLockedOut)
+            {
+                aspnetUser.UnlockUser();
```
Mohtshum: There is a method with AutoUnlock... but it was using time setting.

kanm2015: looks good to me, for resetting pwd no need to go through AutoUnlock expiry chk

kanm2015:
code review done

DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
galatrash-at-dnn: Why we need to auto-unlock the user here? If Admin has locked out any user, then this user can unlock themselves this way which may create a security issue.

Mohtshum: Actually Admin authorize and lockout mostly occur if someone has attempted with wrong password.

PR #1939, opened 2017-11-22
Bug/dnn 10289

Mohtshum:
DNN-10289: Changes for null exceptions
- Passed the portalsetting argument to method


kanm2015:
looks good, merged

PR #1940, opened 2017-11-23
DNN-9220: No wrap in HTML Editor

Mohtshum:
- this CSS3 checked on all browser. this rule take care of possible word wrap  break point

### DIFF
--- a/Website/Resources/Shared/stylesheets/dnndefault/7.0.0/default.css
+++ b/Website/Resources/Shared/stylesheets/dnndefault/7.0.0/default.css
@@ -223,6 +223,7 @@ small {
     font-size: 12px;
     font-weight: normal;
     line-height: 18px;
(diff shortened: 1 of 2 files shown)


Website/Resources/Shared/stylesheets/dnndefault/7.0.0/default.css
```diff
@@ -231,6 +231,11 @@ small {
     color: #999;
 }
 
+/* CSS3 to wrap at possible break point */
+.Normal {
```
kanm2015: please apply the rule to NormalDisabled & NormalDeleted classes as well

kanm2015: please also update /8.0.0/default.css

kanm2015:
code review done

PR #1941, opened 2017-11-24
DNN-8486: User Profile Image in View Profile Module doesn't use Image…

Mohtshum:
…Handler to render images

- replacing template with ImageHandler based URL to bypass internal logic


kanm2015:
looks good, merged

PR #1942, opened 2017-11-28
DNN-10550: fix for endless exceptions loop to prevent stack overflow.

galatrash-at-dnn:
Also fixed a unit test that causes such a behavior.


kanm2015:
looks good and all unit tests passed, merged

PR #1944, opened 2017-11-30
DNN-10291: Browser autofill feature interrupts the password reset lin…

Mohtshum:
…k/query string

- Disabling the username text box not allowing browser anymore to change the username or email


kanm2015:
code review done

PR #1946, opened 2017-12-01
DNN-10557: Set CommentXML to NULL in case of less than 2000 characters

nimrozm:
Included Updated Stored Procedure Script in 09.02.00.SqlDataProvider.
Older version of stored procedure is available in 07.01.02.SqlDataProvider


Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
```diff
@@ -736,6 +736,96 @@ BEGIN
...
+DECLARE @cdataComment nvarchar(max)
+
+IF EXISTS(SELECT * FROM {databaseOwner}[{objectQualifier}Journal_Comments] WHERE JournalId = @JournalId AND CommentId = @CommentId)
+BEGIN
+	IF (LEN(@Comment) < 2000)
+	BEGIN
+		IF @CommentXML IS NOT NULL
+		BEGIN
+			SET @cxml = @CommentXML
+			SET @cxml.modify('delete /root/comment')
```
kanm2015: From performance perspective, instead of deleting it here, can we prevent API from sending it in the xml?

kanm2015: Also, if comment is the only child node, after deletion it'll be "<root />" in the CommentXML field and the UI will still have problem to display comment.

kanm2015:
Looks like George modified the same file recently, please resolve conflicts. Thx

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
nimrozm: UI is not just looking for some value in CommentXML field but specifically the comment node in CommentXML field. See line 419 in https://github.com/dnnsoftware/Dnn.Evoq.Social/commit/54cc1fc66d4a7a4c89d5f1f2fe70c04177b5a884

kanm2015: Here's js code, since commentXml is not null (root element only), it won't show comment, right?
if (that.commentXml()) {
    return $(that.commentXml()).find("comment").text().replace(/^<\!\[CDATA\[|\]\]>$/g, '');
} else {
    return that.comment();
}

nimrozm: Oops my bad, missed to commit the changes. Here's the updated code for the UI check
https://github.com/dnnsoftware/Dnn.Evoq.Social/commit/fa952c5e9ab3e3097ef52d18e0ab933a63ff27cc

kanm2015: Ye, this works. see my first comment? can you improve the API?

nimrozm: API call only sends old comment in xml form if the logged in user is a moderator otherwise it doesn't. Will fix it in the next commit.
Will still keep the stored proc part as it is as a safety net.

kanm2015: I'm not a fan of putting business logic in stored proc but sure, let's keep this change for now. Ideally, it should reset commentXml to null if there's no child nodes. :)

nimrozm: modified API to reset commentXml to null if there's no child nodes. Removed the logic of deleting comment node from stored procedure.

kanm2015:
code review done

PR #1950, opened 2017-12-07
DNN-10547: added try/catch on GetSystemMessage

kanm2015:


galatrash-at-dnn:
Turns out this first merge is not fixing the issue properly. @kanm2015 will change again and submit a new PR.

PR #1952, opened 2017-12-08
DNN-10143: Fix ambiguous column reference

nimrozm:
Original definition for this stored procedure can be found [here](https://github.com/dnnsoftware/Dnn.Platform/blob/0659a27424fe90a4805e709ba71d303e747b81eb/DNN%20Platform/Modules/DnnExportImport/Providers/DataProviders/SqlDataProvider/09.01.00.SqlDataProvider#L1999)


kanm2015:
code review done

PR #1954, opened 2017-12-11
DNN-9070: SI PB Admin Logs: No Mail From Address (Portal SMTP)

Mohtshum:


DNN Platform/Library/Services/Log/EventLog/LogTypeConfigInfo.cs
```diff
@@ -66,7 +69,17 @@ public DateTime StartDateTime
 
         public bool EmailNotificationIsActive { get; set; }
 
-        public string MailFromAddress { get; set; }
+        public string MailFromAddress
```
Mohtshum: Actually If senderadress is empty then email was not getting shoot. So set it with the current portal and also checked for web application and Scheduler process as well.

Website/Providers/DataProviders/SqlDataProvider/09.02.00.SqlDataProvider
```diff
@@ -904,6 +904,108 @@ BEGIN
 END
 GO
 
+/**************************************************************************/
```
Mohtshum: Please check is it correct place to make db changes

kanm2015:
code review done

PR #1957, opened 2017-12-13
DNN-7812: Redirect to Login page fails when accessing a restricted pa…

Mohtshum:
…ge of not default language

- Using GetCurrentLocale which in turn using DefaultLocale if not finding any


kanm2015:
code review done

PR #1960, opened 2017-12-18
DNN-10564: Using Direct DnnImageHandler.ashx call and removing api ca…

Mohtshum:
…lls for uploaded images


kanm2015:
code review done

PR #1964, opened 2017-12-22
Bug/dnn 10619

ricardoleon:


DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -1704,6 +1712,10 @@ public void MoveTabToParent(TabInfo tab, int parentId)
 
             //Clear the Cache
             ClearCache(tab.PortalID);
+
+            var portalId = GetPortalId(tab.TabID, -1);
```
Mohtshum: Please use  PortalController.Instance.GetCurrentPortalSettings() to get portal ID rather than -1 in above changes as well.

ashishpd:
@zyhfish can you please review @Mohtshum's comments and make changes if you agree.

DNN Platform/Library/Entities/Tabs/TabController.cs
zyhfish: IMO -1 is ok here, as there have other places used -1 as well. shoudn't use GetCurrentPortalSettings as it will be NULL if called from scheduler.

zyhfish:

PR #1965, opened 2017-12-24
DNN-8552: Account Login Module Redirect Issue

Mohtshum:
- Making sure redirect after login only works if request for home page and authentication occured in same lifecycle of request


Mohtshum:
This is not relevant to the DNN-8552

PR #1966, opened 2017-12-26
DNN:10621 User Profile Image thumbnail in the browse popup doesn't us…

Mohtshum:
…e ImageHandler to render images

- for thumbnail in file fileuploader


kanm2015:
code review done

PR #1969, opened 2017-12-29
DNN-10645:  If "Redirect After Login" is set, any page having "Accoun…

Mohtshum:
…t Module" will be inaccessible for logged-in user

- Ensuring redirect after login only works if request for home page and authentication occured in same lifecycle of request


Website/DesktopModules/Admin/Authentication/Login.ascx.cs
```diff
@@ -252,6 +252,8 @@ protected string RedirectURL
             }
         }
 
+        private bool NeedRedirectAfterLogin => LoginStatus == UserLoginStatus.LOGIN_SUCCESS;
```
kanm2015: please take the following status into account,
LOGIN_SUPERUSER
LOGIN_INSECUREHOSTPASSWORD
LOGIN_INSECUREADMINPASSWORD

kanm2015:
code review done

PR #1974, opened 2018-01-03
DNN-10646: Accessing a restricted page while logging in using top nav…

Mohtshum:
…igation "login" link redirecting to home page

- setting correct url to set for account login link

### DIFF
--- a/Website/admin/Skins/Login.ascx.cs
+++ b/Website/admin/Skins/Login.ascx.cs
@@ -145,11 +145,20 @@ namespace DotNetNuke.UI.Skins.Controls
 				        }
 
 				        string returnUrl = HttpContext.Current.Request.RawUrl;
-				        if (returnUrl.IndexOf("?returnurl=") != -1)
-				        {
-					        returnUrl = returnUrl.Substring(0, returnUrl.IndexOf("?returnurl="));
-				        }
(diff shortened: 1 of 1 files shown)


Website/admin/Skins/Login.ascx.cs
```diff
@@ -145,11 +145,20 @@ protected override void OnLoad(EventArgs e)
 				        }
 
 				        string returnUrl = HttpContext.Current.Request.RawUrl;
-				        if (returnUrl.IndexOf("?returnurl=") != -1)
-				        {
-					        returnUrl = returnUrl.Substring(0, returnUrl.IndexOf("?returnurl="));
-				        }
-				        returnUrl = HttpUtility.UrlEncode(returnUrl);
+                        const string returnUrlToken = "?returnurl=";
+
```
mitchelsellers: I would like to see the string manipulation on this control updated to use the overload that doesn't require a ToLowerInvariant call.  The use of ToLower() etc has some memory implications and additionally the lack of using a conversion or case insensitive comparison could impact URL's for the initial match.

tpluscode:
Oh, this one got forgotten. Actually, Ash made the call back in January to not change this behaviour.

@ashishpd, the internal link is https://jira.devfactory.com/browse/DNN-8020

Closing this for now

PR #1977, opened 2018-01-05
DNN-10601: preventing handling external URLs in image handler.

galatrash-at-dnn:


kanm2015:
change looks good, merged

PR #1991, opened 2018-01-18
DNN-10693: adding Appveyor build status.

galatrash-at-dnn:
Also, fixing the bundle packaging and unifying all Newtonsoft.Json package references.


kanm2015:
review done, merged

PR #1997, opened 2018-01-26
DNN-10728: add site path to image thumbnail request path.

zyhfish:


Website/Resources/Shared/components/FileUpload/dnn.FileUpload.js
```diff
@@ -25,6 +25,10 @@
         };
     })();
 
+    var getSiteRoot = function () {
```
kanm2015: do we need this func? what about just dnn.getVar("sf_siteRoot", "/")

kanm2015:
code review done

PR #2006, opened 2018-03-01
DNN-10783: inject analytics content in after page init complete which set to inject at top.

zyhfish:


zyhfish:
The reason i have add on https://dnntracker.atlassian.net/browse/DNN-10783 within description:
"This is related to analytics http module, if you set the config as prepend script into body element, it will cause the control index changed then viewstate becomes invalid after postback, so that all viewstate will lost."

PR #2008, opened 2018-03-05
DNN-8893: clear child files' cache after folder renamed.

zyhfish:


DNN Platform/Tests/DotNetNuke.Tests.Integration/Modules/DigitalAssets/DigitalAssetsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        private readonly int PortalId = 0;
+
+        #endregion
+
+        #region SetUp
+
+        public DigitalAssetsTests()
+        {
+        }
```
tpluscode: Empty constructor is redundant so let's remove it

zyhfish:
changed applied, thx!

PR #2011, opened 2018-03-08
DNN-8504 Fixed number of pages returned by get-portal command

mean2me:


DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -1469,6 +1469,33 @@ public TabCollection GetTabsByPortal(int portalId)
...
         }
 
+        /// <summary>
+        /// Get the actual visible tabs for a given portal id. 
+        /// System Tabs and Admin Tabs are excluded from the result set.
+        /// </summary>
+        /// <param name="portalId"></param>
+        /// <param name="adminTabId"></param>
+        /// <returns></returns>
+        public TabCollection GetTabsByPortal(int portalId, int adminTabId)
```
mihai-cucicea-xo: Shouldn't we adapt the old code using optional parameters? Having definition as `public TabCollection GetTabsByPortal(int portalId, int? adminTabId = null)` and then have an `if` statement before returning to decide if it needs to be filtered or not? This way we avoid code duplication.

DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -1469,6 +1469,33 @@ public TabCollection GetTabsByPortal(int portalId)
...
+        {
+            string cacheKey = string.Format(DataCache.TabCacheKey, portalId);
```
mitchelsellers: This new method is going to introduce an undesired behavior.  We now have two methods, with different parameters using the same cache key to store data.  This could result in the expected/actual results being different.  A different key should be created based on the presence of an AdminTabId.

Also, given the possibility of large portals etc, is there any ability to do the filtering at the DB layer rather than with LINQ after?

mean2me: You're right. I should check if somewhere in the code we really need the full list of pages, including System and Admin pages. In this case, I would keep two methods with a different cache key.

DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -1469,6 +1469,33 @@ public TabCollection GetTabsByPortal(int portalId)
...
+        /// <summary>
+        /// Get the actual visible tabs for a given portal id. 
+        /// System Tabs and Admin Tabs are excluded from the result set.
+        /// </summary>
+        /// <param name="portalId"></param>
+        /// <param name="adminTabId"></param>
+        /// <returns></returns>
+        public TabCollection GetTabsByPortal(int portalId, int? adminTabId)
+        {
+            var tabs = GetTabsByPortal(portalId);
```
tpluscode: I think that @mitchelsellers' idea was to have two separate keys `{ portalId }` and `{ portalId, adminTabId }` rather than retrieving one and filtering in memory. However, for a functionality probably scarcely used I think we can live with that

DNN Platform/Library/Entities/Tabs/ITabController.cs
```diff
@@ -246,6 +246,16 @@ public interface ITabController
...
 
+
+        /// <summary>
+        /// Get the actual visible tabs for a given portal id. 
+        /// System Tabs and Admin Tabs are excluded from the result set.
+        /// </summary>
+        /// <param name="portalId"></param>
+        /// <param name="adminTabId"></param>
+        /// <returns>tab collection</returns>
+        TabCollection GetTabsByPortal(int portalId, int? adminTabId);
```
tpluscode: I find this API a bit misleading:

* The second parameter is there because `AdminTabId` from portal to portal. This is understandable but this method can be abused by passing any number to produce a weird result
* other than in the comment, the method hides the fact that system pages are also excluded

Here are some ideas:

* a more specific name would be more helpful like `GetUserTabsByPortal` or `GetNonSystemTabs`. 
  * I don't necessarily buy into the `*ByWhatever` suffixes but I'm not adamant
* I would remove the parameter and have the `AdminTabId` retrieved inside. `TabCollection GetUserTabsByPortal(int portalId)` or even `TabCollection GetUserTabsByPortal(PortalInfo portal)` would do.   
  * you would then just call it as `GetUserTabsByPortal(this)`
* **OR** just do the filtering inside the `PortalInfo#Pages`? 🤔

mean2me: GetUserTabsByPortal is the one I prefer 🤔 
Reasons: 

- AdminTabId is something detected by on check for "Admin" string in the page name (some improvement here in 9.2.1?)
- Get**UserPage**ByPortal let us undertand we are not including admin and system pages

DNN Platform/Library/Entities/Tabs/TabController.cs
mitchelsellers: I agree @tpluscode I would have expected a separate key and object in cache, just not sure of the usage.  Given the strong usage of cache in other locations though, I would expect this to cache results as well.

tpluscode: Well, we cannot avoid filtering in memory without more changes to CBO. I think the current solution is good enough where we reuse the existing cache entry

PR #2012, opened 2018-03-08
DNN-9776: do not show the user skin objects in custom error page.

zyhfish:


DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -405,5 +405,16 @@ public static void Handle404Exception(HttpResponse response, PortalSettings port
                 response.End();
             }
         }
+
+        /// <summary>
+        /// Detect whether current page is custom error page.
+        /// </summary>
+        /// <param name="portalSettings"></param>
+        /// <returns></returns>
+        public static bool InErrorPageRequest(PortalSettings portalSettings)
```
tpluscode: I find `UrlUtils` an unintuitive location for such method.

What would you think about an extension method like 

```
public static bool IsActiveTabAnErrorPage(this PortalSettings portalSettings)
```

?

zyhfish:
ShowInErrorPage is added for backward compatible, so if customer really want to show those controls in custom error page, they have way to do it.

PR #2017, opened 2018-03-13
DNN-8504 Bugfix: correct CdfVersion number returned by get-portal com…

mean2me:
…mand


zyhfish:
Hi @mean2me , please check the code at https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/DotNetNuke.Web.Client/ClientResourceSettings.cs#L64.

For CrmVersion, it has 2 levels settings(Portal Level/Host Level), if set portal level override host level, then should get the value from PortalSettings table, otherwise should get from HostSettings table.

mean2me:
@zyhfish  It means we should also review the way it's updated. When we update it from ServerSettings, it only affect HostSettings table.

PR #2018, opened 2018-03-13
DNN-8504

mean2me:


zyhfish:
Hi @mean2me , sorry that i didn't clarify what i think need to be done here in message. what i think we doesn't need to do any change in platform, but just call this method from PB side.

after i research more i notice that the GetVersion method always check the setting for current portal, but can't specific the portal id to get setting, that means if we call get-portal with a different portal id, it may return wrong value.

so the easiest way i think is add a method in PB to get portal setting if override set to true, otherwise get from host settings, method have portal id parameter so that can get correct setting. what's your opinion?

Thanks,

mean2me:
Hi @zyhfish , I made some changes so that we can return the proper CdfVersion according to the current Client resource management mode. I prefer to not introduce controller functionalities in PB modules. IMO, it's better to keep changes in the controller side.

PR #2019, opened 2018-03-15
DNN-18077: add sharpziplib 0.81.0 install package for backward compatible.

zyhfish:


iJungleboy:
I believe this doesn't work - because I had tried it before :). As far as I remember, it results in parts of DNN breaking. I don't know which parts, but as an example the extension area of DNN broke when we tested it. Basically anything that does reflection runs into issues when the very same namespace is included in two separate DLLs :(.

iJungleboy:
I created a new pull request specifically fixing this issue without adding another DLL - https://github.com/dnnsoftware/Dnn.Platform/pull/2118

PR #2020, opened 2018-03-15
DNN-18120: add the override method for MinifyFile.

zyhfish:


DNN Platform/DotNetNuke.Web.Client/Providers/DnnCompositeFileProcessingProvider.cs
```diff
@@ -34,6 +35,31 @@ public class DnnCompositeFileProcessingProvider : CompositeFileProcessingProvide
...
+                    throw new InvalidOperationException("Cannot read input stream");
+
+                if (stream.CanSeek)
+                    stream.Position = 0;
+
+                var reader = new StreamReader(stream);
+                return reader.ReadToEnd();
+            };
+
+            switch (type)
```
tpluscode: I think `if` statements will be clearer here, getting rid of the ternary operator 

```
if (type == ClientDependencyType.Css && MinifyCss) 
{
    return CssHelper.MinifyCss(fileStream);
}

if (type == ClientDependencyType.Javascript && MinifyJs) 
{
    return JSMin.CompressJS(fileStream);
}

return streamToString(fileStream);
```

tpluscode: In which case you won't need the `Func` either. Just inline the `streamToString`

DNN Platform/DotNetNuke.Web.Client/Providers/DnnCompositeFileProcessingProvider.cs
```diff
@@ -34,6 +35,31 @@ public class DnnCompositeFileProcessingProvider : CompositeFileProcessingProvide
     {
         private readonly ClientResourceSettings clientResourceSettings = new ClientResourceSettings();
 
+        public override string MinifyFile(Stream fileStream, ClientDependencyType type)
+        {
+            Func<Stream, string> streamToString = stream =>
+            {
+                if (!stream.CanRead)
```
tpluscode: Also, please always use braces for a single statement

zyhfish:
Hi @tpluscode , this code was copied from https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Components/ClientDependency/Source/CompositeFiles/Providers/BaseCompositeFileProcessingProvider.cs#L600, which i don't want to do refactor but keep them have same structure, then if ClientDependency.Core upgrade to new version future, we can apply the changes into our provider quickly.

PR #2023, opened 2018-03-19
DNN-18172: clean build warnings.

zyhfish:


DNN Platform/Admin Modules/Dnn.Modules.Console/web.config
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<configuration>
+	<runtime>
+		<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+			<dependentAssembly>
+				<assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30AD4FE6B2A6AEED" culture="neutral"/>
+				<bindingRedirect oldVersion="0.0.0.0-10.0.0.0" newVersion="10.0.0.0"/>
```
tpluscode: No harm in having binding redirects but I looked at the projects and it turns out that we only have version 10 installed of Newtonsoft.Json. Why are there warnings in the first place?

DNN Platform/Components/ClientDependency/Source/ClientDependency.Core.csproj
```diff
@@ -29,7 +29,7 @@
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>bin\ClientDependency.Core.XML</DocumentationFile>
-    <NoWarn>1591,1573</NoWarn>
+    <NoWarn>1591,1573,618</NoWarn>
```
tpluscode: I don't think we should be ignoring obsolete warnings. Any obsolete code should have a recommended replacement, right? If there is non I think we should keep the warnings

DNN Platform/Dnn.AuthServices.Jwt/Dnn.Jwt.dnn
```diff
@@ -31,12 +31,14 @@
         </component>
         <component type="Script">
           <scripts>
-            <basePath>DesktopModules\AuthenticationServices\JWTAuth\Data\Scripts</basePath>
+            <basePath>DesktopModules\AuthenticationServices\JWTAuth</basePath>
```
tpluscode: Oh, how is this related to build warnings? 🙂

DNN Platform/Components/ClientDependency/Source/ClientDependency.Core.csproj
zyhfish: These warnings are comes from ClientDependency.Core project, and it contains several obsoleted warnings, if not ignore this maybe we can't reach the goal "less than 10 warnings.".

DNN Platform/Dnn.AuthServices.Jwt/Dnn.Jwt.dnn
zyhfish: The build scripts will try to copy these files from project directory(without basePath defined), and throw out warnings because these files aren't there.

DNN Platform/Components/ClientDependency/Source/ClientDependency.Core.csproj
tpluscode: So these are actually coming from the 3rd party code we copied to DNN, right?

DNN Platform/Modules/MemberDirectory/DotNetNuke.Modules.MemberDirectory.csproj
```diff
@@ -37,7 +37,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <DocumentationFile>bin\DotNetNuke.Modules.MemberDirectory.XML</DocumentationFile>
-    <NoWarn>1591</NoWarn>
+    <NoWarn>1591,618</NoWarn>
```
tpluscode: And what about the obsolete warnings in this project?

DNN Platform/Admin Modules/Dnn.Modules.Console/web.config
zyhfish: this warning should be caused by there have old Newtonsoft.Json assembly installed in system level.

DNN Platform/Components/ClientDependency/Source/ClientDependency.Core.csproj
zyhfish: Yep.

PR #2024, opened 2018-03-21
DNN-18204: catch CryptographicException as well for backward compability.

zyhfish:


DNN Platform/Library/Entities/Portals/PortalInfo.cs
```diff
@@ -805,7 +806,7 @@ public void Fill(IDataReader dr)
                     ? p
                     : Security.FIPSCompliant.DecryptAES(p, Config.GetDecryptionkey(), Host.Host.GUID);
             }
-            catch(FormatException)
+            catch (Exception ex) when (ex is FormatException || ex is CryptographicException)
```
tpluscode: Very cool! I like the use of C# 6 ❤️.

tpluscode:
I've rebased your changes onto release/9.2 and merged into development

PR #2025, opened 2018-03-26
NOJIRA: fixing NuGet packages references (to enable restore/update. D…

galatrash:
…isable troubling code to pass Jenkins build.

These changes necessary to get Jenkins builds passig so we can move the migration from TeamCity to Jenkins production server.

### DIFF
--- a/DNN Platform/Connectors/Azure/Dnn.AzureConnector.csproj
+++ b/DNN Platform/Connectors/Azure/Dnn.AzureConnector.csproj
@@ -63,6 +63,9 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\Components\WindowsAzure\Microsoft.WindowsAzure.Storage.dll</HintPath>
     </Reference>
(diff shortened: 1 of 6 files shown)


galatrash:
@zyhfish the changes are simple and for nuget references only. I need to get this running to move forward with Jenkins migration. Please review ASAP.

zyhfish:
Hi @galatrash , all PR need be processed by @tpluscode.

galatrash:
@tpluscode , tge removal of the `#if false` should be done by your team. I added it to get the builds passing temporarily, but for sure you need to re-enable this test and fix the compilation error that is seen in  when this is build in TC here https://build.dnnsoftware.com/viewLog.html?buildId=1006350&tab=buildResultsDiv&buildTypeId=DnnPlatform921Public_Compile

tpluscode:
Oh I see, this uses inline `out` variables from C# 7. I know we have this limitation on TC but why cannot we support that on Jenkins?

galatrash:
Because we are compiling in VS 2015 compatibility mode (msbuild 14.0). And also DNN's code main code is still limited to VS 2015 and must not use C# 7 features.

PR #2027, opened 2018-03-30
Bug/dnn 12099

mean2me:


DNN Platform/Library/UI/Skins/Skin.cs
```diff
@@ -414,6 +414,25 @@ private bool ProcessModule(ModuleInfo module)
             return success;
         }
 
+        /// <summary>
+        /// Handle access denied errors by displayign an error message
```
tpluscode: Nit: Typo: "displaying"

DNN Platform/Library/UI/Skins/Skin.cs
```diff
@@ -425,7 +444,7 @@ private bool ProcessMasterModules()
                 // Versioning checks.
                 if (!TabController.CurrentPage.HasAVisibleVersion)
                 {
-                    Response.Redirect(Globals.NavigateURL(PortalSettings.ErrorPage404, string.Empty, "status=404"));
+                    HandleAccesDenied(true);
```
tpluscode: Hm, you created this method with a parameter but this is the only usage?

DNN Platform/Library/UI/Skins/Skin.cs
mean2me: sorry

DNN Platform/Library/UI/Skins/Skin.cs
mean2me: I propose to replace direct call to Response.Redirect with that method, if you agree

tpluscode: I don't mind a method. What I mean is that you added the `bool redirect = false` parameter but this is the only usage, right? It will never be called otherwise. So maybe it should be paremeterless instead if we don't need it now?

mean2me: On lines 455 and 478, error message is displayed without redirect.

tpluscode: Oh, ok. Then we should replace them with a call to `HandleAccesDenied`, right?

mean2me: Just pushed ;)

PR #2028, opened 2018-04-02
DNN-8511: Fix menu style hiding other pages for long page names

ricardodiasc:


bdukes:
I don't think these styles belong in `default.css`, they should probably be in https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Skins/Xcillion/Skin/skin.css

tpluscode:
@ricardodiasc how did you choose those specific values for `max-width`? Experimentally or do they appear somewhere else in our styles?

ricardodiasc:
@tpluscode I made it based on container size from bootstrap that is in use. Since navbar does not set a max-width based on the available space, word-wrap would not work

tpluscode:
What I mean is to use some variable instead of concrete values. But we don't do SASS/SCSS, do we?

ricardodiasc:
There is no SASS/SCSS/LESS in content editing. We could use css-variables but is not supported by IE11 https://caniuse.com/#feat=css-variables

PR #2029, opened 2018-04-02
DNN-18552 move ClientDependency to a submodule

tpluscode:


bdukes:
Is there a plan to handle the DNN-specific modifications to the CDF?

tpluscode:
Not a the moment, unless you'd want to start discussions with the authors about what and why we've modified. Some of it could be merged back I suppose. Personally, I'm not aware of the context of most of those changes though. I don't know the first thing about why they were introduced.

PR #2032, opened 2018-04-03
(DNN-10795) - All pages except home page return 404

cameron-moore:
I have witnessed that occasionally on app pool recycle all,except the home page, will return 404 until the application pool is recycled a second time.

I've reviewed the code & believe that the root cause of the issue is due to the fact that the code that builds the tab index, portalDepths dictionary & tabPaths dictionary is not thread safe. I can see code in the method TabIndexController.FetchTabDictionary is using SharedDictionary classes to store the tab dictionaries, however the code is not thread safe when adding the dictionaries to the cache. Therefore when multiple threads are executing the FetchTabDictionary method it's possible for an empty dictionary to be added to the cache.

To resolve this issue the code has been updated so that only one thread can add the dictionaries to cache at a time.

### DIFF
--- a/DNN Platform/Library/Entities/Urls/TabIndexController.cs
+++ b/DNN Platform/Library/Entities/Urls/TabIndexController.cs
@@ -1392,6 +1392,26 @@ namespace DotNetNuke.Entities.Urls
             return rewritePath;
         }
 
+        /// <summary>
+        /// Returns whether the portal specified exists in the Tab index or not.
+        /// </summary>
+        /// <param name="portalDepths">The current portalDepths dictionary.</param>
+        /// <param name="portalId">The id of the portal to search for.</param>
+        /// <returns></returns>
+        private static bool PortalExistsInIndex(SharedDictionary<int, PathSizes> portalDepths, int portalId)
+        {
+            bool result = false;    
+
+            if (portalDepths != null)
+            {
+                using (portalDepths.GetReadLock())
+                {
+                    result = portalDepths.ContainsKey(portalId);
+                }
+            }
+            return result;
+        }
+
         /// <summary> 
         /// Gets the Tab Dictionary from the DataCache memory location, if it's empty or missing, builds a new one 
         /// </summary>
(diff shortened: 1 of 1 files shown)


tpluscode:
The dictionary implementation is the problem? Maybe it's worth considering to replace `SharedDictionary` with [`ConcurrentDictionary`](https://msdn.microsoft.com/en-us/library/dd287191(v=vs.110).aspx)

cameron-moore:
Hi Tomaz,

The issue is not with the type of dictionary used but with the fact that multiple threads can add the dictionary to cache at the same time. The SharedDictionary used is already thread safe. It’s the code that is creating the dictionary and adding to cache that needs to be changed not the dictionary it itself.

Kind regards,
Cameron.

mitchelsellers:
@cameron-moore  Can we get the CLA completed for this

ohine:
@cameron-moore The easiest way to get the CLA completed is to push another change into your fork/branch. Maybe update a code comment in this file? The other option would be to close and resubmit the fix. Sorry for the issue, we promise it's just a one time thing the first time you submit a change to the project.

cameron-moore:
Hi guys,

CLA has now been completed.

Cheers,
Cameron.

ohine:
@cameron-moore Thank you for getting that done! We will review and get this hopefully into v9.3.0.

mitchelsellers:
@cameron-moore What testing has been done on this change?  Do you happen to know of a quick/easy way to recreate the contention issues that happened with the old way?

cameron-moore:
@mitchelsellers  Because it's a thread contention issue it's quite difficult to reproduce the issue on demand because it only occurs when multiple requests are being made to the site simultaneously and the tab dictionaries are yet to be stored in cache, so it's only going to happen when the DNN instance is first loading up, or just after the dictionaries are removed from cache. 
You could perhaps setup something up where you script many Curl processes to request a page at the same time. The application pool should be spun down prior to making the requests. Even this though, is not guaranteed to re-produce the issue reliably.

When I was investigating I stepped through the code and updated the tab dictionary to be an empty collection just prior to being cached, to simulate the contention. Doing this I was able to reproduce the same behavior as we were seeing in our production DNN instance, where all requests to the site result in a 404. 

Changes made in this pull request were also applied to our production instance of DNN, running a couple of hundred portals. Prior to applying the change in our production environment we were seeing all pages 404ing at least once a week. Each time this would happen we'd need to recycle the application and things would start working again. We've not seen any repeats of the issue in production, where all pages 404, since this change was applied.

Hope this helps.

mitchelsellers:
I have been testing this fix on a few installations, and it does appear to work.  Following Microsoft guidance this appears to be sound as well.  This issue impacts Axure installations more than others.

@dnnsoftware/tag what do we think about including this in 9.3.0 given that we are doing extended testing?  My vote is to include this.  @daguiler and @ashishpd would like your specific weigh in as well

robaxx:
FYI - Cameron's fix here has completely resolved this problem on an 8.0.4 instance of mine that was having multiple daily outages over the past year. This should have been integrated ten years ago. Let's make it happen now.

cameron-moore:
@mitchelsellers Feel free to delete Fixed-DLLs folder. It shouldn't have been merged. I just added that in my fork so that if people want to download a pre-compiled version of the DotNetNuke.dll, that has the  changes, they can get it from my fork. It shouldn't be in the the main DnnPlatform project code.

rudgr:
We've been experiencing this issue on multiple DNN installs, has this fix been included in DNN 9.3.0 RC?

https://github.com/dnnsoftware/Dnn.Platform/releases

cameron-moore:
@rudgr  I downloaded & took a look at the source package for 9.3.0 RC. It looks like the code has been included in the 9.3.0 RC release but just it's been missed from the release notes.

valadas:
It should be yes.

rudgr:
awesome!!!!!!!!! :)

mitchelsellers:
I updated the release notes to call this out, apologies that we missed it @cameron-moore

ohine:
Sorry @cameron-moore, I completely spaced out on this PR and adding the noteworthy tag to it. It's a new addition to our process and we just need to do better next time around. Thanks for letting us know! and thanks again for taking the time to research the issue and submit a fix!!

PR #2033, opened 2018-04-04
DNN-17683:  SI: DDR Menu: Exclude Nodes

Mohtshum:
- Simplify the matching logic for filtering nodes
- Add new method to remove the nested or child node by recursion

### DIFF
--- a/DNN Platform/Modules/DDRMenu/MenuBase.cs
+++ b/DNN Platform/Modules/DDRMenu/MenuBase.cs
@@ -197,22 +197,15 @@ namespace DotNetNuke.Web.DDRMenu
 			    }
 				else
 				{
-					var nodeText2 = nodeText;
-					filteredNodes.AddRange(
-						RootNode.Children.FindAll(
-							n =>
-							{
-								var nodeName = n.Text.ToLowerInvariant();
-								var nodeId = n.TabId.ToString();
-								return (nodeText2 == nodeName || nodeText2 == nodeId);
-							}));
+					var nodeText2 = nodeText;                    
+                    filteredNodes.Add(RootNode.FindByNameOrId(nodeText2));
 				}
 			}
 
             // if filtered for foksonomy tags, use flat tree to get all related pages in nodeselection
 		    if (flattenedNodes.HasChildren())
 		        RootNode = flattenedNodes;
-			RootNode.Children.RemoveAll(n => filteredNodes.Contains(n) == exclude);
+			RootNode.RemoveAll(filteredNodes);
 		}
 
         private void FilterHiddenNodes(MenuNode parentNode)
--- a/DNN Platform/Modules/DDRMenu/MenuBase.cs
+++ b/DNN Platform/Modules/DDRMenu/MenuBase.cs
@@ -197,15 +197,21 @@ namespace DotNetNuke.Web.DDRMenu
 			    }
 				else
 				{
-					var nodeText2 = nodeText;                    
-                    filteredNodes.Add(RootNode.FindByNameOrId(nodeText2));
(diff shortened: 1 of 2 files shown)


DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -197,22 +197,15 @@ private void FilterNodes(string nodeString, bool exclude)
...
-							}));
+					var nodeText2 = nodeText;                    
+                    filteredNodes.Add(RootNode.FindByNameOrId(nodeText2));
 				}
 			}
 
             // if filtered for foksonomy tags, use flat tree to get all related pages in nodeselection
 		    if (flattenedNodes.HasChildren())
 		        RootNode = flattenedNodes;
-			RootNode.Children.RemoveAll(n => filteredNodes.Contains(n) == exclude);
```
tpluscode: I see that the `exclude` parameter is not longer used in this method but it is called with both `true` and `false` in `MenuBase#PreRender`. If you're certain that your changes are correct for all cases please refactor the calling code too.

DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -197,22 +197,15 @@ private void FilterNodes(string nodeString, bool exclude)
...
-					var nodeText2 = nodeText;
-					filteredNodes.AddRange(
-						RootNode.Children.FindAll(
-							n =>
-							{
-								var nodeName = n.Text.ToLowerInvariant();
-								var nodeId = n.TabId.ToString();
-								return (nodeText2 == nodeName || nodeText2 == nodeId);
-							}));
+					var nodeText2 = nodeText;
```
tpluscode: This variable can be inlined.

DNN Platform/Modules/DDRMenu/MenuBase.cs
Mohtshum: Good catch that was surely missed out, thnx!

DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -197,22 +197,21 @@ private void FilterNodes(string nodeString, bool exclude)
...
-								var nodeName = n.Text.ToLowerInvariant();
-								var nodeId = n.TabId.ToString();
-								return (nodeText2 == nodeName || nodeText2 == nodeId);
-							}));
+                    filteredNodes.Add(RootNode.FindByNameOrId(nodeText));
 				}
 			}
 
             // if filtered for foksonomy tags, use flat tree to get all related pages in nodeselection
 		    if (flattenedNodes.HasChildren())
```
tpluscode: I know it's unrelated but please add braces here

DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -197,22 +197,21 @@ private void FilterNodes(string nodeString, bool exclude)
...
 		        RootNode = flattenedNodes;
-			RootNode.Children.RemoveAll(n => filteredNodes.Contains(n) == exclude);
+            if (exclude)
+            {
+                RootNode.RemoveAll(filteredNodes);
```
tpluscode: I think this condition is reversed. And it is suspicious that this call doesn't actually take the `exclude`  as parameter to pass down to it's children.

Mohtshum: Actually in older code deletion was not recursive due to which issue was comming for true case i.e for EXCLUDE=true (when EXCLUDE is set) and when false things are working fine for other case

PR #2035, opened 2018-04-09
DNN-7984 Removed useless code preventing users to upload files into a…

mean2me:
… subfolders of the User folder

### DIFF
--- a/DNN Platform/DotNetNuke.Web/InternalServices/FileUploadController.cs
+++ b/DNN Platform/DotNetNuke.Web/InternalServices/FileUploadController.cs
@@ -429,17 +429,7 @@ namespace DotNetNuke.Web.InternalServices
 
                 var effectivePortalId = isHostPortal ? Null.NullInteger : portalId;
 
-                // Check if this is a User Folder                
-                int userId;
                 var folderInfo = folderManager.GetFolder(effectivePortalId, folder);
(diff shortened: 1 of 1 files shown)


mitchelsellers:
Do you have recreation steps for how this was failing before?  Looking at Ticket DNN-7984 that appears to have been closed back in 2015 and was not related to this issue (https://dnntracker.atlassian.net/browse/DNN-7984)

tomepoh:
I am not sure this is the right area to post this but adding an existing module to a page will move all the modules already on the page to the location of the new module. Is there any fix for this in the next release? 
Thank you,
Tom
[Managed IT Services](https://www.bizconnectors.com)

mean2me:
Hello @mitchelsellers , we recently migrated to a new JIRA platform, so we have new IDs. That issue is still on the old JIRA and not related to the current fix.

tpluscode:
@mitchelsellers the issue being fixed is tracked as [DNN-10682](https://dnntracker.atlassian.net/browse/DNN-10682) in the old Jira

tpluscode:
I'm not sure this is something we are aware of @tomepoh. See if you can find a related issue on https://dnntracker.atlassian.net/. 

If not then do report an issue there keeping in mind that we'll need some more information: your running version and anything you think is specific to your problem.

PR #2036, opened 2018-04-11
DNN-19181: clear output cache on all web heads.

zyhfish:


DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -2339,6 +2339,7 @@ private static bool IgnoreRequestForInstall(string physicalPath, string refererP
         private static bool IgnoreRequestForWebServer(string requestedPath)
         {
             if (requestedPath.ToLowerInvariant().IndexOf("synchronizecache.aspx", StringComparison.Ordinal) > 1
+                || requestedPath.ToLowerInvariant().IndexOf("synchronizeoutputcache.aspx", StringComparison.Ordinal) > 1
```
tpluscode: Let's put the `synchronizeoutputcache.aspx` as a static readonly field of `MemoryCacheSynchonizationHandler`

tpluscode: Also `requestedPath.IndexOf("synchronizeoutputcache.aspx", StringComparison.OrdinalIgnoreCase)` will be cleaner

DNN Platform/Library/Services/OutputCache/Providers/MemoryCacheSynchonizationHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+// DEALINGS IN THE SOFTWARE.
+#endregion
+
+using System.Web;
+using DotNetNuke.Common.Utilities;
+using DotNetNuke.Entities.Host;
+
+namespace DotNetNuke.Services.OutputCache.Providers
+{
+    public class MemoryCacheSynchonizationHandler : IHttpHandler
```
tpluscode: Why are we actually using http handlers and not controllers? Arguably this doesn't even belong in DNN Library...

DNN Platform/Library/Services/OutputCache/Providers/MemoryProvider.cs
```diff
@@ -175,6 +187,99 @@ public override bool StreamOutput(int tabId, string cacheKey, HttpContext contex
...
+                _registeredWaitHandle?.Unregister(asynchronousResult.AsyncWaitHandle);
+            }
+        }
+
+        private static void TimeoutCallback(object state, bool timedOut)
+        {
+            if (timedOut)
+            {
+                var request = (HttpWebRequest)state;
+                request?.Abort();
```
tpluscode: You don't need the null coalescing operator here. If you fear that can will fail you should do a say cast.

DNN Platform/Library/Services/OutputCache/Providers/MemoryProvider.cs
```diff
@@ -175,6 +187,99 @@ public override bool StreamOutput(int tabId, string cacheKey, HttpContext contex
...
+                    //build url
+                    var url = "http://" + synchronizeUrl;
+                    if (HostController.Instance.GetBoolean("UseSSLForCacheSync", false))
+                    {
+                        url = "https://" + synchronizeUrl;
+                    }
+                    var serverUrl = string.Format(url, server.Url, dataParam);
+
+                    //Create HttpWebRequest
+                    var request = WebRequest.Create(serverUrl) as HttpWebRequest;
```
tpluscode: Conversely you safe cast here but don't do null checks laster on :)

DNN Platform/Library/Services/OutputCache/Providers/MemoryProvider.cs
```diff
@@ -175,6 +187,99 @@ public override bool StreamOutput(int tabId, string cacheKey, HttpContext contex
...
+                                                                            && s.IISAppName == thisServer.IISAppName)
+                                                                            && (s.ServerGroup == thisServer.ServerGroup || string.IsNullOrEmpty(thisServer.ServerGroup))
+                                                                            && !string.IsNullOrEmpty(s.Url));
+                //Iterate through the servers
+                foreach (var server in enabledServers)
+                {
+                    var dataParam = Host.DebugMode ? tabId.ToString() : UrlUtils.EncryptParameter(tabId.ToString(), Host.GUID);
+                    //build url
+                    var url = "http://" + synchronizeUrl;
+                    if (HostController.Instance.GetBoolean("UseSSLForCacheSync", false))
```
tpluscode: I cannot find this key, is this a completely new setting?
Looks like something that should be derived from an existing setting instead?

DNN Platform/Library/Services/OutputCache/Providers/MemoryProvider.cs
```diff
@@ -175,6 +187,99 @@ public override bool StreamOutput(int tabId, string cacheKey, HttpContext contex
...
+                return;
+            }
+
+            var servers = ServerController.GetServers();
+            if (servers.Count > 1)
+            {
+                var thisServer = servers.Single(s => s.ServerName == Globals.ServerName && s.IISAppName == Globals.IISAppName);
+
+                //Get the servers
+                var enabledServers = ServerController.GetEnabledServers().Where(s => !(s.ServerName == thisServer.ServerName
```
tpluscode: I know not everyone love LINQ but this is hell of lot more readable

```c#
var enabledServers = from s in ServerController.GetEnabledServers()
                     where !(s.ServerName == thisServer.ServerName && s.IISAppName == thisServer.IISAppName)
                     where s.ServerGroup == thisServer.ServerGroup || string.IsNullOrEmpty(thisServer.ServerGroup)
                     where !string.IsNullOrEmpty(s.Url)
                     select s;
```

mitchelsellers:
Question regarding this commit.  This changes the cache behavior between the other providers.  A provider that is server aware I think should be a distributed/new provider rather than an update to the existing provider.  Those that are working in a load balanced environment with a MemoryBasedOutputCacheProvider in use wouldn't know that this was going on.  This also goes against the behavior of the different caching providers

mitchelsellers:
Sorry for the second post on this, but I had a bit more time to review this commit and have come across a few additional concerns regarding this provider, the other providers and the situation this creates

*Existing Providers*

Currently, in DNN Platform, we do not have ANY caching providers that support web-farm enabled operations.  This means that the existing providers for ModuleCache and OutputCache are already set up for single-server operations only.

*Existing Web-Farm Providers*

Within the Evoq product line, a WebRequestCachingProvider does exist, this creates a provider that is aware of the complex server environments and keeps things in sync.  You must opt-in to that service by changing your caching provider to using this one.  Requests were made to bring this provider to the community.  If this is coming, then we might be able to bypass this concern.  If it isn't this solution creates a single provider that is web-farm capable, but the other providers are not, this would leave a system in a very large disarray.

*Parity to Provider Function*

Considering the above, the existing providers that are capable of synchronization are named in a manner that will ensure everyone is aware that they are working in the desired fashion.  Taking a MemoryCachingProvider (1 variant) and making it implement web-request backed synchronization calls while leaving another provider that is memory based that doesn't do this, results in a disparity in the transparency and consistency of the product.

*Proposed Solution*

In a perfect world, my recommendation for this would be to do the following.

1 Revert this change to the existing Memory provider
2 Create a new WebRequestOutputCachingProvider with this functionality
3 Backport, as part of this, change the existing WebRequestCachingProvider to provide similar functionality for the other cache store.

This solution would provide a consistent approach, with similar outlines and dependencies and could easily be understood by the community and users.

DNN Platform/Library/Services/OutputCache/Providers/MemoryProvider.cs
mitchelsellers: This is a setting that currently only exists at the Evoq level, as it is part of the WebRequestCachingProvider

nvisionative:
@mitchelsellers so glad you chimed in here.  We concur with your assessment and would love to see your recommendations followed.

PR #2037, opened 2018-04-11
Profile properties view-able/update-able by security roles

AlwaysVbNet:
Would like to have Profile Properties made available to security roles. For example a new custom profile property 'Company Size' should be made available to be viewed and updated only by users in the 'Company' security role.
Is this possible for this functionality to be added?


tpluscode:
Please be advised that a Pull Request (which I have to close because it tries to merge what we don't want merging) is not the right place to discuss feature requests.

Do feel free to reach out to us on the official issue tracker http://dnntracker.atlassian.net.

We will be also happy to accept a PR with the changes you need to `development` branch.

PR #2039, opened 2018-04-26
DNN-19842: use settimeout to avoid strict mode from jquery.

zyhfish:


Website/Resources/Search/SearchSkinObjectPreview.js
```diff
@@ -180,7 +182,9 @@
                                 href = $wrap.parent().next().attr('href');
                             }
                             var code = href.replace('javascript:', '');
-                            eval(code);
+                            setTimeout(function(){
+                                eval(code);
```
ricardodiasc: Same here.

Website/Resources/Search/SearchSkinObjectPreview.js
```diff
@@ -96,7 +96,9 @@
 	                    var $searchButton = $wrap.parents('.SearchContainer').length ? $wrap.parent().next() : $wrap.next();
                         var href = $searchButton.attr('href');
                         var code = href.replace('javascript:', '');
-                        eval(code);
+                        setTimeout(function(){
+                            eval(code);
```
ricardodiasc: Hi @zyhfish , this eval code is about extracting a javascript function to execute. We could replace:
-  var href = $searchButton.attr('href');
to
-  $searchButton.[0].click();

It would execute the same function without eval();

Website/Resources/Search/SearchSkinObjectPreview.js
```diff
@@ -180,7 +182,9 @@
                                 href = $wrap.parent().next().attr('href');
                             }
```
ricardodiasc: To find the $searchButton, I believe it just needs:
var $searchButton = $wrap.next();
if(!$searchButton)  $searchButton = $wrap.parent().next()

PR #2047, opened 2018-05-02
DNN-20032: render the provider/priority value into html content.

zyhfish:


DNN Platform/Library/UI/Modules/ModuleHost.cs
```diff
@@ -435,16 +435,20 @@ private void RestoreCachedClientResourceRegistrations(string cachedContent)
             foreach (Match match in matches)
             {
                 cachedContent = cachedContent.Replace(match.Value, string.Empty);
-                switch (match.Groups[1].Value.ToUpperInvariant())
+                var dependencyType = match.Groups[1].Value.ToUpperInvariant();
+                var filePath = match.Groups[2].Value;
+                var forceProvider = match.Groups[3].Value;
+                var priority = Convert.ToInt32(match.Groups[4].Value);
```
tpluscode: The priority is optional in the regular expression, right? The conversion will fail in such case

bdukes:
@tpluscode the comment is parsed when the HTML comes from the cache

DNN Platform/Library/UI/Modules/ModuleHost.cs
zyhfish: Hi @tpluscode , its not optional, will always render from the DnnJsInclude/DnnCssInclude control.

zyhfish:
Thanks @bdukes  for the explanation,

DNN Platform/Library/UI/Modules/ModuleHost.cs
tpluscode: You're right sorry, I was misled by the question mark in groups, like `(\d+?)`. Apparently this acts just like `(\d+)`

bdukes: But this _will_ error for JS libraries, won't it?  Or is the updated Regex not going to match JS libraries, and this will stop them from being registered via cached HTML? @tpluscode @zyhfish

zyhfish: @bdukes , thanks for the review, i missed that control, updated in PR #2048. thanks,

tpluscode: Let me ask another question. Can the `<!--CDF()-->` comment be also something coming from 3rd party controls?

zyhfish: I don't think 3rd party controls will output these contents, they should use dnn controls to register the scripts/styles.

tpluscode: Yes,let's improve that in #2048.

I think a single regex will be enough. You can wrap the new params in an optional, non-capturing group and check whether matches `[2]` and `[3]` were successful: https://regex101.com/r/hODQ7F/1

zyhfish: Hi @tpluscode , thanks for the suggestion, PR #2048 updated, please review, thx.

PR #2049, opened 2018-05-04
DNN-20057: uninstall old javascript libraries.

zyhfish:


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3205,9 +3205,12 @@ private static void UpdateFoldersForParentId()
         }
 
 
-        private static void UninstallPackage(string packageName, string packageType, bool deleteFiles = true)
+        private static void UninstallPackage(string packageName, string packageType, string version = "", bool deleteFiles = true)
```
mitchelsellers: Ideally, adding the new parameter to the end of the method signature would be helpful to avoid the more cluttered code later in the process, as it would no longer require the use of a "string.empty" call in the version for the one package that was deleted

tpluscode:
@bdukes, this comes from a customer who got jQuery 1 marked as potential security vulnerability. 

In general I'm in favour of having the latest and greatest but is there a risk of this breaking modules which rely on old jQuery?

bdukes:
@tpluscode since DNN requests jQuery, there probably aren't many scenarios where the 1.x version would get included anyway, so I'm not worried too much about it, I just wanted to understand the reasoning (and make sure y'all understood the ability to have multiple versions of a JS library installed).

tpluscode:
Ok, let's merge then. I suppose that a specific user/customer who'd need an older version has the option to install manually. Will that be possible?

bdukes:
Yes, exactly

IrishChieftain:
We have 8.x for our Intranet and 9.x for our public site. We have a scenario where jQuery 1.9.1 is causing security issues on 9.x. We also have custom modules on both versions and have come up against issues where Bootstrap enters the mix as it requires jQuery version 1.9.1 or higher, but lower than version 3 (9.x uses a jQuery lower than 3.3.7 which introduced a fix for this) ... bottom line is that different modules need to be able to select different versions of jQuery.

PR #2056, opened 2018-05-09
Fix path to resource file construction. Fixes #2055

### LINKED ISSUE #2055: Path construction in LocalizationProvider.SaveString is wrong
The current code uses the following:

```cs
string filePath = HostingEnvironment.MapPath("~/" + Globals.ApplicationPath + resourceFileName);
```

When not in a virtual directory Globals.ApplicationPath is usually "/". So it doesn't add anything. But in a virtual dir it is "/foobar" while HostingEnvironment.MapPath adds the foobar as well. And given the lack of trailing slash it is wrong regardless. This code should be:

```cs
string filePath = HostingEnvironment.MapPath("~/" + resourceFileName);
```

I'll submit a PR for this.

Impact: this code is used only by editing certain list entries (!) that are pulled from resource files (countries, regions).

I have personally created this code and this error. I am not sure why it has gone unnoticed for 4 years, but my guess is that no one uses this method often.

donker:
See issue 2055 on Github for details

### DIFF
--- a/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
+++ b/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
@@ -144,7 +144,7 @@ namespace DotNetNuke.Services.Localization
                         break;
                 }
                 resourceFileName = resourceFileName.TrimStart('~', '/', '\\');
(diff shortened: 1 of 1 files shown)


donker:
I've double checked this.

PR #2067, opened 2018-05-22
Move AppVeyor config to the repository

tpluscode:
I moved the AppVeyor configuration to the repo, exported it from the current CI setup.

Until now it was checked to ignore `appveyor.yml` and skip PR builds, which I re-enabled and we should now be seeing build status reports from AppVeyor on commits.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,29 @@
+version: 1.0.{build}
+pull_requests:
+  do_not_increment_build_number: true
+skip_tags: true
+max_jobs: 1
+configuration: Release
+platform: Any CPU
+nuget:
+  disable_publish_on_pr: true
+before_build:
+- cmd: .nuget\NuGet.exe restore DNN_Platform.sln
+build:
+  project: DNN_Platform.sln
+  parallel: true
+  verbosity: minimal
+test:
+  assemblies:
+    only:
+    - '**\Tests\DotNetNuke.Tests.Authentication\bin\$(configuration)\DotNetNuke.Tests.Authentication.dll'
+    - '**\Tests\DotNetNuke.Tests.Content\bin\$(configuration)\DotNetNuke.Tests.Content.dll'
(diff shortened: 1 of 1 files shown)


ohine:
@tpluscode Is this still needed for the AppVeyor builds?

bdukes:
I'd prefer to config in the repo, so long as this is still the latest config

ohine:
@bdukes me too, I just wanted to triple check before starting the review/merge.

tpluscode:
Yes, it is the same config.

ohine:
@tpluscode can you modify appveyor to save the artifacts generated by the build?

tpluscode:
@ohine which artefacts do you have in mind? At this point this is not a complete packaging build. Our Jenikins (and TC before) handles Persona Bar and other dependencies using the CI tools' respective features. 

We need to first migrate complete build scripts to upstream projects, figure out a way to share them with Dnn.Platform build and then have packaging scripts executed on AppVeyor

ohine:
I'm not completely familiar with AppVeyor, I only set it up once for CK Editor. Can we easily have it run the build install package bat script and then save the generated zip? Or would we need to rewrite the script completely? 

If it needs a total rewrite, we should just wait for the CI tasks to be completed. I don't think it makes sense to do anything time consuming with AppVeyor since we've moving to VSTS in the near future. I was looking to test one of the pull requests yesterday and thought it was configured to save the zip artifact post build.

tpluscode:
> Can we easily have it run the build install package bat script and then save the generated zip?

Arguably AppVeyor is quite a bit simpler from VSTS.

In terms of DNN, it should be easy. These build scripts, in the form of msbuild projects, exist in folder `./Build/BuildScripts/`. It's the same scripts which have been used for years. A rewrite may be due but they do work.

Again, the problem is getting modules which are built separately to bundle in the installation package. This is where I think this should be independent of the CI.

mitchelsellers:
@ohine and @ahoefling do we still need to process this PR?  Or should this be closed without merge due to the transition to VSTS?

PR #2068, opened 2018-05-22
Remove references to ClientDependency.Core csproj

tpluscode:


bdukes:
I think this change is correct, but I still can't build locally because the NuGet restore doesn't work, so I'm looking into that

PR #2069, opened 2018-05-22
Fix name of test project

tpluscode:
Previously, I abbreviated the project `DotNetNuke.Tests.AspNetClientCapabilitiesProvider` to `DotNetNuke.Tests.AspNetCPP` which shortens the path length.

Needless to say it should be `AspNetCCP` instead


mitchelsellers:
@dnnsoftware/tag could someone else review this item?  I believe this should be addressed

bdukes:
👍 merged

PR #2070, opened 2018-05-22
DNN-20316: Cannot create new user with email address as username

Mohtshum:
- Pull string to static field to be used in other project as well

### DIFF
--- a/DNN Platform/Library/Entities/Users/IUserController.cs
+++ b/DNN Platform/Library/Entities/Users/IUserController.cs
@@ -107,5 +107,12 @@ namespace DotNetNuke.Entities.Users
         /// in background threads or scheduler jobs. 
         /// </remarks>
         string GetUserProfilePictureUrl(int portalId, int userId, int width, int height);
+
+        /// <summary>
+        /// Check username against
+        /// </summary>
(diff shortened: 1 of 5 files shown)


mitchelsellers:
When looking at the space checking why not use string.IsNullOrWhiteSpace() to check since it captures more options of whitespace?

Mohtshum:
Please review changes

DNN Platform/Library/Entities/Users/IUserController.cs
```diff
@@ -107,5 +107,12 @@ public interface IUserController
         /// in background threads or scheduler jobs. 
         /// </remarks>
         string GetUserProfilePictureUrl(int portalId, int userId, int width, int height);
+
+        /// <summary>
+        /// Check username against
```
tpluscode: Nit: Please complete the comment ✍️

Mohtshum: Thnx, too fast too furious, how could I missed that 😮

Mohtshum:
Changes pushed

tpluscode:
Looking good otherwise @mitchelsellers ?

PR #2074, opened 2018-05-24
DNN-20580: import pages in order.

zyhfish:


DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -143,24 +143,27 @@ private void ProcessImportPages()
...
             CheckPoint.TotalItems = CheckPoint.TotalItems <= 0 ? exportedTabs.Count : CheckPoint.TotalItems;
             if (CheckPointStageCallback(this)) return;
             var progressStep = 100.0 / exportedTabs.OrderByDescending(x => x.Id).Count(x => x.Id < _totals.LastProcessedId);
 
+            var index = 0;
             foreach (var otherTab in exportedTabs)
             {
                 if (CheckCancelled(_exportImportJob)) break;
-                if (_totals.LastProcessedId > otherTab.Id) continue; // this is the exported DB row ID; not the TabID
+                if (_totals.LastProcessedId > index) continue; // this is the exported DB row ID; not the TabID
```
tpluscode: What is the initial value of `LastProcessedId`? It looks to me like there is no relation between that and `index`, which could cause unexpected results

PR #2075, opened 2018-05-24
RFA/DNN-20510 Removed session invalidation at SignIn

mean2me:


tpluscode:
LGTM. 

@mitchelsellers @bdukes @galatrash please have a look. This was part of a security fix introduced some time last year on top of keeping track of session cookies (#1945). The changes prevented a possible cookie replay attach.

We think this particular call is a bit much, because it causes module which rely on session being kept when user logs in (for example shopping cart). 

Arguably, such functionality can be implemented in other ways but on the flip side keeping session when user logs in doesn't seem to impact security.

DNN Platform/Library/Security/PortalSecurity.cs
```diff
@@ -616,8 +616,6 @@ public string Remove(string inputString, ConfigType configType, string configSou
 
         public void SignIn(UserInfo user, bool createPersistentCookie)
         {
-            InvalidateAspNetSession(HttpContext.Current);
-
```
galatrash: I don't remember the exact detail, but there was a reason for adding this line of code when I made that change. Before I added it, there were cases which allowed reusing the session after logging out and this line prevented that from happening. Therefore, without this line, there could be some security issues. More testing needed.

PR #2076, opened 2018-05-25
Don't hardcode NuGet source

bdukes:
With a clean clone, there are still issues running `./Build_Install_Package.bat`, the first of which is an error resolving the new ClientDependency NuGet package from #2068.  Currently, this PR fixes that NuGet issue by adjusting the `nuget restore` call done in the build process to use the `nuget.config` file instead of passing a `-Source` URL directly.

At this point, I'm now getting errors from the `DotNetNuke.Modules.DDRMenu` and `DotNetNuke.Modules.HtmlEditorManager` projects:
>C:\Code\Dnn.Platform\Platform.testtest\DNN Platform\Modules\DDRMenu\DotNetNuke.Modules.DDRMenu.csproj(205,3): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v15.0\WebApplications\Microsoft.WebApplication.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.


bdukes:
So, it looks like on my local machine, that file should be found from `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\Microsoft\VisualStudio\v15.0\WebApplications` instead of `C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v15.0\WebApplications\Microsoft.WebApplication.targets`, which ultimately stems from using `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe` instead of `C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe` to run this process.

Currently, when building packages, the script attempts to find msbuild in this order:

1. `%ProgramFiles(x86)%\MSBuild\15.0\Bin\MSBuild.exe`
2. `%ProgramFiles(x86)%\MSBuild\14.0\Bin\MSBuild.exe`
3. `%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe`
4. `%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\MSBuild.exe`
5. `%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe`

I propose that if there is an msbuild available in the environment, we should use that, otherwise it should look for the version bundled with Visual Studio before the standalone MSBuild.  I don't know that any scenario is foolproof, unless maybe we search specifically for `Microsoft.WebApplication.targets` under the path before we settle on one?

Thoughts @tpluscode @galatrash @ahoefling @mitchelsellers?

galatrash:
This is tricky abit. When I created this batch file I tried under different environment but was not exhaustive. I didn't want to look in the registry which requires elevated rights; I wanted it to be simple CLI job. If you have a better way to do it, please do and submit a PR.

ahoefling:
@bdukes :
>I propose that if there is an msbuild available in the environment, we should use that, otherwise it should look for the version bundled with Visual Studio before the standalone MSBuild. I don't know that any scenario is foolproof, unless maybe we search specifically for Microsoft.WebApplication.targets under the path before we settle on one?

I agree that we should pull msbuild out of the environment. We should pick a minimum version of visual studio and support that version to the latest. 

I have always found it weird that the script looks for a particular version of msbuild and tries to fall through to the next. I'm sure there was a great reason for it but this has the potential to cause a lot of build side-effects. 

* For example what do you do if someone installs Visual Studio into a non-standard path?

bdukes:
I pushed the change to the build scripts which uses the MSBuild available in the environment, and only looks in those specific paths if it isn't defined.  I think there may still be something off with the script, because it generates `/Build/BuildScritps/_Install.zip` instead of `/Artifacts/DNN_Platform_9.2.1.123-456_Install.zip`, but it's closer

bdukes:
This has been approved by two reviewers, is there anything holding up a merge?

PR #2078, opened 2018-05-27
DNN-18327: Fix the search for field definitions with spaces

### LINKED ISSUE #2096: Handling underscores and spaces in property names
## Description

We've identified some issues with regard to how profile properties are handled:

1. Property name cannot contain spaces because they are [replaced with underscores][adapter] and the profile page tries to find a property which doesn't exist
1. There cannot be two properties which differ only in spaces vs underscores (`"Hire date" ~ "Hire_date"`) because they will generate the same web control identifiers on profile page\

[adapter]: https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/UI/WebControls/PropertyEditor/Adapters/CollectionEditorInfoAdapter.cs#L276

## Steps to reproduce

- Log in
- Go to Persona Bar > Settings > Site Settings > Site Behavior > User Profiles > + Add Field
- Set Field Name* to : **HireDate**
- Set Data Type* to : **Date**
- Set Property Category* to : **Date**
- Set Read-Only to **true/enabled**
- Set Visible to **true/enabled**
- Set Default Visibility to : **All Users**
- Press Next
- If you have Content Localization enabled, for each Language you must set the name of the field again. I only have English so for Localized Content, I set the field name to : **HireDate**
- Next, repeat the steps above but instead of HireDate, add a new field called **Hire Date** and save.
- You should now have two new User Profile Fields: **HireDate** and **Hire Date**
- Next, go to Persona Bar > Manage > Users > Add User > create a user with the default settings
- On the Users list click the Ellipsis menu (... Menu) to the right of the user you created
- Click Login as User
- From the top-right corner of the screen, click on your user name > **Edit Profile**
- Expand the new user profile field category you created (in my case **Date** category)
- You will see that HireDate is read-only as and works as intended
- You will see that Hire Date is not read-only although we set it as such and can be edited

## Proposed solution

Short of changing how those names are handled within C# code and passed around we came up with another solution to limit how the profile can be used:

### Treat properties as equal but stripping underscores and replacing with spaces

| name given | name saved |
| ------------- | ------------- |
| "Hire date" | "Hire date" |
| "Hire_date" | "Hire date" |
| "   Hire date  " | "Hire date" |
| "&#95;Hire date&#95;" | "Hire date" |

This means that if there already is a property called `Hire date` then attempting to add `_Hire_date_` would replace it and not create a new one, as it now happens when names are 100% equal. 

**NOTE** I find this behaviour kind of surprising and it should warn the user about what is happening but it's a separate discussion.

ricardodiasc:
When adding new fields to user profile, if the new field has spaces, it will not identify the read-only properties correctly.

fixes #2096

### DIFF
--- a/DNN Platform/Library/UI/WebControls/PropertyEditor/ProfileEditorControl.cs
+++ b/DNN Platform/Library/UI/WebControls/PropertyEditor/ProfileEditorControl.cs
@@ -26,7 +26,6 @@ using System.Web.UI;
 using DotNetNuke.Common.Lists;
 using DotNetNuke.Entities.Portals;
 using DotNetNuke.Entities.Profile;
-using DotNetNuke.Entities.Users;
 using DotNetNuke.Security;
 
 #endregion
@@ -35,6 +34,7 @@ using DotNetNuke.Security;
 namespace DotNetNuke.UI.WebControls
 // ReSharper restore CheckNamespace
 {
+    
     /// -----------------------------------------------------------------------------
     /// Project:    DotNetNuke
     /// Namespace:  DotNetNuke.UI.WebControls
@@ -50,7 +50,9 @@ namespace DotNetNuke.UI.WebControls
     [ToolboxData("<{0}:ProfileEditorControl runat=server></{0}:ProfileEditorControl>")]
     public class ProfileEditorControl : CollectionEditorControl
     {
-		#region Protected Methods
+
+        
+        #region Protected Methods
 
         /// -----------------------------------------------------------------------------
         /// <summary>
@@ -73,8 +75,9 @@ namespace DotNetNuke.UI.WebControls
 
             foreach (FieldEditorControl editor in Fields)
             {
-                //Check whether Field is readonly
-                string fieldName = editor.Editor.Name;
+                //Check whether Field is readonly. Replace "_" with " " (Spaces) to find the correct definitions.
+                string fieldName = editor.Editor.Name.Replace("_", " ");
+
                 ProfilePropertyDefinitionCollection definitions = editor.DataSource as ProfilePropertyDefinitionCollection;
                 ProfilePropertyDefinition definition = definitions[fieldName];
 
--- a/DNN Platform/Library/UI/WebControls/PropertyEditor/ProfileEditorControl.cs
+++ b/DNN Platform/Library/UI/WebControls/PropertyEditor/ProfileEditorControl.cs
@@ -50,8 +50,6 @@ namespace DotNetNuke.UI.WebControls
     [ToolboxData("<{0}:ProfileEditorControl runat=server></{0}:ProfileEditorControl>")]
     public class ProfileEditorControl : CollectionEditorControl
     {
-
-        
         #region Protected Methods
 
         /// -----------------------------------------------------------------------------
--- a/Website/Providers/DataProviders/SqlDataProvider/09.02.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.02.01.SqlDataProvider
@@ -102,6 +102,111 @@ ELSE
 		AND    ModuleId = @ModuleId
 GO
 
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}AddPropertyDefinition]') AND type in (N'P', N'PC'))
+	DROP PROCEDURE {databaseOwner}[{objectQualifier}AddPropertyDefinition]
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}AddPropertyDefinition]
+	@PortalId int,
+	@ModuleDefId int,
+	@DataType int,
+	@DefaultValue nvarchar(max),
+	@PropertyCategory nvarchar(50),
+	@PropertyName nvarchar(50),
+	@ReadOnly bit,
+	@Required bit,
+	@ValidationExpression nvarchar(2000),
+	@ViewOrder int,
+	@Visible bit,
+    @Length int,
+    @DefaultVisibility int,
+	@CreatedByUserID int
+
+AS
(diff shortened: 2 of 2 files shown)


tpluscode:
Hi @ricardodiasc. The proposed solution is not good, because it will cause an analogous problem whenever a property is created with underscores.

Currently it's so that property `Hire Date` is converted to `Hire_Date` and thus cannot be found.
With your change it will be the opposite, with property like `Hire_Date`, the code will look for `Hire Date` and the same behaviour will occur.

----

Instead we should not conflate property names which are essential human-readable with their WebControl counterparts. After all the underscores are introduced for a reason, I guess 🤔.

See [this line](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/UI/WebControls/PropertyEditor/Adapters/CollectionEditorInfoAdapter.cs#L276) in `CollectionEditorInfoAdapter` where the replacement occurs.

I propose to refactor the [`EditorInfo`](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/UI/WebControls/PropertyEditor/EditorInfo.cs) type so that it holds both the control name and profile property name (new class property). The new property would be used in `ProfileEditorControl` for lookup.

```diff
-        public string Name { get; set; }
+        public string Name
+        {
+            get { return PropertyName.Replace(" ", "_"); }
+
+            set { PropertyName = value; }
+        }
+
+        public string PropertyName { get; set; }
```

ricardodiasc:
Hi @tpluscode, as we discussed, since this EditorInfo.Name is used in other parts for display a label before it gets replaced with underscores, we cannot replace the behavior get/set from this property. Adding a new property with underscores need some refactoring to make correct usage from other parts of the code.

Even if we do this, we still have this problem if the user type inserts property like "Hire Date" and "Hire_Date" and if we assign one as read-only, the two will be read-only since the spaces get replaced by underscores.
I altered the install/upgrade script from the database to consider "Hire Date" and "Hire_Date" as they are the same property so it will not conflict in code and for the user.,

DNN Platform/Library/UI/WebControls/PropertyEditor/ProfileEditorControl.cs
```diff
@@ -73,8 +73,9 @@ protected override void CreateEditor()
 
             foreach (FieldEditorControl editor in Fields)
             {
-                //Check whether Field is readonly
-                string fieldName = editor.Editor.Name;
+                //Check whether Field is readonly. Replace "_" with " " (Spaces) to find the correct definitions.
+                string fieldName = editor.Editor.Name.Replace("_", " ");
```
tpluscode: I think that combined with the modified SP this is still not good @ricardodiasc. Consider this

1. User creates the property `Hire_date`
1. `editor.Editor.Name` will be the same
1. the code above will replace underscores with space so that `fieldName  == "Hire date"`
   * profile property will not be found

If we don't want to change much C# code the the only way to be fool proof is to ensure that the property added to the database does not include any underscores in the first place.

The stored procedure should not only match by the normalised name but also stored that normalised name in the table. Also leading and trailing whitespace would probably have to be removed if it's not done already (after underscores are replaced).

Please see my description in #2096

Website/Providers/DataProviders/SqlDataProvider/09.02.01.SqlDataProvider
```diff
@@ -102,6 +102,115 @@ ELSE
...
+AS
+	DECLARE @PropertyDefinitionId int
+
+	SET @PropertyName = Replace(LTrim(RTrim(Replace(@PropertyName,'_',' '))), '_', ' ')
+
+
+	SELECT @PropertyDefinitionId = PropertyDefinitionId
+		FROM   {databaseOwner}{objectQualifier}ProfilePropertyDefinition
+		WHERE  (PortalId = @PortalId OR (PortalId IS NULL AND @PortalId IS NULL))
+			AND PropertyName = @PropertyName
```
tpluscode: I think this condition should also normalise the name stored in database. Otherwise it would not find a match when the existing property already contains some underscores

Website/Providers/DataProviders/SqlDataProvider/09.02.01.SqlDataProvider
```diff
@@ -102,6 +102,115 @@ ELSE
...
+	@ViewOrder int,
+	@Visible bit,
+    @Length int,
+    @DefaultVisibility int,
+	@CreatedByUserID int
+
+AS
+	DECLARE @PropertyDefinitionId int
+
+	SET @PropertyName = Replace(LTrim(RTrim(Replace(@PropertyName,'_',' '))), '_', ' ')
```
tpluscode: Is it necessary to to `REPLACE` twice?

sleupold:
IMO we need to prevent spaces (and other special cases) in any property name, For UI, they can be used in localized caption.

ricardodiasc:
@sleupold  , you mean to show some error when trying to insert any special character on a new property?

sleupold:
yes - or simply not accepting any special character or blank in the field.

Website/Providers/DataProviders/SqlDataProvider/09.02.01.SqlDataProvider
ricardodiasc: Unfortunately yes. Because of the test cases To work with both cases: "_Hire Date_" and "  Hire_Date  ".

ricardodiasc:
@sleupold, I did a comment on #2096 where I believe this could be treated. A new pull request is on: https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions/pull/579.

tpluscode:
@dnnsoftware/tag dnnsoftware/Dnn.AdminExperience.Extensions#579 got merged into development of PB Extensions.

Do we want this fix also on 9.2.1 release? If so, we should also merge those changes into the `release/1.6.0` branch.

PR #2079, opened 2018-05-28
DNN-19429: include the script in script block control.

zyhfish:


Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
```diff
@@ -39,39 +40,41 @@
...
-<script type="text/javascript">
-	/*globals jQuery, window, Sys */
-	(function ($, Sys) {
-		function setUpLogin() {
-			var actionLinks = $("a#dnn_ctr<%#ModuleId > Null.NullInteger ? ModuleId.ToString() : ""%>_Login_Login_DNN_cmdLogin");
-			actionLinks.click(function () {
-				if ($(this).hasClass("dnnDisabledAction")) {
-					return false;
-				}
+<dnn:DnnScriptBlock runat="server">
```
tpluscode: Could you explain to me this control

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
```diff
@@ -39,39 +40,41 @@
...
 		
-		$(document).ready(function () {
-		    $('.dnnLoginService').on('keydown', function (e) {
-				if ($(e.target).is('input:text,input:password') && e.keyCode === 13) {
-					var $loginButton = $('#dnn_ctr<%#ModuleId > Null.NullInteger ? ModuleId.ToString() : ""%>_Login_Login_DNN_cmdLogin');
-					if ($loginButton.hasClass("dnnDisabledAction")) {
-						return false;
-					}
+            $(document).ready(function () {
+                $(document).on('keydown', '.dnnLoginService', function (e) {
```
tpluscode: Why is it important to handle this event on document level?

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
zyhfish: Hi @tpluscode , no special to this control, just include the <%=...%> block in a server control so it becomes a child control of this one but not the login control, then modify the control collection of login control won't be failed.

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
zyhfish: because login control support partial render, when it get updated in ajax panel, the DOM will refreshed, if we handle the event in this control level, the event will lost after postback.

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
```diff
@@ -39,39 +40,41 @@
...
+                    if ($(e.target).is('input:text,input:password') && e.keyCode === 13) {
+                        var $loginButton = $('#dnn_ctr<%=ModuleId > Null.NullInteger ? ModuleId.ToString() : ""%>_Login_Login_DNN_cmdLogin');
```
tpluscode: And what is the significance of changing the data-binding syntax here?

zyhfish: if using <%#..., then must call DataBind from codebind to let the databinding effective, using <%=... doesn't need that. if calling DataBind manually, it may cause issue by life cycle changed.

PR #2084, opened 2018-05-31
Fix CDF settings tests

tpluscode:
This fixes two failing tests which were caused by #2018 (see [here](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Entities/Portals/PortalSettingsController.cs#L238-L250))

In short, the `CdfVersion` was previously only taken from portal settings and now it will be taken from host settings as fallback.


tpluscode:
FYI, the two other tests have been identified and we're workmen on them too

bdukes:
Rebased as 2d87f8c6f9442bfcdae0952e7d5233f1ba29c91c

PR #2088, opened 2018-06-01
Installation resource files reviewed

perezandres:
Added missing translations and some mistakes corrected.

### DIFF
--- a/Website/Install/App_LocalResources/Installwizard.aspx.es-ES.resx
+++ b/Website/Install/App_LocalResources/Installwizard.aspx.es-ES.resx
@@ -1,110 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
-    Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
-    associated with the data types.
-    
-    Example:
-    
-    ... ado.net/XML headers & schema ...
-    <resheader name="resmimetype">text/microsoft-resx</resheader>
-    <resheader name="version">2.0</resheader>
-    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
-    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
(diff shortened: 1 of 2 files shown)


perezandres:
I Mitch. I think that I edited both files with the language editor inside DNN, but maybe I used Peter's DNN Translator, I'm not really sure right now... Which is the best option? I don't mind if I must review them ;-)

EPTamminga:
Peter’s editor is a far better solution if you have tondo a serious amount of translations and you really want to manage the process.

ohine:
@perezandres @EPTamminga I'm not familiar with the structure of the resource files. I know the system works without any issues I've stumbled on if the schema is removed. My assumption is less repetitive junk in the xml files, and the less ram DNN might consume. I'm OK with this as long as others who are much more familiar with the ins/outs of dnn's localization / resource file system to give it the final OK before we merge it into the platform. 

Can @erikvb or @donker or both take a look and just confirm that the format of the translation is acceptable? A good way to compare is to check the ignore-whitespace differences
https://github.com/dnnsoftware/Dnn.Platform/pull/2088/files/949102e3fe4a8d138becdc0ca5a05826cf3a8a35?utf8=%E2%9C%93&diff=unified&w=1

perezandres:
No problem for me if I have to re-generate both files...

donker:
My DNN Translator removes the schema junk from the resource file (why MS insists on including it in every f*cking resx file is beyond me). A schema is not necessary unless some software wishes to "validate" the XML. But DNN doesn't do that. It knows the format and loads the XML as it finds it.

sleupold:
IMO all translations for the official languages (French, German, Italian, Dutch and Spanish) should be done or coordinated by the official translators, otherwise there is a risk of changes being applied forth and back during install and LP upgrade.

perezandres:
That's a good point, Sebastian. Who's the coordinator for Spanish?

sleupold:
@perezandres the Spanish translation is managed by @vmasanas (Vicenç Masanas)

perezandres:
I've worked previously with Vicenç, translating DNN resources since version 6. I'll contact him for next contributions ;-)

sleupold:
great, thanks a lot!

valadas:
I do not know who are officially maintaining those languages, but if nobody maintains french, i offer myself.

sleupold:
French is maintained by Benoît Sarton from BSI.fr

PR #2089, opened 2018-06-02
The default.aspx should support Async

jbrinkman:
Currently it is not possible to make async calls during a page lifecycle event like On_Load.  Adding Async="true" allows extension controls to utilize async await for handling asynchronous calls.

### DIFF
--- a/Website/Default.aspx
+++ b/Website/Default.aspx
@@ -1,4 +1,4 @@
-<%@ Page Language="C#" AutoEventWireup="True" Inherits="DotNetNuke.Framework.DefaultPage" CodeBehind="Default.aspx.cs" %>
+<%@ Page Language="C#" AutoEventWireup="True" Inherits="DotNetNuke.Framework.DefaultPage" CodeBehind="Default.aspx.cs" Async="true" %>
 <%@ Register TagPrefix="dnncrm" Namespace="DotNetNuke.Web.Client.ClientResourceManagement" Assembly="DotNetNuke.Web.Client" %>
 <%@ Register TagPrefix="dnn" Namespace="DotNetNuke.Common.Controls" Assembly="DotNetNuke" %>
 <asp:literal id="skinDocType" runat="server" ViewStateMode="Disabled"/>


mitchelsellers:
What testing has been done on the impacts of this?  @bdukes I think it was you that discussed this option  a while back.  Thoughts on the risk associated with this pull request?

ohine:
I haven't tested this yet, but we will need a github issue created as well with details on the async change.

bdukes:
I had done some testing at some point, and came away with the impression that there are cases where components loaded by the async page would need to be aware that the page is `async`, but I don't remember the specifics.  It may be that I had tested with some bad code (like warned against in https://www.hanselman.com/blog/TheMagicOfUsingAsynchronousMethodsInASPNET45PlusAnImportantGotcha.aspx)

Or perhaps issues like mentioned in https://docs.microsoft.com/en-us/aspnet/web-forms/overview/performance-and-caching/using-asynchronous-methods-in-aspnet-45#registerasynctask-notes:
>Methods hooked up with `RegisterAsyncTask` will run immediately after PreRender. You can also use async void page events directly… The downside to async void events is that developers no longer has full control over when events execute. For example, if both an .aspx and a .Master define `Page_Load` events and one or both of them are asynchronous, the order of execution can't be guaranteed. The same indeterminiate order for non event handlers (such as `async void Button_Click`) applies.

Also possible that I was testing before DNN had upgraded to .NET 4.5 and the `UseTaskFriendlySynchronizationContext` setting wasn't set.

All told, we need to do some testing and verification, but it would be wonderful if this didn't cause big problems.

ahoefling:
While this may not be very helpful to the discussion but I remember having a conversation with someone in the DNN TAG group about async in DNN as it was not supported so we could not implement it in the DNN MVC Platform. 

- Does adding this change allow us to implement async/await in the MVC or Web API Platform?
- Will this change break any of the existing module platforms?

mitchelsellers:
@jbrinkman @bdukes or @ahoefling do any of you have time to dedicate to test this in more detail?  I would like to include it, but am concerned for the gotchas that could come.

bdukes:
IMO we should flip this switch on the 10.x branch, and see how things go

ohine:
We are bypassing the cla-bot requirement since @jbrinkman most definitely has a CLA on file and proceeding to merge this into development targeting the v9.4.0 release. Thanks for the contribution @jbrinkman!

bdukes:
This PR was rolled back by #2810, because it introduced issues on multi-lingual sites.  DNN currently relies on `CultureInfo.CurrentCulture` which comes from the current thread, and turning on `Async` introduces multiple threads to handle a single request.  Those addition threads do _not_ inherit the culture of the thread that creates them.  So, for instance, on a multi-lingual site, all URLs in the menu will point to the en-us site, regardless of the selected language of the user.

PR #2097, opened 2018-06-07
Issue and PR templates

tpluscode:
@dnnsoftware/tag I added stubs of PR and issue templates which will be displayed whenever either one is created.


bdukes:
LGTM

tpluscode:
You're right @valadas, we'll have to take care of it too.

That said, PRs are always appreciated :)

PR #2098, opened 2018-06-07
DNN-20914: save module permissions into database.

### LINKED ISSUE #2099: Export/Import not importing module permissions for custom roles
## Steps to reproduce

On Source: 
1. On Source Create a couple of custom roles. 
2. Add permissions to a couple of modules for those roles. Create new modules if possible. 
3. Add permissions to a page for those roles. Create a new page if possible. 
4. Create a Full export of the site. You can choose to exclude extensions or assets since these are not needed. 

On Target: 
3. Review permission settings for modules and pages. 

---------------------------------------------------------------------------------------------------- 
Note: I have reviewed the contents of the export in the dnndb file and the permissions for the custom roles are present there as well as the default (Admin,CM,CE) permissions. 

## Expected behavior

It should import both the page and module permissions.

## Current behavior

The custom role permissions exist on target site for pages; however, no permissions are added for modules and it only inherits the default ones from a page.

If the import is run a second time in the source site, the following error will be thrown: 

```
Dnn.ExportImport.Components.Engines.ExportImportEngine - System.NullReferenceException: Object reference not set to an instance of an object. 
   at Dnn.ExportImport.Components.Services.PagesExportService.<>c__DisplayClass38_1.<ImportModulePermissions>b__0(ModulePermissionInfo x) 
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate) 
   at Dnn.ExportImport.Components.Services.PagesExportService.ImportModulePermissions(ModuleInfo localModule, ExportModule otherModule, Boolean isNew) 
   at Dnn.ExportImport.Components.Services.PagesExportService.ImportTabModulesAndRelatedItems(TabInfo localTab, ExportTab otherTab, Boolean isNew) 
```

It appears to be an issue in this line of code: 
https://github.com/dnnsoftware/Dnn.Platform/blob/a23a02a99e7f2e5becb952e8f9d180a4a5f5c9be/DNN%20Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs#L958 

since this second time there is a local reference of that same module; however, the modulepermission entries is still null as no permissions are added in the db for this module.

zyhfish:
fixes #2099

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -475,6 +475,7 @@ namespace Dnn.ExportImport.Components.Services
                             RoleName = other.RoleName,
                             ModuleDefID = Util.GeModuleDefIdByFriendltName(other.FriendlyName) ?? -1,
                             PermissionKey = other.PermissionKey,
+                            PermissionName = other.PermissionName,
                             AllowAccess = other.AllowAccess,
                             PermissionID = permissionId.Value
                         };
@@ -999,6 +1000,7 @@ namespace Dnn.ExportImport.Components.Services
                             RoleName = other.RoleName,
                             Username = other.Username,
                             PermissionKey = other.PermissionKey,
+                            PermissionName = other.PermissionName,
                             AllowAccess = other.AllowAccess,
                             PermissionID = permissionId.Value
                         };
@@ -1015,15 +1017,17 @@ namespace Dnn.ExportImport.Components.Services
                             local.RoleID = roleId.Value;
                         }
 
-                        other.LocalId = localModule.ModulePermissions.Add(local);
-                        var createdBy = Util.GetUserIdByName(_exportImportJob, other.CreatedByUserID, other.CreatedByUserName);
-                        var modifiedBy = Util.GetUserIdByName(_exportImportJob, other.LastModifiedByUserID, other.LastModifiedByUserName);
-                        UpdateModulePermissionChangers(local.ModulePermissionID, createdBy, modifiedBy);
+                        other.LocalId = localModule.ModulePermissions.Add(local, true);
 
                         Result.AddLogEntry("Added module permission", $"{other.PermissionKey} - {other.PermissionID}");
                         count++;
                     }
                 }
+            }
+
+            if (count > 0)
+            {
+                ModulePermissionController.SaveModulePermissions(localModule);
             }
 
             return count;


DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -1015,17 +1017,19 @@ private int ImportModulePermissions(ModuleInfo localModule, ExportModule otherMo
                             local.RoleID = roleId.Value;
                         }
 
-                        other.LocalId = localModule.ModulePermissions.Add(local);
-                        var createdBy = Util.GetUserIdByName(_exportImportJob, other.CreatedByUserID, other.CreatedByUserName);
-                        var modifiedBy = Util.GetUserIdByName(_exportImportJob, other.LastModifiedByUserID, other.LastModifiedByUserName);
-                        UpdateModulePermissionChangers(local.ModulePermissionID, createdBy, modifiedBy);
```
tpluscode: Why remove these lines?

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -1015,17 +1017,19 @@ private int ImportModulePermissions(ModuleInfo localModule, ExportModule otherMo
...
+                        other.LocalId = localModule.ModulePermissions.Add(local, true);
```
tpluscode: Why is it important to add the second parameter?

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
zyhfish: Because the ModulerPermissionId always be -1 until it saved into database, so actually these code do nothing. similar code commented in SaveTabPermissions method as well.

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
zyhfish: when add the permission into collection, second parameter will make sure the permission is unique, and won't cause index duplicated error when save data into db.

sleupold: to prevent duplicate entries in the database, use a unique constraint (or index) and proper merge statement in add/update stored procedure

affizafazreen:
Code review SLA exceeded for 15d. @tpluscode  we need to move this issue to merge / or reject this

affizafazreen:
reminder  @tpluscode this PR has breached 24h Code Review SLA for Customer Defect DNN-20914 ( Severity-High) .. proceed with central process cc @ashishpd

PR #2100, opened 2018-06-08
*DO NOT MERGE* - introducing ConcurrentDictionary in SharedDictionary class

mean2me:
**Description:**
To prevent performance problem on large sites.

## Summary
Please, let me know if you see anything wrong in the proposed change. I just took care of introducing ConcurrentDictionary in place the non-thread-safe Dictionary previously used. Actually SharedDictionary is wrapping a ConcurrentDictionary now.

I also created a fake Lock controller to avoid read locks, since ConcurrentDictionary is supposed to be safe for read access by concurrent threads.

This is just a test branch, not expected to be merged.

### DIFF
--- a/DNN Platform/Library/Collections/FakeDisposable.cs
+++ b/DNN Platform/Library/Collections/FakeDisposable.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotNetNuke.Collections.Internal
+{
+    class FakeDisposable : ISharedCollectionLock
+    {
+
+        private bool disposedValue = false;
+
+        protected virtual void Dispose(bool disposing)
+        {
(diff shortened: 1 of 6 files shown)


mitchelsellers:
What is the purpose of the FakeLockController.  If this is purposful, and not designed as a Fake for testing, I believe at a minimum it should NOT be named Fake*.  

For example if the purpose is to prevent locks, naming it DisabledLockController, or NonLockingLockController would be more appropriate.  As the term "Fake" typically only relates to items added for unit testing

tpluscode:
> If this is purposful, and not designed as a Fake for testing, I believe at a minimum it should NOT be named Fake*.

I would typically call such class a `NullLockController`, as reference to the [null object pattern](https://sourcemaking.com/design_patterns/null_object).

mitchelsellers:
Since this is a "do not merge" pull request I am going to close this.  If it is still needed please let me know.

PR #2112, opened 2018-06-11
Added browser section to issue template

tpluscode:
## Summary

I figure that it's enough to have one section for browser / OS.


tpluscode:
cc @dnnsoftware/tag

PR #2113, opened 2018-06-12
#2053 - Update Contributing Document

### LINKED ISSUE #2053: Instructions to correctly build dnn from repo and submit pull requests need updating
We need clear instructions on how to build the project now that it is split into multiple repositories.

There is already very clear instructions on how to do it from the source zip package but how should the folder structure be made locally so that when building persona bar extensions in release mode, the dll gets in the proper folder to test on the development website. Same for the others.

I don't know if this is clear, but we need to discuss this, I can write up the a good blog post or .md file here, but I need some quick answers on the stuff I am more confused about.

ohine:
## Summary

This updates our contributing guide to point people to the current issue tracker (it was still linked to jira). I also added commands required to get git setup for the first time and to help other users rebase their forks to the current code base. 

If @dnnsoftware/tag has any other suggestions, lets hear them so we can help everyone participate together!

Fixes #2053


CONTRIBUTING.md
```diff
@@ -60,12 +61,6 @@ appropriate to start the first line of a commit with '(doc)' instead of
 a ticket number. 
 
 ````
-    (doc) Add documentation commit example to CONTRIBUTING
```
bdukes: Why were the first lines of the example commit messages removed?  I think these either need to be complete examples (as they previously were) or they need to be removed altogether and replaced with a link to documentation on writing a good commit message.

valadas:
I like it, but it would also be nice to show how to build from sources, if people only get this repo, they are not getting a working Dnn, so how should this be done, is there a way to get this + persona bar + pb module, etc. into a single location, build from github sources and then have all that in visual studio to edit, build test.

Or do people just get the zipped sources from Dnn Nightly buil and then move over the change into the VS Project cloned from here?

bdukes:
Agreed, @valadas.  We're working on getting the build and packaging more streamlined, but until then, there should probably be a note to start with a package from [the nightly feed](http://www.dnnsoftware.com/platform/build/nightly-builds) (is this still being updated, with the changes being made to the CI?).  

Once we have the packaging of the various components coming together, we can hopefully update to a much simpler process.

CONTRIBUTING.md
ohine: It seemed like boilerplate text which didn't make sense to me. If it does make sense to keep, I'll add it back. L31-L38 and L63-L68 are the blocks I'm referencing.

bdukes: It is a little confusing, but it's supposed to be an example commit message (for a commit about adding the commit message example, which is probably why it gets confusing).  Maybe using a more concrete example would be better.

ohine: How is the latest update?

bdukes: perfect

Issue #2116, opened 2018-06-12
New feature: Package dependency on database version


bdukes:
A package should be able to indicate that it requires a particular version of SQL Server.  Currently, packages can specify dependencies in their manifest. This suggestion is to add a new type of dependency.

The most straightforward would be to add `<dependency type="SqlServer">2016</dependency>`.  Because DNN's data provider is abstract with multiple possible implementations, it may make sense to be less SQL Server specific.  Perhaps a better approach would look more like `<dependency type="database" version="2016">SqlDataProvider</dependency>`.

Open questions:
 * What should the name of the dependency be?
 * How should the version be specified (e.g. 2017 or 14.0.1000.169)?
 * Does this need to be an additional feature that any data provider could implement (e.g. would an `OracleDataProvider` be able to provide an implementation which determined the database version)?
 * Do we want to handle this in a specific way during the installation and upgrade wizards (e.g. allow skipping packages whose dependencies aren't met, show warnings, etc.)?

sleupold:
we should not request a specific version of SQL server, but a compatibility level for the database, which determines the language features available, see http://www.sqlmvp.org/sql-server-database-compatibility-level-performance/
Note: the compatibility level may be modified by SQL, Turbo Scripts always sets it to max value supported by the server, but this requires DBOwner permission.

PR #2118, opened 2018-06-12
Fix breaking dll rename of SharpZipLib - issue #2095

### LINKED ISSUE #2095: Dependencies to old SharpZipLib don't get resolved properly
Because of the update in 9.2, any module using SharpZipLib before will be incompatible, and updating makes it incompatible with previous DNNs. 

This should be fixed. There is a way to intercept DLL resolves and I have the code for it ready in 2sxc, which I'll contribute in a few days.

iJungleboy:
## Summary
Fixes #2095 by registering an assembly-resolver that is used by .net when an assembly isn't found. So if something tries to get SharpZipLib.dll and fails, this resolver will kick in and tell .net that it's in the new assembly `ICSharpCode.SharpZipLib`. 

I also added a readme.md explaining it some more, and the "register assembly handler" must happen very early, which is why I needed to put it into the application start.


DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,54 @@
...
+            AlreadyRun = true;
+
+            try
+            {
+                // Adds an AssemblyResolve handler to redirect all attempts to load a specific assembly name to the specified version.
+                AppDomain.CurrentDomain.AssemblyResolve += Handler;
+            }
+            catch
+            {
+                /* ignore */
```
tpluscode: In what circumstances is an exception here possible? Swallowing it will mean that the fix will not fire, right?

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        private static Assembly Handler(object sender, ResolveEventArgs args)
+        {
+            // only check for access to old SharpZipLib
+            if (!args.Name.StartsWith(OldName))
+                return null;
```
tpluscode: Please add braces

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,54 @@
...
+        private const string OldName = "SharpZipLib";
+        private const string NewName = "ICSharpCode.SharpZipLib";
+
+        /// <summary>
+        /// Registration call - should only be called once
+        /// Has extra security to prevent it from running multiple times
+        /// </summary>
+        public static void RegisterSharpZipLibRedirect()
+        {
+            if (AlreadyRun) return;
```
tpluscode: Please add braces

bdukes:
@tpluscode it doesn't look like #2019 was ever merged

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,54 @@
...
+            catch
+            {
+                /* ignore */
+            }
+        }
+
+        private static Assembly Handler(object sender, ResolveEventArgs args)
+        {
+            // only check for access to old SharpZipLib
+            if (!args.Name.StartsWith(OldName))
```
bdukes: please add `StringComparison.Ordinal` as second parameter to `StartsWith`

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// This class is in charge of fixing an assembly rename which was necessary when DNN
+    /// upgraded from an older SharpZipLib. 
+    /// </summary>
+    /// <remarks>
+    /// final solution taken from https://raw.githubusercontent.com/2sic/2sxc/master/2sxc%20Dnn/Dnn920/SharpZipLibRedirect.cs
+    /// read the readme.md in this folder for more info
+    /// </remarks>
+    public class SharpZipLibRedirect
```
bdukes: I'm not real excited about this being `public`, seems like it will confuse folks that there's a `DotNetNuke.Services.Zip` namespace that doesn't have anything useful for them inside of it.  Should this just live next to (or inside of) `DotNetNukeHttpApplication`?

DNN Platform/Library/Services/Zip/readme.md
```diff
@@ -0,0 +1,11 @@
+﻿# Zip Services in DNN
+
+DNN uses SharpZipLib as the zip system of choice. Because of this, it is bundled with DNN.
```
bdukes: I would prefer that this be included as comments in the `SharpZipLibRedirect` class, rather than as a markdown document.

tpluscode:
Oh, true. Good :)

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
bdukes: Is there a reason this is using `StartsWith` instead of `Equals`?

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
mitchelsellers: I agree, internal would be better if we can

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: Yes, this is to ensure that if anything at all goes wrong when registering this, it won't break DNN.

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: is that the DNN convention? or your personal preference? just to be sure...

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: DNN convention?

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: ok, will do. I think I originally used the name without DLL, so equals should be ok. i'll change it.

DNN Platform/Library/Services/Zip/readme.md
iJungleboy: I prefer markdown - because often explanations will need multiple code files, but I'll put the same info into the class file as well - ok?

iJungleboy:
Thanks for the thorough review! I'm updating everything as discussed, after testing I'll republish...

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: ok great suggestion!

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: I found out that it needs starts with, because the real string looks something like this `SharpZipLib, Version=0.81.0.1407, Culture=neutral, PublicKeyToken=null`

iJungleboy:
Changes made and published, pls re-check, thx!

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
tpluscode: Not a DNN convention, given that no official conventions exist. It is however a sensible choice always to add curly brackets and also a [StyleCop inspection](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1503.md), enabled by default I think

DNN Platform/Library/Services/Zip/readme.md
mitchelsellers: For consistency with the rest of the project I would ONLY have this in the code comments.  Multi-line XML comments can support HTML formatting and will be automatically built into the documentation.  THis is used universally across the entire solution

ohine:
I'm guessing we need to update this function as well with both assembly names? or change the `StartsWith` to `Contains`? Right @bdukes?

https://github.com/dnnsoftware/Dnn.Platform/blob/6387dd1a7e55cdd994d567cdb76733fbd63819dc/DNN%20Platform/Library/Framework/Reflections/TypeLocator.cs#L109

DNN Platform/Library/Services/Zip/readme.md
iJungleboy: @mitchelsellers ok, I just moved all the docs into code-comments only

iJungleboy:
@tpluscode @bdukes @ohine I believe I did everything you guys asked for. Ok for now, or anything still pending?

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// redirect it to the correct (new) assembly.
+    ///
+    /// Once this remapping has been completed, this code will not be used again until the next
+    /// restart of the DNN application. 
+    /// 
+    /// final solution taken from https://raw.githubusercontent.com/2sic/2sxc/master/2sxc%20Dnn/Dnn920/SharpZipLibRedirect.cs
+    /// </remarks>
+    internal class SharpZipLibRedirect
+    {
+        internal static bool AlreadyRun { get; private set; }
```
tpluscode: Hm, I notice this changed from `public` to `internal`. But does it really have to be accessible from anywhere? I'd expect such property to be `private`

iJungleboy:
@bdukes and @ohine I don't understand that part about the other code. Should I do something there? I've never used/seen that code before.

DNN Platform/Library/Services/Zip/SharpZipLibRedirect.cs
iJungleboy: I would leave it in for debugging - in case this is ever questioned. But I can change it if that is preferred.

tpluscode: No strong opinion here. Let's leave it as is

ohine:
@iJungleboy I've never seen that code either, I was researching the SQL dependency issue (#2116) and stumbled across that function in the process. We should add both of the assembly names to that function. If you can do it in this pull request that would be great, if not I'll start up another pull request with that change.

iJungleboy:
@ohine I would rather not touch that - I don't know anything about it and it's not related to the sharpziplib issue directly - ok?

I believe we've covered everything - shall we finish the pull @bdukes  and @tpluscode ?

mitchelsellers:
I think that we need determine WHY those DLL's are excluded from the scanning process before we fully implement this.  For a DLL to be explicitly listed here my guess is that there was an issue.

@ohine or @bdukes any idea on how we might track this down?

ohine:
Maybe @ashishpd might know some history on that function/class and it's purpose in the platform?

https://github.com/dnnsoftware/Dnn.Platform/blob/6387dd1a7e55cdd994d567cdb76733fbd63819dc/DNN%20Platform/Library/Framework/Reflections/TypeLocator.cs#L69-L124

ohine:
Created PR #2126 with the fix to the `CanScan` function.

PR #2126, opened 2018-06-21
add new sharpziplib assembly name to ignore list

ohine:
## Summary
Added the updated `SharpZipLib` assembly name to the ignore list of assemblies to scan for Dnn features and extension points. 

## Details
In v9.2.0 the assembly was updated but the `CanScan` function wasn't. This pull request updates Dnn to be aware of both assemblies since in v9.2.1 we are attempting to solve the breaking change for any module developers who were directly referencing `SharpLibZip`


ohine:
#2118

PR #2137, opened 2018-06-26
build script changes for vsts

ohine:
#### Description
This change adds a cake build task which calls the existing msbuild task for packaging the platform. The new features are only run from the cake task, we also download the branch from PB, CDF, CK repositories, compile, and package those extensions as well. 

After we get the new cake task fully verified and working to generate usable install/upgrade packages. We will then clean up the old tasks and remove any old stale code. 

#### Related Pull Requests:
- https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions/pull/582
- https://github.com/dnnsoftware/Dnn.AdminExperience.Library/pull/155

#### Testing:
Because this change spans multiple repositories, testing is a bit tricky. The best way would be to pull down platform and apply this pull request locally. Then you'll want to edit the cake file and point it to my repos. 

https://github.com/nbrane/Dnn.Platform/blob/devops/build.cake#L79-L81

Change dnnsoftware to nbrane. Save the file. Then open up a powershell terminal and run. 
- npm config set registry https://www.myget.org/F/dnn-software-public/npm/
- .\build.ps1 -target Build

The other option would be to pull down everything locally, and then publish it to github and adjust the urls to your branches. If you have any questions or problems, let me know and I'll be happy to help figure things out.

This will then go fetch the latest of everything, compile it all in release mode, and you should have a usable Install and Upgrade zip under the Artifacts folder.


ohine:
Thanks @bdukes!

PR #2143, opened 2018-07-02
upgrade .net framework to v4.6.2 

ohine:
## Summary
This pull request updates the framework for all projects to v4.6.2.


bdukes:
This is a breaking change, right?  It will need to wait until version 10 (and we're talking about requiring 4.7.2 for DNN 10).

ohine:
@bdukes I don't think it's a "breaking" change, it was a clean upgrade and we didn't have to change any core code. It's more of a change in the base requirements, I think if we happen to do a 9.3 maybe it could be included in that release to make the update less of a jump for v10. 

I'll have one for v4.7.2 shortly as well. This is part of a research project I'm working on, and figured might as well submit the pull request for each version and then we can further discuss when to upgrade to what and merge or close any of the pull requests.

mitchelsellers:
At this point in time I would not address this upgrade.  I am ok I believe with setting a requirements change in version 10.0.0 that we will require 4.7.2 at that time.

ohine:
No worries if this doesn't make it into the v9.x branch. I completely understand. Feel free to close it, or leave it open just incase things change as we fill out the roadmap.

mitchelsellers:
Based on prior discussion we are going to look at using the 4.7.2 one for DNN 10.0.  Closing this request

PR #2151, opened 2018-07-12
DNN-21637: add config key.

zyhfish:


bdukes:
@ohine agreed, but you need to have a module to run `IUpgradeable` code, so doing this for a library or provider isn't really something the platform supports.  We should probably figure out a good way to allow that (it's a scenario we've run into a couple of times, as well)

PR #2153, opened 2018-07-12
Fixes #DNN-8338 SI: Full Exports Do Not Update Existing HTML Pro Modules

### LINKED ISSUE #2176: Full Exports Do Not Update Existing HTML Pro Modules
## Description 

When a full export is imported into a target which, pre-existing HTML Pro module content is not updated. New HTML Pro modules are added along with their content as are other content such as new files, user and tab permission changes etc etc.

The export/import feature has 2 different modes Full and Differential. 

The Full mode would export all of the data selected; while the differential one would only export the data that has changed since the last Full/Differential Export.

## Steps to reproduce

1. Install 2 Evoq 9.2 sites (Source and Target) 
2. Create a new page with an html module with content on source portal. 
3. Do a full export of that Portal. (Note that exporting the new Page and enabling content is enough and it is not necessary to export everything) 
4. Copy the Files into target site App_data/ExportImport 
5. Run the Import into the Target Portal. 
6. Verify that the new page with all the content is added to the target site. 
7. Go to source target and modify the content of the HTML module 
8. Create another Full Export of the source Portal (Page and Content) 
9. Copy the Files into target site App_data/ExportImport 
10. Run the Import in the target Portal. 

Result: The content has not changed. 

11. Go to source target and modify the content of the HTML module once again. 
12. Create another Differential Export of the source Portal (Page and Content) 
13. Copy the Files into target site App_data/ExportImport 
14. Run the Import in the target Portal. 

Result: The content has been updated.

## Current result

The Html pro module content is not updated using the full mode.

## Expected result

It should work with both differential and full.

## Affected version

* [x] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

mean2me:
Fixes #2176

## Summary
Removed useless code generating a null reference exception while performing a full import.
Actually no functional changes will be introduced. Just avoid a Null Reference exception generated by a piece of that did nothing.

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -959,69 +959,40 @@ namespace Dnn.ExportImport.Components.Services
             {
                 var userId = UserController.GetUserByName(_importDto.PortalId, other.Username)?.UserID;
                 var roleId = Util.GetRoleIdByName(_importDto.PortalId, other.RoleID ?? noRole, other.RoleName);
-                var local = localModulePermissions.FirstOrDefault(
-                    x => x.PermissionCode == other.PermissionCode && x.PermissionKey == other.PermissionKey
-                    && x.PermissionName.Equals(other.PermissionName, StringComparison.InvariantCultureIgnoreCase) &&
-                    x.RoleID == roleId && x.UserID == userId);
+                var permissionId = DataProvider.Instance().GetPermissionId(other.PermissionCode, other.PermissionKey, other.PermissionName);
 
-                var isUpdate = false;
-                if (local != null)
+                if (permissionId != null)
                 {
-                    switch (_importDto.CollisionResolution)
+
+                    var local = new ModulePermissionInfo
                     {
-                        case CollisionResolution.Overwrite:
-                            isUpdate = true;
-                            break;
-                        case CollisionResolution.Ignore:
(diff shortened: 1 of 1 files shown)


tpluscode:
@mean2me please complete the CLA

mean2me:
@tpluscode  Yes, there was some useless code, probably an incomplete implementation. I removed that code, since it didn't affect the returned value but, actually, was responsible of a NullReference Exception due to which the import failed.

mitchelsellers:
@mean2me Can you please sign the CLA?

ohine:
Here is the successful results from a manually queued build of the DNN[CI] vsts task. 
https://dotnetfoundation.visualstudio.com/DNN/_build/results?buildId=3842&_a=summary

tpluscode:
> Yes, there was some useless code, probably an incomplete implementation

Is there anyone how knows that part of Platform to confirm we really were free to just get rid of that code? @zyhfish @dnnsoftware/tag

mean2me:
@tpluscode that piece of code actually was incomplete and did nothing but raising a null reference exception :)

tpluscode:
@mitchelsellers what kind of information are you asking about? What do you think is missing from #2176 ?

mitchelsellers:
@tpluscode Does this look good to you?  I don't see any issues, but I don't have an easy way to validate the impact of an Evoq solution

PR #2156, opened 2018-07-12
Mention security issue in issue template

bdukes:


valadas:
Just got an idea, Github supports multiple issue templates, you can try that on most https://github.com/dnncommunity repositories. There is one for a bug, another template for an improvement, feature request, etc. I suggest adding a template for a security related issue, that template could have in uppercase a clear message to not submit this issue on github and give the security email address instead.

Issue #2158, opened 2018-07-13
Managing version numbers


tpluscode:
## Description 

Currently every time a new release is prepared a bunch one files must be manually modified. For example the commit for 9.2.1 was 961ba3b. It changes a bunch of `*.dnn` files and adds an extra `.sqldataprovider`.

This is tedious and can be complete unnecessary. We will be submitting a PR for the 9.2.2 but going forward this could and should be automated:

1. Version numbers in manifests should be updated automatically on CI during build
1. The SQL script could be avoided by having each individual module set its correct version based on what it have in the manifest.

ohine:
This is about 85% complete with the new VSTS CI pipeline.

valadas:
I think this is now done @ohine right? I am closing but if I am wrong, please reopen.

PR #2159, opened 2018-07-13
DNN-21505: Upgrade page is displayed after install Evoq Engage 9.2.2.7-2

Mohtshum:
Fixes DNN-21505
- Need to make changes for versions 

-  *.dnn files 
-  solutioninfo.cs 
-  add 09.02.20.sqlprovider

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
@@ -1,0 +1,81 @@
+/************************************************************/
+/*****              SqlDataProvider                     *****/
+/*****                                                  *****/
+/*****                                                  *****/
+/***** Note: To manually execute this script you must   *****/
+/*****       perform a search and replace operation     *****/
+/*****       for {databaseOwner} and {objectQualifier}  *****/
+/*****                                                  *****/
+/************************************************************/
+
+/* Update version number in DesktopModules */
+/*******************************************/
+DECLARE @version NVARCHAR(8)
+SET @version = '09.02.02'
+
+UPDATE {databaseOwner}{objectQualifier}DesktopModules SET Version = @version WHERE ModuleName='Authentication'
+UPDATE {databaseOwner}{objectQualifier}DesktopModules SET Version = @version WHERE ModuleName='Portals'
+UPDATE {databaseOwner}{objectQualifier}DesktopModules SET Version = @version WHERE ModuleName='HostSettings'
+UPDATE {databaseOwner}{objectQualifier}DesktopModules SET Version = @version WHERE ModuleName='Scheduler'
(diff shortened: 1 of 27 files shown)


tpluscode:
Related to #2158, this correctly sets us up for 9.2.2 release @dnnsoftware/tag.

Without these changes (the SQL Provider file actually), Platform goes nuts and tries to perform an upgrade right after installation

Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
```diff
@@ -0,0 +1,81 @@
...
+/*****              SqlDataProvider                     *****/
+/*****                                                  *****/
+/*****                                                  *****/
+/***** Note: To manually execute this script you must   *****/
+/*****       perform a search and replace operation     *****/
+/*****       for {databaseOwner} and {objectQualifier}  *****/
+/*****                                                  *****/
+/************************************************************/
+
+/* Update version number in DesktopModules */
```
mitchelsellers: Do we have a history as to WHY we are doing this?  The installation of the various module packages update the version to the version that is referenced in the install package.  This is essentially duplicating the work of the installer.

tpluscode: @zyhfish @ashishpd, any insights?

mitchelsellers: I do know that we have to have a 09.02.02.sqldataprovider file as things get weird without it.  I'm curious on the actual content/process within

zyhfish: each release need a sqldataprovider file for it, even it may contains nothing. AFAIK currently we nee update legacy module's version in each upgrade.

bdukes: Are these extensions that are getting included automatically and aren't part of any extension package?

zyhfish: these are the core data providers, doesn't belong to any extension package.

affizafazreen:
hello there - can you help to expedite this review - it is blocking our 9.2.2 testing cc @ashishpd

Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
sleupold: I don't see a need to update version number of components, which haven't been modified.
Of course, they can't be applied to previous versions, but there is no need to update its registration.

But this has been done like this since DNN 7.0 AFAIR, IMO we shouldn't change before 10.0.1

PR #2168, opened 2018-07-18
CI: Adjustments to cake build script

ohine:
This PR enables the web deploy package, and it also creates a new task so we can generate builds of just platform or build all of the related repositories. 

The source packaging still has some issues, I'm hoping to get those solved soon.


tpluscode:
I don't think it's a good idea to be downloading and building latest `development` of PB and other repositories. It will cause inconsistent and accidental builds.

For example, now we are preparing 9.2.1. Say it's been release and we're already working on 9.3+. If you wanted to go back to 9.2.1 you will end up mixing 9.2.1 of Platform with the latest of Persona Bar.

The default should be instead to download tagged commit from those repositories with the option to select latest. I think it should be easily achievable by hacking the GitHub urls.

ohine:
Downloading the latest and getting a development environment configured is one of the many pieces the overall CI/VSTS work is attempting to solve. This specific pull request actually disables that functionality from the default build and makes it an optional step, this way we can have the PR verification CI task only compile the platform, and another task call the same script with a different target to build and test everything.

Remember it's a work in progress. We are currently attempting to mimic the internal builds from jenkins but without direct access to that system. The next pull request will add variables so we can switch which branches for each of the external extensions. This will help when we are getting ready to release 9.2.2 and until we have specific branches created for that release, development is the branch to target for the platform and all the related repositories.

We are also in the middle of discussions on how to package up and consume releases from the various repositories, but since that's something that is new and isn't part of the current release pipeline, it's safe to say those enhancements will come after the 9.2.2 release.

bdukes:
Is this an expected error?
>YARN : error An unexpected error occurred: "https://registry.yarnpkg.com/dnn-webpack-externals: Not found".

ohine:
As for the yarn error, you need to run ‘npm config set registry https://www.myget.org/F/dnn-software-public/npm/’ first. I need to get this added into the cake script, just not sure the exact best spot for it yet. Right now VSTS sets it as one of the first commands before calling cake.

tpluscode:
> As for the yarn error, you need to run ‘npm config set registry https://www.myget.org/F/dnn-software-public/npm/’ first

Easiest way is to add an `.npmrc` file to do repository. If memory serves me right it should be simply

```
registry=https://www.myget.org/F/dnn-software-public/npm/
```

PR #2173, opened 2018-07-19
DNN-21290: Import/Export Circular Dependency in pages cause pages not…

### LINKED ISSUE #2172: Import/Export Circular Dependency in pages cause pages not to be imported
## Steps to reproduce

On Source: 
1. Create 2 pages (these are gonna be the child pages). 
2. Create 1 page (this is the parent page with a large tabid) 
3. Nest/Move the 2 pages from step 1 into the parent page created. 
4. Go to the page settings for the parent page, and change the Page type to existing, make it point to one the child pages. 
4. Create a Full export of the site. You can choose to exclude extensions/Content Library and assets since these are not needed. 

On Target: 
3. Check whether the pages are created. 

## Current result

The pages are not imported since the child pages require the parent page to be added first; nevertheless, the parent page is a reference to the child page which has not been imported yet so all pages fail to import.

## Expected result

While this is an edge case, there should be a way to simply add one page before the other one in case of a circular dependency.

## Affected version

* [x] 9.2

## Workaround

Set the parent pages as normal pages during the export process, and set them as references to existing pages once the Import is completed (doable if this is the case for less than 10 pages)

tpluscode:
## Summary

fixes #2172 

This has been verified to work in client's patched 9.2.0 installation.

@Mohtshum if possible, please share any helpful insight about implementation details.

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -82,6 +82,8 @@ namespace Dnn.ExportImport.Components.Services
         private IList<int> _exportedModuleDefinitions = new List<int>();
 
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(ExportImportEngine));
+        private Dictionary<Guid, int> _partialImportedTabs = new Dictionary<Guid, int>();
+        private Dictionary<int, bool> _searchedParentTabs = new Dictionary<int, bool>();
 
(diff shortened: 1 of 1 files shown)


Mohtshum:
Current implementation is to check for parent recursively, by iterating through each tab either by checking its parent id or its tab-type

ohine:
Here is the successful results from a manually queued build of the DNN[CI] vsts task. https://dotnetfoundation.visualstudio.com/DNN/_build/results?buildId=3843&_a=summary

PR #2175, opened 2018-07-20
DNN-21870: only clean user count cache during create user process.

zyhfish:


ashishpd:
This PR is to resolve a major perf problem where anytime a user gets created, DNN flushes the entire entity cache. This causes slow down for sites that have large objects in the entity cache.

tpluscode:
Reported as #2178

Issue #2133, opened 2018-06-25
JWT ExtendToken: passing RenewToken (rtoken) has no use


erw13n:
## Description 
I check the source code of JWT ExtendToken, and it seem sent any renewalToken will just pass-in and return with exact same value.
## Steps to reproduce
Call API of JWT ExtendToken, and pass any data on rtoken
## Current result
It return accessToken and the exact same word we type on RenewalToken

![image](https://user-images.githubusercontent.com/792847/41873926-a1322e50-78f0-11e8-9f21-d3e72c824c4c.png)

Behnam-Emamian:
I am fixing it.

Behnam-Emamian:
waiting for more approvals for merging to master

contact2adil:
is it fixed yet? I am facing the same issue but no solution yet.

valadas:
It looks like it was merged in the 9.3.0 milestone, you can give it a try (only on a non production environment) using the release candidate available at https://github.com/dnnsoftware/Dnn.Platform/releases 

But keep in mind:
- This release is not production ready yet
- There are known issues with RC0 (mainly UI issues with some persona bar components)
- If you see RC1 is out (should be very shortly, try that instead).

valadas:
I am assuming this is now fixed and closing, if not, please open a new issue.

PR #2179, opened 2018-07-20
renewal token must be validate before renewing access token.

Behnam-Emamian:
Fix issue #2133

## Summary

1. renewal token must be validate before renewing access token.
2. making error messages clear

### DIFF
--- a/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
+++ b/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
@@ -214,7 +214,7 @@ namespace Dnn.AuthServices.Jwt.Components.Common.Controllers
             if (ptoken == null)
             {
                 if (Logger.IsTraceEnabled) Logger.Trace("Token not found in DB");
-                return EmptyWithError("token-not-found");
+                return EmptyWithError("not-found");
             }
 
             if (ptoken.RenewalExpiry <= DateTime.UtcNow)
@@ -226,26 +226,26 @@ namespace Dnn.AuthServices.Jwt.Components.Common.Controllers
             if (ptoken.RenewalHash != GetHashedStr(renewalToken))
             {
                 if (Logger.IsTraceEnabled) Logger.Trace("Invalid renewal token");
-                return EmptyWithError("bad-renewal-token");
+                return EmptyWithError("bad-token");
             }
 
             if (ptoken.TokenHash != GetHashedStr(rawToken))
             {
                 if (Logger.IsTraceEnabled) Logger.Trace("Invalid access token");
-                return EmptyWithError("bad-access-token");
+                return EmptyWithError("bad-token");
             }
 
             var userInfo = TryGetUser(jwt, false);
             if (userInfo == null)
             {
                 if (Logger.IsTraceEnabled) Logger.Trace("User not found in DB");
-                return EmptyWithError("user-not-found");
+                return EmptyWithError("not-found");
             }
 
             if ((ptoken.UserId != userInfo.UserID))
             {
                 if (Logger.IsTraceEnabled) Logger.Trace("Mismatch token and user");
-                return EmptyWithError("bad-token-user");
+                return EmptyWithError("bad-token");
             }
 
             return UpdateToken(renewalToken, ptoken, userInfo);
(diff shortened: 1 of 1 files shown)


DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
```diff
@@ -224,17 +223,29 @@ public LoginResultData RenewToken(HttpRequestMessage request, string renewalToke
...
+                return EmptyWithError("bad-access-token");
+            }
+
             var userInfo = TryGetUser(jwt, false);
             if (userInfo == null)
             {
-                if (Logger.IsTraceEnabled) Logger.Trace("Token not found in DB");
-                return EmptyWithError("not-found");
+                if (Logger.IsTraceEnabled) Logger.Trace("User not found in DB");
+                return EmptyWithError("user-not-found");
```
mitchelsellers: By disclosing this are we giving a malicious user more information than they need?

DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
```diff
@@ -224,17 +223,29 @@ public LoginResultData RenewToken(HttpRequestMessage request, string renewalToke
...
             }
 
-            if ((ptoken.TokenHash != GetHashedStr(rawToken)) || (ptoken.UserId != userInfo.UserID))
+            if ((ptoken.UserId != userInfo.UserID))
             {
-                if (Logger.IsTraceEnabled) Logger.Trace("Mismatch in received token");
-                return EmptyWithError("bad-token");
+                if (Logger.IsTraceEnabled) Logger.Trace("Mismatch token and user");
+                return EmptyWithError("bad-token-user");
```
mitchelsellers: Same question here, I'm inclined to say it should just be "bad-token"

nvisionative:
I agree @mitchelsellers - “bad-token” seems appropriate.  Thanks for the contribution @Behnam-Emamian !

DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
Behnam-Emamian: Ok, I see

You wanted to hide the main reason of problem?

as I can see the secret which jwt token has been encrypted is unique for each user in portal, so at the end of day the attacker can find just its secret.

I think we can get more benefit for debug instead of users attacks.

DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
Behnam-Emamian: var secret = ObtainSecret(sessionId, portalSettings.GUID, userInfo.Membership.LastPasswordChangeDate);

secret is compliantly unique for each user in each portal so I think there is no reason for worry about the disclose the details.

Behnam-Emamian:
another question is why we need to save token and information in Database? why we need a session id?

JWT is a self contained solution, it has all the information in payload all the time. why we need to save it in database and why we make a session for it?

donker:
"why we need to have another token for renewal?"
This is the JWT standard that specifies that, no? I haven't seen a JWT implementation that doesn't use separate refresh tokens.

"why we need to save token and information in Database?"
I'm not sure I understand the question but here goes: Given that the entire site is recycled every so often we don't have any info on the server about the user's access to data. This is why we need to verify the token. To verify we need the token at the server.

"why we make a session for it?"
This would be to speed things up I believe. If you want to avoid having to go through token verification for every request you can keep a session alive with a session ID.

Behnam-Emamian:
1. I agree. this is not a JWT standard
2. because we use dynamic password for generating JWT token, so every time we need to make password again for validation so it is better to save whole token in database for validation. it is not a part of standard JWT implementation too ;-)
3. infect it is decrease the speed, because all the time we need 1 fetch from database to validate the token but if we use static password, we just need to calculate that is this the valid token or not?

Behnam-Emamian:
it seems the target of this module was making Mobile application by JWT, not implementing pure JWT.

for my usage, I mean calling external web services, I need pure standard JWT implementation which I have done ;-)

PR #2182, opened 2018-07-24
JWT Audience

### LINKED ISSUE #2181: JWT Audience is not implemented
## Description 
I have an architect of having several microservices which can provide some services to DNN. I want to have a JWT just working for each of my microservices not having one general JWT that can work everywhere.

## Steps to reproduce
Call JWT api, there is no way to set Audience.

## Expected result
I need to have a parameter in JWT api to get the audience and make a aud claim in JWT.

Behnam-Emamian:
Fixes #2181


donker:
Given that method signatures are changed this should go into at least a minor release. Not a revision.

From the tests it appears we need to examine file upload more closely.

valadas:
@ohine this is blocked from VSTS check but it passes Azure pipeline, can we remove that check or will you be able when ready to merge this to override the check ?

valadas:
@dnnsoftware/tag Any concerns with this PR for Dnn10? It has been here for some time and stalebot picked it up as stale. I don't know enough about JWT to review it.

PR #2195, opened 2018-08-01
Defect/dnn 21337 SI: Option to use Site Timezone For Page Versioning Timestamps

### LINKED ISSUE #2200: Option to use Site Timezone For Page Versioning Timestamps
## Description 

Page Versioning Timestamps appear in Database/Server Time Zone

## Steps to reproduce

Go to page versions

## Current result

time listed for each version is stamped with server time, it needs to be in Site's time zone

dnn-timezone-issues

## Expected result

Everything on the UI should be displayed in user's timezone. There are APIs available in DNN Platform to convert any time to a user's timezone.

## Affected version

* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

mean2me:
## Summary
Go to page versions, time listed for each version is stamped with server time, it needs to be in Site's time zone

## Current
Page Versioning Timestamps appear in Database/Server Time Zone

## Desired
Everything on the UI should be displayed in user's timezone. There are APIs available in DNN Platform to convert any time to a user's timezone.

fixes #2200


tpluscode:
fixes #2200

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +119,34 @@ private static string GetTabVersionsCacheKey(int tabId)
...
+                }
+                return tabVersion;
+            });
+        }
+
+        private DateTime ApplyDaylightSavingHour(TimeZoneInfo info, DateTime localTime)
+        {
+            if(info.IsDaylightSavingTime(localTime))
+            {
+                return localTime.AddHours(1);
```
tpluscode: I'm pretty sure this is incorrect. `TimeZoneInfo` is perfectly aware of daylight saving in specific time zones

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersion.cs
```diff
@@ -51,6 +51,11 @@ public class TabVersion : BaseEntityInfo
         /// </summary>
         public DateTime TimeStamp { get; set; }
 
+        /// <summary>
+        /// Creation date displayed in current user's preferred timezone
+        /// </summary>
+        public DateTime CreateOnUserLocalDate { get; set; }
```
tpluscode: I don't think we want this here. Looks more like a UI-specific issue. Instead of adding a property I would change how the dates are presented on the view.

@dnnsoftware/tag ?

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +119,34 @@ private static string GetTabVersionsCacheKey(int tabId)
...
         }
+
+        private IEnumerable<TabVersion> ApplyUserLocalTime(IEnumerable<TabVersion> tabVersions, TimeSpan userUtcOffset, TimeZoneInfo userTimeZone)
+        {
+            return tabVersions.Select((tabVersion) =>
+            {
+                var serverUtcOffset = TimeZoneInfo.Local.BaseUtcOffset;
+                if (serverUtcOffset.CompareTo(userUtcOffset) >= 0)
+                {
+                    tabVersion.CreateOnUserLocalDate = ApplyDaylightSavingHour(userTimeZone, tabVersion.CreatedOnDate.Subtract(serverUtcOffset).Add(userUtcOffset));
```
tpluscode: There is also a `Timestamp` property. Which is used on the versions UI?

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +119,34 @@ private static string GetTabVersionsCacheKey(int tabId)
...
+                }
+                else
+                {
+                    tabVersion.CreateOnUserLocalDate = ApplyDaylightSavingHour(userTimeZone, tabVersion.CreatedOnDate.Add(serverUtcOffset).Subtract(userUtcOffset));
+                }
+                return tabVersion;
```
tpluscode: The currently displayed time is UTC or server time?

Either way you do not have to manually calculate the times. Please check [this guide][tz] for timezone-related APIs. It's a simple built-in call to convert from one timezone to the other.

[tz]: https://docs.microsoft.com/en-us/dotnet/standard/datetime/converting-between-time-zones#converting-between-any-two-time-zones

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
mean2me: Timestamp is not used in UI. I checked before to apply the fix :)

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +119,15 @@ private static string GetTabVersionsCacheKey(int tabId)
         {
             return string.Format(DataCache.TabVersionsCacheKey, tabId);
         }
+
+        private IEnumerable<TabVersion> ApplyUserLocalTime(IEnumerable<TabVersion> tabVersions, TimeSpan userUtcOffset, TimeZoneInfo userTimeZone)
```
tpluscode: `userUtcOffset` parameter is now unused

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -42,17 +43,23 @@ public TabVersion GetTabVersion(int tabVersionId, int tabId, bool ignoreCache =
         
         public IEnumerable<TabVersion> GetTabVersions(int tabId, bool ignoreCache = false)
         {
+            var user = UserController.Instance.GetCurrentUserInfo();
+            var userTimeZone = user.Profile.PreferredTimeZone;
+            var userUtcOffset = userTimeZone.BaseUtcOffset;
```
tpluscode: `userUtcOffset` is unused

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +119,15 @@ private static string GetTabVersionsCacheKey(int tabId)
         {
             return string.Format(DataCache.TabVersionsCacheKey, tabId);
         }
+
+        private IEnumerable<TabVersion> ApplyUserLocalTime(IEnumerable<TabVersion> tabVersions, TimeSpan userUtcOffset, TimeZoneInfo userTimeZone)
+        {
+            return tabVersions.Select((tabVersion) =>
+            {
+                tabVersion.CreateOnUserLocalDate = System.TimeZoneInfo.ConvertTime(tabVersion.CreatedOnDate, userTimeZone);
```
tpluscode: Because `CreatedOnDate` is a `DateTime` the conversion can produce incorrect result (please see *Remarks* section on [this page](https://msdn.microsoft.com/en-us/library/bb382835(v=vs.110).aspx).

Hence my earlier question. What is the representation of the times in the database? They are server time, right? But what is their `DateTimeKind`?

Specifically, if it's `Local` then the source timezone should be retrieved and passed explicitly to the [`TimeZoneInfo.ConvertTime Method (DateTime, TimeZoneInfo, TimeZoneInfo)`](https://msdn.microsoft.com/en-us/library/bb382770(v=vs.110).aspx) overload

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -187,6 +187,7 @@ private object GetRoleNull(int RoleID)
 
         public abstract IDataReader ExecuteSQL(string sql);
         public abstract IDataReader ExecuteSQL(string sql, int timeoutSec);
+        public abstract T ExecuteSQL<T>(string sql);
```
tpluscode: Hm, this is a very open method, easy to shoot oneself in the foot this way.

I'd rather you do one of the following (firs option preferred)

1. use `IDataReader ExecuteSQL(string sql)` and retrieve the result from the reader
2. add a stored procedure ad use the `ExecuteScalar` method

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3353,6 +3354,14 @@ public virtual DateTime GetDatabaseTime()
             return ExecuteScalar<DateTime>("GetDatabaseTime");
         }
 
+        public virtual DateTimeOffset GetDatabaseTimeOffset() {
+            var dateTimeString = ExecuteSQL<DateTimeOffset>("SELECT SYSDATETIMEOFFSET()").ToString();
+            DateTimeOffset dateTimeOffset;
+            DateTimeOffset.TryParse(dateTimeString, out dateTimeOffset);
```
tpluscode: You initially get a `DateTimeOffset` only to stringify and parse again. Is that necessary?

DNN Platform/Tests/DotNetNuke.Tests.Web/InternalServices/TabVersionControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+using DotNetNuke.Entities.Tabs.TabVersions;
+using DotNetNuke.Tests.Utilities.Mocks;
+using DotNetNuke.Entities.Users;
+using DotNetNuke.Entities.Controllers;
+using DotNetNuke.Common;
+
+namespace DotNetNuke.Tests.Web.InternalServices
+{
+    [TestFixture]
+    class TabVersionControllerTests
```
tpluscode: Please make `public`

DNN Platform/Tests/DotNetNuke.Tests.Web/InternalServices/TabVersionControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        private Mock<ICBO> _mockCBO;
+        private Mock<IUserController> _mockUserController;
+        private Mock<IHostController> _mockHostController;
+
+        private const int UserID = 1;
+        private const int TabID = 99;
+
+        private int OffsetHours { get; set; }
```
tpluscode: Please remove field and make parameter of `GetMockedUser ` method.

DNN Platform/Tests/DotNetNuke.Tests.Web/InternalServices/TabVersionControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        private void SetupCBO()
+        {
+            _mockCBO = new Mock<ICBO>();
+            _mockCBO.Setup(c => c.GetCachedObject<List<TabVersion>>(It.IsAny<CacheItemArgs>(), It.IsAny<CacheItemExpiredCallback>(), It.IsAny<bool>()))
+                .Returns(() =>
+                {
+                    return GetMockedTabVersions();
+                });
```
tpluscode: This can be simply `.Returns(GetMockedTabVersions);`

DNN Platform/Tests/DotNetNuke.Tests.Web/InternalServices/TabVersionControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        [Test]
+        [TestCase(-11, "2018-08-15 1:00:00.0")]
+        [TestCase(-2, "2018-08-15 10:00:00.0")]
+        [TestCase(12, "2018-08-16 0:00:00.0")]
+        public void GetTabVersions_Verify_User_Preferred_TimeZone(int offsetHours, string expectedDateTimeString)
+        {
+
+            // Arrange
+            OffsetHours = offsetHours;
```
tpluscode: Also, please call it `userPreferredTimeZone` to be clear on its meaning.

DNN Platform/Tests/DotNetNuke.Tests.Web/InternalServices/TabVersionControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            SetupUserController();
+
+            // Act
+            var tabVersionController = new TabVersionControllerTestable();
+            var tabVersions = tabVersionController.GetTabVersions(TabID);
+            var tabVersion = tabVersions.FirstOrDefault();
+            var expectedDate = DateTime.Parse(expectedDateTimeString);
```
tpluscode: Parsing datetimes from string is prone to fail on machines with different locale.

Please consider replacing `[TestCase]` with [`[TestCaseSource]`](https://github.com/nunit/docs/wiki/TestCaseSource-Attribute) where you would create appropriate values with `new DateTime()`

DNN Platform/Library/Data/DataProvider.cs
mean2me: Is it worth for such a small feature improvement?

tpluscode: You mean the stored procedure @mean2me? It's probably not worth it.

Unless you have a better suggestion I would prefer the former. Instead of adding a new overload please use the `IDataReader` one above

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3353,6 +3353,26 @@ public virtual DateTime GetDatabaseTime()
...
+                        return reader.GetDateTimeOffset(0);
+                    } else
+                    {
+                       throw new Exception("Error retrieving database UTC offset");
+                    }
+                }
+            } catch (Exception)
+            {
+                throw;
+            }
```
tpluscode: This `try..catch` is redundant.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +117,26 @@ private static string GetTabVersionsCacheKey(int tabId)
         {
             return string.Format(DataCache.TabVersionsCacheKey, tabId);
         }
+        
+        protected virtual TimeZoneInfo GetDatabaseDateTimeOffset()
```
tpluscode: One last comment. I think this method should be renamed to `GetDatabaseTimeZone` and moved. Maybe also to `DateUtils`?

tpluscode:
> The conversion to user time should be performed in UI level or close to it

It was my initial thought too but in the end you would probably just end up with similar boilerplate in other parts of UI. Would you agree? 

Ideally, we would have `DateTimeOffset` in the DB and DTO classes so that it's always clear what is the exact point in time. Otherwise it's easy to make wrong assumptions and meaningless conversions.

A good thing to do would be to establish a common way for performing this operations. Currently the https://github.com/dnnsoftware/Dnn.Platform/pull/2195/files#diff-1d95514c463d6eb8296c986a69e4b15dR129 method is too specific to tab versions.

Any other thought @dnnsoftware/tag ?

DNN Platform/Library/Data/SqlDataProvider.cs
```diff
@@ -442,13 +442,13 @@ public override IDataReader ExecuteSQL(string sql, int timeoutSec)
         {
             return ExecuteSQLInternal(ConnectionString, sql, timeoutSec);
         }
-
```
sleupold: please exclude these non beneficial file changes from your pull request

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersion.cs
sleupold: There might be use for a function, but not for a property

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -112,6 +117,26 @@ private static string GetTabVersionsCacheKey(int tabId)
...
+        {
+            var dateTimeOffset = DateUtils.GetDatabaseDateTimeOffset();
+            var offset = dateTimeOffset.Offset;
+            var id = string.Format("UTC {0}", offset.ToString());
+            return TimeZoneInfo.CreateCustomTimeZone(id, offset, id, id);
+        }
+
+        private IEnumerable<TabVersion> ApplyUserLocalTime(IEnumerable<TabVersion> tabVersions, TimeZoneInfo userTimeZone)
+        {
+            var serverTimeZone = GetDatabaseDateTimeOffset();
```
sleupold: why not just using TimeZoneInfo.Convert to convert from one timezone to another?

sleupold:
@tpluscode, @mean2me
We should limit database access to as least queries and interfaces as possible. 
I don't see a need for method GetDatabaseTimeOffset in DataProvider as this maybe easily calculated by from `GetDatabaseLocalTime() - GetDatabaseUtcTime() `, which are using cached values (unless we are rewriting these methods to use GetDatabaseTimeOffset instead).
In DNN 6, we intentionally removed previous UserTime class to avoid unnecessary code, as .Net supports easy calculations between timezones using timeZoneInfo.ConvertTime, see GetProperty in DotNetNuke.Services.Tokens.PropertyAccess.DateTimePropertyAccess.cs class for reference (although it needs to be updated to use Datatabase time instead of DateTime.Now(), but that is another issue.

sleupold:
Providing correct timezone calculations has always been a challenge in DNN - see the old forums module. 
We need to be aware there are four different time zones usually involved:
- UTC
- server time zone
- Portal default timezone
- user timezone
In DNN 6, we provided methods to provide now and UTCNow from the database, to make sure, we always get the same time, despite obtaining now in the platform code or in a stored procedure - even if both are on different servers with non-synched clocks. Therefore GetDatabaseLocalTime() should always be preferred over DateTime.Now() and  GetDatabaseUtcTime() over DateTime.UTCnow()!
In general, it is preferrable to store all timestamp values in the database as UTC, which allows moving the database from one timezone to another or from a local server to azure without the need to correct all timezone values.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
mitchelsellers: I second this thought

mean2me: What if Database is on a remote server with a different TZ?

mean2me: @sleupold @mitchelsellers @tpluscode 
We also have to consider the CreatedOnDate property, inherited from BaseEntityInfo, is a read only property. It's never set from code. Its value comes from a database "NOW()" function. That's reason why I took care of using Database timezone offset instead of the server's one. Whatever we do with that *CreatedOnDate* property, it's meaningful only if we know the Database locale time settings. IMHO.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -42,17 +43,21 @@ public TabVersion GetTabVersion(int tabVersionId, int tabId, bool ignoreCache =
...
             {
                 DataCache.RemoveCache(tabCacheKey);
             }
-            
-            return CBO.GetCachedObject<List<TabVersion>>(new CacheItemArgs(tabCacheKey,
+            var tabVersions = CBO.Instance.GetCachedObject<List<TabVersion>>(new CacheItemArgs(tabCacheKey,
                                                                     DataCache.TabVersionsCacheTimeOut,
                                                                     DataCache.TabVersionsCachePriority),
-                                                            c => CBO.FillCollection<TabVersion>(Provider.GetTabVersions(tabId)));            
+                                                            c => CBO.FillCollection<TabVersion>(Provider.GetTabVersions(tabId)), false);
```
daguiler: What's the purpose of this `false` argument @mean2me ?

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -108,10 +113,11 @@ private void ClearCache(int tabId)
             DataCache.RemoveCache(GetTabVersionsCacheKey(tabId));
         }
 
-        private static string GetTabVersionsCacheKey(int tabId)
+        public static string GetTabVersionsCacheKey(int tabId)
```
daguiler: Unused, should be reverted

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -3353,6 +3353,20 @@ public virtual DateTime GetDatabaseTime()
...
+        public virtual DateTimeOffset GetDatabaseTimeOffset()
+        {
+            using (var reader = (SqlDataReader)ExecuteSQL("SELECT SYSDATETIMEOFFSET()"))
+            {
+                if(reader.HasRows && reader.Read())
+                {
+                    return reader.GetDateTimeOffset(0);
+                } else
+                {
+                    throw new Exception("Error retrieving database UTC offset");
```
daguiler: I don't think this will ever happen and even if it does, let's keep the original exception info and stack trace better

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersion.cs
daguiler: And also this is not used anywhere now.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -42,17 +43,21 @@ public TabVersion GetTabVersion(int tabVersionId, int tabId, bool ignoreCache =
         
         public IEnumerable<TabVersion> GetTabVersions(int tabId, bool ignoreCache = false)
         {
-            //if we are not using the cache, then remove from cacehh and re-add loaded items when eeded later
+            var user = UserController.Instance.GetCurrentUserInfo();
+            var userTimeZone = user.Profile.PreferredTimeZone;
```
daguiler: Not used.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
```diff
@@ -42,17 +43,21 @@ public TabVersion GetTabVersion(int tabVersionId, int tabId, bool ignoreCache =
...
+
+            //if we are not using the cache, then remove from cache and re-add loaded items when eeded later
```
daguiler: There is still one typo in the comment

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
sleupold: @mean2me 
Timestamps should always be obtained from SQL Server, therefore there is a function to get it in server time or UTC. In general, UTC is the better choice, however, for historical reasons, DNN always uses local SQL Server time.

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
mean2me: @daguiler it's used in lines 50 and 113

DNN Platform/Library/Data/SqlDataProvider.cs
mean2me: @sleupold I think it's worthy for readability to add one line space between two methods definitions

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
tpluscode: Daniel meant to make it `private` again

DNN Platform/Library/Entities/Tabs/TabVersions/TabVersionController.cs
tpluscode: Previously no value was passed, which is equivalent to `true`. Is this change intentional?

sleupold:
@tpluscode 
there is no adjustment of date values, if the database is moved to a server in a different time zone. The best would be migrating all datetime columns to UTC, however, this would be a big change,

sleupold:
we removed some methods, which were misnamed and resulted in issues like the wrong time displayed by Forums module (see dnnsoftware.com). Of course, there should have been better guideline provided. 
Note: this is the associated blog post: https://www.dnnsoftware.com/community-blog/cid/135312/enhanced-time-zone-support-in-6

EPTamminga:
I support the principle that all date/time fields in the database should store the date/time in UTC.

sleupold:
>           I support the principle that all date/time fields in the database should store the date/time in UTC.

We would need a separate issue for this. However, given the number of date fields, this would be a bigger effort.

valadas:
> Of course, there should have been better guideline provided.
> Note: this is the associated blog post: https://www.dnnsoftware.com/community-blog/cid/135312/enhanced-time-zone-support-in-6

I would have been awesome to have read this blog post a couple months ago when I replaced a lot of deprecated API usages in a ton of modules. There should have been a short note on the deprecated Time Zone methods and properties with a link to this blog post.

sleupold:
@valadas 
I can't but agree that source code comments would have been helpful.

PR #2197, opened 2018-08-01
DNN-22117: add sp GetFilesRecursive to get files from folder and all sub folders in one time.

### LINKED ISSUE #2201: Improve search times when searching images
## Description 

...when portal has thousands of folders/files.

## Steps to reproduce

1. install engage 9.2.0 and login as host; 
2. add a new page and edit the default html module, open images dialog; 
3. type something in the search box.

Screencast: https://www.screencast.com/t/d4A2NQOmKAm

## Current result

the search will be very slow to response.

## Expected result

search result should be display quickly.

## Affected version

* [x] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

zyhfish:
fixes #2201


sleupold:
Does it really make sense to pass the folder path instead of the folderid?

sleupold:
Another question: is there any benefit in sorting the files by folderID and fileName?
it sounds more logical to me sorting it by folderpath (which is used for WHERE as well) and fileName - which should result in better performance.

Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
```diff
@@ -76,6 +76,89 @@ UPDATE {databaseOwner}{objectQualifier}Packages SET Version = @version WHERE Nam
...
+            F.EnablePublishPeriod,
+            F.StartDate,
+            F.EndDate,
+            F.CreatedByUserID,
+            F.CreatedOnDate,
+            F.LastModifiedByUserID,
+            F.LastModifiedOnDate,
+            F.PublishedVersion,
+            F.ContentItemID,
+            F.HasBeenPublished
```
tpluscode: Do we need the `IF...ELSE` statement if there is this flag? I mean, I see that the selects are from to different tables. Could you explain that to me?

tpluscode:
> Does it really make sense to pass the folder path instead of the folderid?
This would not even require the PortalID to be passed

So, I'm not sure about that last one. Wouldn't skipping `PortalID` make it potentially possible to retrieve files from wrong portal?

sleupold:
@tpluscode the sproc does not query from different tables but uses different views - one filters by state and date to make sure to return currently published files only.

sleupold:
@tpluscode folderID is always portal specific. 
IMO we should keep sproc signatures similar, i.e. GetFilesRecursive should use same signature as GetFiles sproc (another option would be adding a flag includeSubfolders to GetFiles parameter list and select appropriately inside the sproc. This would keep the data layer in API a bit more compact.

tpluscode:
> IMO we should keep sproc signatures similar, i.e. GetFilesRecursive should use same signature as GetFiles sproc

Consistency is always good. @zyhfish does it fit you idea for the store procedure?

> another option would be adding a flag includeSubfolders to GetFiles parameter list

that would be a breaking change to consumers and thus not something we'd want to do. Unless you mean to add an overload instead of a new method

```diff
public virtual IDataReader GetFiles(int folderId, bool retrieveUnpublishedFiles = false)
{
-    return ExecuteReader("GetFiles", folderId, retrieveUnpublishedFiles);
+    return GetFiles(folderId, false, retrieveUnpublishedFiles);
}

+public virtual IDataReader GetFiles(int folderId, bool recursive, bool retrieveUnpublishedFiles = false)
+{
+    if(recursive)
+    {
+        return ExecuteReader("GetFilesRecursive", folderId, retrieveUnpublishedFiles);
+    }
+    
+    return ExecuteReader("GetFiles", folderId, retrieveUnpublishedFiles);
+}
```

sleupold:
"that would be a breaking change to consumers and thus not something we'd want to do. Unless you mean to add an overload instead of a new method" - The stored procedure is not considered as part of the API - 3rd party developers should not call them inside their own sprocs nor from their code behind, but use the DNN API instead. For the data method, we may just add an overload with the additional parameter instead of a new method.

tpluscode:
I understood you suggested changing the signature of existing `GetFiles` method. That is the breaking change I referred to

mitchelsellers:
Historically the only protected signatures are those of the C# API's and not those of the Stored Procedures.

For example  in this change since the method signature remains `public virtual IDataReader GetFiles(int folderId, bool retrieveUnpublishedFiles = false)` it is fine to change the SP, even if you change parameters.

I'm ok with the proposed change to add the new method, however, if doing that we should be marking that overload as deprecated, to be removed in version 11.0.

tpluscode:
Ok, let's see if we're on the same page here @sleupold @mitchelsellers 

1. change the new method to overload `GetFiles`
1. change the signature to be using `folderId` like the existing one

Regarding deprecation, I'm not convinced that we will want to remove the current overload.

mitchelsellers:
@tpluscode Affirmative

sleupold:
as far as I can see, there is still a new stored procedure instead of merging both into one.

mitchelsellers:
I sill would like to see the older method marked as deprecated in 11.0 if we can

tpluscode:
> as far as I can see, there is still a new stored procedure instead of merging both into one (GetFiles with (optional) parameter (includeChilds Bit = 0)

@sleupold, I understood to only merge C# methods and not SPs. @zyhfish what do you say?

> I sill would like to see the older method marked as deprecated in 11.0 if we can

we can. is it really going to be removed though?

sleupold:
Deprecated methods will be removed one time in the future (DNN 11.0.0). 
The current overload should be removed, if it is covered by the other one (using an optional parameter).
Regarding the sproc, I prefer a single one, to make sure the output fields are always the same.
I created a performant sample:

[Indexed_Journal_Get.sql.txt](https://github.com/dnnsoftware/Dnn.Platform/files/2409369/Indexed_Journal_Get.sql.txt)

We might avoid the four branches, if we are always using portalID and filepath, the perf decrease should be minimal. What's more important for performance is the order of rows in the result set. is FolderID and FileName really beneficial - I would assume FolderPath and Filename would be better?

sleupold:
Did you check the best order of files - is it FolderID, FileName or FolderPath, FileName?

tpluscode:
@zyhfish please see Sebastian's comment

zyhfish:
This is same order with previous GetFiles sp, no change.

sleupold:
@zyhfish 
previous getFiles returned files from a single folder only, folderid was used in oder by to match the index.
If there is no specific order required, remove "order by". if the order should be "folderpath, filename", use these columns in "order by"

tpluscode:
Wish we actually had change history of stored procedures...

@sleupold, the previous version of `GetFiles` stored procedure was in 8.0.4 and indeed comes with the exact same `ORDER BY` clause:

https://github.com/dnnsoftware/Dnn.Platform/blob/development/Website/Providers/DataProviders/SqlDataProvider/08.00.04.01.SqlDataProvider#L1399

sleupold:
@tpluscode 
you are correct - and wrong :)
the previous version of GetFiles included an "ORDER BY FolderID, FInleName" - but "FolderID was used here just to meet one of the indexes - the previous version were always returning files from a single Folder only. However, this time we are retrieving files by FolderPath and FileName. In order to return ordered by FolderID and FileName, an additional SORT is perfomed as last action. 
However, if this sorting is not used or a different sort is needed (if files from a subtree are retrieved), we should make sure, the result is returned in this order to avoid an extra sort in .Net code. Be aware, that the new call may return a huge number of files, where sorting may affect performance significantly (I am not aware of any sort algorithm using less than O(N logN).

zyhfish:
should the order change to "FolderPath, FileName"?

sleupold:
@zyhfish 
if order "FolderPath, FileName" fits to your call of method GetFileVersionContent (order, how you are processing the results), yes.
(AFAIK this is currently used by Evoq products only)

valadas:
I think [FolderPath], [FileName] makes sense for usability. If a module needs it just by FileName, at least the list would already be partially sorted, which saves some cycles in re-sorting. However if sorting by FolderPath is decided to cost too much for performance (I am no sql expert), then I would just sort only on FileName since I don't see any use case where having the results by FolderId useful.

sleupold:
@valadas 
We are not talking about usability, but performance. Ideally any database call should return the rows in the same order as being used by the API to be displayed in UI, only for small numbers a sort can be done in API without significant impact. In this case, a recursive call might be used to return all files of a portal, returning a large number of files (i know clients with tens of thousands and there might be some with hundreds of thousands or even millions of records. Transferring such a large number of records to the API takes its time, but a resort will most likely result in a timeout (fastest known sort algorithms take O(N*logN), i.e. effort for sort increases much faster than number of records.
SQL Server is optimized to handle large number of records in storage and sorting (using parallelism), if proper indexes have been added, a sort might be avoided at all during retrieval. 
For the current stored procedure, requesting files from a single folder don't care whether the result is ordered by FolderID + FileName, FolderName + FileName or just FileName, as I don't expect ten thousands of files being typically stored in a single folder.  However, if we return files from a folder with all subtree, files should be returned in proper order, according how they are used for display or export. I assume FolderPath + FileName would be best (especially, no subfolder content is returned before the content of its parent and folders may be created on the fly, e.g. upon import), but this needs to be verified.

Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
```diff
@@ -148,7 +148,7 @@ AS
             F.HasBeenPublished
         FROM {databaseOwner}[{objectQualifier}vw_PublishedFiles] F            
         WHERE ((@Recursive = 1 AND ISNULL(F.PortalId, -1) = @PortalId AND F.Folder LIKE @FolderPath + '%') OR F.FolderID = @FolderID)
-        ORDER BY [FolderID], [FileName]
+        ORDER BY [Folder], [FileName]
```
sleupold: @zyhfish Thank you!

sleupold:
@tpluscode yes. I will review indexes - we might modify or add one or two, but this should be a separate PR

valadas:
@sleupold I totally agree that performance is key here and I understand this will return a huge amount of records in many instances. We are saying about the same thing. My point was that I am not SQL server performance expert and IF for some reason a decision was made that sorting by FolderName and FileName was not a viable option performance wise, THEN I would see not good use of sorting by FolderId and Filename and that in that situation I think only Filename would be fine. That being said if those who know more than me about query performance agree that FolderName, Filename is fine, I totally agree with that. With proper indexes, I guess this is the best option.

PR #2207, opened 2018-08-03
Improve user feedback for cleanup node in installer

### LINKED ISSUE #2206: Installer should not tell a user it's an error if not being able to delete a file/folder is an error
## Description 
The CleanUp installer will run through CleanUp files and try to process each line and delete what is mentioned. There is a try-catch around every delete and this will not fail the installer. But the code still adds an "error" rather than a "warning" to the user. This confuses users who will assume the install has failed.

## Steps to reproduce
Include a cleanup file with non-existent files/folders to delete

## Current result
Shows error in report

## Expected result
Shows *warning* in report

## Affected version
All AFAIK

## Affected browser
N/A

donker:
Instead of showing an install as having an error, failing to clean up a file/folder should show a warning as it is not fatal to the installation of an extension.

Fix #2206

### DIFF
--- a/DNN Platform/Library/Services/Installer/Installers/CleanupInstaller.cs
+++ b/DNN Platform/Library/Services/Installer/Installers/CleanupInstaller.cs
@@ -75,7 +75,7 @@ namespace DotNetNuke.Services.Installer.Installers
             }
             catch (Exception ex)
             {
-                Log.AddWarning(string.Format(Util.CLEANUP_ProcessComplete, ex.Message));
+                Log.AddWarning(string.Format(Util.CLEANUP_ProcessError, ex.Message));
                 //DNN-9202: MUST NOT fail installation when cleanup files deletion fails
                 //return false;
             }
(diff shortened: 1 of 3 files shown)


donker:
Ultimately the try-catch shouldn't be around the entire cleanup file as it is now. It should be on each line. But I didn't want to create any risk with this change of changing behavior. So left that as is for now.

DNN Platform/Library/Services/Installer/Installers/CleanupInstaller.cs
```diff
@@ -75,7 +75,7 @@ private bool ProcessCleanupFile()
             }
             catch (Exception ex)
             {
-                Log.AddFailure(ex);
+                Log.AddWarning(string.Format(Util.CLEANUP_ProcessComplete, ex.Message));
```
bdukes: is this supposed to be using the `CLEANUP_ProcessError` field you defined below?

donker: Yes. I'll adjust

PR #2212, opened 2018-08-07
DNN-22224: Registration form registering is failing when Use email as…

### LINKED ISSUE #2213: Cannot create new user with email address as username
## Steps to reproduce

1. Add User- username=email at Users > Persona Bar

## Current result

Not able to add

## Affected version

* [x] 9.2.1

Mohtshum:
… username

fixes #2213

### DIFF
--- a/Website/DesktopModules/Admin/Security/Register.ascx.cs
+++ b/Website/DesktopModules/Admin/Security/Register.ascx.cs
@@ -569,7 +569,7 @@ namespace DotNetNuke.Modules.Admin.Users
 
 		    if (_IsValid)
 		    {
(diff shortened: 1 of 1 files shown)


Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -569,7 +569,7 @@ private bool Validate()
 
 		    if (_IsValid)
 		    {
-                var name = User.Username;
+                var name = User.Username ?? User.Email;
```
Mohtshum: Due to option (Use  Email as Username) is enabled

PR #2237, opened 2018-08-16
Fixed issue where the browser covered the region and country dropdowns

### LINKED ISSUE #2146: Broswer form fill suggestions hide the country and region suggestions
## Description 
When trying to enter a country and region, the browser may offer suggestions from previous addresses, this hides the Dnn suggested countries and regions and without dismissing the browser behaviour and not noticing the issue, it is not possible to save the value.

## Steps to reproduce
- Use some saved auto-fill addresses in Chrome
- Create a Dnn user
- Try to edit the user profile region and country
- Notice the suggestions are hidden by the browser suggestions menu

## Current result
Dnn suggestions are hidden by browser suggestions

## Expected result
Setting the fields attribute autocomplete="off" would probably bypass this behaviour and show only the Dnn suggestions.

## Affected version

* [x] 9.2
* [x] 9.1.1
* [?] 9.1
* [?] 9.0

## Affected browser

* [x] Chrome
* [?] Firefox
* [?] Safari
* [?] Internet Explorer
* [?] Edge

valadas:
Closes #2146 

## Summary
Switched the country and region fields so that the country is first. This field populated the region dropdown so it needs to be filled first in order to see a dropdown of the regions available in that country. Without this, the entered region disappears when the user selects a country after the region.

Implemented a workaround to prevent browsers to display their suggestions above the Dnn built-in ones. Due to a bug in chrome autocomplete="off" does not work, to really make sure that the browser does not try to autocomplete, we need a random value in the autocomplete parameter so that the browser had no fields to suggest in memory.

## How to test
- make sure to use chrome latest release, and in the browser settings make sure you have a couple of form fill records with addresses.

Before this fix:
- Try to enter a region
- Enter a country that has regions available (US or Canada should do)
- Notice the browser covers the Dnn dropdown
- Notice the typed region disappears for a dropdown with nothing selected after selecting a country.

After this fix:
- Add a new website and notice after the website creation, the profile editor now has the country above the region and the browser no longer suggest regions and countries on those fields.
- For existing users, I thought it could break custom profile forms to force the country above the region (them may not even use both). So for those users I would add in the release notes that they can switch the order of these fields to prevent the problem.

### DIFF
--- a/DNN Platform/Website/Templates/Blank Website.template
+++ b/DNN Platform/Website/Templates/Blank Website.template
@@ -156,15 +156,15 @@
     </profiledefinition>
     <profiledefinition>
       <propertycategory>Location</propertycategory>
-      <propertyname>Region</propertyname>
-      <datatype>Region</datatype>
+      <propertyname>Country</propertyname>
+      <datatype>Country</datatype>
       <length>0</length>
       <defaultvisibility>2</defaultvisibility>
     </profiledefinition>
     <profiledefinition>
       <propertycategory>Location</propertycategory>
-      <propertyname>Country</propertyname>
-      <datatype>Country</datatype>
+      <propertyname>Region</propertyname>
+      <datatype>Region</datatype>
       <length>0</length>
       <defaultvisibility>2</defaultvisibility>
     </profiledefinition>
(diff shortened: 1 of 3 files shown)


donker:
I agree that having region first while controlled by country is unpractical. Even though it runs against logic that you go from specific (street) to general (country) and then back to specific (region).

PR #2241, opened 2018-08-16
DNN-21554 Orphaned Data in aspnet_Users & aspnet_Membership Tables

mikebigun:
### Problem: 
In some occasions, when user deleted permanently, it is not removed from aspnet_Users and aspnet_Membership tables. The impact of this orphaned data is that we can not create another user with the same username.

### Rootcause: 
Membership deletion is possible if user does not belong to any portal. Cached user entry prevents us to check actual data from data store. Non empty value returned from GetUserByUserName even if user is already removed from all portals.

### Solution: 
Making permanent deletion we should not be care of the cache. Need to make sure we send request to data store and obtaining actual data.

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -1161,7 +1161,7 @@ namespace DotNetNuke.Security.Membership
         
         /// -----------------------------------------------------------------------------
         /// <summary>
-        /// GetUserByUserName retrieves a User from the DataStore
+        /// GetUserByUserName retrieves a User from the DataStore. Supports user caching in memory cache.
         /// </summary>
         /// <remarks>
         /// </remarks>
@@ -1176,10 +1176,7 @@ namespace DotNetNuke.Security.Membership
                     DataCache.UserCacheTimeOut, DataCache.UserCachePriority),
                 _ =>
                 {
-                    using (var dr = _dataProvider.GetUserByUsername(portalId, username))
-                    {
-                        return FillUserInfo(portalId, dr, true);
-                    }
+                    return GetUserByUserNameFromDataStore(portalId, username);
                 });
             return objUserInfo;
         }
@@ -1580,7 +1577,7 @@ namespace DotNetNuke.Security.Membership
                 _dataProvider.RemoveUser(user.UserID, user.PortalID);
 
                 //Prior to removing membership, ensure user is not present in any other portal
-                UserInfo otherUser = GetUserByUserName(Null.NullInteger, user.Username);
+                UserInfo otherUser = GetUserByUserNameFromDataStore(Null.NullInteger, user.Username);
                 if (otherUser == null)
                 {
                     DeleteMembershipUser(user);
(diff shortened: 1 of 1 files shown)


sleupold:
we might include a SQL statement into the upgrade script to batch remove orphaned data from the aspnet tables

mikebigun:
@sleupold @tpluscode 

1. No. Foreign key does not have 'delete on cascade' option enabled for the aspnet tables. ASP.NET membership tables are created directly from [InstallCommon.sql](https://github.com/dnnsoftware/Dnn.Platform/blob/c8e04ee1beae9394daa440710eaaa9cc2fd2bf81/Website/Providers/DataProviders/SqlDataProvider/InstallCommon.sql#L95) and [InstallMembership.sql](https://github.com/dnnsoftware/Dnn.Platform/blob/c8e04ee1beae9394daa440710eaaa9cc2fd2bf81/Website/Providers/DataProviders/SqlDataProvider/InstallMembership.sql#L69). I would prefer to not modify existent schema to make one time orphaned data deletion. Should I alter aspnet tables for that?
2. Agree. Will be corrected.
3. Agree. Will be corrected.
4. We can manage this as stored procedure (just to be able to pass application name as parameter to satisfy point 2). Still I have concern in adding orphaned data deletion to deleteuser sp. It has a single responsibility. I would avoid the mess by keeping orphaned users deletion as a standalone query or stored procedure. Also, periodic deletion will just hide the problem.  One run on upgrade will clear old orphaned entries and the cache fix will prevent recurrence.
5. I will avoid temp tables.

Website/Providers/DataProviders/SqlDataProvider/09.02.02.SqlDataProvider
```diff
@@ -76,32 +76,56 @@ UPDATE {databaseOwner}{objectQualifier}Packages SET Version = @version WHERE Nam
...
+			DELETE u FROM dbo.aspnet_Users u
+			INNER JOIN orphaned_aspnet_users o ON u.UserId = o.UserId
+		END TRY
+		BEGIN CATCH
+			SELECT ERROR_MESSAGE() AS ErrorMessage
+			IF @@TRANCOUNT > 0  
+				ROLLBACK TRANSACTION
+		END CATCH
+	IF @@TRANCOUNT > 0 
+	    COMMIT TRANSACTION
```
tpluscode: Hm, if there were no errors, I think you should always commit. Otherwise won't the transaction be rolled back automatically when connection closes?

mikebigun: @tpluscode 
The construction is following:
BEGIN TRANSACTION - TRY/CATCH - COMMIT
I'm always commit if there are no errors. If errors, transaction rolled back, so nothing to commit at the end.
Removing rollback command or doing commit in a try block will not change the essence of current implementation. All commands are called explicitly, so we know what happens on every step.

tpluscode: You're right. I misunderstood what I was looking at. Thanks

mikebigun:
@sleupold  

Script has been moved to 9.3.0. 

I did a very quick test using my local machine.

- ~320K users, ~10K orphaned users = ~4 sec
- ~1M users, ~30K orphaned users = ~21 sec

mikebigun:
Hi @sleupold 

Is there any conclusion on this?

sleupold:
for a one time run, it would have been sufficient to perform the statement in the sproc inside the upgrade script. but might be helpful in the future (I'll include something similar in my TurboScripts as well :) )

PR #2247, opened 2018-08-17
DNN-22612: update the tab info in cache object instead of clear whole cache during workflow process.

### LINKED ISSUE #2246: whole portal cache will be cleared if tab's workflow triggered and impact the performance
Reproduce Steps:
1. create a sn evoqite with thousands pages; 
2. add a new page, set its workflow to "Content Approval"; 
3. browse to the page within edit mode, refresh page several times to confirm it load quickly; 
4. add a new module into page. 
5. refresh page.

Current Behavior:
the page will load slowly first time after add a module to page, or submit/publish page which will trigger page workflow changes.

Expected Behavior:
page should always load quickly.

zyhfish:
Fix #2246 .

### DIFF
--- a/DNN Platform/Library/Entities/Tabs/TabController.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabController.cs
@@ -1002,6 +1002,16 @@ namespace DotNetNuke.Entities.Tabs
             CacheController.FlushPageIndexFromCache();
         }
 
+        public void RefreshCache(int portalId, int tabId)
+        {
+            var portalTabs = GetTabsByPortal(portalId);
+            if (portalTabs.WithTabId(tabId) != null)
+            {
+                var updateTab = GetTab(tabId, portalId, true);
+                portalTabs.RefreshCache(updateTab);
+            }
+        }
+
         /// <summary>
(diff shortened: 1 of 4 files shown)


DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -1002,6 +1002,16 @@ public void ClearCache(int portalId)
             CacheController.FlushPageIndexFromCache();
         }
 
+        public void RefreshCache(int portalId, int tabId)
+        {
+            var portalTabs = GetTabsByPortal(portalId);
+            if (portalTabs.WithTabId(tabId) != null)
+            {
+                var updateTab = GetTab(tabId, portalId, true);
+                portalTabs.RefreshCache(updateTab);
```
tpluscode: I'm not convinced about `RefreshCache` of `TabCollection`.

I think I would rename it to be only removal and add the tab again here, inside `TabController`

zyhfish: Hi @tpluscode , TabCollection have several internal list and dictionaries, it can't access outside, so we have to provide this method in TabCollection.

PR #2251, opened 2018-08-17
Update User.ascx.cs fix null reference error using email as username setting

thabaum:
Unable to register using email as username.  This checks to see if the portal setting is false when it should be checking if it is true.

### DIFF
--- a/Website/DesktopModules/Admin/Security/User.ascx.cs
+++ b/Website/DesktopModules/Admin/Security/User.ascx.cs
@@ -1,6 +1,6 @@
 #region Copyright
 // 
-// DotNetNuke® - http://www.dotnetnuke.com
+// DotNetNukeÂ® - http://www.dotnetnuke.com
 // Copyright (c) 2002-2018
 // by DotNetNuke Corporation
 // 
@@ -267,7 +267,7 @@ namespace DotNetNuke.Modules.Admin.Users
             var user = User;
 
             // make sure username is set in UseEmailAsUserName" mode
-            if (PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false))
+            if (PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, true))
             {
                 user.Username = User.Email;
                 User.Username = User.Email;


thabaum:
My first pull request... how exciting :)  Thank you Mitchel for taking time helping me understand the issue.  Should build as I just changed a value from false to true.  I have not tested it as I am still getting familiar with a new IDE environment and I need to set it up more still and practice... I hope this works as I am desperate for a fix to be applied soon.

bdukes:
@thabaum I don't think this is going to address your issue.  That third argument to [`PortalController.GetPortalSettingAsBoolean`](https://www.dnnsoftware.com/dnn-api/?topic=html/6c36273c-7d30-8f2c-38d4-e48c59af158d.htm) is the default value when the setting isn't set, not the value that's being compared to.

thabaum:
OK Let me think this through:
current code asks this question
if (PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false))

So if in my DB Registration_UseEmailAsUsername is set to true would this expression fail to execute the next lines of code in the if statement?

bdukes:
If the value in the database is _true_, `PortalController.GetPortalSettingAsBoolean` will return `true`, and you'll enter the `if` block.  If it's _false_ in database, you'll skip the `if` block.  If it's empty in the database, it'll return `false` and skip the `if` block (your PR changes this to enter the `if` block when it's empty in the database)

thabaum:
well my statement is TRUE and it is not entering the block when it should, right?

thabaum:
Maybe the issue is in the Register module UI code?  When I hit the register button it does nothing but throw an exception, even on a fresh installation of DNN 9.2.1.

Can you link me to registration UI code page.  The order of fields displayed is not correct on my installations as Mitchel pointed out as well and I just feel it is related to something not being passed to set the email as username.

DISPLAY NAME
PASSWORD
RE-ENTER PASSWORD
USERNAME
EMAIL ADDRESS

is the order being displayed maybe this has something to do with it.

When set to username as email I get this exception on all installations even a fresh one.

Message:Object reference not set to an instance of an object.

StackTrace:

InnerMessage:Object reference not set to an instance of an object.

InnerStackTrace:

at DotNetNuke.Entities.Users.UserController.IsValidUserName(String userName) at DotNetNuke.Modules.Admin.Users.Register.Validate() at DotNetNuke.Modules.Admin.Users.Register.registerButton_Click(Object sender, EventArgs e) at System.Web.UI.WebControls.LinkButton.OnClick(EventArgs e) at System.Web.UI.WebControls.LinkButton.RaisePostBackEvent(String eventArgument) at System.Web.UI.Page.ProcessRequestMain(Boolean includeStagesBeforeAsyncPoint, Boolean includeStagesAfterAsyncPoint)

So it looks like it is checking if hte username is valid but it is not getting one to check?

bdukes:
@thabaum That code is in https://github.com/dnnsoftware/Dnn.Platform/blob/5be4d592ec1803dca5ac81ce7d9ec346decf0b1f/Website/DesktopModules/Admin/Security/Register.ascx.cs#L572-L583, but it looks like the culprit is probably in the `User` property from the base class.  It's initialized in https://github.com/dnnsoftware/Dnn.Platform/blob/5be4d592ec1803dca5ac81ce7d9ec346decf0b1f/DNN%20Platform/Library/Entities/Modules/UserModuleBase.cs#L303, but maybe the values are supposed to be coming via the `userForm` control? https://github.com/dnnsoftware/Dnn.Platform/blob/5be4d592ec1803dca5ac81ce7d9ec346decf0b1f/Website/DesktopModules/Admin/Security/Register.ascx.cs#L337

thabaum:
OK thanks, I am still learning where everything is this saves a lot of time for me to poke around here see if I can figure something out.  I will probably just learn 1000 ways not to do something but I need to...

thabaum:
Line 583 var valid = UserController.Instance.IsValidUserName(name);
when using this function from https://github.com/dnnsoftware/Dnn.Platform/blob/5be4d592ec1803dca5ac81ce7d9ec346decf0b1f/Website/DesktopModules/Admin/Security/Register.ascx.cs

Is the statement above calling the function below and if so should the variable used actually be named userName for it to function right or can you call it whatever you like? Im trying to break the rust off my brain here and get use to .NET along with DNN classes so help here is much appreciated.

line 564 public bool IsValidUserName(string userName)

From

https://github.com/dnnsoftware/Dnn.Platform/blob/5be4d592ec1803dca5ac81ce7d9ec346decf0b1f/DNN%20Platform/Library/Entities/Users/UserController.cs

bdukes:
You can call the variable whatever you'd like

PR #2268, opened 2018-08-23
DNN-23121: sanitize the properties.

zyhfish:


zyhfish:
we can't save these into db, as if some code read it and use decode in future, the issue will occur again. we should remove them for safety.

mitchelsellers:
@zyhfish In that case sanitization into the DB done via Html.Encode would be preferred.

zyhfish:
Hi @mitchelsellers , generally we shouldn't save those scripts, as its just like a bomb, we don't know who will read and how to display it. we should always remove them before saving into db.

tpluscode:
I agree. HTML encoding should be preferred instead of removing content

zyhfish:
code updated.

PR #2270, opened 2018-08-23
Managing version

### LINKED ISSUE #2269: VSTS is not producing correct versions numbers
## Description of bug

Currently we have release version 9.2.2 on a branch and development brach is the future 9.3.

This somewhat related to #2158

## Current result
 
The CI assigns next in sequence number for 9.2.2 instead. For example https://dotnetfoundation.visualstudio.com/DNN/_build/results?buildId=4307

I understand that the base version is meant to be configured on VSTS @ohine? Or is it only take from the [cake script](https://github.com/dnnsoftware/Dnn.Platform/blob/development/build.cake#L10)?  

## Expected result

Build number should be automatically calculated based on branch/tag being built. At the time of writing it would be:

```
development   -> 9.3.0.*-*
release/9.2.2 -> 9.2.2.*-*
v9.2.2-rc0    -> 9.2.2-rc0
v9.2.1        -> 9.2.1
```

tpluscode:
fixes #2269 

## Summary

In this PR I remove any maintenance actions required to manage the build version. All will be calculated during build based on the branch

More details in source code comments


gitversion.yml
```diff
@@ -0,0 +1,7 @@
+assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{CommitsSinceVersionSource}'
+branches:
+  development:
+    regex: development      
+    tag: ''
+    is-mainline: true
+    source-branches: []
```
tpluscode: Versions will be applied to `SolutioInfo.cs` as follows

```c#
[assembly: AssemblyVersion("9.2.1.0")]
[assembly: AssemblyInformationalVersion("9.2.1-managing-version.1+670.Branch.managing-version.Sha.52eace732928da5053e2f597bf5c42322405dd4a")]
[assembly: AssemblyFileVersion("9.2.1.670")]
```

It's important to note that `[AssemblyVersion]` will be the same for all builds of given release. This way it will be easier to debug with local builds. Without assembly loader complaining about mismatched versions.

The `assembly-file-versioning-format` pattern above is used to feed the file version.

The informational version generated by gitversion includes extensive metadata.

tpluscode:
There is nothing needed to build and run unit tests. The assembly version will simply default to `0.0.0.0`.

To get packages locally it's best to use cake scripts. When building through VS it's possible to run `build.ps1 -Target UpdateDnnManifests` to set the numbers (maybe that target should be renamed `UpdateVersions`?

However, let's still take an extra moment to consider the implication if those changes. Gitversion is sometimes a bit mercurial and requires strictly following the merges. For this reason (I think, unclosed `release/9.2.1` actually) it seems to insist on calling `development` 9.2.1 even though there is a newer branch already. Hopefully this would be resolved once 9.2.2 is merged. For that reason I'm adding a `next-version` property to the config file. Worst case, it would be the source of truth which we'd have to edit once in a while

Second, I'd like to address the fourth, revision part of the version. I discussed with Oliver that we may want it to keeps similar scheme to what we have always had. The difference is though that this number is not strictly a sequence and I have seen issues where it gets unexpectedly reset.

DNN Platform/Library/Application/Application.cs
```diff
@@ -230,7 +231,9 @@ public virtual Version Version
         {
             get
             {
-                return Assembly.GetExecutingAssembly().GetName().Version;
+                var assemblyLocation = Assembly.GetExecutingAssembly().Location;
+                var fileVersion = FileVersionInfo.GetVersionInfo(assemblyLocation).FileVersion;
+                return new Version(fileVersion);
```
tpluscode: I realised that the version shown in Persona Bar is taken from `[AssemblyInfo]` so this will break with my changes (will always show `9.x.y (0)`). To fix I changed

PR #2271, opened 2018-08-27
Fix Newtonsoft.Json package install during DNN Upgrade.

### LINKED ISSUE #2121: NewtonSoft
The upgrade package does not include the newtonsoft.json.dll library which causes the upgrade to break if you do not already have v10.0.0.0 installed of the lib. The main install package contains it though

dimarobert:
Fixes #2121 

## Summary

The Config component in the manifest of Newtonsoft.Json package is not needed. The binding redirect is ensured by the AssemblyInstaller through the [AddOrUpdateBindingRedirect()](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs#L194) method call inside the InstallFile() method.
This also broke the DNN upgrade from a version prior to 9.2.0 when the web.config contained additional `<codebase>` assembly bindings before the one with the `<bindingRedirect>`.
Moreover besides that, it also updated the web.config with the binding redirect twice, in both scenarios (clean install and DNN upgrade).

### DIFF
--- a/DNN Platform/Components/Newtonsoft/DotNetNuke.Newtonsoft.Json.dnn
+++ b/DNN Platform/Components/Newtonsoft/DotNetNuke.Newtonsoft.Json.dnn
@@ -24,25 +24,6 @@
             </assembly>
           </assemblies>
         </component>
-        <component type="Config">
-          <config>
-            <configFile>web.config</configFile>
-            <install>
-              <configuration>
-                <nodes>
-                  <node path="/configuration/runtime/ab:assemblyBinding/ab:dependentAssembly[ab:assemblyIdentity/@name='Newtonsoft.Json']" action="update" targetpath="/configuration/runtime/ab:assemblyBinding/ab:dependentAssembly[ab:assemblyIdentity/@name='Newtonsoft.Json']/ab:bindingRedirect" collision="save"  nameSpace="urn:schemas-microsoft-com:asm.v1" nameSpacePrefix="ab">
-                    <bindingRedirect oldVersion="0.0.0.0-32767.32767.32767.32767" newVersion="10.0.0.0" />
-                  </node>
-                </nodes>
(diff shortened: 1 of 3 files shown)


bdukes:
@dimarobert I've pushed a change to your PR branch so that the code which updates binding redirects won't incorrectly target the `codeBase` elements, can you try your upgrade scenarios again and let me know if that addresses the issue regardless of the order of the elements in the `web.config`?  Thanks!

dimarobert:
@bdukes I ran the install again with the reversed order binding redirects and the result was:
``` XML
<dependentAssembly>
        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" />
        <codeBase version="8.0.0.0" href="bin\NewtonSoft.Json-8.0.0.0\Newtonsoft.Json.dll" />
      </dependentAssembly>
      <!--Upgraded by DotNetNuke.Newtonsoft.Json version 10.0.3 - Date: 8/31/2018 9:14:32 AM-->
      <!--<dependentAssembly xmlns="urn:schemas-microsoft-com:asm.v1"><assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" /><bindingRedirect oldVersion="0.0.0.0-7.0.0.0" newVersion="7.0.0.0" /></dependentAssembly>-->
      <dependentAssembly>
        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" />
        <bindingRedirect oldVersion="0.0.0.0-32767.32767.32767.32767" newVersion="10.0.0.0" />
      </dependentAssembly>
```

So now the result is the same in all three scenarios. I think we are good to go.

PR #2274, opened 2018-08-28
Use master portalID set by GetEffectivePortalId to which user is belong

Mohtshum:
Fix [#50](https://github.com/dnnsoftware/Dnn.AdminExperience/issues/50)

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -1045,8 +1045,9 @@ namespace DotNetNuke.Entities.Users
                     //send email notification to portal administrator that the user was removed from the portal
                     SendDeleteEmailNotifications(user, portalSettings);
                 }
-                DataCache.ClearPortalCache(portalId, false);
-                DataCache.ClearUserCache(portalId, user.Username);
+
+                DataCache.ClearPortalCache(user.PortalID, false);
+                DataCache.ClearUserCache(user.PortalID, user.Username);
 
                 //also clear current portal's cache if the user is a host user
                 if (portalSettings.PortalId != portalId)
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -1045,8 +1045,9 @@ public static bool DeleteUser(ref UserInfo user, bool notify, bool deleteAdmin)
                     //send email notification to portal administrator that the user was removed from the portal
                     SendDeleteEmailNotifications(user, portalSettings);
                 }
-                DataCache.ClearPortalCache(portalId, false);
-                DataCache.ClearUserCache(portalId, user.Username);
+
+                DataCache.ClearPortalCache(user.PortalID, false);
```
mitchelsellers: As we are making changes in this section, is there a reason that we clear the entire portal cache when we delete a user?  I thought we had a recent change to avoid this. @dnnsoftware/tag

tpluscode: Indeed. You are referring to #2175. We could apply an analogous change here

Mohtshum: Thnx @tpluscode. Let me check this, if it works then would push those changes

Mohtshum:
Working !!! .Changes push as suggested by @mitchelsellers  👍

ohine:
Can we get this and #2175 refactored to use a shared function? I really don’t want to start clearing one off cache keys all over the application. @tpluscode @mitchelsellers @mohtshum

ohine:
@Mohtshum Thanks for doing that! It looks great. I just approved #2285

PR #2275, opened 2018-08-28
Fixed issue Redirect After Login failed for culture specific portalalias

### LINKED ISSUE #2267: Redirect after login
## Description of bug
When having more than one language activated the redirect after login doesn't work

## Steps to reproduce
1. Go to security;
2. Set a redirect after login page;
3. Go to language;
4. Enable a second language, for example dutch;
5. login with a registered user;
6. You wil see it doesn't redirect to the page but goes to the homepage;
7. Deactivate the second language;
8. You will see the redirect does work again

## Current result
After login with a registered user it doesn't redirect to the redirect page

## Expected result
It should redirect to the page set under security - redirect after login page

## Error log
no errors in the log

## Affected version
* [x] 9.2.1

## Affected browser
All browsers:
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

skamphuis:
Fixes #2267 

Issue is reported in 9.2.1, but also occurs in 9.2.2 RC.
## Summary
Added` .TrimEnd('/')` to the following line, because a portalalias for a specific culture will not have a trailing /, while a returnurl will. Thus the `.Equals` did fail and the redirect to the configured tab did not occur.
`                var isDefaultPage = redirectURL == "/"
                    || (alias.Contains("/") && redirectURL.TrimEnd('/').Equals(alias.Substring(alias.IndexOf("/", comparison)), comparison));
`
## How to test
The issue was easily reproducible from the issue description.

### DIFF
--- a/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
+++ b/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
@@ -200,8 +200,9 @@ namespace DotNetNuke.Modules.Admin.Authentication
 
                 var alias = PortalAlias.HTTPAlias;
                 var comparison = StringComparison.InvariantCultureIgnoreCase;
+                // we need .TrimEnd('/') because a portlalias for a specific culture will not have a trailing /, while a returnurl will.
                 var isDefaultPage = redirectURL == "/"
-                    || (alias.Contains("/") && redirectURL.Equals(alias.Substring(alias.IndexOf("/", comparison)), comparison));
+                    || (alias.Contains("/") && redirectURL.TrimEnd('/').Equals(alias.Substring(alias.IndexOf("/", comparison)), comparison));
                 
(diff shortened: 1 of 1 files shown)


sleupold:
shouldn't all Aliases have a trailing slash '/' - IMO this would be a more consistent solution

skamphuis:
@sleupold That's a different discussion, imho. People built their modules on the way it is now, so changing that would have a massive impact.

PortalAliasses don't have a trailing slash, and never had. This issue was "created" in DNN 9(.2?). It worked before with the same Portaliasses, so it should be fixed. I think this issue will be a pain for a lot of multi lingual websites.

valadas:
@dnnsoftware/tag Please review when possible.

ohine:
@dnnsoftware/tag Please review

PR #2277, opened 2018-08-30
Fixed: dragdrop fails for modules with friendlyname starting with number

### LINKED ISSUE #2276: Module DragDrop fails when Friendlyname starts with number
## Description of bug
As of DNN 8 (8.0.4 tested), DragDrop of a module on a page is not working anymore for modules that have a friendlyname starting with a number.
There's also a [Jira ticket](https://dnntracker.atlassian.net/browse/DNN-8944) on this error.

## Steps to reproduce
**Please note:** 2sic have made a workaround for this error, so 2sexy content works anyway.

1. Install a module with a friendlyname starting with a number, like [40Figners Filecuumcleaner](https://www.40fingers.net/Products/DNN-Filecuumcleaner)
2. Put the module on a page
3. Try to drag the module to another pane.
4. Have a look at the network tab in your browser's console to notice 40 being used as moduleid in the call
5. The page refreshes with the module back on its original spot.

## Current result
After drag-drop of a module, the page refreshes and the module is back in it's prvious spot.

## Expected result
Obviously one would expect the module to be on the new spot after the page refreshes.

## Screenshots
none

## Error log
From the network tab in the browser's console:

Form data |  
-- | --
**ModuleId | 40**
ModuleOrder | 2
Pane | TopPaneFullWidth
TabId | 1313

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0
* [x] 8.0.4

## Affected browser
Not depending on browser type or version

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

skamphuis:
Fixes #2276

## Summary
In dnn.DragDrop.js I changed the regular expression that's used to get the moduleId from a class attribute.
It used to work in cases like this:
`class="DnnModule DnnModule-OpenContent DnnModule-4084"`.

but would fail in cases like this:
`class="DnnModule DnnModule-40FingersWebshopSearch DnnModule-3767"`

With the fix, the latter also works.

This is the same solution @iJungleboy proposed in a previous Jira ticket mentioned in the issue. I tried and tested the solution on DNN 8.0.4, 9.2.1, 9.2.2RC.

### DIFF
--- a/Website/Resources/Shared/scripts/dnn.dragDrop.js
+++ b/Website/Resources/Shared/scripts/dnn.dragDrop.js
@@ -38,7 +38,7 @@
         var $module;
 
         var getModuleId = function ($mod) {
-            var result = $mod.attr("class").match(/DnnModule-([0-9]+)(?:\W|$)/);
+            var result = $mod.attr("class").match(/\bDnnModule-([0-9]+)\b/);
             return (result && result.length === 2) ? result[1] : null;
         };
 
--- a/Website/Resources/Shared/scripts/dnn.dragDrop.js
+++ b/Website/Resources/Shared/scripts/dnn.dragDrop.js
@@ -38,7 +38,7 @@
         var $module;
 
         var getModuleId = function ($mod) {
-            var result = $mod.attr("class").match(/DnnModule-([0-9]+)/);
+            var result = $mod.attr("class").match(/DnnModule-([0-9]+)(?:\W|$)/);
             return (result && result.length === 2) ? result[1] : null;
         };
 


Website/Resources/Shared/scripts/dnn.dragDrop.js
```diff
@@ -38,7 +38,7 @@
         var $module;
 
         var getModuleId = function ($mod) {
-            var result = $mod.attr("class").match(/DnnModule-([0-9]+)/);
+            var result = $mod.attr("class").match(/DnnModule-([0-9]+)(?:\W|$)/);
```
bdukes: to solve @iJungleboy's edge cases, we can use [`\b`](https://www.regular-expressions.info/wordboundaries.html) to match the word boundary, i.e. `/\bDnnModule-([0-9]+)\b/`

You can see a test at https://jsbin.com/gocokiw/1/edit?js,output

skamphuis: @bdukes Thanks. I tested your proposed solution. Learned something new again :-). I committed the change.

iJungleboy: @bdukes amazing - i've been regexing for forever and never knew about `\b` - very cool! 👍

PR #2281, opened 2018-08-30
Implement Parent by Url

### LINKED ISSUE #2280: Import/Export Circular Dependency in pages cause pages not to be imported
## Description of bug

Workaround: Set the parent pages as normal pages during the export process, and set them as references to existing pages once the Import is completed (doable if this is the case for less than 10 pages) 

## Steps to reproduce

On Source:
1. Create 2 pages (these are gonna be the child pages).
2. Create 1 page (this is the parent page with a large tabid)
3. Nest/Move the 2 pages from step 1 into the parent page created.
4. Go to the page settings for the parent page, and change the Page type to existing, make it point to one the child pages.
4. Create a Full export of the site. You can choose to exclude extensions/Content Library and assets since these are not needed.

On Target:
3. Check whether the pages are created.

## Current result
The pages are not imported since the child pages require the parent page to be added first; nevertheless, the parent page is a reference to the child page which has not been imported yet so all pages fail to import. 

## Expected result
All pages should be imported to target site no matter having cyclic dependency as far as fulfilling parent tab condition

## Affected version
* [x] 9.2.2

Mohtshum:
Fixes #2280 

## Summary
Implement parent tab case by URL

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -168,7 +168,7 @@ namespace Dnn.ExportImport.Components.Services
                 _totals.LastProcessedId = index++;
                 CheckPoint.StageData = JsonConvert.SerializeObject(_totals);
             }
-
+            _searchedParentTabs.Clear();
             ReportImportTotals();
         }
 
@@ -182,7 +182,7 @@ namespace Dnn.ExportImport.Components.Services
                   && (t.CultureCode ?? "") == (otherTab.CultureCode ?? ""));
 
             var isParentPresent = IsParentTabPresentInExport(otherTab, exportedTabs, localTabs);
-            _searchedParentTabs.Clear();
+            
             if (localTab != null)
             {
                 localTab.TabSettings.Remove("TabImported");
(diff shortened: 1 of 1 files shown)


DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -1876,12 +1888,28 @@ private void SetPartialImportSettings(ExportTab exportedTab, TabInfo localTab)
...
 
-        private void AddIntoPartialImportedTabs(ExportTab tabToExport)
+        private void CheckForPartialImportedTabs(ExportTab tabToExport)
         {
             var exportTabParentId = tabToExport.ParentId.GetValueOrDefault(Null.NullInteger);
-            if (!_partialImportedTabs.ContainsKey(tabToExport.LocalizedVersionGuid) && exportTabParentId != -1)
+
+            if (exportTabParentId == -1)
+            {
+                if (int.TryParse(tabToExport.Url, out exportTabParentId))
```
tpluscode: This is a bit weird. Why parse `Url` into string?

Mohtshum: It is by designed when URL property of page is type of number then that number(tabId) represent the parent Id of the page.

tpluscode:
Will have QA test various export/import scenarios @mitchelsellers

PR #2290, opened 2018-09-05
Portal Creation Overwrites Portal Folder

### LINKED ISSUE #2292: Portal Creation Overwrites Portal Folder
## Steps to reproduce

1. Login as a host user in your site. 
2. Go to `Manage > Sites`, Add new site. 
3. Give it any name, but for the Directory section use `Portals/2`, create the site. 
4. Go to the new created site, and under assets create a folder, and add an asset. 
5. Verify that these exists in the file system 
6. Create a new site, but this time do not modify the Directory, it will create it using `Portals/[Portal ID]` 
7. Visit the Assets for the first portal created and resync the file system.

## Current result

All files/folders created are gone. This happens because the Portal first created was PortalID 1, but we created it so it uses a folder with the name '2'; therefore, once PortalID 2 was created it overwrote the folder entirely, and at this point both Portals share the same folder.

## Expected result

The user should get a warning if the Portal folder already exists which should stop users from overwriting it. 

This is a niche issue; however, it can cause the loss of data, if a user uses some odd enumeration for their folders.

## Affected versions

* [x] 9.2.1

mean2me:
## Summary
Avoided creation of portals using an existing home folder.

fixes #2292


DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -163,6 +163,12 @@ private static void CreateDefaultPortalRoles(int portalId, int administratorId,
             }
             string mappedHomeDirectory = String.Format(Globals.ApplicationMapPath + "\\" + homeDirectory + "\\").Replace("/", "\\");
 
+            if(Directory.Exists(mappedHomeDirectory))
```
tpluscode: Maybe this check should actually be done by reading `HomeDirectory` column from `Portals` table? @dnnsoftware/tag

tpluscode: Now that I think about it, I would even enforce uniqueness of the home directory. I think it makes no sense for multiple portals to share one and would have strange side-effects

tpluscode: On third thought, checking the directory does also make sense. 

Question is what should happen if no portal exists in the database using a given directory but it exists in filesystem. It should not happen but could be possible I expect. Would it be fine to overwrite then?

mitchelsellers: With regards to your third thought, the way this would happen is a deleted portal that didn't delete cleanly with a custom home directory

tpluscode:
I have to close this PR. We need those changes as a hot fix to v9.2.1. Will reopen once we have a correct branch

Issue #2234, opened 2018-08-14
VSTS Unit Tests Failing when using Sql CE as Data Provider


mtrutledge:
## Description 
The current VSTS build fails when testing the DataContext and using the PetaPoco connection strings because the VSTS sytem cannot find the System.Data.SqlServerCe.4.0 data provider.

## Steps to reproduce
Build with Unit Tests on in VSTS

## Current result
DataContext tests fail

## Expected result
All tests should pass

## Affected version
9.2.2.66

* [ ] 9.2.1
* [ ] 9.2.0
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

ohine:
#2291 fixed a few more of these.

valadas:
I am closing this assuming it is either solved because of the two merged PRs or invalid since we now use Azure Devops, if I am mistaken, please reopen

PR #2291, opened 2018-09-05
CI: add cake task to create the localdb database with basic test data req…

ohine:
Initial script to create a test database to help run unit tests in the solution. 

To test you can run the following command `.\build.ps1 -Target CreateDatabase`

This is also not on the default build, I'm going to manually trigger a build which will run the `BuildWithDatabase` task and review the logs from VSTS.


ohine:
#2234

PR #2293, opened 2018-09-06
DNN-23121: encode the value.

zyhfish:


tpluscode:
Thie PR replaces #2268 which was based on 9.3 instead of 9.2.2

PR #2296, opened 2018-09-07
After export with Content Localization site language flags disappears from pages

### LINKED ISSUE #2294: After export with Content Localization site language flags disappears from pages
## Description of bug

After export with Content Localization site language flags disappears from pages

## Steps to reproduce

1. Install a site 
2. Create a page with an html module 
3. Enable 3 languages: en-US, fr-FR, it-IT 
4. Enable content localization 
5. Enable and mark all pages as translatable 
6. User can see flags on the top left and change page language using those 
7. Perform a full export of the site 
8. Create a new site# 
9. Enable 3 languages: en-US, fr-FR, it-IT 
10. Enable content localization 
11. Enable and mark all pages as translatable 
12. Import export package

## Current result

flags gone and user cannot longer see translated pages 
as workaround admin must disable and re-enabled content localization making again all pages as translatable

## Expected result

flags still works, user can see the new page added and navigate to that in different languages

## Additional context

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

ESW issue DNN-20856

mean2me:
## Summary
Fixed by ensuring that target site will activate and publish exported languages before to import data.

fixes #2294


DNN Platform/Modules/DnnExportImport/Components/Services/PortalExportService.cs
```diff
@@ -208,6 +209,15 @@ public override int GetImportTotal()
...
             var localLanguages = CBO.FillCollection<Locale>(DotNetNuke.Data.DataProvider.Instance().GetLanguages());
+
+            // ensure exported languages are enabled in target site
+            foreach(var lang in portalLanguages)
+            {
+
+                LocaleController.Instance.ActivateLanguage(portalId, lang.CultureCode, true);
+                LocaleController.Instance.PublishLanguage(portalId, lang.CultureCode, true);
+            }
+
```
mitchelsellers: I believe that this will fail, if the language isn't actually installed to the target portal.  In the below loop you are going through and looking for local portal support, I would prefer to activate & publish in that loop, after it was validated that the value was created?  Or am I missing something?

mean2me: @mitchelsellers localPortalLanguages is populated with installed languages, so looping through will only activate existing languages.

daguiler: @mean2me, did you test the scenario described by @mitchelsellers? Can you confirm it actually works?

DNN Platform/Modules/DnnExportImport/Components/Services/PortalExportService.cs
```diff
@@ -30,6 +30,7 @@
 using Dnn.ExportImport.Dto.Portal;
 using DotNetNuke.Services.Localization;
 using DataProvider = Dnn.ExportImport.Components.Providers.DataProvider;
+using DotNetNuke.Entities.Portals;
```
daguiler: Not needed

mitchelsellers:
@mean2me @daguiler Can we get an update on this one so we can either approve it, or close it?

DNN Platform/Modules/DnnExportImport/Components/Services/PortalExportService.cs
daguiler: @mean2me, if you look in the loop right below your change (which btw is essentially the same loop), you'll notice that the source language is validated to exist in the destination portal:
```c#
var localLanguageId = localLanguages.FirstOrDefault(x => x.Code == exportPortalLanguage.CultureCode)?.LanguageId;
var existingPortalLanguage = localPortalLanguages.FirstOrDefault(t => t.LanguageId == localLanguageId);
if (existingPortalLanguage != null)
{
    // now we know the language exists on the destination portal
}
```
Concern here is that `portalLanguages` is populated with source languages, not destination languages, and thus you should first check if the language exists, and only then activate/publish the language in the destination portal.
Otherwise you're just assuming that the language will exist.
Essentially, remove your loop and move the activation/publication lines to the existing loop, after above validation.
@mitchelsellers correct me if I'm wrong here please.

ohine:
@mitchelsellers @daguiler Do you guys want to finish up the review on this PR so we can merge for v9.3.0? Or should I update the issue to v9.4.0?

daguiler:
@mitchelsellers, @ohine, I've been testing this and, at this point, I don't feel comfortable merging this pull request. I was still able to reproduce the issue after applying these changes, so I think we need to keep working on the root cause.

mitchelsellers:
@daguiler Works for me.  I'm going to close this PR for now, once a proper solution is done, either re-open this or create a new one.

PR #2297, opened 2018-09-07
Bugfix: avoid creation of portals using an existing home folder.

### LINKED ISSUE #2292: Portal Creation Overwrites Portal Folder

mean2me:
fixes #2292

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -163,6 +163,12 @@ namespace DotNetNuke.Entities.Portals
             }
             string mappedHomeDirectory = String.Format(Globals.ApplicationMapPath + "\\" + homeDirectory + "\\").Replace("/", "\\");
 
+            if(Directory.Exists(mappedHomeDirectory))
+            {
+                message += string.Format(Localization.GetString("CreatePortalHomeFolderExists.Error"), homeDirectory);
+                throw new Exception(message);
+            }
+
             message += CreateProfileDefinitions(portalId, mergedTemplatePath);
 
             if (String.IsNullOrEmpty(message))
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -163,6 +163,12 @@ private static void CreateDefaultPortalRoles(int portalId, int administratorId,
             }
             string mappedHomeDirectory = String.Format(Globals.ApplicationMapPath + "\\" + homeDirectory + "\\").Replace("/", "\\");
 
+            if(Directory.Exists(mappedHomeDirectory))
+            {
+                message += string.Format(Localization.GetString("CreatePortalHomeFolderExists.Error"), homeDirectory);
+                throw new Exception(message);
```
tpluscode: I'm not convinced about a basic `Exception`. At least `InvalidOpertionException`? Or maybe some type DNN-specific?

tpluscode:
@mitchelsellers I'm not sure what was the conclusion from the last conversation on the other PR. Do you think that checking for folder existence would be enough?

PR #2329, opened 2018-09-22
Journal_GET SP Performance: Timing out when 1K+ friends.

mikebigun:
### Problem
When User has 1K+ relationships, dbo.[Journal_Get] stored procedure fails with timeout exception.

### Rootcause
Apart of getting journal item, stored procedure validates whether user has appropriate permissions for the particular journal when @SecurityCheck is true. In other case, stored procedure joins all permissions with the main selection. It produces as many records as permission check returns.

### Sample
1K+ relationships produces 1K+ permisions with dbo.[Journal_User_Permissions] (...) function.
Following construction returns single journal multiplied to 1K+ permissions when @SecurityCheck = 0. At the end, 1K+ same records returned.

```
SELECT j.*
FROM Journal j
INNER JOIN Journal_Security js ON js.JournalId = j.JournalId
INNER JOIN Journal_User_Permissions(...) t ON t.seckey = js.SecurityKey OR @SecurityCheck = 0
WHERE j.JournalId = @JournalId
```

### Solution

Single record is expected to be returned form the stored procedure. 
'Left join' will resolve above issue. Solution is to rewrite the query as following:

```
SELECT j.*
FROM Journal j
LEFT JOIN Journal_Security js ON js.JournalId = j.JournalId
LEFT JOIN Journal_User_Permissions(...) t ON t.seckey = js.SecurityKey
WHERE j.JournalId = @JournalId AND 
(@SecurityCheck = 0 OR (@SecurityCheck = 1 AND t.seckey IS NOT NULL))
```
To execute the query provided in PR takes a seconds with a huge amount of data. Tested on 2K relationships, 100K users and 100K journals.


sleupold:
just replacing inner joins with left joins usually doesn't improve performance, it is more likely the change from ON to WHERE causing the improvement, but this should be further checked. 

Note: just testing on one sample doesn't mean, it will be working the same in all conditions and on any version of SQL Server.

sleupold:
there is different logic to execute the select statement, depending whether parameter @SecurityCheck is 0 or 1, which will result in different query plans. I suggest splitting the procedure into separate branches, making it easier for the query optimizer to cache the plan, like I am doing it in my version of the sproc (I am using an inline function to encapsulate the security check, which makes it easier to test and read the code, but that is just a preference).

mikebigun:
I have tested every part of the query separately. As a standalone parts everything works pretty fast with a huge data.
Combining all in one makes the query heavy and causes exceptions.

The following negatively effects on performance:
1. CASE conditions 
2. Sub-queries

Multiplying above to a thousands of records causes timeouts. In our particular case LEFT join does the trick. We are sure we are dealing with a single record. 

In particular, customer have following data size (shared by customer):
Journal - 61k 
Users - 37k 
Journal_Data - 57k 
Journal_Security - 122k 
UserRelationships (Friends for that User) - 1k

Running original query on this size ~ more than 2 mins
Running modified query on this size ~ 2-3 sec

sleupold:
@mikebigun 
please be aware, you were testing it against a single data set - other large installations may have different characteristics, based how the site is used. And another version of SQL Server may use a different approach (query plan analyzer has been rewritten for SQL Server 2014 and is used depending on compatibility level. 
The bad performance of the original query is caused a) by an inefficient Function using temp tables which b) is called, even if not needed. (btw. be aware that function calls are not included in query plans).
To avoid a) I am using an IF @securityCheck = 1 THEN SELECT … (from all tables) ELSE SELECT … (without Permissions and function call)

sleupold:
btw, what about performance, if SecurityCheck = 1 ? I assume the query needs as long as before. 
We should consider redesign this crappy table Journal_Security, by replacing SecurityKey column by individual UserID, GroupID, RoleID, RelationID columns (nullable with filtered index on it). What do you think?
I'll try that in my scripts, using calculated persisted columns.

sleupold:
[Indexed_Journal_Get.sql.txt](https://github.com/dnnsoftware/Dnn.Platform/files/2408012/Indexed_Journal_Get.sql.txt)
I just created a new version of the procedure and underlying function for my TurboScripts, just to provide you with an Idea. It has not been tested, please be careful on production sites.

mikebigun:
> you were testing it against a single data set - other large installations may have different characteristics, based how the site is used. And another version of SQL Server may use a different approach.

I'm wondering if we should expect other results in a later sql versions. But, to make sure we are fine, I tried it on 2012 and latest 2016. No issues.
Instead of considering all possible characteristics (not really sure what do you mean, sounds like a very general statement including customer hardware and etc..), I would focus on a concrete database objects related to the problematic procedure.

> btw, what about performance, if SecurityCheck = 1? I assume the query runs as long as before.

No. The query runs in a seconds on a big installation when SecurityCheck = 1. That's because join condition `t.seckey = js.SecurityKey` is activated. Skipping the linkage produces one record of table 1 multiplied to all records of table 2 and so on. This is a root of customer problem!

> Permission check is called, even if not needed

Fully agree with you. Will be fixed.

> I just created a new version of the procedure and underlying function for my scripts

It's a great redesign approach to make performance better. Still, do we really solve performance issue?
Let's agree, our case is different. We are not dealing with a long running queries.


To summarize, I would not consider my fix as an improvement to existent performance. It is more likely logical fix that makes our query aligned with what it should do. 
The root cause of timeouts is not temp tables or case conditions, but wrong usage of permission check. Timeouts were gone once query started to return consistent results.

CC: @tpluscode

ohine:
@mikebigun Can you run some tests on the dnnsoftware dataset? I'm guessing that's also quite large and might show a different set of issues? 

@sleupold if you have any concrete examples which we can add to the tests to validate which method of improves this procedure and works better in a wider range of installs, that would be helpful as well.

sleupold:
I don't really have a site atm with large use of journal. However, main cause for bad performance is this weird design of table Journal_Security with the multipurpose column SecurityKey (nvarchar), which contains values like "E" (everyOne), "C" (registered Users), "R71" (RoleID 71), "U1440" (UserID 1440) and "F1470" (friends of UserID 1470). This should be normalized, i.e. replaced by columns UserID (int), RelUserID (Int), RoleID (int), EveryOne (bit) and Registered (bit), properly using foreign keys with cascading delete. This would avoid orphaned records and be evaluated much faster than the slow function Journal_User_Permissions, with its temp table.

ohine:
@sleupold Pull requests are welcomed to further improve the platform. Those items are out of scope for the work that @mikebigun is working on in this pull request.

mikebigun:
> Can you run some tests on the dnnsoftware dataset? 

@ohine, may I ask you provide more details. Is this database I can download or connect with?

mikebigun:
@ohine @sleupold @tpluscode 

Team, I have finally got access to a copy of production dnnsoftware database. 
I found following datasets: 
Journal - 28607
Users - 1089475
Journal_Data - 2624
Journal_Security - 62350
UserRelationships - 3904
 
I've selected a user with a maximum relationships (285 friends/followers) - UserID = 267593

`EXEC [dbo].[Journal_Get] 0, 267593, 35465`

Original query returns 160 identical rows.
Modified one - 1 row.

Execution time of the modified SP is less than a second.

sleupold:
@mikebigun 
Would you mind to test `EXEC [dbo].[Journal_Get] 0, 267593, 35465, 0, 0, 1` as well, thank you!

mikebigun:
@sleupold 
Yes, sorry, not all scenarios I included in my previous comment. Of course, I did validation for the cases when security check is enabled as well as different combination of other parameters (IncludeAllItems, IsDeleted).

I did not found the case when execution time of SP exceeds 1 sec.

PR #2336, opened 2018-09-25
Language code has been removed before calling the providers of applic…

### LINKED ISSUE #2335: Multilanguage AUM pass incorrect urlParms to TransformFriendlyUrlToQueryString
## Description of bug

urlParms

{string[1]}

string[]
[0] "param1" string

urlParms

{string[3]}

string[]
[0] "en-us" string
[1] "pagename" string
[2] "param1" string

## Steps to reproduce

## Current result

## Expected result

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1

Mohtshum:
…ation

Fixes #2335 

## Summary
Removing language code before calling providers in application

### DIFF
--- a/DNN Platform/Library/Entities/Urls/RewriteController.cs
+++ b/DNN Platform/Library/Entities/Urls/RewriteController.cs
@@ -1392,7 +1392,11 @@ namespace DotNetNuke.Entities.Urls
                 absoluteUri = absoluteUri.Replace(queryString, "");
             }
             absoluteUri = HttpUtility.UrlDecode(absoluteUri); //decode the incoming request
-            string rewritePath = GetTabFromDictionary(absoluteUri, queryStringCol, settings, result, parentTraceId);
+
+            // below 810 check is taking care of lanugage code
+            var absoluteUriWithLanguageCode = Regex.Replace(absoluteUri, $"\\/{result.CultureCode}", string.Empty, RegexOptions.IgnoreCase);
+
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Entities/Urls/RewriteController.cs
```diff
@@ -1392,7 +1392,11 @@ internal static void GetUrlWithQuerystring(HttpRequest request, Uri requestUri,
                 absoluteUri = absoluteUri.Replace(queryString, "");
             }
             absoluteUri = HttpUtility.UrlDecode(absoluteUri); //decode the incoming request
-            string rewritePath = GetTabFromDictionary(absoluteUri, queryStringCol, settings, result, parentTraceId);
+
+            // below 810 check is taking care of lanugage code
+            var absoluteUriWithLanguageCode = Regex.Replace(absoluteUri, $"\\/{result.CultureCode}", string.Empty, RegexOptions.IgnoreCase);
```
tpluscode: I think this variable should be called `*WithoutLanguageCode`?

tpluscode: Are we certain that the regular expression will always match successfully? What I mean is a possibility that the URI will contain a language code different from the one in `result.CultureCode`?

tpluscode:
Closed as it is superseded by #2355

PR #2340, opened 2018-09-25
DNN-8503: fix vw_Modules casting code

sleupold:
undo changes from DNN 08.00.00.21.sqldataprovider

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
@@ -107,8 +107,8 @@ GO
 
 CREATE VIEW {databaseOwner}[{objectQualifier}vw_Modules]
 AS
-    SELECT
         M.PortalID AS [OwnerPortalID],
+        DM.PackageID,
         T.PortalID,
         TM.TabID,
         TM.TabModuleID,
@@ -123,13 +123,9 @@ AS
         TM.Color,
         TM.Border,
         {databaseOwner}{objectQualifier}FilePath(T.IconFile) AS IconFile,
-        M.AllTabs,
         TM.Visibility,
-        TM.IsDeleted,
         TM.Header,
         TM.Footer,
-        M.StartDate,
-        M.EndDate,
         TM.ContainerSrc,
         TM.DisplayTitle,
         TM.DisplayPrint,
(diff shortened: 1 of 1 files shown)


nvisionative:
@sleupold this looks like it is referencing an old JIRA issue.  Have you created a GitHub issue?  For a quick reference on the process, [see this document](https://github.com/dnnsoftware/Dnn.Platform/blob/development/CONTRIBUTING.md).

sleupold:
@nvisionative sorry, it would not be polite to blame so for this issue.

mitchelsellers:
@sleupold Thanks for submitting this one.  Do you happen to have a compare of this script from the current 9.2.2 version to this one?  or can you at least highlight the reason for the reversion?

I will review this in detail in the next few days, but this will help me get that done sooner.

sleupold:
I just reverted iconFile retrieval  {databaseOwner}{objectQualifier}FilePath(T.IconFile) AS IconFile, 
because the previous version 
' SELECT     IsNull(Folder, '') + [FileName]
                            FROM          dbo.[vw_Files]
                            WHERE      FileId = CAST(SUBSTRING(TM.IconFile, 8, 10) AS Int)) ELSE COALESCE (TM.IconFile, '') END AS IconFile '
fails, if the value on M.IconFile is a URL, which cannot be converted to an integer.

sleupold:
Note: using a function with a query might not be the fastest solution, but in a typical scenario, where only a smaller part (maybe 30%) are using an icon, I tested it to be the fastest one.

mitchelsellers:
@sleupold Thanks for the detail on this, overall this does look cleaner as well.  I'll work to get this one reviewed ASAP.  The current implementation did have detailed performance review completed, granted it might have missed the IconFile usage testing.

mitchelsellers:
Just to update this PR since discussions have been happening outside of this.

The change here was completed back in 8.0.0, and we don't have any reported issues surrounding this issue.  It is true that prior sites, with some old data included, might encounter an error, but no-one has been able to recreate an error with currently enabled & possible settings in the DNN Interface.

If we can duplicate/create an issue, in 9.2.2 that this will resolve, we can reconsider, but at this time it doesn't seem to solve a supported issue at the moment, and the change was implemented for performance reasons.

sleupold:
@mitchelsellers this PR fixes an error reported by multiple users after the change had been introduced in DNN 8.0.0, which had been submitted by you to "improve performance", though my tests indicated a performance decrease besides the error. if we are now supporting fileid's only, we should convert the field to int to eliminate the costly data type conversion. But of course, this decision is up to you. 
I have no problems to close the issue (my users already got the fix in my TurboScripts).

mitchelsellers:
@sleupold I managed to drill into the history on this a bit more and found details on exactly what was done when this change was introduced.  The change that was introduced in DNN 8.0.0 was tied to DNN Bug DNN-8950.

The changes completed at that time were two fold.

1. The stored procedure GetModuleByDefinition was modified to ensure that in all situations it used the vw_modules view for data retrieval
2. The view vw_Modules was updated to include the logic that handled both file ID's and paths in the field for icon file (Case when tm.IconFile LIKE 'fileid=%' 
			then (SELECT IsNull(Folder, '') + [FileName] FROM dbo.[vw_Files]
			 WHERE fileid = CAST(SUBSTRING(tm.IconFile, 8, 10) AS Int))
			 else Coalesce(tm.IconFile,'')
			 end as IconFile,)

As such, now I'm further confused on the reversion on this change.  As this was noted as tied to a revert of the change, but that change was a multi-part change.  We are sure tahat this doesn't impact the functionality of the GetModuleByDefinition call in which this change was introduced for?

sleupold:
@mitchelsellers please read https://dnntracker.atlassian.net/browse/DNN-8503.
I ran tests on the several solutions with a typical scenario of ~10000 modules and about 20% having set the icon file and noticed the function running nearly as fast as a joined table and faster as the subquery used in your solution.

sleupold:
@mitchelsellers with this pull request being cancelled, I would need someone else to submit pull requests for my contributions.

PR #2341, opened 2018-09-25
DNN-9651: drop duplicate index and unused column

sleupold:
Removes the duplicate index `IX_Users_LowerEmail` and removes the calculated column `LowerEmail` column because it's not utilized in the platform.

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
@@ -101,6 +101,14 @@ END
 
 GO
 
+/***** DNN-9651: Drop duplicate index and unused calculated column *****/
+IF Exists (SELECT * FROM sys.indexes WHERE OBJECT_ID = OBJECT_ID(N'{databaseOwner}[{objectQualifier}Users]') AND name = N'IX_{objectQualifier}Users_LowerEmail')
+	DROP INDEX IX_{objectQualifier}Users_LowerEmail ON {databaseOwner}[{objectQualifier}Users];
+GO
+
+IF Exists (SELECT * FROM sys.columns WHERE OBJECT_ID = OBJECT_ID(N'{databaseOwner}[{objectQualifier}Users]') AND name = N'LowerEmail')
+	ALTER TABLE {databaseOwner}[{objectQualifier}Users] DROP COLUMN LowerEmail;
+GO
 /************************************************************/
 /*****              SqlDataProvider                     *****/
 /************************************************************/


nvisionative:
@sleupold this looks like it is referencing an old JIRA issue. Have you created a GitHub issue? For a quick reference on the process, [see this document](https://github.com/dnnsoftware/Dnn.Platform/blob/development/CONTRIBUTING.md).

sleupold:
@nvisionative this is by design, I don't want to blame anyone

mitchelsellers:
@sleupold Is the primary concern here the duplicate index?  If so, can we leave that as the only change to speed the approval of this submission as I agree 100% that this index is duplicated as it is reported by SQL Azure on all of our databases

The removal of the column, however, is a breaking change, potentially, as the field could be used by other projects.

sleupold:
@mitchelsellers the column is not surfaced in the API

ohine:
If the column is not used anywhere, should we also remove this reference to it in the export user dto object? 

https://github.com/dnnsoftware/Dnn.Platform/blob/584b0f54bd6b6d237594dd9f96053f14db0b1a53/DNN%20Platform/Modules/DnnExportImportLibrary/Dto/Users/ExportUser.cs#L54

mitchelsellers:
@sleupold See Oliver's comment, would you have time to make that change?

sleupold:
@mitchelsellers 
there might be other calls in 3rd party modules as well, though it's not very likely. 
Fixing it, would improve the database, but as this would require additional arguments, I suggest just closing this issue, any objections?

mitchelsellers:
This DTO should be a nonbreaking change. But...@ohine thoughts?  

Removing the column is the right call.  But we do want it totally removed to not add technical debt.

ohine:
I don't think we need to close the issue. Just need to figure out if `LowerEmail` is really used and wired up anywhere. We could flag it as obsolete and then remove it in v11, but still remove it from the DB in v9.3 and warn people that if they did use the dto object or query directly off that field in any custom modules or reports that it would be a breaking change for them?

valadas:
I am of the opinion that if third party modules query that directly from the database without using the API, that would be bad practice on the module side right?

As for the DTO what about marking the LowerEmail deprecated but for now just return the email lowered just in case and do nothing on set.

sleupold:
@valadas the database is not an API and there is no guarantee to keep this field. In API we could keep the property LowerEmail but calculate LowerEmail := Email.ToLower and mark as deprecated.

ohine:
As for querying the DB directly, while it's not the best practice. It's still done. I know of 100's if not 1000's of sql reporting modules which are going directly to the DB. I think we should just mention that it's removed and could potentially be a breaking change if you are bypassing the api and going directly to the database. I'd much rather warn someone then have them come and complain after upgrading.

valadas:
@sleupold that is what I meant too, remove from the database only but still return the calculated value until this method is removed. and also mark it deprecated. I am not sure how DTO works but comparing that to PetaPoco when a field is marked [IgnoreColumn] so it does not read it from the database but can still return something to other code that calls it. Not sure what is the equivalent of that in a DTO.

@ohine I am all for documenting those changes in the release notes. There has been so many undocumented changes in the past, documenting these type of changes would be really helpful for developers to keep up with changes.

sleupold:
resubmitted

mitchelsellers:
Closing this PR as the other PR was successfully approved

PR #2344, opened 2018-09-26
Clear user cache after Updates for Approval or LockedOut

### LINKED ISSUE #2343: User Lockout value is not reflected until the cache is manually cleared.
## Description of bug
User Lockout value is not reflected until the cache is manually cleared.

## Steps to reproduce
1. Create a new user and authorize it. 
2. Open an incognito window and fail the password 5 times (default number of attempts before getting locked out) 
3. Check the User Properties Manage > Users > Gear Icon for the User.

## Current result
It says the user is not locked out because this data is loaded from Cache. If you check the aspnet_membership table in the database, the user will already have the lockedout flag set to true.

## Expected result
The cache should be cleared after the user is locked out

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x ] 9.1
* [x] 9.0

Mohtshum:
Fixes #2343 

## Summary
Clear user cache after Updates for Approval or LockedOut on found user

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -1767,10 +1767,7 @@ namespace DotNetNuke.Security.Membership
             //Global Data Store but not in the Local DataStore ie. A shared Global Data Store
 
             //Initialise Login Status to Failure
-            loginStatus = UserLoginStatus.LOGIN_FAILURE;
-
-            DataCache.ClearUserCache(portalId, username);
-            DataCache.ClearCache(GetCacheKey(username));
+            loginStatus = UserLoginStatus.LOGIN_FAILURE;       
 
             //Get a light-weight (unhydrated) DNN User from the Database, we will hydrate it later if neccessary
             UserInfo user = (authType == "DNN")
@@ -1794,7 +1791,7 @@ namespace DotNetNuke.Security.Membership
                     }
                     else
                     {
-                        loginStatus = UserLoginStatus.LOGIN_USERLOCKEDOUT;
+                        loginStatus = UserLoginStatus.LOGIN_USERLOCKEDOUT;                      
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -1767,10 +1767,7 @@ public override void UpdateUsersOnline(Hashtable userList)
             //Global Data Store but not in the Local DataStore ie. A shared Global Data Store
 
             //Initialise Login Status to Failure
-            loginStatus = UserLoginStatus.LOGIN_FAILURE;
-
-            DataCache.ClearUserCache(portalId, username);
-            DataCache.ClearCache(GetCacheKey(username));
+            loginStatus = UserLoginStatus.LOGIN_FAILURE;
```
mitchelsellers: By removing this call at the beginning it is possible that when beginning the login call you will get a cached user object, which might not have been properly invalidated, thus allowing the user to be identified as a non-deleted user.  I don't believe it would be safe to remove this call here?  Can you confirm the validation that this did not cause issues?

DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -1794,7 +1791,7 @@ public override void UpdateUsersOnline(Hashtable userList)
                     }
                     else
                     {
-                        loginStatus = UserLoginStatus.LOGIN_USERLOCKEDOUT;
+                        loginStatus = UserLoginStatus.LOGIN_USERLOCKEDOUT;                      
                     }
```
Mohtshum: DataCache.ClearUserCache(portalId, username);
this is also working if I put it here. So What I'm thinking that put back those at the beginning and one here

mitchelsellers: @Mohtshum Yes, I think it should, but I don't believe that we can guarantee that to be the case, login should ALWAYS be pulling a non-cached user.  So It think re-adding the ones at the beginning and then one here would be the best option to do this quickly

Mohtshum: Did revert and put the cache clear call after Lockedout and approval call

PR #2347, opened 2018-09-27
Set host user parameter to be add in default portal at the time of in…

### LINKED ISSUE #2345: Host user entry not found in UserPortal against default portal(site)
## Description of bug
Host user entry not found in UserPortal making it not searchable by Email (there could be other issues as well). Mean while when create a portal or site and same host for that newly created portal then a entry in Userportal is added, but somehow no entry for default site

## Steps to reproduce
Install any DNN product. And after successful installation, go to 
    Settings > SQL Console and run following query
    select * from userportals;
    

## Current result
you won't see any record of 
    userid (1) and portalid (0)

## Expected result
    There should be a record for 
    userid (1) and portalid (0)

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1

Mohtshum:
…stallation

Fixes #2345 

## Summary
Add optional parameter to AddPortal call during installation, and pass the host user into it.

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/Upgrade.cs
+++ b/DNN Platform/Library/Services/Upgrade/Upgrade.cs
@@ -3937,7 +3937,7 @@ namespace DotNetNuke.Services.Upgrade
         /// <remarks>
         /// </remarks>
         /// -----------------------------------------------------------------------------
-        [Obsolete("Deprecated in DNN 9.3.0. Use the overloaded one with the 'superUser' parameter instead.")]
+        [Obsolete("Deprecated in DNN 9.3.0, will be removed in 11.0.0. Use the overloaded method with the 'superUser' parameter instead.")]
         public static int AddPortal(XmlNode node, bool status, int indent)
         {
             return AddPortal(node, status, indent, null);


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3797,7 +3797,7 @@ public static void AddModuleToPages(string tabPath, int moduleDefId, string modu
         /// <remarks>
         /// </remarks>
         /// -----------------------------------------------------------------------------
-        public static int AddPortal(XmlNode node, bool status, int indent)
+        public static int AddPortal(XmlNode node, bool status, int indent, UserInfo superUser = null)
```
ohine: You are going to need to mark the old AddPortal function obsolete and removed in v11. Otherwise the changes look good! Thanks for tracking down the issue and getting a fix submitted!

sleupold:
A portal has an admin, but host user is not a portal property.

I could imagine setting a host being set automtically to the admin user of the first portal if there is no superuser yet.

mitchelsellers:
@Mohtshum Just confirming that the issue here is that with a default install, with a single user added the superuser isn't searchable in the Users section, even under SuperUsers correct?

sleupold:
@mitchelsellers In DNN 9.2.0 we switched search in user management back to database and superusers are identified by isSuperuser flag only and are never displayed as site users, even if having a userPortals record.

Mohtshum:
@mitchelsellers, I found this issue while working on #2345, enabling UseEmailasUserName not working for host user, as [Call to GetUserByEmail](https://github.com/dnnsoftware/Dnn.Platform/blob/a03b5589a7e1a0fdded59ec8bf033ed1e9eca5cd/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs#L269) is returning null for host user only (which is using stored procedure namely GetSingleUserByEmail).

And when installed fresh DNN.Platform without wizard verified that host user entry is not entered in UserPortals table

DNN Platform/Library/Services/Upgrade/Upgrade.cs
Mohtshum: @ohine  Please review changes.

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3928,6 +3930,19 @@ public static int AddPortal(XmlNode node, bool status, int indent)
...
         }
 
+        /// -----------------------------------------------------------------------------
+        /// <summary>
+        ///   Obsolete, AddPortal manages the Installation of a new DotNetNuke Portal
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        /// -----------------------------------------------------------------------------
+        [Obsolete("Deprecated in DNN 9.3.0. Use the overloaded one with the 'superUser' parameter instead.")]
```
mitchelsellers: Can we get one small change to this please, it should read.

Deprecated in DNN 9.3.0, will be removed in 11.0.0. Use the overloaded method with the 'superUser' parameter instead.

I'll make sure to update the PR documentation on this.  We just want to note, when it was deprecated when it will be removed, and the workaround.

ohine:
@Mohtshum This looks good, I updated the message slightly to include the version we will be removing this function from the platform. @mitchelsellers if you can update your review and process the merge that would be great!

Mhtshum:
@mitchelsellers changes have been pushed by @ohine. Please review and if OK then merge

Mhtshum:
@sleupold, Actually when you are installing via a wizard then host user is added into the userportal table and when installing without wizard then this issue is occurring.

tpluscode:
@sleupold I don't think that there is a misunderstanding after all. @Mhtshum is referring to an unattended installation, without wizard. in such case the superuser is indeed created but not associated with a portal.

sleupold:
@tpluscode Superusers are by default not associated with any website (portal).
However the website needs a dedicated admin user, who should be member of the administrators role.
For unattended installations, DNN used to create a superuser "host" and a site assigned user "admin".

Issue #2353, opened 2018-10-02
Show Module Labels


tpluscode:
## Description of problem
When a page is in edit mode, it would be helpful to have a small label to indicate the name/type of module(s) on the page, and/or a visual indicator for empty modules on a page.

## Description of solution
Sample Workflow:
A new module is placed on the page > the page refreshes > it is difficult to discern where the empty module is sitting on the page and what type of module is in the pane

## Description of alternatives considered
@mikebigun proposes two solutions. One could be to show the label centred:

![image-2018-09-26-18-02-56-236](https://user-images.githubusercontent.com/588128/46337509-fc171800-c62d-11e8-942b-2a54ec4c2cb4.png)

or a vertically next to module control menus;

<img width="1920" alt="screen shot 2018-10-02 at 10 29 50" src="https://user-images.githubusercontent.com/588128/46337648-53b58380-c62e-11e8-9f41-5717b3cb7dcc.png">

PR #2354, opened 2018-10-02
Show Module Labels

mikebigun:
fixes [2353](https://github.com/dnnsoftware/Dnn.Platform/issues/2353)

Similar indicator is already shown for the shared modules. It is placed vertically in a right corner. To stay with the same design, module label is added in a same way. 
I'm open to discuss more ideas.

Short DEMO: [mp4](https://drive.google.com/file/d/1C2EFMusxS2npDRwDpN0qfoFsKir4NpEZ/view)

### DIFF
--- a/Website/admin/Menus/ModuleActions/ModuleActions.ascx
+++ b/Website/admin/Menus/ModuleActions/ModuleActions.ascx
@@ -58,7 +58,8 @@
                     supportsMove: <% = SupportsMove.ToString().ToLower() %>,
                     supportsQuickSettings: supportsQuickSettings,
                     displayQuickSettings: displayQuickSettings,
-                    isShared : <% = IsShared.ToString().ToLower() %>
+                    isShared : <% = IsShared.ToString().ToLower() %>,
+                    moduleTitle: '<% = Localization.GetSafeJSString(ModuleTitle) %>'
                 }
             );
         }
--- a/Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs
+++ b/Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs
@@ -84,6 +84,8 @@ namespace DotNetNuke.Admin.Containers
             return Localization.GetString(key, Localization.GlobalResourceFile);
         }
 
+        protected string ModuleTitle { get; set; }
+
         protected override void OnInit(EventArgs e)
         {
             base.OnInit(e);
@@ -117,7 +119,7 @@ namespace DotNetNuke.Admin.Containers
             {
                 SupportsQuickSettings = false;
                 DisplayQuickSettings = false;
-
+                ModuleTitle = ModuleContext.Configuration.ModuleTitle;
                 var moduleDefinitionId = ModuleContext.Configuration.ModuleDefID;
                 var quickSettingsControl = ModuleControlController.GetModuleControlByControlKey("QuickSettings", moduleDefinitionId);
 
--- a/Website/admin/Menus/ModuleActions/ModuleActions.css
+++ b/Website/admin/Menus/ModuleActions/ModuleActions.css
@@ -178,10 +178,11 @@ div.actionMenu ul.dnn_mact > li.actionQuickSettings div.qsFooter a.secondarybtn 
   text-align: center;
   font-size: 13px;
 }
-div.actionMenu ul.dnn_mact li.dnn_shared {
+div.actionMenu ul.dnn_mact li.dnn_menu_label {
   display: inline-block;
+  pointer-events: none;
 }
-div.actionMenu ul.dnn_mact li.dnn_shared > div {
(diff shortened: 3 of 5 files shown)


ohine:
I'm 50/50 on the UI, not enough to say I don't approve, just enough to mention some concerns before we merge it into the platform.

The vertical text is difficult to read and now on hover, we're adding elements all over and it's not really consistent and appears cluttered to me.

What about adding an (i) info icon in the top left corner. Hovering or clicking (for touch devices) would toggle a tooltip, this would display pane name, module type, and I'm thinking maybe module title if display container is false. This would allow us to continue to add additional details if we determine to later and keep a consistent and non-cluttered interface for now. 

Thanks a trillion for the video, it helps a lot to provide feedback without having to setup a test site against this pull request!!

EPTamminga:
Why wait for hovering and not directly display pane name? 
Having the pane name of all panes visible in edit mode is essential if you are moving modules around, especially if you have a page skin that has a lot of panes.

ohine:
While I know this is mostly off topic, I'm also not a fan of the 3 module actions buttons. I think if we had one larger icon and then consolidated all the modules actions under a single menu which expanded like it originally did, I would be much happier and it would continue cleaning up the edit interface quite a bit.

ohine:
@EPTamminga I'll probably be the ONLY one that's a fan of bringing back layout mode, but that was ideal for being able to see the skin layout. I don't think the info is required after the first few times you are editing a new skin or site, the hover effect to display that works as you can now at least identify what pane is what. 

Plus I had plans for improving layout mode it to add more functionality into the skinning engine by putting the ui to swap out styles/panes/rows/columns/toggle sko's, etc in layout mode. It's always annoyed me we have 19 variations of a container which is only changing the H tag, or a color. These should be handled by switching into layout mode and having additional buttons to control the actual layout of the site.

EPTamminga:
@ohine Having 3 different buttons with (seemingly random distributions of) actions is also not my favourite UI solution. But seeing the names of all the panes in edit mode helps a lot when people are moving modules around, especially if you do not use drag/drop, but choose a pane that is listed in the action menu.

ohine:
@EPTamminga I know personally I need to see the pane names for maybe a couple hours after dealing with a new skin, after that it's easy to picture where a pane is positioned on the page. Seeing them all the time I think isn't right. Maybe we could toggle them to appear if you expand the move module menu?

EPTamminga:
@ohine That might be true if you manage a few websites regularly, but not if you manage dozens of different formatted site only once a month or so, where sites can have page skins with 20 or more panes.

mitchelsellers:
Just my $0.02 here.  This implementation is an extension of an already existing feature, showing shared module names.  Although the UI/UX isn't perfect here this is using existing styles and behaviors and not introducing new paradigms.

There are other changes in play that deal with Pane Names, I'd have to go back and find it, but there is one.  

The mobile user is a concern, but not at the moment as the mobile user cannot even enter edit mode so they will never see these actions, or be forced to hover interact with it.

EPTamminga:
IF there is an existing feature that can cover it, I am in favour of reusing that i.s.o. introducing new stuff.

mikebigun:
@ohine, @EPTamminga , @mitchelsellers, @tpluscode 

Team, thanks all of you for the valuable inputs and sharing your user experience. 
At the end we should make some conclusion on it.
As per @mitchelsellers, this is an extension to the already existing feature. Simplest and fastest solution. It does not improve UI/UX significantly, but helps our customers easily navigate over the panes. Originally, that was the purpose of this change.
Since at this moment of time we do not have a clear consensus, I'd go with the current implementation.

sleupold:
I remember, Torsten Weggen (@weggetor) had some nice ideas, when the action module was designed a few years ago

weggetor:
I think Sebastian refers to this: http://www.bitboxx.net/Blog/Post/285/DnnImagehandler-Hot-or-not#/

The original DNN imagehandler contained the generation of a ModuleInfo Image. 
image

This generated image was removed by the the corp.

mikebigun:
@ohine, @EPTamminga , @mitchelsellers, @tpluscode

Team, so one more idea from sleupold and weggetor. Any comments? I just want to make some conclusion that will be accepted by everyone.

tpluscode:
Guys, are we fine with the @mikbigun's proposal? I'd like us to move this forward.

If there are ideas for long-term improvement then I'd rather they were discussed as an issue in its own right.

sleupold:
@tpluscode I just checked "shared module" on a site - is there a reason, this is not displayed on all tabmodules or is this just a bug?

valadas:
I really like the [i] informations popup idea, I think that would cover it for modules in any pane size (some skins have very small panes). This would leave some room for possibly showing pane names in edit mode.

bdukes:
We're merging this as-is for now, and can tackle rearranging the module actions in a separate issue, see #2433.

PR #2355, opened 2018-10-03
Don't send language params to TransformFriendlyUrlToQueryString

### LINKED ISSUE #2335: Multilanguage AUM pass incorrect urlParms to TransformFriendlyUrlToQueryString

Mhtshum:
Fixes #2335 

## Summary

change variable name


DNN Platform/Library/Entities/Urls/RewriteController.cs
```diff
@@ -1416,6 +1420,11 @@ internal static void GetUrlWithQuerystring(HttpRequest request, Uri requestUri,
             return result.DoRewrite;
         }
 
+        public static string GetUrlWithLanguageCode(string absoluteUri, string cultureCode)
```
daguiler: Can we use the [InternalsVisibleTo] attribute instead of making this public?

@dnnsoftware/approvers

tpluscode: @dnnsoftware/approvers ☝️

bdukes: 👍 sounds like the better approach

Mhtshum:
Changes pushed as @daguiler, though attribute already present in AssemblyInfo.cs

DNN Platform/Library/Entities/Urls/RewriteController.cs
```diff
@@ -43,8 +44,9 @@
 
 #endregion
 
+[assembly: InternalsVisibleTo("DotNetNuke.Tests.Web")]
```
tpluscode: Please put it in `AssemblyInfo.cs`

DNN Platform/Library/Entities/Urls/RewriteController.cs
```diff
@@ -1392,7 +1392,11 @@ internal static void GetUrlWithQuerystring(HttpRequest request, Uri requestUri,
                 absoluteUri = absoluteUri.Replace(queryString, "");
             }
             absoluteUri = HttpUtility.UrlDecode(absoluteUri); //decode the incoming request
-            string rewritePath = GetTabFromDictionary(absoluteUri, queryStringCol, settings, result, parentTraceId);
+
+            // below 810 check is taking care of lanugage code
+            var absoluteUriWithoutLanguageCode = GetUrlWithoutLanguageCode(absoluteUri, result.CultureCode);
```
tpluscode: Good rename :+1:. Haven't noticed that before

sleupold:
I haven't reviewed the full code, but please be aware that the culture code in the URL should be part of the portal Alias and might be modified by the admin (e.g. www.domain.com/en or en.domain.com)

Mhtshum:
@sleupold Right below there is already a logic, to add again the language code which I try to mention by in comment //810

DNN Platform/Library/Entities/Urls/RewriteController.cs
```diff
@@ -1392,7 +1392,11 @@ internal static void GetUrlWithQuerystring(HttpRequest request, Uri requestUri,
...
+            
+            string rewritePath = GetTabFromDictionary(absoluteUriWithoutLanguageCode, queryStringCol, settings, result, parentTraceId);
```
mitchelsellers: By making the change in the way that you get the re-write path doesn't this change the actual content that is obtained?  It might be ok, but I want to make sure that it is correct. before we go.

Also, this change has a potential to break the API for users that have worked around.  I would prefer to NOT include this breaking change in 9.3, but will defer to the rest of the @dnnsoftware/tag

Mhtshum: @mitchelsellers  In fact this change, is intended to correct the content, when multilingual option is enable. If you check #2335, TransformFriendlyUrlToQueryString is getting extra or additional parameter of language code, which is the problem or issue.

mitchelsellers: @dnnsoftware/tag Thoughts?  I'm not ok with this.

nvisionative: This can indeed be a breaking change if people are not following the practice pointed out by @sleupold above.  This doesn't sound like a good idea for 9.3 IMO.

DNN Platform/Tests/DotNetNuke.Tests.Web/UrlRewrite/RewriteControllerUnitTests.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace DotNetNuke.Tests.Web.UrlRewrite
+{
+    [TestFixture]
+    class RewriteControllerUnitTests
+    {
+        [TestCase("http://www.dnn.com/en-us/aboutus", "en-us", "http://www.dnn.com/aboutus")]
+        [TestCase("http://www.dnn.com/en-Us/aboutus", "en-Us", "http://www.dnn.com/aboutus")]
+        [TestCase("http://www.dnn.com/aboutus", "en-Us", "http://www.dnn.com/aboutus")]
+        public void RemoveLanguageCodeFromAbsoluteURL(string absoluteURL, string languageCode, string expectedURL)
```
Mhtshum: @mitchelsellers @sleupold  I'm not able to understand the breaking changes please guide by pointing out the missing case

DNN Platform/Tests/DotNetNuke.Tests.Web/UrlRewrite/RewriteControllerUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [TestFixture]
+    class RewriteControllerUnitTests
+    {
+        [TestCase("http://www.dnn.com/en-us/aboutus", "en-us", "http://www.dnn.com/aboutus")]
+        [TestCase("http://www.dnn.com/en-Us/aboutus", "en-Us", "http://www.dnn.com/aboutus")]
+        [TestCase("http://www.dnn.com/aboutus", "en-Us", "http://www.dnn.com/aboutus")]
+        [TestCase("https://dnn.com/en", "en-Us", "https://dnn.com/en")]
+        [TestCase("https://www.dnn.com/en", "en-Us", "https://www.dnn.com/en")]
+        [TestCase("http://en.dnn.com/", "en-Us", "http://en.dnn.com/")]
```
Mhtshum: @sleupold , @mitchelsellers Please check added cases.

daguiler:
I think I found a better way to fix this issue. Check this out: #2759

mitchelsellers:
Closing this in favor of the other PR

Issue #2357, opened 2018-10-03
Ability to specify Terms and Privacy pages just like other default pages


donker:
## Description of problem
We can specify login, registration, 404 etc pages, but not "terms of use" or "privacy statement". It is no longer sufficient to keep these texts in resx files for sites that need to comply with GDPR. In these cases you need to have much more tailored texts. Instead of making an editor for the resource text, it's preferable to be able to manage this as an entire page.

## Description of solution
Provide two new default pages in the site settings and read that in the skin objects. If not specified the site behaves as before.

valadas:
We also need to consider upgrades. Maybe the simplest would be to use the resx key if no page is specified, if a page is specified than use that... If we go with this approach, then explaining this in the label help text would be nice.

valadas:
Oh, as I see the pull request, it looks it was implemented that way.

ohine:
Closing since PR was merged.

PR #2352, opened 2018-10-01
Merge after 9.2.2

bdukes:
This is a redo of #2311

PR #2358, opened 2018-10-03
New default pages for terms and privacy

donker:
Issue description here:

https://github.com/dnnsoftware/Dnn.Platform/issues/2357

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/09.02.02.01.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.02.02.01.SqlDataProvider
@@ -1,0 +1,281 @@
+/************************************************************/
+/*****              SqlDataProvider                     *****/
+/*****                                                  *****/
+/*****                                                  *****/
+/***** Note: To manually execute this script you must   *****/
+/*****       perform a search and replace operation     *****/
+/*****       for {databaseOwner} and {objectQualifier}  *****/
+/*****                                                  *****/
+/************************************************************/
+
+/* Implement new default pages */
+/*******************************************/
+
+ALTER TABLE {databaseOwner}{objectQualifier}PortalLocalization
+ADD TermsTabId [int] NULL, PrivacyTabId [int] NULL
+GO
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}{objectQualifier}UpdatePortalInfo') AND type in (N'P', N'PC'))
+DROP PROCEDURE {databaseOwner}{objectQualifier}UpdatePortalInfo
+GO
+
+CREATE PROCEDURE {databaseOwner}{objectQualifier}UpdatePortalInfo
+	@PortalID				INT,
+	@PortalGroupID			INT,
+	@PortalName				NVARCHAR(128),
+	@LogoFile				NVARCHAR(50),
+	@FooterText				NVARCHAR(100),
+	@ExpiryDate				DATETIME,
+	@UserRegistration		INT,
+	@BannerAdvertising		INT,
+	@Currency				CHAR(3),
+	@AdministratorId		INT,
+	@HostFee				MONEY,
+	@HostSpace				INT,
+	@PageQuota				INT,
+	@UserQuota				INT,
+	@PaymentProcessor		NVARCHAR(50),
+	@ProcessorUserId		NVARCHAR(50),
+	@ProcessorPassword		NVARCHAR(50),
+	@Description			NVARCHAR(500),
+	@KeyWords				NVARCHAR(500),
+	@BackgroundFile			NVARCHAR(50),
+	@SiteLogHistory			INT,
+	@SplashTabId			INT,
+	@HomeTabId				INT,
+	@LoginTabId				INT,
+	@RegisterTabId			INT,
+	@UserTabId				INT,
+	@SearchTabId			INT,
+    @Custom404TabId			INT,
+    @Custom500TabId			INT,
+    @TermsTabId				INT,
+    @PrivacyTabId			INT,
+	@DefaultLanguage		NVARCHAR(10),
+	@HomeDirectory			VARCHAR(100),
+	@LastModifiedByUserID	INT,
+	@CultureCode			NVARCHAR(50)
+
+AS
+
+	UPDATE {databaseOwner}{objectQualifier}Portals
+		SET    
+		   PortalGroupID		= @PortalGroupID,
+		   ExpiryDate			= @ExpiryDate,
+		   UserRegistration		= @UserRegistration,
+		   BannerAdvertising	= @BannerAdvertising,
+		   Currency				= @Currency,
+		   AdministratorId		= @AdministratorId,
+		   HostFee				= @HostFee,
+		   HostSpace			= @HostSpace,
+		   PageQuota			= @PageQuota,
+		   UserQuota			= @UserQuota,
+		   PaymentProcessor		= @PaymentProcessor,
+		   ProcessorUserId		= @ProcessorUserId,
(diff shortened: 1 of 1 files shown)


mitchelsellers:
@donker It appears that your changes include some other items that are merged in from other items.  I thik we need to cleanup the tree/commit history on this as it isn't reviewable at this time.  @ohine any idea what happened?

sleupold:
I am missing code to update PortalLocalization table, view vw_Portals and GetPortals procedure.

sleupold:
At some point, we should consider moving all these tabids to a SpecialTabs table (PortalID, CultureCode TabDefinitionID, TabID, controlled by a SpecialPageDefinition table (TabDefinitionID, TabName, required, isPortalSpecific, IsLocalized, DefaultTabPath)

ohine:
It appears this is based from the v9.2.2 release and then targets development? I’m hopeful after we merge #2352 into development it’ll clean things up. If not I’ll manually clean up the PR since it’s an issue due to issues earlier in the week when attempting to get development current with the fixes pushed to v9.2.2 during the RC process. I apologize for this and we will make sure for the next release this task is done prior to the final releasing being marked public.

donker:
> 
> 
> At some point, we should consider moving all these tabids to a SpecialTabs table (PortalID, CultureCode TabDefinitionID, TabID, controlled by a SpecialPageDefinition table (TabDefinitionID, TabName, ModuleDefinitionID, required, isPortalSpecific, IsLocalized, DefaultTabPath)

I forgot to include the sql script. Should now be there.

donker:
> 
> 
> I am missing code to update PortalLocalization table, view vw_Portals and GetPortals procedure.

Seems like the sql script didn't get uploaded. Now it's there.

ohine:
@donker I'll see if I can fix it, we need to change that script to v9.3.0....

DNN Platform/Tests/DotNetNuke.Tests.Core/Controllers/Search/InternalSearchControllerTests.cs
```diff
@@ -250,7 +250,7 @@ private IDataReader GetPortalCallBack(int portalId, string culture)
             table.Rows.Add(portalId, null, "My Website", "Logo.png", "Copyright 2011 by DotNetNuke Corporation", null,
                 "2", "0", "2", "USD", "0", "0", "0", "0", "0", "1", "My Website", "DotNetNuke, DNN, Content, Management, CMS", null,
                 "1057AC7A-3C08-4849-A3A6-3D2AB4662020", null, null, null, "0", "admin@change.me", "en-US", "-8", "58", "Portals/0",
-                null, homePage.ToString("D"), null, null, "57", "56", "-1", "-1", "7", "-1", "2011-08-25 07:34:11", "-1", "2011-08-25 07:34:29", culture);
+                null, homePage.ToString("D"), null, null, "57", "56", "-1", "-1", "7", null, null, "-1", "2011-08-25 07:34:11", "-1", "2011-08-25 07:34:29", culture);
```
mitchelsellers: Is the expected value actually null here, or -1 like the other values?

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/JwtAuthMessageHandlerTests.cs
```diff
@@ -176,7 +176,7 @@ private static IDataReader GetPortalCallBack(int portalId, string culture)
                 "2", "0", "2", "USD", "0", "0", "0", "0", "0", "1", "My Website",
                 "DotNetNuke, DNN, Content, Management, CMS", null, "1057AC7A-3C08-4849-A3A6-3D2AB4662020",
                 null, null, null, "0", "admin@change.me", "en-US", "-8", "58", "Portals/0", null,
-                homePage.ToString(), null, null, "57", "56", "-1", "-1", "7", "-1", "2011-08-25 07:34:11",
+                homePage.ToString(), null, null, "57", "56", "-1", "-1", null, null, "7", "-1", "2011-08-25 07:34:11",
```
mitchelsellers: Same question here

Website/DotNetNuke.Website.csproj
```diff
@@ -3288,6 +3288,7 @@
     <Content Include="Providers\DataProviders\SqlDataProvider\09.02.00.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.02.01.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.02.02.SqlDataProvider" />
+    <Content Include="Providers\DataProviders\SqlDataProvider\09.02.02.01.SqlDataProvider" />
```
mitchelsellers: This should be 09.03.00.01

Website/admin/Skins/Privacy.ascx.cs
```diff
@@ -70,7 +70,7 @@ protected override void OnLoad(EventArgs e)
                 {
                     hypPrivacy.Text = Localization.GetString("Privacy", Localization.GetResourceFile(this, MyFileName));
                 }
-                hypPrivacy.NavigateUrl = Globals.NavigateURL(PortalSettings.ActiveTab.TabID, "Privacy");
+                hypPrivacy.NavigateUrl = PortalSettings.PrivacyTabId == Null.NullInteger ? Globals.NavigateURL(PortalSettings.ActiveTab.TabID, "Privacy") : Globals.NavigateURL(PortalSettings.PrivacyTabId);
```
mitchelsellers: Based on this check, the prior checks with Null vs -1 should be updated to be consistent

DNN Platform/Tests/DotNetNuke.Tests.Core/Controllers/Search/InternalSearchControllerTests.cs
donker: Disclaimer: I have not been able to get testing running on my VS Community 2017 locally. So I'm reaching in the dark. In any case: both -1 and null should return the same behavior in the site. These are nullable fields (during an upgrade existing portals will have null on these fields) so I figured it'd be the same.

Website/DotNetNuke.Website.csproj
donker: I can rename that. Just need to figure out how my local build-to-test process works after that.

Website/admin/Skins/Privacy.ascx.cs
donker: The values get read as NullIntegers when null in the DB AFAIK. I'd need to get the tests working to confirm.

Website/DotNetNuke.Website.csproj
sleupold: IMO we don't care whether it is named 09.03.00.sqldataprovider or 09.03.00.01.sqldataprovider, though I prefer am the second option, reserving the first of upgrade of version numbers.
Once the SQL files are restructured, we may omit the subversion files.

Website/admin/Skins/Privacy.ascx.cs
sleupold: I agree with Peter, there is no difference for the .Net code, as null values read from the database are converted to Null.NullInteger = -1. This is one of our old burdens, as early versions of .Net did not support nullable data types. From a database perspective, it is preferable to save unset values as Null instead of a fake value, as a) this is more obvious to be not a value and b) you may place a FK on the column, e,g, to Tabs.TabID in this case. 
Note: in general, no value is saved as -1 - however, for roleID, UserID and PortalID, -1 has a specific meaning (All Users, unauthenticatedUser, Host Portal)
During a cleanup process of the database I plan to replace inappropriate -1 values by Null.

donker:
Superceded by https://github.com/dnnsoftware/Dnn.Platform/pull/2369

PR #2365, opened 2018-10-05
DNN23855_Clear user cache onlogin failure

### LINKED ISSUE #2343: User Lockout value is not reflected until the cache is manually cleared.

Mhtshum:
Fixes #2343

## Summary
After max invalid attempts reached user still Unlock in Admin UI, though membership table showing Locked out true.

### DIFF
--- a/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
+++ b/DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
@@ -707,6 +707,12 @@ namespace DotNetNuke.Security.Membership
             return settings[settingKey] == null ? string.Empty : settings[settingKey].ToString();
         }
 
+        private bool IsUserLocked(string userName)
+        {
+            var aspnetUser = System.Web.Security.Membership.GetUser(userName);
+            return aspnetUser?.IsLockedOut ?? false;
+        }
+
         #endregion
 
         #region Public Methods
@@ -1830,9 +1836,7 @@ namespace DotNetNuke.Security.Membership
                         }
                     }
 
-                }
-
-                DataCache.ClearUserCache(portalId, username);
+                }                
 
                 //Verify User Credentials
                 bool bValid = false;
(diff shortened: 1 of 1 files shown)


tpluscode:
@Mhtshum, you reference an issue which was already fixed by another by another PR. please advise

Mhtshum:
Actually, QA point out issue, so that PR was not successfully resolving the issue, as upon maxattempts i.e. 5. Admin UI still showing user unlock true. And upon sixth time it was showing true, which was not required.

DNN Platform/Library/Security/Membership/AspNetMembershipProvider.cs
```diff
@@ -1841,7 +1839,11 @@ public override void UpdateUsersOnline(Hashtable userList)
                 {
                     //Clear the user object
                     user = null;
-                }                              
+
+                    //Clear cache for user so that locked out & other status could be updated
+                    DataCache.ClearUserCache(portalId, username);
+                    DataCache.ClearCache(GetCacheKey(username));
```
tpluscode: What do those two calls clear? Untrained eye could think that the first call should be enough to "clear user cache"

tpluscode: Also, why is cache cleared inside the `if (!bValid)` block?

Mhtshum: Its mean Username or Email exist but it has failure because of password mismatched or any other reason, clearing these cache will make framework to retrieve fresh info regarding lockedout or other.

PR #2368, opened 2018-10-06
Checking for option of EnableEmailasUsername before validating user

### LINKED ISSUE #2367: When Username field is changed to Email address then system allows user to login into the application by entering Username into the email address field.
## Description of bug
For login page, when Username field is changed to Email address then system allows user to login into the application by entering Username into the email address field.

This will happen only one time.

## Steps to reproduce
1. Login as the superuser when user can login into the application using username.
2. Change the settings, to use the email address as username.
3. Logout from application.
4. Now Enter the username of newly created user/existing user in the email address field. (Refer Precondition) In my case its content editor.
5. Enter the password for newly created user.(content editor)
6. Hit the login button. 

## Current result
System allows user to enter into the application when we enter Username into the Email address field of login page.
This happen only once.
After that if we logged out and try to login in the application again by entering Username into the Email address field then system gives error of Invalid Username/Password. 

## Expected result
System should not allow user to enter into the application when we enter Username into the Email address field of login page. 

## Known Issue
When EnableEmailasUsername, host user not able to sign in if installation is done without wizard. That issue has been reported in  #2345 

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

Mhtshum:
Fixes #2367 

## Summary
If EnableEmailasUsername is set, then verify user obtained by GetbyEmail must not be null.

### DIFF
--- a/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
+++ b/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
@@ -264,17 +264,29 @@ namespace DotNetNuke.Modules.Admin.Authentication.DNN
 
                 //DNN-6093
                 //check if we use email address here rather than username
-                if(PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false))
+                UserInfo userByEmail = null;
+                var emailUsedAsUsername = PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false);                
+
+                if (emailUsedAsUsername)
                 {
-                    var testUser = UserController.GetUserByEmail(PortalId, userName); // one additonal call to db to see if an account with that email actually exists
-                    if(testUser != null)
+                    // one additonal call to db to see if an account with that email actually exists
+                    userByEmail = UserController.GetUserByEmail(PortalId, userName);                     
+
+                    if (userByEmail != null)
                     {
-                        userName = testUser.Username; //we need the username of the account in order to authenticate in the next step
+                        //we need the username of the account in order to authenticate in the next step
+                        userName = userByEmail.Username; 
                     }
                 }
 
-				var objUser = UserController.ValidateUser(PortalId, userName, txtPassword.Text, "DNN", string.Empty, PortalSettings.PortalName, IPAddress, ref loginStatus);
-				var authenticated = Null.NullBoolean;
+                UserInfo objUser = null;
+
+                if (!emailUsedAsUsername || userByEmail != null)
+                {
+                    objUser = UserController.ValidateUser(PortalId, userName, txtPassword.Text, "DNN", string.Empty, PortalSettings.PortalName, IPAddress, ref loginStatus);
+                }
+
+                var authenticated = Null.NullBoolean;
 				var message = Null.NullString;
 				if (loginStatus == UserLoginStatus.LOGIN_USERNOTAPPROVED)
 				{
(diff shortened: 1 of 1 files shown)


Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
```diff
@@ -264,17 +264,29 @@ private void OnLoginClick(object sender, EventArgs e)
 
                 //DNN-6093
                 //check if we use email address here rather than username
-                if(PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false))
+                UserInfo userByEmail = null;
+                var isUserEmailAsUserName = PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false);
```
tpluscode: Readability nit: I would rename to `emailUsedAsUsername`

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
```diff
@@ -264,17 +264,29 @@ private void OnLoginClick(object sender, EventArgs e)
...
-                if(PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false))
+                UserInfo userByEmail = null;
+                var isUserEmailAsUserName = PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false);                
+
+                if (isUserEmailAsUserName)
                 {
-                    var testUser = UserController.GetUserByEmail(PortalId, userName); // one additonal call to db to see if an account with that email actually exists
-                    if(testUser != null)
+                    // one additonal call to db to see if an account with that email actually exists
+                    userByEmail = UserController.GetUserByEmail(PortalId, userName);
```
tpluscode: Good rename here ✨

Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
```diff
@@ -264,17 +264,29 @@ private void OnLoginClick(object sender, EventArgs e)
...
+                        userName = userByEmail.Username; 
                     }
                 }
 
-				var objUser = UserController.ValidateUser(PortalId, userName, txtPassword.Text, "DNN", string.Empty, PortalSettings.PortalName, IPAddress, ref loginStatus);
-				var authenticated = Null.NullBoolean;
+                UserInfo objUser = null;
+
+                if (!isUserEmailAsUserName || userByEmail != null)
+                {
```
tpluscode: I don't understand why you added this logical statement. Please explain

Mhtshum: When username not email and isUserEmailasUsername(isUserEmailasUsername) also true, was going into this logic

tpluscode: I don't follow. In every case we want to validate the user credential. 

With current code here's what we get

|| isUserEmailAsUserName | userByEmail | userName is | ValidateUser called? | | 
|--| -- | -- | -- | -- | -- | 
|1| true | null | original | no  | ❌ |
|2| true | not null | `userByEmail.Username` | yes | ✅ |
|3| false | null | original | yes | ✅ |
|4| false | not null | original |yes | ✅ |

I'm concerned with row `1`. Please correct me if I'm wrong. I read here that when we enable emails as user names but fail to find a user with that email, the code will never call `UserController.ValidateUser`.

This will be correct if the `Registration_UseEmailAsUserName` means that only email is allowed? Users cannot log in with both email and actually username? @dnnsoftware/tag

Mhtshum: Yes, exactly It is the case that if the Registration_UserEmailAsUsername is true than login only by email not by username, which is #2367 is about that user was getting logged-in by username once and if you check application is turning username into the email when this UseEmailasusername is true

tpluscode: It could make sense but those are significant changes and I don't think worth the trouble and risk

bdukes:
I assume the build is broken because of #2370, not the changes to this PR, right?

tpluscode:
> I assume the build is broken because of #2370, not the changes to this PR, right?

That's right @bdukes. @Mhtshum please merge in master to get us a green build

Mhtshum:
Please @bdukes  and @tpluscode  merged it as PR is raised by me.

PR #2370, opened 2018-10-07
CI: Fixing broken builds due to change in build number variable

ohine:
#2270 changed the `buildNumber` variable to `GetBuildNumber()`, and #2323 was merged which was submitted and merged using the original variable.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -222,7 +222,7 @@ Task("CreateNugetPackages")
 		//basic nuget package configuration
 		var nuGetPackSettings = new NuGetPackSettings
 		{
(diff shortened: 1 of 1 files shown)


ohine:
> It's not ideal to submit PRs directly from between `development` branches because that way they will diverge once it's merge. Especially awkward if we rebase/squash

My fault. I didn't realize I was back on development. I can redo this on another branch, or I can rebase my development after this is merged... right?

tpluscode:
Yea, you can always fix you fork's develop to match origin

PR #2371, opened 2018-10-07
Bug/dnn 9651: drop duplicate index and unused column 

sleupold:
## Summary
I am trying to resubmit https://github.com/dnnsoftware/Dnn.Platform/pull/1696 - or at least, half of it.

Fixing a long outstanding and reported issue, documented in Jira, which failed to be accepted due to the transition. Unnecessary calculated column (LowerEmail) was added  in DNN 7.2.0  and an index (on Email column, not LowerEmail). Since we are always (sic!) using a case insensitive collation, all search or filter ignore case of any character, i.e. Index 'LowerEmail' is already covered by Index 'Email' - and the unused calculated field and index are just useless and hurting performance (this happens, when .Net developers don't take the time to get familiar with SQL and the database server).


sleupold:
@ohine this is a fix for a long outstanding and reported issue, documented in Jira, which failed to be accepted due to the transition. If you'd understand SQL, you'd see that in DNN 7.2, an unnecessary calculated column was added (LowerEmail) and an index (on Email column, not LowerEmail). Since we are always (sic!) using a case insensitive collation, all search or filter ignore case of any character, i.e. Index 'LowerEmail' is already covered by Index 'Email' - and the unused calculated field and index are just useless and hurting performance.

ohine:
@sleupold I understand SQL great, the point is we shouldn't have to physically read the code when writing the release notes. **You need to describe it**. That's part of the pull request process that you helped define in the TAG meetings. 

Having a title from a 3rd party system is pointless for us. In addition, you have given other pull requests a hard time for doing the exact same thing. 

It really shouldn't be that difficult to properly describe in English what your suggested change is doing.

sleupold:
it seems to be useless to provide pull request here, sorry

ohine:
It's too bad that you still continue refusing to work as a team player @sleupold. It really isn't that hard to follow the guidelines and provide the basic documentation on your suggested change.

donker:
> If you'd understand SQL, you'd see that ...

That's uncalled for Sebastian. Oliver is trying to do his work.

PR #2375, opened 2018-10-08
DNN-25715: return the interface list to match previous api.

### LINKED ISSUE #2374: CKEditor file browse page throw exception
1. install platform latest build and login as host;
2. edit one of the html module on home page;
3. enter the full edit window, click the Image button, then click "Browse Server" button.

Expected Result:
the file browse page works correctly.

Actually Result:
image

zyhfish:
Fix #2374 .

### DIFF
--- a/DNN Platform/Library/Services/FileSystem/FolderManager.cs
+++ b/DNN Platform/Library/Services/FileSystem/FolderManager.cs
@@ -619,7 +619,8 @@ namespace DotNetNuke.Services.FileSystem
         {
             Requires.NotNull("folder", folder);
 
-            return CBO.Instance.FillCollection<FileInfo>(DataProvider.Instance().GetFiles(folder.FolderID, retrieveUnpublishedFiles, recursive));
+            return CBO.Instance.FillCollection<FileInfo>(DataProvider.Instance().GetFiles(folder.FolderID, retrieveUnpublishedFiles, recursive))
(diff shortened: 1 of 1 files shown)


tpluscode:
This is not a good fix, we should not be catering for bad code in CKEditor. 

The `GetFiles` method returns `IEnumerable<IFileInfo>`. The problem is not the use of `IFileInfo` vs `FileInfo`. The problem is that CKEditorProvider attempts to [cast the collection back to a concrete list](https://github.com/DNN-Connect/CKEditorProvider/blob/development/Browser/Browser.aspx.cs#L273)

```c#
var files = (List<IFileInfo>)FolderManager.Instance.GetFiles(currentFolderInfo);
```

The path of least resistance is to change the usage of `GetFiles` in the provider so that it recreates the list there. Otherwise it's assuming the returned object is a `List<>` which is programming to a concrete type and not to the interface.

```c#
var files = FolderManager.Instance.GetFiles(currentFolderInfo).Cast<IFileInfo>().ToList();
```

I vote to close @dnnsoftware/tag

valadas:
I agree the fix should be done in the provider as @tpluscode suggests.

mitchelsellers:
I agree as well, closing.

zyhfish:
Changes made in CK editor provider side: https://github.com/DNN-Connect/CKEditorProvider/pull/94

PR #2388, opened 2018-10-11
Member Directory Search: adding unused text to page

### LINKED ISSUE #2391: Member Directory: Basic Search: GetSuggestions / AutoComplete adding text to page
## Description of bug
Provide a clear and concise description of what the bug is.

## Steps to reproduce
Install latest version of Evoq Engage

1. Create some users in your site (about 10). Let's call one 'Test Test' 
2. Create a new page and add the Member Directory. 
3. In the Basic Search start typing 'Test' You will get the suggestion for 'Test Test' 
4. Press Enter/Search 
5. Refresh Directory - Icon to the left of the search bar

## Current result
The User Name text is also added to the bottom of the page after the </form> tag. You either get the 'Test Test' suggestion or the number of suggestions returned such as. '1 Ocurrence ...' 

screenshot-1-3

Additionally, once the directory is refreshed, the page height increases and the text is pushed top

screenshot-2

## Expected result
The text should not be added to the page and only under the search bar for suggestions. 

This text is added during the AJAX response to getSuggestions in MemberDirectory.js

## Affected version

* [x] 9.2.2
* [x] 9.2.1

mikebigun:
fixes #2391

### Summary

When searching member directory users, once autocomplete popup opened, application is adding the text next to the main form.

### Steps to reproduce

1. Create multiple dnn users
2. Create a new page
3. Add Member Directory module to a newly created page, click on publish
4. Start typing user name in a search box
5. Observe the text "..results are available, use up and down key" which is added at the bottom of the page

### Rootcause
Text which is added is a part of [Live Region](https://www.oreilly.com/library/view/javascript-cookbook/9781449390211/ch14s09.html). It is used to voice additional info by screen readers. 
Live Region is marked with `.ui-helper-hidden-accessible` style, but it is not defined on a module level css. Similarly to other modules, we should hide helper text `display:none;`

### Sample of the issue
image

### DIFF
--- a/DNN Platform/Modules/MemberDirectory/module.css
+++ b/DNN Platform/Modules/MemberDirectory/module.css
@@ -340,3 +340,6 @@ ul.ui-autocomplete li{margin:0;padding:0;list-style:none;}
 .dnnForm .dnnFormItem .mdFilterLists input, .dnnForm .dnnFormItem .mdFilterLists select { width: 45%;}
 .dnnForm .dnnFormItem .mdFilterBy { background-color: #F0F0F0; margin-left: 36%; width: 50%;}
 
+.ui-helper-hidden-accessible {
+	display:none;
+}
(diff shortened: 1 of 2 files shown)


mikebigun:
@tpluscode, you are right. Instead of working with this issue tonight, I had to go to bed :)

mikebigun:
`.ui-helper-hidden-accessible` should be defined on a module level css. I found in a same way it is done for a couple of other modules (Journal, ActivityStream, SocialLibrary). I.e. [Journal module](https://github.com/dnnsoftware/Dnn.Platform/blob/3c5357f5794151bc1cfa7ba9e5ae9b4a9c78f81d/DNN%20Platform/Modules/Journal/module.css#L281).
In all cases we're just hiding helper text.

Result:
image

PR #2417, opened 2018-10-19
- Replacing http to https if SSL is setup

### LINKED ISSUE #2416: SSL Offloading: Canonical Link still shows as HTTP.
SSL Offloading: Canonical Link still shows as HTTP.

## Description of bug
In SSLOffloading environment, when adding new alias, then automatically generated urls are shown with http

## Steps to reproduce
1. Set up SSL Offloading for your evoq site. Use a different alias/binding than the one current used such as proxytest.local - https://dnnsupport.dnnsoftware.com/hc/en-us/articles/360004929513-How-to-to-emulate-the-behavior-of-a-Load-Balancer-on-a-single-local-machine 

2.Go to the DNN site and add the proxytest.local domain as a new alias. DO NOT mark it as primary.
3. In DNN go to Settings > Security > More, and enable SSL and add the offload header value: "X-Forwarded-For"
4. Visit the site using https://proxytest.local which should take you to the DNN site.
5. Inspect element to find the canonical reference

## Current result
The canonical reference uses HTTP

## Expected result
The canonical link should be generated using HTTPS which is the default behavior when not using SSL Offloading and simply SSL.

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

Mhtshum:
Fixes #2416 

## Summary
Replace http with https when SSL is setup

### DIFF
--- a/Website/Default.aspx.cs
+++ b/Website/Default.aspx.cs
@@ -658,6 +658,11 @@ namespace DotNetNuke.Framework
                 {
                     var originalurl = Context.Items["UrlRewrite:OriginalUrl"].ToString();
                     CanonicalLinkUrl = originalurl.Replace(PortalSettings.PortalAlias.HTTPAlias, primaryHttpAlias);
+
+                    if (UrlUtils.IsSecureConnectionOrSslOffload(Request))
+                    {
+                        CanonicalLinkUrl = CanonicalLinkUrl.Replace("http://", "https://");
+                    }
                 }
             }
 


Website/Default.aspx.cs
```diff
@@ -658,6 +658,11 @@ protected override void OnInit(EventArgs e)
                 {
                     var originalurl = Context.Items["UrlRewrite:OriginalUrl"].ToString();
                     CanonicalLinkUrl = originalurl.Replace(PortalSettings.PortalAlias.HTTPAlias, primaryHttpAlias);
```
tpluscode: Is this the only location where `CanonicalLinkUrl` is being set? Why only when `primaryHttpAlias != null`?

Website/Default.aspx.cs
```diff
@@ -658,6 +658,11 @@ protected override void OnInit(EventArgs e)
                 {
                     var originalurl = Context.Items["UrlRewrite:OriginalUrl"].ToString();
                     CanonicalLinkUrl = originalurl.Replace(PortalSettings.PortalAlias.HTTPAlias, primaryHttpAlias);
+
+                    if (UrlUtils.IsSecureConnectionOrSslOffload(Request))
+                    {
+                        CanonicalLinkUrl = CanonicalLinkUrl.Replace("http://", "https://");
+                    }
```
tpluscode: Wouldn't we want this to be outside of the current `if` statement?

daguiler: Well, since this is the only place where the CanonicalLinkUrl is set, I think it makes sense the way it is.

PR #2420, opened 2018-10-22
DNN-26075: sort the files list to alphabetically order.

### LINKED ISSUE #2419: Install packages don't have alphabetically file-lists on nas system
## Description
website files hosted on a some certain nas system, which will cause Directory.GetFiles not return alphabetically file-lists.

## Steps to reproduce
install the website which files hosted on a special nas system.

## Current result
the packages install order will be randomly but not in alphabetically order.

## Expected result
packages should install with in alphabetically order.

## Affected version
* [x] All Versions

zyhfish:
Fix #2419 .


tpluscode:
Please have a look @dnnsoftware/approvers.

IMO it doesn't hurt. In most cases the array will already be sorted and the additional operation will be cheap

Issue #2185, opened 2018-07-27
LockRecursionException thrown from various locations


tpluscode:
## Description 

We've just had a report from an Evoq customer who have a busy site and observe intermittent downtime caused by "thread-safe" code. Quick google finds a whole lot of similar reports dating back to 2011. Let's put this one to bed once and for all.

https://www.dnnsharp.com/helpcenter/url-adapter/recursive-read-lock-acquisitions-not-allowed-in-this-mode
http://www.dnnsoftware.com/forums/forumid/108/threadid/412632/scope/posts
http://www.dnnsoftware.com/answers/help-with-error-recursive-read-lock-acquisitions-not-allowed-in-this-mode
https://dnntracker.atlassian.net/browse/DNN-24111

Looking at the most recent report we received, out of 57k exceptions logged (40 per minute!) there are two prime sources which fail to acquire a lock:

* [`DotNetNuke.UI.Containers.Container.InvokeContainerEvents`][1] (22k)
* [`DotNetNuke.ComponentModel.SimpleContainer.GetComponentType`][2] (34k)

[1]: https://github.com/dnnsoftware/Dnn.Platform/blob/f3dde5a325b3043ebf508ad3721c5e789283f65b/DNN%20Platform/Library/UI/Containers/Container.cs#L501
[2]: https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/ComponentModel/SimpleContainer.cs#L138

### `InvokeContainerEvents`

The exception is logged in various places in the UI lifecycle. For example

```
DotNetNuke.Services.Exceptions.Exceptions - System.Threading.LockRecursionException: Recursive read lock acquisitions not allowed in this mode.
   at System.Threading.ReaderWriterLockSlim.TryEnterReadLockCore(TimeoutTracker timeout)
   at System.Threading.ReaderWriterLockSlim.TryEnterReadLock(TimeoutTracker timeout)
   at DotNetNuke.Collections.Internal.ReaderWriterLockStrategy.GetReadLock(TimeSpan timeout)
   at DotNetNuke.UI.Containers.Container.InvokeContainerEvents(ContainerEventType containerEventType)
   at System.Web.UI.Control.InitRecursive(Control namingContainer)
   at System.Web.UI.Control.AddedControl(Control control, Int32 index)
   at DotNetNuke.UI.Skins.Pane.InjectModule(ModuleInfo module)
```

[...]

PR #2423, opened 2018-10-22
Recursive read lock acquisitions not allowed

mikebigun:
fixes [issue 2185](https://github.com/dnnsoftware/Dnn.Platform/issues/2185)


mitchelsellers:
Pending customer verification per @daguiler will follow-up again no later than 1/1/2019

daguiler:
@mitchelsellers We had preliminary confirmation from customer three weeks ago that this worked, and I'm touching base with them again to confirm there were no negative side effects.

daguiler:
We got confirmation from 2 customers that this patch works. They have patched their production environments and they've been stable for several weeks now.
I think we should merge this @dnnsoftware/approvers

daguiler:
Looks like we have a few outdated unit tests breaking the build. Will take care.

PR #2456, opened 2018-11-06
Defect/dnn 26006

### LINKED ISSUE #2457: Activity feed notification sent using the Site's UI language
## Description of bug

Activity feed notification emails are sent with a link using the UI language setting at the moment the post was generated, regardless of the site default language settings or the user's default language settings.

## Steps to reproduce

1. Enable a second language in a DNN site. 
2. Add two test users. 
3. Make the users follow each other's update in the Activity Feed. 
4. Change the UI language to the secondary language. 
5. Create a new post in the activity feed. 
6. The respective notification email will have a link to the post using the language from the UI at the moment of the post creation, and not the default language of the site, or from the user's preferred language. 

## Current result

Activity feed notification emails are sent with a link using the UI language setting at the moment the post was generated, regardless of the site default language settings or the user's default language settings.

## Expected result

The expected result is that a subscribed user who views a notification will view the related content in the language of their preference or in the site's default language.

## Screenshots
test4-mail
test3-mail
test2-mail
test1-mail

## Error log

## Additional context

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [ ] 9.0

mean2me:
Read default language from PortalSettings instead of current session

fixes #2457


DNN Platform/Library/Common/Globals.cs
```diff
@@ -811,7 +811,7 @@ internal static string GetCultureCode(int TabID, bool IsSuperTab, PortalSettings
                 }
                 if (string.IsNullOrEmpty(cultureCode))
                 {
-                    cultureCode = Thread.CurrentThread.CurrentCulture.Name;
+                    cultureCode = PortalController.Instance.GetCurrentPortalSettings().DefaultLanguage;
```
tpluscode: Should it be portal settings or user's personal settings? @mean2me

mean2me: portal settings is correct, IMO, since user's personal settings could be any language and, for example, in this case all Notifications will be sent according to that language, regardless of the recipient of that notification.

tpluscode: I see, that makes sense. And is there a way to send them in the recipient's language?

mean2me: @tpluscode Yes, with some code refactoring. There is a PR limited to Evoq/Social at the moment.

tpluscode:
Please review @dnnsoftware/approvers

Issue #2474, opened 2018-11-12
Unable to create a user using domain\username format


mikebigun:
## Description of bug
The customer is not able to add a user with the `domain\username` format as username. The following Error Message is displayed: "The username specified is invalid. Please specify a valid username."

## Steps to reproduce
1. Set up a clean DNN 9.2 install. 
2. Go to `Persona Bar > Manage > Users`. 
3. Click the `Add User` button. 
4. Fill details for the new user, using the `domain\username` format. 
5. Click `Save` button. 
6. An error message is displayed saying: "The username specified is invalid. Please specify a valid username."

## Expected result
To allow the customer to be able to add a user using the `domain\username` format as a valid username in the default user creation form.

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

PR #2475, opened 2018-11-12
Unable to create a user using domain\username format

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2474

### DIFF
--- a/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
+++ b/Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
@@ -378,6 +378,23 @@ CREATE PROCEDURE {databaseOwner}[{objectQualifier}Journal_ListForGroup]
 	ORDER BY RowNumber ASC;
 GO
 
+IF EXISTS (SELECT * FROM dbo.sysobjects
+	WHERE id = object_id(N'{databaseOwner}[{objectQualifier}HostSettings]') and OBJECTPROPERTY(id, N'IsUserTable') = 1)
+BEGIN
+	IF NOT EXISTS (SELECT * FROM {databaseOwner}[{objectQualifier}HostSettings] WHERE [SettingName] = N'UsernameUnallowedCharacters')
+		INSERT INTO {databaseOwner}[{objectQualifier}HostSettings] 
+			([SettingName]
+			,[SettingValue]
+			,[SettingIsSecure]
+			,[CreatedByUserID]
+			,[CreatedOnDate]
+			,[LastModifiedByUserID]
+			,[LastModifiedOnDate])
+		VALUES (N'UsernameUnallowedCharacters', N'!"#$%&''()*+,/:;<=>?[\]^`{|}', 0, -1, GETDATE(), NULL, NULL)
+END
+
+GO
+
 /************************************************************/
 /*****              SqlDataProvider                     *****/
 /************************************************************/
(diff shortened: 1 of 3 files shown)


valadas:
I am totally not sure here, but could this have been intentional so that active directory users could only register accounts through the active directory provider?

Kind of a way to reserve \ for active directory to prevent a user name already being taken when adding active directory to a site.

mikebigun:
The main problem is that current implementation blocks customers in few use causes when user is created using 'Add new user' form.
List of unallowed ascii characters was introduced in https://github.com/dnnsoftware/Dnn.Platform/pull/1930 and changed registration behavior. Migration from 9.1.1 to 9.2.2 makes customer unable to proceed with a few critical scenarios.
To my opinion this is kind of regression when some use cases were not considered implementing this change. 

Another option is to add additional optional UI field for the domain input.

sleupold:
I would prefer having this regex being stored in the database (HostSettings) or a config file instead of hard coded in the API. Opinions?

mikebigun:
@sleupold great idea! I'll change it in that way if no more options/ideas from community.

mikebigun:
@sleupold Am I right that the latest script is 09.03.00.01?
https://github.com/dnnsoftware/Dnn.Platform/blob/development/Website/Providers/DataProviders/SqlDataProvider/09.03.00.01.SqlDataProvider

Or new host settings should go to 09.03.00?

sleupold:
@mikebigun 
great, thank you.
it doesn't matter, which file you are using. I suggest 09.03.00, the idea of incremental upgrade files never took off and I am having a better strategy for the future,

Website/Providers/DataProviders/SqlDataProvider/09.03.00.SqlDataProvider
```diff
@@ -378,6 +378,23 @@ CREATE PROCEDURE {databaseOwner}[{objectQualifier}Journal_ListForGroup]
 	ORDER BY RowNumber ASC;
 GO
 
+IF EXISTS (SELECT * FROM dbo.sysobjects
```
sleupold: I'd Omid this check - if HostSettings table is missing, the script should fail, because there is a serious problem.

valadas:
Looks like a good solution, allows people who have special needs to override and does not break anything for the others.

However, this same string is used in https://github.com/dnnsoftware/Dnn.Platform/pull/1930/files#diff-d8bb34eba9af04d7a2a2ba4f19dfa0a2

So to prevent having different behaviours when creating and editing usernames (what this change was originally made for), I would also use that host setting there instead of the hardcoded string.

Finally, maybe we should deprecate Globals.glbUserNameRegEx since that is only an empty regex... Opinions?

mikebigun:
Hi @valadas it looks like you are checking outdated version. See below
https://github.com/dnnsoftware/Dnn.Platform/blob/2f03c8341ea263616cfee3ea3a14b631f42b94ea/DNN%20Platform/Library/Security/Membership/AspNetMembershipProvider.cs#L760

We use same validation from the UserController. So, no issues.

sleupold:
@valadas I would not deprecate Globals.glbUserNameRegEx, even if it is an empty string. This could remain as "documented" default, if not overridden in HostSettings.

valadas:
@mikevigun sorry for that , I did a search and failed to notice it returned an old commit. You are right.

@sleupold I am not familiar enough with the active directory providers to have an opinion on that. My guess is that you can have regular users and AD users and that the AD provider has its own UI where it can set those requirements. But like I said, not familiar enough so my guess could be totally wrong.

However if wrong I still think this PR is an improvement for now and we can create a new issue for the regex?

mikebigun:
@valadas fully agree with you. For the new scenarios related to AD we should go with another issue.
With this one, we should unblock those customers who has an issues during registrations after migration from 9.1.1 to 9.2.2. 

CC: @daguiler

daguiler:
Do you approve @valadas ?

PR #2482, opened 2018-11-19
Customer defect/dnn 26844

mean2me:
Total results count was inconsistent for paginated results.


DNN Platform/Library/Services/Search/Internals/SearchSecurityTrimmer.cs
```diff
@@ -178,11 +178,12 @@ private void PrepareScoreDocs()
                         continue;
                     }
 
-                    _scoreDocs.Add(scoreDoc);
-                    if (++collectedSoFar == pageSize)
+                    if (++collectedSoFar >= pageSize)
```
daguiler: There is a bug here. The `collectedSoFar` counter is incremented but then the document might not be added. This fails when pageSize is 1 for example. So this should be compared using `>` instead of `>=`. But that would change the meaning of `collectedSoFar`, so to preserve the correct meaning I suggest following implementation:
```c#
                    if (collectedSoFar >= pageSize)
                    {
                        continue;
                    }

                    _scoreDocs.Add(scoreDoc);
                    collectedSoFar++;
```

daguiler:
You're right @bdukes, we should have created an issue for this PR. I'll take care of it.

daguiler:
Created https://github.com/dnnsoftware/Dnn.Platform/issues/2490

Issue #2487, opened 2018-11-25
Account Login Module: Login Button Disabled


mikebigun:
## Description of bug
Login button in the DNN Login module gets disabled (grayed out) after multiple failed login attempts.

## Steps to reproduce
1. Login to the site as Superuser/Host 
2. Create a new test page 
3. Add the `Account Login` module to the newly created test page 
4. Logout from the site 
5. Go to the newly created test page and login via the login module you added 
6. Simulate a failed login attempt by entering either an incorrect username or password 
7. The login button gets disabled on the second failed login attempt; it is grayed out and is not clickable 
8. Refresh the page and see the login button is clickable again (not grayed out) 

DEMO is available [here](https://www.useloom.com/share/1af37607476244cdaecb8521fc9b0315).

## Current result
Login button in the DNN Login module gets disabled (grayed out) after multiple failed login attempts.

## Expected result
Login button should not be disabled after successive failed login attempts.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

PR #2488, opened 2018-11-25
Account Login Module: Login Button Disabled

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2487

Login form action buttons (login, cancel, register, reset password) gets disabled every time user clicks on a particular action link. This behavior is expected and needed to avoid double click to raise same request multiple times.
When post back request completed the state of the buttons is not restored. In this PR it is fixed so all action links are enabled at the end of every post back request.

### DIFF
--- a/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
+++ b/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
@@ -44,14 +44,33 @@
     <script type="text/javascript">
         /*globals jQuery, window, Sys */
         (function ($, Sys) {
-            function setUpLogin() {
-                var actionLinks = $("a#dnn_ctr<%=ModuleId > Null.NullInteger ? ModuleId.ToString() : ""%>_Login_Login_DNN_cmdLogin");
-                actionLinks.click(function () {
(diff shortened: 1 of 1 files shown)


Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
```diff
@@ -44,26 +44,44 @@
...
-                    actionLinks.addClass("dnnDisabledAction");
+            const disabledActionClass = "dnnDisabledAction";
+            const actionLinks = $('a[id^="dnn_ctr<%=ModuleId > Null.NullInteger ? ModuleId.ToString() : ""%>_Login_Login_DNN"]');
+            function isActionDisabled($el) {
+                return $el && $el.hasClass(disabledActionClass);
+            }
+            function disableAction($el) {
+                if ($el == null || $el.hasClass(disabledActionClass)) {
+                    return;
+                }
```
daguiler: @mikebigun You can use your new function above to reduce redundancy:
```javascript
if (isActionDisabled($el))
{
    return;
}
```

kurtwilbies:
Actually the whole module needs a rewrite. There are too may issues.

Issue #2424, opened 2018-10-24
Send An email to new user is not working


Icidis:
## Description of bug
When creating a new user account as a SuperUser/Admin and checking the "Send An Email To New User." checkbox, no email is sent to the email I specify.

Once the user is created, sending password reset links, etc. work fine. Emails are sent and received correctly so it appears the SMTP settings are correct.

## Steps to reproduce
List the steps to reproduce the behavior:
1. Go to 'Manage Users'
2. Click on 'Add User'
3. Fill in user details
4. Check the "Send An Email To New User" checkbox
5. Click "Save"

## Current result
The user is created successfully but no email is send to the user.

## Expected result
User gets created and registration email is sent to the users' email address.

## Additional context
SMTP is configured to use SendGrid smtp but password reset etc do work without any issues.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2

## Affected browser

* [x] Chrome
* [x] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [x] Edge

sleupold:
Which registration type ate you using - public, verified or private?

Icidis:
Hi Sebastian (@sleupold),

The "User Registration" setting is set to "None"

Tychodewaard:
I have a customer who has 'verified' as user registration with the same issue.

sleupold:
AFAIR there has been already a PR merged in this area, I am not sure, whether it fixes this issue as well. It would be great if anyone could test it in a nightly build.

Icidis:
I've upgraded the site from version 9.2.2 to the Nightly Build [9.3.0-alpha.31](https://dotnet.visualstudio.com/DNN/_build/results?buildId=6055&view=results) and the same issue persists.

jacques-hoventer:
Is there any known workaround for this?

mikebigun:
Dear community, 

Email notification is not handled for the User Registration `None` type. Initially, if `None` is set, we shouldn't allow anyone to create users. 
Nevertheless, I found property `IgnoreRegistrationMode` which overwrites above logic and makes us ALWAYS possible to create new users. See below:

https://github.com/dnnsoftware/Dnn.AdminExperience/blob/95513df2b8440bc0d37316545d6dd327a3f30506/Extensions/Manage/Dnn.PersonaBar.Users/Components/Contracts/RegisterationDetails.cs#L41

[...]

PR #2492, opened 2018-11-29
Initial New User Email Not Sending At Time of Creation

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2424

This is alternative way to fix above issue proposed in https://github.com/dnnsoftware/Dnn.AdminExperience/pull/174

As per @sleupold , we need to move email notifications from UI to core part.
Once this will be approved and merged, we can remove email notifications from UI and replace it with updated controller method to let notifications to be send to their recipients.

FYI: @Icidis

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserEventArgs.cs
+++ b/DNN Platform/Library/Entities/Users/UserEventArgs.cs
@@ -26,6 +26,6 @@ namespace DotNetNuke.Entities.Users
     public class UserEventArgs : EventArgs
     {
         public UserInfo User { get; set; }
-        public bool SendEmailNotification { get; set; }
+        public bool SendNotification { get; set; }
     }
 }
--- a/DNN Platform/Library/Entities/Users/UserEventArgs.cs
+++ b/DNN Platform/Library/Entities/Users/UserEventArgs.cs
@@ -26,6 +26,6 @@ namespace DotNetNuke.Entities.Users
     public class UserEventArgs : EventArgs
     {
         public UserInfo User { get; set; }
-        public bool SendNotification { get; set; }
+        public bool SendEmailNotification { get; set; }
     }
 }
--- a/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
+++ b/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
@@ -6,15 +6,15 @@ using static DotNetNuke.Common.Globals;
 
 namespace DotNetNuke.Entities.Users
 {
-    public class UserEmailRegistrationNotifier
+    public class UserRegistrationEmailNotifier
     {
-        private static PortalSettings _portalSettings = PortalSettings.Current;
+        private static PortalSettings __portalSettings = PortalSettings.Current;
 
-        public UserEmailRegistrationNotifier()
+        public UserRegistrationEmailNotifier()
         {
         }
 
-        public static void EmailToAdministrator(UserInfo user)
+        public static void NotifyAdministrator(UserInfo user)
         {
             // avoid self-notification (i.e. on site installation/super user creation)
             var currentUser = UserController.Instance.GetCurrentUserInfo();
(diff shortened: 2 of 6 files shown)


mikebigun:
@sleupold let me know please if this is what you expected.

DNN Platform/Library/Entities/Users/UserEventArgs.cs
```diff
@@ -26,6 +26,6 @@ namespace DotNetNuke.Entities.Users
     public class UserEventArgs : EventArgs
     {
         public UserInfo User { get; set; }
-        public bool SendNotification { get; set; }
+        public bool SendEmailNotification { get; set; }
```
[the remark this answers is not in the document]
mikebigun: No. By default it is false. I'm going to enable it in admin experience when `CreateUser` is invoked on a new user registration. So, email will be sent to admin and registered user.
Also, since we moved this to core part, this flag becomes available to other 3rd part applications and api consumers, so they can use overloaded `CreateUser` method to enable notifications as well.

mikebigun:
@daguiler @sleupold can we have this approved please? On top of this change I'll have to create one more PR in admin experience to use updated api method.

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
```diff
@@ -0,0 +1,59 @@
...
+using DotNetNuke.Common.Utilities;
+using DotNetNuke.Entities.Portals;
+using DotNetNuke.Services.Mail;
+using static DotNetNuke.Common.Globals;
+
+namespace DotNetNuke.Entities.Users
+{
+    public class UserRegistrationEmailNotifier
+    {
+        private static PortalSettings __portalSettings = PortalSettings.Current;
```
bdukes: Having this field be `static` will cause it to never be updated, correct?  It seems like we should just use `PortalSettings.Current` throughout, and/or require a `PortalSettings` instance in the constructor.

DNN Platform/Library/Entities/Users/UserEventArgs.cs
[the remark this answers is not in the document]
sleupold: If UserEventArgs is part of the API, renaming a property would be a breaking change

sleupold: if UserEventArgs is part of the public API, this would be a breaking change

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
mikebigun: great observation.. fixed now!

DNN Platform/Library/Entities/Users/UserEventArgs.cs
mikebigun: fixed!

mikebigun:
Hi @sleupold 

> IMO we need to make sure to keep previous behavior during upgrade scenarios, i.e. insert appropriate portal setting according to previous registration type.

Not really got the idea. Do you mean portal settings gets overwritten and we need restore them after upgrade? Aren't they preserving during upgrade?

sleupold:
@mikebigun previously, emails were sent under specific conditions to users - we need to make sure, this is kept, if we switch the single param method to "false" for the second param - or am I missing sth? if necessary, we may need to set the "send email to user" setting accordingly to the registration type.

mikebigun:
Well, so all previous conditions are kept. Previously, there was a possibility to send registration email from admin experience only. We were sending email using particular body depending on registration type. 
Now, all this staff is moved to api. 
New overloaded method will be invoked in admin experience with a `sendEmailNotification=true`. For now, it is going to be only one consumer of this method. For the rest app, this flag is `false` by default.
But having this method available gives ability to other 3rd party apps to start using it in the next releases. All setting conditions are included in API, so 3rd party tools don't need to care about it. Email will be sent conditionally, depending on settings/registration type. This is a part of API now.

PR #2494, opened 2018-11-30
Resolve UserProfile Loading Errors From Unsecure pages

mean2me:
When we have SSL enabled on single pages we get same-origin errors on loading AdminExperience -> Users -> UserProfile, since it's loaded into an *iframe* from an unsecure source.
This change will detect this specific case and will force iframe (or popup) source to be loaded as a secure page.

**Steps to reproduce**

1. Install DNN under HTTPS
2. Activate **SSL Enabled** and **SSL Enforced** in Settings > Security > More
3. Create a blank page.
4. Make sure **Secure Connection** for the page is enabled in Page Settings > Advanced > More
5. Browse to the page (should navigate under HTTPS)
6. Open PersonaBar > Manage > Users
7. Click on **Profile Settings** icon of any user

Current Behavior:
User profile is not displayed and browser shows following error in console:
```
Mixed Content: The page at 'https://platform-930.dnndev.me/Secure' was loaded over HTTPS, but requested an insecure resource 'http://platform-930.dnndev.me/Host/Superuser-Accounts/ctl/Edit/mid/353/UserId/1/editprofile/true/portalid/0?popUp=true'.
This request has been blocked; the content must be served over HTTPS.
```


DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -1533,6 +1533,10 @@ private static bool CheckForDebug(HttpRequest request, NameValueCollection query
                                              portalSettings.ActiveTab.TabName + " IsSecure: " +
                                              portalSettings.ActiveTab.IsSecure.ToString());
 
+                    // this covers the specific case of popups and iframes opened from a secure tab
+                    // e.g: AdminExperience -> Users -> UsersProfile 
+                    // example url: https://YOUR_HOST/Admin/User-Accounts/ctl/Edit/mid/464/UserId/4/editprofile/true?popUp=true
+                    var isPopUpFromSecurePage = result.IsSecureConnection && UrlUtils.IsPopUp(url);
```
mitchelsellers: Can we confirm, or document here what the Result & Url values are?  It appears from this that the URL is the destination, Result is current page?

DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -1554,8 +1558,10 @@ private static bool CheckForDebug(HttpRequest request, NameValueCollection query
                     //check ssl enforced
                     if (portalSettings.SSLEnforced)
                     {
+                        var isSecureConnectionOnUnsecurePage = !portalSettings.ActiveTab.IsSecure && result.IsSecureConnection;
+                        
                         //check page is not secure, connection is secure
-                        if (!portalSettings.ActiveTab.IsSecure && result.IsSecureConnection)
+                        if (isSecureConnectionOnUnsecurePage && !isPopUpFromSecurePage)
```
mitchelsellers: Is there any other way to prevent this check from being added to all requests?  A higher level change possibly?

mitchelsellers:
@mean2me for all future Pull Requests can we please get detailed titles that are helpful to all reviewing requests.  References to internal ESW tickets that we do not have exposure to make management of the PR's as well as release notes complicated.  I've updated the title on this one for you.

@ashishpd @daguiler can you communicate this to the rest of the team?

mean2me:
@mitchelsellers  You're right. Actually I forgot to adjust PR title. 🙇 👍

DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
mean2me: @mitchelsellers, probably it's possible if we plan an accurate refactoring. Actually that part of the project seems to be quite redundant and easy to corrupt. I would focus on the easiest way to fix the bug here. Any deep refactoring of the core is welcome but needs time and a big effort.

daguiler: To optimize this, instead of initializing `isPopUpFromSecurePage` every time at the top, we can remove that declaration and also the `isSecureConnectionOnUnsecurePage` declaration, and just do:
```c#
if (result.IsSecureConnection && !portalSettings.ActiveTab.IsSecure && !UrlUtils.IsPopUp(url))
```
This way we call `UrlUtils.IsPopUp` only when necessary.

DNN Platform/Library/Properties/AssemblyInfo.cs
```diff
@@ -37,7 +37,7 @@
 [assembly: AssemblyDescription("Open Source Web Application Framework")]
 [assembly: CLSCompliant(true)]
 
-[assembly: AssemblyStatus(ReleaseMode.Alpha)]
+[assembly: AssemblyStatus(ReleaseMode.Stable)]
```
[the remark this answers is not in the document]
daguiler: @mean2me please remove this change.

DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
mean2me: @daguiler Yes, same for me. Just preferred initialization for easy code readability.

DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -1554,8 +1554,11 @@ private static bool CheckForDebug(HttpRequest request, NameValueCollection query
                     //check ssl enforced
                     if (portalSettings.SSLEnforced)
                     {
-                        //check page is not secure, connection is secure
-                        if (!portalSettings.ActiveTab.IsSecure && result.IsSecureConnection)
+                        // Prevent browser's mixed-content error in case we open a secure PopUp or a secure iframe 
+                        // from an unsecure page
+                        if (!portalSettings.ActiveTab.IsSecure && 
+                            result.IsSecureConnection &&
+                            UrlUtils.IsPopUp(url))
```
daguiler: @mean2me I think now the `UrlUtils.IsPopUp(url)` part of the condition got reversed. It should be negated to be equivalent to your initial implementation:
```c#
if (!portalSettings.ActiveTab.IsSecure && result.IsSecureConnection && !UrlUtils.IsPopUp(url))
```

mean2me: @daguiler  I think my initial implementation is the right one. 
Please, have a look at this spreadshit I created: https://docs.google.com/spreadsheets/d/1IPE7QsC1WPiLLlhC4A0j459rw3RN7oTfRh6jv_2YtAQ/edit?usp=sharing
The column **previous** implements a formula reproducing my initial pull request and, actually, it's the expected condition.

mean2me: @daguiler I think current implementation is correct, since it reproduce expected results. If you apply your suggestion, we get a wrong condition.
Please, have a look at this spreadsheet: https://docs.google.com/spreadsheets/d/1IPE7QsC1WPiLLlhC4A0j459rw3RN7oTfRh6jv_2YtAQ/edit?usp=sharing

daguiler: @mean2me Latest results are different from your original results. This is because your latest implementation is not equivalent to your original implementation, or the implementation suggested by me.

Your original implementation was:
```c#
isPopUpFromSecurePage = result.IsSecureConnection && UrlUtils.IsPopUp(url)
isSecureConnectionOnUnsecurePage = !portalSettings.ActiveTab.IsSecure && result.IsSecureConnection

if (isSecureConnectionOnUnsecurePage && !isPopUpFromSecurePage)
{
}
```
Decomposing and simplifying above code, it boils down to:
```c#
if (!portalSettings.ActiveTab.IsSecure && result.IsSecureConnection && !UrlUtils.IsPopUp(url))
{
}
```
Notice the last part is negative, not positive.

Also in your spreadsheet, you are missing the most important combination according to the original implementation: false, true, false.

Here's why I say latest implementation is wrong and my suggestion is equivalent to the original:

dnn-26576

The important condition (to enter the `if` statement) is highlighted in blue.

**So 2 options now:**

1) Original implementation was correct --> Take my suggestion
2) Original implementation was not correct --> Leave as is

Please confirm. Thanks!

mean2me: @daguiler 
The combination I want to be true is (following above columns' order): FALSE, TRUE, TRUE.
It means: 

- *ActiveTab.IsSecure = FALSE*, so I'm opening a *popup* or an *iframe* from an *unsecure page*.

- *IsSecureConnection = TRUE*: PopUp/iFrame content is secure

- *IsPop = TRUE*: It returns true for both *PupUps and iFrames*.

mean2me: @daguiler you're right, I checked the history of the file and I could see my misunderstanding.
Fixed now.

daguiler:
This works as expected now @mean2me. Approving.

daguiler:
@mitchelsellers I added repro steps to the PR's description.

PR #2497, opened 2018-12-03
Module Permissions: wrong behavior on add/remove users when Inheritance is disabled

mean2me:
When disabling the "Inherit View permissions from Page" in any module, the Add User procedure for the module starts to fail, allowing only 2 users to be added to the module at any point. 
Trying to add a 3rd user. This action will randomly remove the view mark from one of the previously added users. 
Marking the view option again for the disabled user, and clicking update would still make the disabled user to disappear from the ModulePermission table in the DB and, subsequently, from the UI.
Also, removing users from modules permissions always remove the first item in the list, regardless of the one we choose.

### DIFF
--- a/DNN Platform/Library/Security/Permissions/ModulePermissionCollection.cs
+++ b/DNN Platform/Library/Security/Permissions/ModulePermissionCollection.cs
@@ -181,13 +181,15 @@ namespace DotNetNuke.Security.Permissions
 
         public void Remove(int permissionID, int roleID, int userID)
         {
+            var idx = 0;
             foreach (PermissionInfoBase permission in List)
             {
                 if (permission.PermissionID == permissionID && permission.UserID == userID && permission.RoleID == roleID)
                 {
-                    List.Remove(permission);
+                    List.RemoveAt(idx);
                     break;
                 }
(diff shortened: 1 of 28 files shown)


DNN Platform/Library/Security/Permissions/ModulePermissionCollection.cs
```diff
@@ -181,13 +181,15 @@ public void Remove(ModulePermissionInfo value)
...
+            var idx = 0;
             foreach (PermissionInfoBase permission in List)
             {
                 if (permission.PermissionID == permissionID && permission.UserID == userID && permission.RoleID == roleID)
                 {
-                    List.Remove(permission);
+                    List.RemoveAt(idx);
                     break;
                 }
+                idx++;
```
daguiler: @mean2me, the index is incremented even when the item was removed from the list

daguiler: Ah never mind, did not see the `break` statement.

daguiler:
Hi @dnnsoftware/approvers, please review.
Thanks!

PR #2503, opened 2018-12-05
Change constraing in FK_Folders_ContentWorkflows

mean2me:
Default it was "ON UPDATE NO ACTION" and it was blocking the possibility to rename global assets folders.


mitchelsellers:
Rather than starting a full review a few notes/discussion points on this.

1. We cannot have the "stable" flag included in this PR
2. Can I get detail on WHY you would set the ID to null on an update?  This doesn't match any convention that we have today.
3. Any change like this will also need to get an upgrade script to handle this on an update

mean2me:
@sleupold @mitchelsellers  I based it on what already happens when we create a new folder in Global Assets.
We actually create a record with a NULL WorkflowId but when we update this record, that constraint blocks the update. I just assumed that if we can create a record with a NULL WorkflowId we must be able to update that record keeping the same value, IMHO.

mitchelsellers:
This change wouldn't/shouldn't fix anything that would be an issue with updating a folder to change the workflow, it would be a change on the other side.

I guess, the real question here is can we get detailed recreation steps to create the current issue, and to see how this fix resolves the issue?

It doesn't make sense to me that this would fix anything, it seems to me that it would hurt.

mean2me:
@mitchelsellers  actually the issue has been raised on Evoq, the big change is there.
When we perform an update, we get this error:
`The UPDATE statement conflicted with the FOREIGN KEY constraint "FK_Folders_ContentWorkflows". The conflict occurred in database "DB_NAME", table "dbo.ContentWorkflows", column 'WorkflowID'. The statement has been terminated.`

mean2me:
@mitchelsellers My question is: is it allowed to have an entry in Folders table with a null WorkflowId ? In case it's wrong, it's a design issue and we should think about a better solution, of course.

mitchelsellers:
What is the exact update you are trying to do?  That error should only be happening if you are changing the workflowid?  Is that really happening?  Or if you are updating the folder to have a non-existent FK value

mean2me:
@mitchelsellers  I got the error when updating a Folder passing a NULL WorkflowId. Anyway, I'm going to test my fix again, on a fresh Evoq installation, using only my fixes on Evoq side. I'll let you know soon what happened. In case it works fine, we can cancel this PR.

sleupold:
of course, you may have a folder with no workflow - this applies to all folders in DNN platform. it looks like you are trying to update this Null value with an invalid WorkflowId - most likely, you are hooked by the horrible Null.integer = -1 fault inside DNN.

sleupold:
PS: it might help to improve your SQL UPDATE statement with an NullIf(value, -1)

mitchelsellers:
I have a feeling that @sleupold is right.

I just verified, null is valid, and in many of my 9.2.2 installations for platform.  And you can update it to, and from null, but trying to use -1 (Null.NullInteger in DNN's api will error).

sleupold:
@mitchelsellers I love your feeling ;)

mean2me:
@mitchelsellers  @sleupold 
I've just tested my changes on Evoq. Actually my development instance was quite messed up. All worked fine without this change on a Evoq 9.3.0, so I'm gonna close this PR.
Thanks a lot for your comments.

PR #2509, opened 2018-12-12
Force user logout after password changed in other place.

zyhfish:


DNN Platform/HttpModules/Membership/MembershipModule.cs
```diff
@@ -256,5 +257,38 @@ public static void AuthenticateRequest(HttpContextBase context, bool allowUnknow
...
+            {
+                if (user == null || user.IsDeleted || user.Membership.LockedOut
+                    || !user.Membership.Approved && !user.IsInRole("Unverified Users")
+                    || !user.Username.Equals(context.User.Identity.Name, StringComparison.InvariantCultureIgnoreCase))
+                {
+                    return true;
+                }
+
+                // if user's password changed after the user cookie created, then force user to login again.
+                var authCookie = context.Request.Cookies[FormsAuthentication.FormsCookieName];
```
mitchelsellers: I am concerned with regards to the performance of this change.  As a HttpModule this is going to execute once for every managed request, including .axd and other style files.  This introduces a number of string operations, an additional decryption, and a lookup validation.  

I would prefer that we find another method to accomplish this goal.  Additionally, this is a change in behavior that could be hard to diagnose for users.

DNN Platform/HttpModules/Membership/MembershipModule.cs
```diff
@@ -256,5 +257,38 @@ public static void AuthenticateRequest(HttpContextBase context, bool allowUnknow
                 context.Items.Add("UserInfo", new UserInfo());
             }
         }
+
+        private static bool NeedLogout(HttpContextBase context, UserInfo user)
```
mitchelsellers: Recommend changing this to RequiresLogout

DNN Platform/HttpModules/Membership/MembershipModule.cs
zyhfish: Hi @mitchelsellers , the requests filter was done at begining of AuthenticateRequest method, it won't process for axd or other files.

mitchelsellers: @zyhfish Ok, this still is a per-page-load situation, I really don't want this overhead and processing on every request

zyhfish: Update the code to get issue date from identity instead of from cookie.

DNN Platform/HttpModules/Membership/MembershipModule.cs
zyhfish: Updated.

DNN Platform/HttpModules/Membership/MembershipModule.cs
daguiler: Can you please revisit @mitchelsellers? The only performance degradation I see now is the `try...catch` block.

mitchelsellers:
@dnnsoftware/tag thoughts on this one?  This is a breaking change in behavior

ohine:
On new installs this behavior is great, but for any upgraded sites I think they should have to enable that functionality through settings exposed under Security > Member Accounts > Member Management.

zyhfish:
Hi @ohine , add the setting to control this behavior, and only enable it in new installation. after this PR merged, i will try to submit PR in AE repo for the UI part, thx

mitchelsellers:
Per review in the TAG group today.  We would also want to address invalidation of any issued JWT issue

mitchelsellers:
THis PR is now out of sync, can it get Re-based against Development?

mtrutledge:
@zyhfish What is the status for the change request that @mitchelsellers made for this change?

zyhfish:
PR https://github.com/dnnsoftware/Dnn.AdminExperience/pull/939 created for the UI part.

Issue #2513, opened 2018-12-13
Performance problems when huge number of portal aliases is in use


mikebigun:
## Description of bug
Customer complains about huge performance degradation when around ~5K portals/aliases is in use with a combination of ~2K requests executed at the same time. Whole system becomes very slow and sometimes is coming up to an unresponsive state.
Tracing the system customer identified the problem on a web server. Using various tools identified that many threads are locked out due to very low capability to validate portal aliases during the incoming requests.

## Steps to reproduce
Issue is reproducible having following database characteristics:
440k users 
500k files 
4885 portal alias 
2000+ portals 
Also, need to simulate 2K+ requests to see the problem. 

## Current result
System is extremely slow.

## Expected result
Need to improve performance. 

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

PR #2514, opened 2018-12-13
Performance problems when huge number of portal aliases is in use

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2513

The main concern I see is how the alias validation is implemented on incoming requests. There is a regex expressions used to validate alias. There are multiple purposes to use such regular expressions: 
a) validate whether requested URL has a proper format and b) check if alias provided in URL exists in the system. For that, we are caching list of all aliases available in system and generating list of accordingly created NON-COMPILED regex objects (one regex per one alias). Having ~5K aliases, system generates ~5K expressions by replacing _ALIAS_ placeholder in a pattern by alias value. On every request, system iterates over ~5K expressions in a loop and checks whether requested URL matches to one of them. Does not look like a really good idea in terms of performance. Now let's imagine ~2K requests is coming at the same time and for every request we should apply ~5K expressions. On top of that, no one regex is [compiled](https://docs.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions), this is impacting system a lot.

The idea is to move from generating a huge number of expressions and keeping all of them in a cache to a single COMPILED regex. Iterating over aliases is enough to satisfy our purposes. This approach will bring a lot of benefits. We'll not need to store all regex in a cache, hence it will not require us to search it there on every iteration. At the end, we don't need to apply every regex to a single url (this is the most important thing). 

Moving to compiled regex is recommended for a frequent use. Moreover, it is senseless to validate the whole URL format if it does not contain expected alias.


DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -152,31 +153,53 @@ internal override void RewriteUrl(object sender, EventArgs e)
...
+                    var queryIndex = urlToMatch.IndexOf("?", StringComparison.InvariantCultureIgnoreCase);
+                    if (queryIndex >= 0 && queryIndex < aliasIndex)
+                    {
+                        // alias is in the query string, go to the next alias
+                        continue;
+                    }
+                    // we are fine here, lets prepeare URL to be validated in regex
+                    urlToMatch = urlToMatch.Replace(alias, "_ALIAS_");
+                }
+                // check whether requested url has a right URL formal containing exsting alias
```
bdukes:
```suggestion
                // check whether requested URL has the right URL format containing existing alias
```

daguiler:
@bdukes According to preliminary tests, this is now 5x faster.
Maybe you can paste the results here @mikebigun ?

mikebigun:
To see whether new implementation brings any improvements or not and how it is valuable, there was a separate tool created to simulate and compare behaviors to find out the best approach. 
I used similar caching mechanism in the tool and used same regular expressions. After numerous of testings, I got following results (see screenshot below).
Of course, this is kind of workaround and can not be treated as an official metric. Still this is the way of how I validated proposed solution.
Test 1 - is an old implementation with a cache and applying regex on every iteration
Test 2 - is a new version (similar to one I submitted), non-compiled regex is used
Test 3 -  is a new version (similar to one I submitted), COMPILED regex is used

Summary: I expect 5x faster URL validation.

image-2018-12-13-15-14-18-427

mitchelsellers:
I'm in 100% agreement with @bdukes on this.

On the surface, things make sense.  But I'd really want to see real-world performance numbers on this to prove this out..  As the impacts of a behavior change here could be catastrophic.

Honestly, I'd like to see a testing plan for this that includes a very large number of unit tests to validate inputs & outputs.  In addition to the true performance validation as the diagnostics of issues here could be very very hard.

mikebigun:
@mitchelsellers @bdukes @daguiler 

### Testing environment details
Windows Server 2012 R2 Standard, Intel(R) Xeon(R) CPU E5-2670 v2 @ 2.50GHz 2.49GHr, 15 GB RAM

### DotNetNuke
Evoq edition, v.9.1.1.129

### Preconditions
Portals created: 10, aliases: 5000. No other tables were modified. We want to isolate alias validation functionality, so this data is enough to proceed.

Load testing is performed to determine system behavior and compare responsiveness of an application using following approaches to validate alias:
- using set of cached regular expressions (current implementation)
- using compiled regex

Purpose is to determine if the implementation proposed in this PR is more efficient under heavy load.

### Load test strategy
- 10 sites
- 5000 aliases
- 260 concurrent users
- Each user hits a randomly selected site/page
- Test ends when 2000 requests have been processed

### Conclusion

Two charts below show the number of requests and response time (vertical axis) needed to complete each of consecutive request. It can be seen that most requests using current implementation requires more time to be completed (red line) vs. compiled regex (blue line). 

single compiled regex_time vs cached regular expressions_time

Using 260 concurrent requests, we can observe following responsiveness. Please, note, elapsed time measured below is a difference between time when request was sent and time when response has been fully received. Chart shows new implementation can process requests faster than current one. Considering results below, we can say it is possible to process 2X more concurrent requests now at the same time which is the goal of proposed changes (of course depending on hardware and environment infrastructure).

elapsed time in milliseconds

### Load testing tools

VS2017, [Apache JMeter](https://jmeter.apache.org/)

### Chart sources

[Spreadsheets](https://docs.google.com/spreadsheets/d/1kdnx8f2F8q-s7g3k_sFXRM9skZl_DR_INagdCBGrSq0/edit#gid=822426883)

daguiler:
I'd like to add some unit testing results.

I executed all tests in the [DotNetNuke.Tests.Urls](https://github.com/dnnsoftware/Dnn.Platform/tree/development/DNN%20Platform/Tests/DotNetNuke.Tests.Urls) assembly (1596) and **all of them still pass**.

Coverage is particularily good in the affected classes:
* [AdvancedUrlRewriter.GetPortalAlias](https://github.com/dnnsoftware/Dnn.Platform/blob/8b7bfd8ca84587d13e5f74b870367fba948ce4bf/DNN%20Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs#L153): 90%
* [CacheController.GetCustomAliasesFromCache](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Entities/Urls/CacheController.cs): 100%
* [TabIndexController.BuildPortalAliasesDictionary](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Entities/Urls/TabIndexController.cs): 86.57%

Some screenshots:

dnn-27498-00

dnn-27498-01

dnn-27498-02

dnn-27498-03

mitchelsellers:
This feedback looks great.

@bdukes any thoughts on testing?

daguiler:
@mitchelsellers, @bdukes there seems to be an issue with case sensitivity that we are trying to identify and fix.

sleupold:
looks fine to me, though I am curious about the effect on performance in real live scenarios.

daguiler:
Case sensitivity issue was found and fixed by @mikebigun and new integration tests were added to cover mixed-cased URLs. All 1991 URL tests pass:

dnn-27498-multicase

PR #2517, opened 2018-12-23
Module's permission: user can manage module's settings

mean2me:
## Steps to reproduce
In DNN 9.2.x, create a test user. 
1. Open a page in Edit mode. 
2. Add a Module to the page (HTML). 
3. Edit the module permissions. 
4. Add the test user to the module and grant the "Manage Settings" permission. 
5. Save the settings for the Module. 
6. Log in as the user and try to manage the module.

## Current Behavior
Currently, only the "Help" option is displayed when the test user is logged in and tries to manage the module settings.

## Desired Behavior
User must be able to access module's settings menu

## note
To avoid any breaking change with previously existing features, I may have created some redundance. Please let me know if you see any possible optimization. What confused me is the permission key "EDIT" used to allow many permissions, including settings management. To avoid any mistake, I created a specific permission key using the proper key "MANAGE".


ohine:
The view/edit permission keys are traditionally the only keys exposed for platform. The additional settings are utilized in evoq and custom permissions providers to enable the full set of permissions. 

I’m not sure if we want to include the manage permission by default in platform at this point. We will need a few more @dnnsoftware/approvers to chime in on this one.

mean2me:
@ohine I totally agree. At the moment this PR won't introduce breaking changes to existing features but yes, it expose settings to users with "MANAGE" permissions.

kurtwilbies:
Well, I disagree. Using an adv. permission model is something the community wants. I think w'll have to change out mind set. Its not correct to say: we do not enhance DNN because EVOQ has those features/extension points. DNN and EVOQ has to be the same (like almost all CMS, there is no real difference between Open Source and Lincensed apps, except SUPPORT).

kurtwilbies:
Something completly unrelated, but I dont use facebook or twitter. Where is the 9.3.0 RC download?

valadas:
@kurtwilbies At the top of the readme.md file, there is an Azure Pipelines badge, if you click it you will be taken there and you will see a "Download Artifacts" button. That will download the Dnn "nightly build" as of the lastest merged commit.

ohine:
@mean2me I believe this will be a breaking change as it's currently coded in this PR. 

I've dug into the permissions quite a bit with one of my open source providers. For upgrades, I seem to recall users that were granted `EDIT` will then not be able to `MANAGE` any settings because the key was swapped out, the only users that would be able to get into settings would be admins or hosts. If we do decide to include this additional permission, we will need to account for this case. After a few other approvers chime in, I can do some testing to verify if my memory is correct on this or not. 

What I've had planned for the permissions model in DNN for quite a few years, but just never really wrote down or had the time to implement in my provider. We really need a security configuration wizard of sorts. Dnn superusers should be able to enable each of the additional permissions if they require that functionality. Not every site needs add/edit/manage/view/import/etc, but it would be nice if someone did want one of the more advanced permissions that they'd be able to turn it on. 

I'll also add I'm not a fan of enabling all of them for every installation by default. View/Edit is nice and works in most cases, but for a more complicated site I'll usually spin off a custom build of my provider which enables the permissions that client site requires.

mitchelsellers:
I have to agree with @ohine on this.

As this PR is at the moment I believe would be a breaking change for upgrades.  I agree with @kurtwilbies that it would be great to get more enhanced permission set in the core, however, this isa  very limited in scope, port of the advanced provider from EVOQ and doesn't set the proper stage for implementation with regards to existing users & installations. 

This is especially true for upgrades.  And at a minimum will introduce risk to the those migrating.

@kurtwilbies re 9.3.0 RC, it hasn't been posted yet, but when it does it will be here: https://github.com/dnnsoftware/Dnn.Platform/releases

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
```diff
@@ -577,11 +577,9 @@ private void LoadActions(HttpRequest request)
             }
             else
             {
-                if (!Globals.IsAdminControl() && ModulePermissionController.HasModuleAccess(SecurityAccessLevel.Admin, "DELETE,MANAGE", Configuration))
```
mitchelsellers: DELETE is removed from the list here, this would be a breaking change for others that expected this to behave as it does now.

kurtwilbies:
I agree. Focus is to bring DNN to a next level.

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
mean2me: it's a mistake on my side. I didn't want to remove it actually. 
Gonna restore it.

mitchelsellers:
This is a fix for the Evoq platform and should be introduced in the Evoq code base

sleupold:
@mitchelsellers @ohine 
it would be great to get an (optional) granular permission provider included in DNN platform as well to test the permissions appropriately, as currently, this is possible only, if there is a wiring 1:1 from internal to displayed permission in UI

mitchelsellers:
@sebastain I’d love that!!

sleupold:
@mitchelsellers I got the advancedPermissionProvider to run on DNN platform, but it turned out, that the community version of the PersonaBar (editbar) ignores the permissions (other than the Evoq versions). We would need to fix these issues, before beginning to expose advanced permissions in DNN Platform

PR #2522, opened 2018-12-28
SQL Script fix for Portal Information

### LINKED ISSUE #2521: Sites updated date is wrong
@valadas commented on [Tue Dec 25 2018](https://github.com/dnnsoftware/Dnn.AdminExperience/issues/269)

## Description of bug
Unless one of Jesus miracles was building websites, the updated date in Sites list is wrong... :)

## Steps to reproduce
1. Go to Manage -> Sites
2. Notice the updated date

## Current result
2018 years ago

## Expected result
Something like yesterday or last week or last month depending on the last edit.

## Screenshots
image

## Error log
Not sure if related or a separate issue (I will log a separate issue just in case too):

```
AbsoluteURL:/API/PersonaBar/Sites/GetPortals
ExceptionGUID:e5bfde54-47e8-4215-9064-80875b863512
AssemblyVersion:
PortalId:-1
UserId:-1
TabId:-1
RawUrl:
Referrer:
UserAgent:
ExceptionHash:vov9OXnxtFF5jN4485rhZVXyNPs=
Message:TermsTabId
StackTrace:
at System.Data.ProviderBase.FieldNameLookup.GetOrdinal(String fieldName) at System.Data.SqlClient.SqlDataReader.GetOrdinal(String name) at System.Data.SqlClient.SqlDataReader.get_Item(String name) at DotNetNuke.Entities.Portals.PortalInfo.Fill(IDataReader dr) at DotNetNuke.Common.Utilities.CBO.FillObjectFromReader(Object objObject, IDataReader dr)
InnerMessage:
InnerStackTrace:
Source:System.Data
FileName:
FileLineNumber:0
FileColumnNumber:0
Method:
Server Name: DanielValadasPC
```

## Additional context

## Affected version
9.3.0.130

---

@valadas commented on [Fri Dec 28 2018](https://github.com/dnnsoftware/Dnn.AdminExperience/issues/269#issuecomment-450296763)

Looks like the date comes wrong from the platform GetProtalsByName api. Moving the issue to Platform.

mitchelsellers:
Fixes #2521 by re-building the view for the PortalsDefaultLanguage.


mitchelsellers:
@dnnsoftware/tag I believe this is the right way to fix this, it is a breaking change though for anyone that has already tested a 9.3.0 upgrade, but I believe that would be the case no matter what.  If I need to change let me know

sleupold:
I doubt that this is the proper fix. it looks like a data call in code behind is not matching columns properly, SQL is always matching by column names.

valadas:
Well this would have been broken only on 9.3 and there was no production release of 9.3, so I think this is good in my opinion.

mitchelsellers:
Sebastian if the order changes of an underlying table a view built with Select * can get out of sync. 

There are two fixes. 

1) Refresh the view 
2) Recreate the view 

Given the prior processes in DNN I used option 2, since that is how we do it. 

You can recreate this issue easily to confirm. We have also tested on multiple 9.3.0 builds that this resolves

sleupold:
@mitchelsellers 
instead of recreating the view, you could exec sp_refreshsqlmodule '{databaseOwner}[{objectQualifier}vw_PortalsDefaultLanguage]', however, it would be better to generally avoid using "SELECT *" in views, see https://sqlblog.org/2009/10/10/bad-habits-to-kick-using-select-omitting-the-column-list, https://stackoverflow.com/questions/128412/sql-query-select-from-view-or-select-col1-col2-coln-from-view etc.

mitchelsellers:
@sleupold Yes, looking back at the history, execution of sp_refreshsqlmodule is not something we have done, and if an error occurs we wouldn't have a handle.  Thus why I went the other route.

I left it as SELECT * for simplicity, if you want to update this PR to list all of the columns, go ahead.  But I'm looking to keep the changes as minimal as possible for review etc at this point.

I'd like in 9.4, or 10.0 to review & remove all SELECT * usage across the board as a dedicated effort, etc.

valadas:
Cool, can someone merge then so we have a working nightly build.

mitchelsellers:
Given that this already had the needed approvals, from not myself, I merged this change

PR #2523, opened 2018-12-29
DNN-27515 : skip anti-forgery-token validation for already authentica…

anishmishra978:
DNN-27515 : handled HttpAntiForgeryException to skip anti-forgery-token validation for already authenticated users.

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/ValidateAntiForgeryTokenAttribute.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/ValidateAntiForgeryTokenAttribute.cs
@@ -26,6 +26,8 @@ using DotNetNuke.Web.Api.Internal;
 using System.Threading;
 using System.Web.Http.Controllers;
 using System.Web.Http.Filters;
+using System.Web.Mvc;
+using System.Web;
 
 namespace DotNetNuke.Web.Api
 {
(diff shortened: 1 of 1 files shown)


mitchelsellers:
This change would introduce a security issue, just because you are authenticated does NOT mean that we skip AntiForgeryToken validation.

sleupold:
I also fear, #2524 is hard to verify, as there is no UI in DNN platform

valadas:
I agree with the security concerns. Also in the issue the message is: **The provided anti-forgery token was meant for user "host", but the current user is ""**

Correct me if I am wrong, but the anti-forgery token is not specially created for any specific user right? And I did not find that string while searching the repository, I think this fix should be done on the actual module where it happens.

donker:
This rests IMHO on a misunderstanding of what the AFT is for. It is specifically meant for those cases where the user is authenticated! So this is equivalent to ripping out the AFT altogether.

With SPA and React/etc apps you need to jump through a few hoops sometimes to ensure you pass the AFT validation. But it is doable. Until now I have not seen anything that outweighs the security benefits you get from the AFT system.

valadas:
@mitchelsellers as for #2524 is this something that needs to be done on the platform or would that be a problem on that specific module then?

mitchelsellers:
@valadas If this is an issue with Content Personalization it would need to be handled by ESW inside of the Evoq code-base.

The Anti-Forgery system, as is, has been tested and validated to be working so I do not believe there is any task for DNN Platform with regards to this scenario.

PR #2527, opened 2019-01-04
Modules > ModuleCreator > fixed path error

mean2me:
On creating a new module with Module Creatore, we are getting a wrong path error when we module type is "Web".
This pull request fixes this issue.


DNN Platform/Admin Modules/Dnn.Modules.ModuleCreator/Templates/Web/Module - HTML/template.ascx
```diff
@@ -33,13 +33,15 @@
     {
         base.OnLoad(e);
 
+        const string ModuleName = "_OWNER_._MODULE_";
+
         if (!Page.IsPostBack)
         {
             //Load the HTML file
-            var path = Server.MapPath(ModulePath) + "_CONTROL_.html";
+            var path = Server.MapPath(ModuleName) + "_CONTROL_.html";
```
daguiler: A separator between folder and file is missing here.

```suggestion
            var path = Path.Combine(Server.MapPath(ModuleName), "_CONTROL_.html");
```

daguiler:
Hi @dnnsoftware/approvers, any concerns with this one?

PR #2530, opened 2019-01-08
Force site's root folder created with standard folder type.

zyhfish:


bdukes:
I thought it was a supported scenario to have the site root on remote storage, and that folks were using that when on cloud hosting. It seems like we'd need to be able to support that if we're ever going to be able to run in a docker container

sleupold:
this should be possible, but I fear, it has (nearly) never been used.
We need to review, which subfolders need to be local (moved to 0-system folder) and which can be requested from external data sources.

zyhfish:
@bdukes , the site root folder need to be always standard, then you can create sub-folders with any folder types, such as remote storage. but if the root folder created as other folder types, then if you try to create sub folders, you will not able to change its folder type.

mitchelsellers:
It seems to me that this is continuing to support a bug in the platform with regards to the way that folder types are implemented.  For a true situation deploying to the cloud this root folder SHOULD support being defined as a different folder type.

I would prefer that we find a solution for the issue that arises when the root folder is not a standard folder, rather than perpetuate the continuance of this bug that can limit implementation progress in the cloud.

daguiler:
Even in the scenario of deploying to the cloud, wouldn't the root folder still be standard? I mean, isn't it the local file system?
Then if you want to connect to an external folder, hosted somewhere else, then you would create a mapping.

bdukes:
All of the data within a portal's home directory can be offloaded, none of it should need to be local (everything that must be local is in the `-System` folder for the portal).  If you want to offload some of the directories, there's a good chance you just want to offload them all, so that no real content is living inside of the DNN file system.

mitchelsellers:
Given the desire to support root file offloading to the cloud, and existing installations closing this PR for now as there has not been comment or action on this.

Should a new direction be needed etc, we can reopen

PR #2532, opened 2019-01-09
Add setting to display search result for users in specific roles

### LINKED ISSUE #2533: add feature to search results module
add a configuration option to allow results scope to only include users from selected roles, in the Search results Settings for the Search module.

Currently, when searching for users, the results show all users including administrators and other registered users. We would like to limit our search results to only show users in the selected role.

image

zyhfish:
fix #2533


sleupold:
the setting caption should mention that only results from indexed "users" records are affected

Issue #2552, opened 2019-01-24
Error during upgrade from 9.0.2 to 9.2.2


Ceten:
## Summary
I'm having problems upgrading an existing website from version 09.00.02 to 09.02.02.
Would it be possible to point me in the right direction to solve this?

## Extended description
I'm getting the error below, apparently while trying to update to 9.1.0, and the update hangs.

## Error log (if necessary)
```
01/24/2019 11:09:51 [INFO] DotNetNuke.Services.Upgrade.Upgrade Start RemoveAdminPages://Admin//SearchAdmin
01/24/2019 11:09:53 [ERROR] DotNetNuke.Services.Upgrade.Upgrade Error: The ''' character, hexadecimal value 0x27, cannot be included in a name. Line 44, position 19.   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.ParseElement()
   at System.Xml.XmlTextReaderImpl.ParseElementContent()
   at System.Xml.XPath.XPathDocument.LoadFromReader(XmlReader reader, XmlSpace space)
   at System.Xml.XPath.XPathDocument..ctor(TextReader textReader)
   at DotNetNuke.Services.Installer.Installers.PackageInstaller..ctor(PackageInfo package)
   at DotNetNuke.Services.Installer.Installer..ctor(PackageInfo package, String physicalSitePath)
   at DotNetNuke.Services.Upgrade.Upgrade.UninstallPackage(String packageName, String packageType, Boolean deleteFiles, String version)
   at DotNetNuke.Services.Upgrade.Upgrade.UpgradeToVersion910()
   at DotNetNuke.Services.Upgrade.Upgrade.UpgradeApplication(String providerPath, Version version, Boolean writeFeedback)
```

## Affected version
* [ ] 9.3.0
* [x] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [x] 9.0

sleupold:
did you unblock the downloaded upgrade package (in file properties), before unzipping the files contained?

Ceten:
No I didn't think of that. Can I unblock, copy the files and try again or is it better to start from a fresh copy of the original website?

sleupold:
if possible, try with a fresh copy and make sure, the ASP.Net account (specified in advanced settings of the IIS App Pool) has write permission for the whole directory of the IIS website. 
[...]

PR #2557, opened 2019-01-27
Log name of package when uninstalling extensions

ohine:
## Summary
When the platform is uninstalling extensions, it doesn't log when and what extensions were uninstalled as part of the upgrade process.

Hopefully this will help troubleshoot similar issues like #2552

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/Upgrade.cs
+++ b/DNN Platform/Library/Services/Upgrade/Upgrade.cs
@@ -3207,6 +3207,8 @@ namespace DotNetNuke.Services.Upgrade
 
         private static void UninstallPackage(string packageName, string packageType, bool deleteFiles = true, string version = "")
         {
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -3207,6 +3207,8 @@ private static void UpdateFoldersForParentId()
 
         private static void UninstallPackage(string packageName, string packageType, bool deleteFiles = true, string version = "")
         {
+	    DnnInstallLogger.InstallLogInfo(string.Concat(Localization.Localization.GetString("LogStart", Localization.Localization.GlobalResourceFile), "Package:", packageName, " Type:", packageType, " Version:", version));
```
mitchelsellers: Could we add the phrase "uninstall" at some point in this.  Right now this would result in an output of

Start Package {Name} Type: {Type} Version: {Version}

Based on the value of "LogStart" being simply the word start.

mitchelsellers:
Changing this to Release 9.3.0 branch as this is necessary for issues with upgrades.

@dnnsoftware/tag Can I get a secondary approval

Issue #2568, opened 2019-01-31
Pages - Localization - Home page tab sequence changes after clicking on gear icon of newly added language


alexnaurea:
## Description of bug
Pages - Localization - Home page tab sequence changes after clicking on gear icon of newly added language.

Actually, when an admin clicks "Make Page Translatable", it adds localized versions of the page for every language. But the new added localized pages are being added at the end of the list. As a result you can see, that Home page menu item at the left side for the original language, but its localized version is on the right side.

## Steps to reproduce
List the steps to reproduce the behavior:
1) Login as SuperUser 
2) Navigate to Persona Bar -> Settings -> Site Settings -> Languages 
3) Allow Content Localization toggle button should be set to OFF and Disable the Content Localization (if its enabled) 
4) Enable the newly added language 
5) Allow Content Localization toggle button is set to ON -> Click on Save 
6) Click on Enable Content Localization 
7) Click on Enable Localized Content 
8) Refresh the Home Page 
9) Navigate to Persona Bar -> Content -> Pages 
10) On Home Page, click on Localization 
11) Click on Make Page Translatable button 
12) View application in new tab using eye icon of both the languages 
13) Click on the gear icon of the Eng language 
14) Click on the gear icon of the newly added language

## Current result
Home page tab sequence changes after clicking on gear icon of newly added language

## Expected result
Home page tab sequence should not change

## Screenshots
![image](https://user-images.githubusercontent.com/45540347/52055732-aae64900-2579-11e9-87b8-ed0f02d035b4.png)

![image](https://user-images.githubusercontent.com/45540347/52055798-d9642400-2579-11e9-8574-6fe2e8d59e72.png)

## Affected version

* [x] 9.3.0

PR #2569, opened 2019-01-31
Create localized versions of the page on the same position in the menu as original version is

alexnaurea:
Fix: https://github.com/dnnsoftware/Dnn.Platform/issues/2568

### DIFF
--- a/DNN Platform/Library/Entities/Tabs/TabController.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabController.cs
@@ -207,7 +207,7 @@ namespace DotNetNuke.Entities.Tabs
             return tab.TabID;
         }
 
-        private void CreateLocalizedCopyInternal(TabInfo originalTab, Locale locale, bool allTabsModulesFromDefault, bool clearCache)
+        private void CreateLocalizedCopyInternal(TabInfo originalTab, Locale locale, bool allTabsModulesFromDefault, bool clearCache, bool insertAfterOriginal = false)
         {
 			try
 			{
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Entities/Tabs/TabController.cs
```diff
@@ -207,7 +207,7 @@ private int AddTabInternal(TabInfo tab, int afterTabId, int beforeTabId, bool in
             return tab.TabID;
         }
 
-        private void CreateLocalizedCopyInternal(TabInfo originalTab, Locale locale, bool allTabsModulesFromDefault, bool clearCache)
+        private void CreateLocalizedCopyInternal(TabInfo originalTab, Locale locale, bool allTabsModulesFromDefault, bool clearCache, bool insertAfterOriginal = false)
```
daguiler: I wonder if we actually need this `insertAfterOriginal` parameter. Do you see any use cases for not adding the translated page at the original position @dnnsoftware/approvers ?

I tested this a bit, and was only able to hit this `CreateLocalizedCopyInternal` method from lines 910 and 929 (the ones modified in this PR), but I wasn't able to hit the other ones, at lines 1083 and 1096.

For the one at 1083, I could only find references coming from deprecated code:
https://github.com/dnnsoftware/Dnn.Platform/blob/8321bdab6d9854a4417321f07532e8c414e8b7a3/DNN%20Platform/Library/Obsolete/TabController.cs#L64
https://github.com/dnnsoftware/Dnn.Platform/blob/8321bdab6d9854a4417321f07532e8c414e8b7a3/DNN%20Platform/Library/Obsolete/TabController.cs#L72

For the one at line 1096, I found this reference:
https://github.com/dnnsoftware/Dnn.Platform/blob/8321bdab6d9854a4417321f07532e8c414e8b7a3/DNN%20Platform/DotNetNuke.Web/UI/RibbonBarManager.cs#L363
which is used also by deprecated code:
https://github.com/dnnsoftware/Dnn.Platform/blob/8321bdab6d9854a4417321f07532e8c414e8b7a3/DNN%20Platform/DotNetNuke.Website.Deprecated/admin/ControlPanel/AddPage.ascx.cs#L94
Is it possible to see this control panel somehow in lastest versions?

mitchelsellers:
@alexnaurea Any update on the comment above?

PR #2573, opened 2019-02-02
Combine Repositories

ohine:
This starts the effort to combine all related repositories into one. There is still a mountain of cleanup and build scripts which we will need to rewrite or adjust. Solutions need to be combined. Issues need to be migrated. Etc... 

Sorry for the short description, Chrome crashed twice when loading the 595,319 additions in this pull request.


ohine:
If anyone else wants to help with this work, ping me and I'll add you as a contributor to my personal fork. I should have another commit shortly which fixes the reference paths on these.

ohine:
All the project based references are updated. Next we need to adjust the nuget references within the newly added projects.

Build scripts are the next huge body of work to be rewritten. I've started that process, but it's a lot to unravel and rewrite. Estimate on that piece of work is about 30hrs in order to be able to start to generate CI packages from this branch.

valadas:
Just a tough but would'nt it be more manageable to first merge common into AE, then merge the projects one at a time, this is a huge PR that way!

valadas:
And maybe we could do this after 9.3.0 is officially released? There might still be a couple things to fix and it may make it hard to merge at the same time?

ohine:
It’s a huge pull request no matter how or when it’s done. It’s going to cause quite a few issues and require everyone to reconfigure their machines that’s for sure. 

As for waiting till v9.3 is out, I don’t think we need to delay development. That’s why we have multiple branches so some of us can continue working on the next release while another is in the final test stages.

PR #2583, opened 2019-02-08
Placeholder to avoid PersonaBar slide effect

daguiler:
## Summary
This is a companion pull request to https://github.com/dnnsoftware/Dnn.AdminExperience/pull/394.


valadas:
I agree with @ohine , @daguiler can you just move the CSS from the other PR into that file and close the other issue. I am all for merging this after that as a temporary fix for the sliding content, we can improve the whole process later but that would make me happy for now and I think it could even be part of the current RC.

daguiler:
Correct, I'll move the CSS classes to AdminExperience repo. Closing this one.

PR #2596, opened 2019-02-25
Add error handling in AssemblyInstaller

### LINKED ISSUE #2595: Exception thrown when native assembly is in assemblies component in extension manifest
## Description of bug
The `AssemblyInstaller` reads version information from provided assemblies to create a binding redirect for any strong-named assemblies.  The code that does this does not have error handling when there are errors reading that information, such as when the assembly file is a native assembly, rather than a .Net assembly.

## Steps to reproduce
1. Create an extension that packages a native assembly, e.g. [bad-assembly-image-demo.zip](https://github.com/dnnsoftware/Dnn.Platform/files/2903010/bad-assembly-image-demo.zip)
2. Install the extension via the Persona Bar, click through the wizard steps

## Current result
An error in the installation log
>Could not load file or assembly 'Microsoft.DiaSymReader.Native.amd64.dll' or one of its dependencies. The module was expected to contain an assembly manifest.

## Expected result
The assembly is installed without any error

## Screenshots
error message screenshot

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

bdukes:
Fixes #2595

### DIFF
--- a/DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
+++ b/DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
@@ -23,11 +23,13 @@
 using System;
 using System.IO;
 using System.Reflection;
+using System.Security;
 using System.Text.RegularExpressions;
 using System.Xml;
 
 using DotNetNuke.Common;
 using DotNetNuke.Data;
+using DotNetNuke.Framework;
 
 #endregion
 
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
```diff
@@ -43,7 +45,9 @@ public class AssemblyInstaller : FileInstaller
     {
         private static readonly Regex PublicKeyTokenRegex = new Regex(@"PublicKeyToken=(\w+)", RegexOptions.CultureInvariant | RegexOptions.Compiled);
 
-		#region "Protected Properties"
+        private static readonly string OldVersion = "0.0.0.0-" + new Version(short.MaxValue, short.MaxValue, short.MaxValue, short.MaxValue);
```
daguiler: Hi @bdukes just curious, did Resharper or VS suggest to make this var static? Not that it's a problem, just wondering what's the benefit of keeping it in memory.

bdukes: It just seemed better to me that there was only one instance, instead of having to create a new `Version` every time we process an assembly.  I don't have any evidence that it makes a difference, either way.  Let me know if you'd prefer it changed back.

daguiler: No, it's fine. My personal preference would be not to create a static variable, but I don't have any evidence to support it either other than static variables are kept in memory.

PR #2598, opened 2019-02-27
Inclusion of NDepend logo on the readme.

mitchelsellers:
The folks at NDepend have graciously contributed a license to their NDepend product for use within Azure DevOps for use by the DNN Platform team to monitor code-quality.  

Adding an attribution to them at the bottom of our ReadMe, just like what was done for the .NET Foundation.


valadas:
Do we do that on the other repositories too ?

mitchelsellers:
@valadas Only need to do it here, as it will be added to the build process on the main DNN.Platform build only.  As it only analyses C# Code.

valadas:
@mitchelsellers  But we also have C# code in the other repositories right ?

PR #2607, opened 2019-03-04
get language from transferred parameter.

### LINKED ISSUE #2606: Globals.GetCultureCode will not work when call from scheduler
## Description of bug
when Globals.GetCultureCode called from scheduler, it will throw out null reference exception, this caused by commit #9ab22767, PortalController.Instance.GetCurrentPortalSettings() doesn't exist in sub thread. i think the code need change to:
`var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
cultureCode = portalSettings?.DefaultLanguage ?? Thread.CurrentThread.CurrentCulture.Name;`

## Affected version
* [x] 9.3.0

zyhfish:
Fix #2606 .

### DIFF
--- a/DNN Platform/Library/Common/Globals.cs
+++ b/DNN Platform/Library/Common/Globals.cs
@@ -811,7 +811,7 @@ namespace DotNetNuke.Common
                 }
                 if (string.IsNullOrEmpty(cultureCode))
                 {
-                    cultureCode = PortalController.Instance.GetCurrentPortalSettings().DefaultLanguage;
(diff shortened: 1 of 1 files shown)


daguiler:
Verified to work ok on latest build

mhessmann:
Hi @zyhfish , can you please have a look at my issue.
I think your commit here is the reason for our problem.

https://github.com/dnnsoftware/Dnn.Platform/issues/2778

Kind regards
Marc

PR #2616, opened 2019-03-08
switch encrypt method.

zyhfish:


DNN Platform/Library/Services/Personalization/PersonalizationController.cs
```diff
@@ -174,9 +174,9 @@ private static string DecryptData(string profileData)
         private static string GetDecryptionkey()
         {
             var machineKey = Config.GetDecryptionkey();
-            var hostGuid = Host.GUID.Replace("-", string.Empty);
-            var key = (machineKey ?? "") + hostGuid;
-            return PortalSecurity.Instance.Encrypt(key, key);
+            var hostGuid = Host.GUID;
```
mitchelsellers: If changing the way this is composed, do away with this declaration as a string it is immutable so no reason to actually allocate this item.  Just use Host.GUID directly below in the String Interpolation call

DNN Platform/Library/Services/Personalization/PersonalizationController.cs
```diff
@@ -174,9 +174,9 @@ private static string DecryptData(string profileData)
...
+            var key = $"{machineKey ?? ""}{hostGuid.Replace("-", string.Empty)}";
+            return FIPSCompliant.EncryptAES(key, key, hostGuid);
```
mitchelsellers: Per the above comment there would be no reason to not just use Host.GUID here

Issue #2621, opened 2019-03-11
RFC: System changes username in a database when "Use Email Address as Username" is enabled


mikebigun:
## Description of bug
When `Use Email Address as Username` is enabled, on a first login it authorizes user with email and replaces username with email address in database. Due to this username gets lost. After disabling this setting we are unable to use username anymore cause it was replaced in db. 

## Steps to reproduce
1. Log in with username: host 
2. In the Settings> Security>Registration Settings enable `Use Email Address as Username` 
3. Log out 
4. Log in with the email address associated with host and log out 
5. Try to login with username: host - not able to do it anymore 
6. Log in with the email address, disable the above option 
7. The username now is the same with the email address 
8. Try to change the username from email to host again - appears a message like the changes were saved but nothing happens 
9. Now you can only log in with the email address as it is set as a username

## Current result
When `Use Email Address as Username` is enabled, it changes username name in database and overwrites it with email address. Due to this, username gets lost and we can't reset it back to use username as login. 

## Expected result
When `Use Email Address as Username` is enabled, we should not be allowed to use username on login, only email address is allowed to be used.
When `Use Email Address as Username` is disabled, system should not authorize us with email address, only username is allowed.
It is important that we should not replace username in database. No data should be lost. 

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

thabaum:
Most all modern sites use email address along with phone number and/or username as well to allow users to log into the website.  I believe there is also an issue with using the unique email address only setting in web.config as this can only be applied to the DNN instance and not the portals or portal groups if added in the future.  You may wish to allow more users to register on seperate websites with seperate credential settings + profiles if you have this setting set no one can do this, however it would be nice to have this settting only applied for each portal or site group.  Because each portal is seperate a unique email for the single site portal only instead of for the entire DNN instance.  

With respect to this issue I have noticed this problem since using and it allowing email address as a username.  This issue has always been a topic of discussion with some of the things that happen when changing this setting back and forth cause.  

Another issues that is related is if you have another user with the email address already created and had the setting to allow users to have more than one email address while using username then switch to email as username... and back again that should mix things up while the setting to allow and not allow duplicate email addresses.  Then it really buggers up switching these settings around.

[...]

PR #2622, opened 2019-03-11
Username is changed to email address in the user profile

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2621

I didn't find any reason why we replace username in database on login event. There is a well-working mechanism to get user object by email, so then we use it's username property to authenticate it.

See demo: [demo](https://drive.google.com/file/d/1PMVgt31AbgT83WX8iKFopL_OUYjgTZyg/view?usp=sharing)

### DIFF
--- a/.nuget/NuGet.Config
+++ b/.nuget/NuGet.Config
@@ -5,6 +5,9 @@
     <add key="enabled" value="true" />
     <add key="automatic" value="true" />
   </solution>
+  <activePackageSource>
+    <add key="All" value="(Aggregate source)" />
+  </activePackageSource>
   <packageSources>
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="MyGet" value="https://www.myget.org/F/dnn-software-public/api/v2" />
--- a/DNN Platform/Library/Common/Globals.cs
+++ b/DNN Platform/Library/Common/Globals.cs
@@ -811,7 +811,7 @@ namespace DotNetNuke.Common
                 }
                 if (string.IsNullOrEmpty(cultureCode))
                 {
-                    cultureCode = PortalController.Instance.GetCurrentPortalSettings().DefaultLanguage;
+                    cultureCode = !string.IsNullOrEmpty(settings.DefaultLanguage) ? settings.DefaultLanguage : Thread.CurrentThread.CurrentCulture.Name;
                 }
             }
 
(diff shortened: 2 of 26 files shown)


mitchelsellers:
I believe that this was added to support a website switching from usernames to email as a username.

daguiler:
Does it mean it was working as designed @mitchelsellers?

If not, and this can be fixed as Mike suggests, we can still include it in this final test round, meaning we could leave the release branch as target branch.

bdukes:
Yes, the current behavior is how it was designed (whether that's the right design can be debated).

IMO, this is neither a showstopper nor a regression, and therefore shouldn't target the release branch.

mikebigun:
I'm in curious of how then this setting expected to work. 
In a current design, there is only one way to switch this option ON, but switching it OFF will not lead to the username usage. Users still have to enter email addresses.
I mean, there should be some reason why don't keep both username and email in database. Why there should be replacement?

valadas:
If I remember correctly, I submitted a fix a long time ago, even before the repo was on GitHub about this. The problem was like @mitchelsellers said that if you switch an existing site to use emails as usernames, the existing users could no longer login. There was also a related problem with duplicated emails already existing in the database when switching to that option. 

If memory serves good, you needed to first require unique emails and there would be a warning if there was duplicated emails, you needed to fix that to actually enable "require unique emails". Then you could only activate "emails as usernames" if "require unique emails" is on.

valadas:
This was one related issue (not exactly this but it had some relation) https://dnntracker.atlassian.net/browse/DNN-5874

I cannot find the exact one for this actual issue but there are many logged here that relate to this https://dnntracker.atlassian.net/browse/DNN-20654?jql=text%20~%20%22email%20as%20username%22%20order%20by%20created%20DESC

daguiler:
In any case I think you should re-target this PR to development branch @mikebigun

mikebigun:
Team, many thanks for explanation. 
Is there way to re-target PR without its reopening?

daguiler:
@mikebigun click on Edit next to the PR title and then a dropdown will show up below it for you to select a different base branch.

mikebigun:
@daguiler done.

daguiler:
Ok I don't think this played out well, I see unrelated changes now in this PR. Please close and open a new one @mikebigun.

mikebigun:
Dear approves, I'm closing this PR in favor of https://github.com/dnnsoftware/Dnn.Platform/pull/2625
New PR is targeted to development branch.

PR #2625, opened 2019-03-12
Username is changed to email address in the user profile

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2621

I didn't find any reason why we replace username in database on login event. There is a well-working mechanism to get user object by email, so then we use it's username property to authenticate it.

See demo: [demo](https://drive.google.com/file/d/1PMVgt31AbgT83WX8iKFopL_OUYjgTZyg/view?usp=sharing)

This is a clone of https://github.com/dnnsoftware/Dnn.Platform/pull/2622. The difference is that this PR is targeted to development branch.

### DIFF
--- a/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
@@ -262,25 +262,25 @@ namespace DotNetNuke.Modules.Admin.Authentication.DNN
 			if ((UseCaptcha && ctlCaptcha.IsValid) || !UseCaptcha)
 			{
 				var loginStatus = UserLoginStatus.LOGIN_FAILURE;
-				string userName = PortalSecurity.Instance.InputFilter(txtUsername.Text,
-										PortalSecurity.FilterFlag.NoScripting |
-                                        PortalSecurity.FilterFlag.NoAngleBrackets |
+				string userName = PortalSecurity.Instance.InputFilter(txtUsername.Text, 
+										PortalSecurity.FilterFlag.NoScripting | 
+                                        PortalSecurity.FilterFlag.NoAngleBrackets | 
                                         PortalSecurity.FilterFlag.NoMarkup);
 
                 //DNN-6093
                 //check if we use email address here rather than username
                 UserInfo userByEmail = null;
-                var emailUsedAsUsername = PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false);
+                var emailUsedAsUsername = PortalController.GetPortalSettingAsBoolean("Registration_UseEmailAsUserName", PortalId, false);                
 
                 if (emailUsedAsUsername)
                 {
                     // one additonal call to db to see if an account with that email actually exists
-                    userByEmail = UserController.GetUserByEmail(PortalId, userName);
+                    userByEmail = UserController.GetUserByEmail(PortalId, userName);                     
 
                     if (userByEmail != null)
                     {
                         //we need the username of the account in order to authenticate in the next step
-                        userName = userByEmail.Username;
+                        userName = userByEmail.Username; 
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


mitchelsellers:
Moving this to version 9.4.0, pending a resolution of an approach that would allow this to be changed at admin-setting time rather than after a failed login.  @donker is going to investigate and create a new issue for this related needed change.

Once a solution for this is identified we will then implement this change

donker:
Having gone through the history of this code and reviewing the general functionality I'd say Mikhail is correct in this observation. The change would be a breaking change, however, so we need to bump the milestone to 10. It may also give us some time to address other parts of this and ensure it works correctly. For example:

- You should not be able to switch to emailasusername if the email addresses are not unique on your site
- If it is part of a portal group /where sso is expected) then this should all work as expected

maduranga001:
I really want to see this feature on next release 9.4.0
I did not realize this as a bug and was finding why my new registration form is not working on the site.
This is a new site and cannot go ahead and go live without fixing the issue.
Please, advise, at least if you have a temp solution. 
I have upgraded the site to 9.3.2 version.
Please, advise.
Thanks!

mitchelsellers:
This will only be going in 10.x. But it doesn’t actually impact the ability for users to login. This change, in fact has the potential to break login. This the reason for it being part of 10.x

maduranga001:
Thank you very much for your reply Mitch.
This is a new dnn website and users need to register from the first day with entering only the email and a password. 
So what do you suggest Mitchel.?
Going for a third party modulevfor the registration and login?
Or any other good solution? 
Any idea abot the releasing date of 10.× version? 
Thanks.

mitchelsellers:
That feature works as is. Nothing is needed

maduranga001:
Means need wait until 10x release? Tnx

mitchelsellers:
No this issue is NOT an indication that you cannot use the email as username feature. 

That feature is fine and is not broken. 

This is a behind the scenes issue. And for a new site is a non issue.

maduranga001:
I have 5 testing users and I made their username also to the email address. Also the superuser too but still register button is not clickable.  But when i change it back to username as normal then it works. 
Please can i know why is this hsppening and how t9 fix this? All my accounts are testing accounts now. So can change anything. 
Really appreciate your advices.
Tnx

mitchelsellers:
Let’s move this discussion to the forums or otherwise to not clutter the conversation here

maduranga001:
Yes mitchel. You are correct. 
Sure. I posted it before 5 days. But not received any reply yet.
https://www.dnnsoftware.com/forums/forumid/89/threadid/547879/scope/posts

Thanks

valadas:
Just commenting this so stalebot does not close it...

thabaum:
Personally I would like to see the ability to use both Username or Email to login.  And also phone number for when SMS 2FA comes into play for a site.

Version 10 should be able to support these changes and allow users to correct as needed.  I will be testing out many of these features continuously over the next month.  I am looking into anything that may relate and this looks related.  I have been posting some minor PR's but I would love to dig into something a little bigger and fully understand the login and registration inside and out.

The issue with Email as Username and keeping sites unique email... this is not working per site but per DNN instance.  I will be testing these out fully but I should be able to allow a user with the same email to register and log into two separate sites when email as username is set for both.  Let alone a series of websites.  And for a Portal Group the scope should change to the primary site for unique emails.

This is what I am concerned with current DNN Login.  The Email as username with many different site configurations on one DNN instance, ability to use multiple account username email or phone to sign in and two-factor authentication.

mitchelsellers:
I'm all for changes with the login process, but we have a severe limitation on the uniqueness of accounts with multi-portal that is a technical limitation that WILL restrict what we can do.  I just encourage you to plan before implementing and to discuss the plan before working hard as there are many land-mines in this area

valadas:
---^^^---
What he said ;)

thabaum:
^^^^ this is what I am trying to do, can we please?  It doesnt break anything for me to change it because DNN CE is broken with respect to this stuff anyhow and I would love to see it all finally get fixed.  I am going to not do some very personal community things so I can devote my time to this community.   I am not DNN corp, I am DNN community just been trying to help for long long long time.  I have had a change in direction that brings me closer here and I am all for understanding DNN top to bottom every layer of the cake.  And I really wish to understand any concerns so I can build upon those as potential solutions that maybe the community can build upon.  But it is hard if these things dont get discussed I appreciate you letting me help, even if it is only for the small stuff.  But I believe the community of devs that are like me are lost as to how we can actually help.  I am trying to help pave the way as I have no clue...

thabaum:
But the Username issue being set to email:

To understand this correctly email address gets put over for the username.  Since when registering you dont enter a username.  This is working right to me.  And if you change your email it should change the username to the new email address.  

And when you use an authentication provider the Username adds the name of the Authentication Provider at the beginning.

Changing this is pretty big deal if this is what is going on I think it works as is.  My issue is more with the portals being separate so that you can have the same email on the same DNN instance just separate sites.

valadas:
I am confused, this is comments on a PR that is already merged, what is the actionable item we are discussing here? The ability to use the same username on multiple sites for different persons? If so then right now this is not possible since user accounts are instance-wide and users are only mapped or not to this or that site. Changing this would mean also changes to site groups.

In my opinion the best course of action is to not use multi-tenant unless you are aware and ok with this behaviour. Any way this is what I do. But if a change is offered for this behaviour I would welcome it providing it does not break any current behaviour for upgrades and I would suggest to open an issue with the proposed solution to agree on it before starting to work on a PR that could be refused for that reason.

thabaum:
So I seen this merged I suppose I misread things a little.  I see it is in the latest release now.

This PR makes it so current usernames don't change but if you register it will still put the email as username in the username field am I correct?

I will test this more tonight to see now that I understand it more.  Thanks for the PR anything that helps out registration system is awesome.  I chose DNN for this purpose of using multi-sites using same login for a project I have been working on.  I really wish to fine tune this so I will pay more attention to things.  Sorry for going off topic.  You guys rock!

PR #2628, opened 2019-03-15
keep exported tab modules in order.

### LINKED ISSUE #2627: imported modules don't keep in order if the tabmodules index broken
## Description of bug
we should not trust the sql index as it maybe broken in some scenarios, need order the export tab modules in codebase to make sure it in correct order.

## Steps to reproduce
1. install 9.2.2 and login as host; 
2. create a new page and add 5 html modules into same pane.(add some special text for them so that we can know the order, such as: test 1, test 2...) 
3. re-order the modules, such as 1-5-4-3-2; 
4. open the website db in SSMS, disable index IX_TabModules_ModuleOrder for TabModules table; 
5. run full export job, create a new empty child site, and import the package; 
6. compare the order between main site and child site.

## Current result
the modules is imported in wrong order.

## Expected result
modules should keep order after imported.

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1

zyhfish:
Fix #2627 .


ohine:
@daguiler Do you want this in v932 or v940?

daguiler:
v9.3.2, but changing base branch did not work as expected. @zyhfish can you fix this PR so that it only includes your changes?

zyhfish:
close this PR as will submit a new PR for 9.3.2.

Issue #2637, opened 2019-03-21
DDRMenu: Add Support for System.Web.WebPages


ahoefling:
Part of: #2599 
Part of #2619 

The DDRMenu is the default templating module in DNN which is a token based templating engine. The module supports many different Template Engines one being Razor. This means the module takes a direct dependency on `DotNetNuke.Web.Razor` which is scheduled to be removed in DNN 11.0 (#2618).

We need to implement a path forward for vendors and developers to continue to use their Razor Templates for DDRMenu in .NET Framework versions of DNN. 

## Proposed Solution
The `DotNetNuke.Web.Razor` Library has something called the `RazorEngine` which is a giant module wrapper around Microsoft's Razor Web Pages. Since there is already a technology out there called [ASP.NET Razor Web Pages (Razor 3)](https://docs.microsoft.com/en-us/aspnet/web-pages/overview/getting-started/program-asp-net-web-pages-in-visual-studio) we can leverage this technology to render our Razor Templates in the DDRMenu.

I have gone ahead and built a prototype of what this could look like in a ASP.NET Web Pages (Razor 3) project.

```c#
var page = (WebPage)WebPageBase.CreateInstanceFromVirtualPath("~/About.cshtml");
var httpContext = new HttpContextWrapper(HttpContext.Current);
var model = new HelloWorld();
var pageContext = new WebPageContext(httpContext, page, model);

var builder = new StringBuilder();
var writer = new StringWriter(builder);
page.ExecutePageHierarchy(pageContext, writer);

var output = builder.ToString();
```

In the code snippet above the `output` variable will be the rendered HTML. This snippet does *not* use any DotNetNuke libraries and was created by file new project.

PR #2638, opened 2019-03-21
DDRMenu: Add Support for Web Pages (Razor 3) Renderer

ahoefling:
Closes: #2637 

## Summary
The DDRMenu depends on `DotNetNuke.Web.Razor` which will be marked as deprecated in 9.x and removed in 11.0. As the platform moves forward we need to give developers and vendors ample time to migrate from the RazorEngine to [Microsoft's Razor 3](https://docs.microsoft.com/en-us/aspnet/web-pages/overview/getting-started/program-asp-net-web-pages-in-visual-studio) Implementation.

This PR adds an additional render support for `System.Web.WebPages.WebPage` which falls back to the RazorEngine if it isn't implemented.

Consider the following DDRMenu Razor Template
```
@using DotNetNuke.Web.DDRMenu;
@using System.Dynamic;
@inherits DotNetNuke.Web.Razor.DotNetNukeWebPage<dynamic>
... omitted code
```

This snippet currently works, and works with the proposed changeset. Developers should start migrating their templates to use System.Web.WebPages.WebPage which will be supported on 11.0 and up (until .NET Core).

```
@using DotNetNuke.Web.DDRMenu;
@inherits System.Web.WebPages.WebPage
... omitted code
```

## Developer Change
As developers migrate their templates we recommend that they use:

```
@inherits System.Web.WebPages.WebPage
```

### DIFF
--- a/DNN Platform/Modules/DDRMenu/DotNetNuke.Modules.DDRMenu.csproj
+++ b/DNN Platform/Modules/DDRMenu/DotNetNuke.Modules.DDRMenu.csproj
@@ -13,6 +13,7 @@
     <SccProvider>
     </SccProvider>
     <UseGlobalApplicationHostFile />
+    <Use64BitIISExpress />
   </PropertyGroup>
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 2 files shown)


ahoefling:
Per discussion with outside of GitHub with @ohine and @mtrutledge
> Let's create follow-up item which will set a standard for unit testing core modules

PR #2639, opened 2019-03-22
Remove version to allow GitVersion to set it at build time

ohine:


ohine:
Please don't merge yet. I have a couple more commits coming.

ohine:
The latest build is looking good, if we can get someone from @dnnsoftware/approvers to merge this (squash or rebase). That would be awesome!

Issue #2615, opened 2019-03-07
Update Assemblies to .NET Framework 4.7.2


ahoefling:
Part of: #2599 

## Description of problem
As DNN approaches a migration to .NET Core all the assemblies will need to implement .NET Standard 2.0 which is not supported by .NET Framework 4.5. The .NET Standard compatibility [document](https://github.com/dotnet/standard/blob/master/docs/versions.md) states that .NET Framework 4.7.2 is recommended for the best .NET Standard 2.0 compatibility.

During DNN Summit 2019 Jon Galloway of Microsoft recommended we update the platform to 4.7.2 as there aren't many significant API changes that would cause breaking changes. 

## Description of solution
Update every project in `DNN_Platform.sln` to build under .NET Framework 4.7.2

| Project | Current Version |
|---------|-------------------|
| Dnn.Modules.Console | 4.5 |
| Dnn.Modules.ModuleCreator | 4.5 |
| Dnn.AzureConnector | 4.5 |
| Dnn.GoogleAnalyticsConnector | 4.5 |
| DnnExportImport | 4.5 |
| DnnExportImportLibrary | 4.5 |
| DotNetNuke.Modules.CoreMessaging | 4.5 |
| DotNetNuke.Modules.DDRMenu | 4.5 |
| DotNetNuke.Modules.DigitalAssets | 4.5 |
| DotNetNuke.Modules.Groups | 4.5 |
| DotNetNuke.Modules.Html | 4.5 |
| DotNetNuke.Modules.HtmlEditorManager | 4.5 |
| DotNetNuke.Modules.Journal | 4.5 |
| DotNetNuke.Modules.MemberDirectory | 4.5 |
| DotNetNuke.Modules.RazorHost | 4.5 |
| DotNetNuke.Web.Deprecated | 4.5 |
| DotNetNuke.Website.Deprecated | 4.5 |
| Dnn.AuthServices.Jwt | 4.5 |
| DotNetNuke.Authentication.Facebook | 4.5 |
| DotNetNuke.Authentication.Google | 4.5 |
| DotNetNuke.Authentication.LiveConnect | 4.5 |
| DotNetNuke.Authentication.Twitter | 4.5 |
| DotNetNuke.Providers.AspNetCCP | 4.5 |
| DotNetNuke.Providers.FolderProviders | 4.5 |
| DotNetNuke.Skins.Xcillion | 4.5 |
| CountryListBox | 4.5 |
[...]

PR #2644, opened 2019-03-25
Upgrade DNN to .NET Framework 4.7.2

ahoefling:
Closes: #2615

## Summary
Upgrades all `.csproj` and `.vbproj` files to use .NET Framework 4.7.2. Verified the build still works and all unit tests are passing locally

### DIFF
--- a/DNN Platform/Admin Modules/Dnn.Modules.Console/web.config
+++ b/DNN Platform/Admin Modules/Dnn.Modules.Console/web.config
@@ -1,11 +1,14 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-	<runtime>
+  <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30AD4FE6B2A6AEED" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-10.0.0.0" newVersion="10.0.0.0" />
       </dependentAssembly>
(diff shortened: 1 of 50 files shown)


DNN Platform/Admin Modules/Dnn.Modules.Console/web.config
```diff
@@ -1,11 +1,14 @@
...
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30AD4FE6B2A6AEED" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-10.0.0.0" newVersion="10.0.0.0" />
       </dependentAssembly>
     </assemblyBinding>
-	</runtime>
-</configuration>
+  </runtime>
+  <system.web>
```
bdukes: if this gets deployed, I think IIS will complain because it's not allowed to have this element declared except in the root `web.config`

ahoefling: I wasn't 100% sure about this, I can go and remove it. Thanks for the feedback

ahoefling:
@bdukes Thanks for the feedback, I made changes as you suggested.

ahoefling:
@mitchelsellers 
> LGTM - Thanks for getting on this one so quickly

No problem at all, another step in the right direction for the platform.

What is the milestone for this? 9.3.1 or 9.4? I want to update the Razor Pages Epic

mitchelsellers:
Milestone will be 9.4.0.  I'll merge it in with one last confirmation with @ohine

Issue #2648, opened 2019-03-27
Resolve Build Warnings on DNN.Platform Solution


mitchelsellers:
When building the DNN Solution in Visual Studio a number of build warnings appear for code-quality related elements.  This clustters the build logs and makes the identification of actual issues complicated.

## Description of bug
Building DNN Platform results in 676 Build Warnings, making it impossible to identify if there are any "actual" issues.

## Steps to reproduce
1. Open the solution
2. Build
3. View Output Window and notice 676 Warnings

## Expected result
Zero build warnings

valadas:
Please keep it Stalebot :)

eng-aomar:
Hello,
Regarding this issue, I have set up and built the dnn.platform successfully, but only 62 warnings appear.
please note that I am using Visual studio 2019 community
![dnn_build](https://user-images.githubusercontent.com/55650010/66452541-91cb5b80-ea69-11e9-8fe0-55571391843f.jpg)

mitchelsellers:
@eng-aomar That is about right.  I'm going to go through this again and fix everything soon.

eng-aomar:
I want to contribute to this issue if it is assigned to nobody.

mitchelsellers:
@eng-aomar We would welcome contributions for this.

Please do include detail comments for any changes made to help with review

valadas:
Just a note @eng-aomar if you do this soon and without breaking changes you could target the release/9.4.x branch for a faster merge. If they are not all solved in one go, any partial fixes are also a welcome step in the right direction :)

eng-aomar:
@valadas  I will do my best.
 Thank you very much, your video titled [Setting up the Tools for DNN Development] on youtube was of great help (https://www.youtube.com/watch?v=zLB6uzdqGF8)

valadas:
Yeah, some things on this playlist are now outdated, but the good news is that it is now easier, we merged the AdminExperience repository into this one, so that part is now easier to test/build. Also you can now just run in powershell ./builds.ps1 to generate all assets into the Artifacts directory. Just ping us here if you run into any issue building.

eng-aomar:
According to the warnings with the code CS0618 such as for example this one "Upgrade.AddPortal(XmlNode, bool, int)' is obsolete: 'Deprecated in DNN 9.3.0, will be removed in 11.0.0. Use the overloaded method with the 'superUser' "
when tracing doing the warning in the install.aspx.cs file
[...]

PR #2649, opened 2019-03-27
Build Warning Reduction (Code Style Standardization)

mitchelsellers:
Associated to #2648 

## Summary
Includes resolution of a number of specific types of build warnings
* IDE0028
* IDE0030
* IDE0031
* IDE0017
* IDE0029
Reduces the number of build warnings from 667 to 469.

## Review Notes

This is designed to help improve the standardization & readability.  With an eventual goal of being able to validate that no additional warnings occur in the future.

### DIFF
--- a/DNN Platform/Admin Modules/Dnn.Modules.Console/Components/ConsoleController.cs
+++ b/DNN Platform/Admin Modules/Dnn.Modules.Console/Components/ConsoleController.cs
@@ -47,10 +47,12 @@ namespace Dnn.Modules.Console.Components
         /// <returns>A list with different icon types</returns>
         public static IList<string> GetSizeValues()
         {
-            IList<string> returnValue = new List<string>();
-            returnValue.Add("IconFile");
-            returnValue.Add("IconFileLarge");
-            returnValue.Add("IconNone");
+            IList<string> returnValue = new List<string>
+            {
+                "IconFile",
+                "IconFileLarge",
+                "IconNone"
+            };
             return returnValue;
         }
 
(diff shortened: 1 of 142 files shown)


DNN Platform/Admin Modules/Dnn.Modules.Console/Components/ConsoleController.cs
```diff
@@ -47,10 +47,12 @@ public class ConsoleController
...
-            IList<string> returnValue = new List<string>();
-            returnValue.Add("IconFile");
-            returnValue.Add("IconFileLarge");
-            returnValue.Add("IconNone");
+            IList<string> returnValue = new List<string>
+            {
+                "IconFile",
+                "IconFileLarge",
+                "IconNone"
+            };
```
ahoefling: I don't have a preference in the style, but I see a lot of changes like this. I am more curious if these were recommendations from tools such as Resharper that made these recommendations. Should we start discussing a style guide that we want to start enforcing?

bdukes:
Those IDE build warnings are built-in style rules in Visual Studio.  They can be [configured via EditorConfig](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017), and we should probably attempt to create a config for the style to which we want to adhere.

ahoefling:
@bdukes this is what I was thinking as well. We should really think about adding an EditorConfig

donker:
@bdukes @ahoefling Are you proposing we hold off on this PR until we have the correct config in the editorconfig? Or do we merge and worry about the config later?

bdukes:
I'm fine with merging as is, we can look at updating configuration in the future

valadas:
Will be resubmitted in Dnn 10

PR #2666, opened 2019-04-04
Increase version column in desktopmodules table to support revision number.

### LINKED ISSUE #2665: Desktop Modules Only Support Versioning of up to 8 Characters
Requesting increase the version column to nvarchar(50), so it can support the version with a revision number.

zyhfish:
Fix #2665


mitchelsellers:
Has this been tested to determine the impacts on other areas that could be impacted?  Initial thoughts indicate testing of.

* Extensions Page
* Upgrade Service
* Extension Validation Service

This is just at first pass

mitchelsellers:
@daguiler we’re my questions tested for impact?

daguiler:
Maybe not fully. I probably clicked the button too quickly @mitchelsellers. But we have time to test/fix/revert if we find anything wrong

PR #2677, opened 2019-04-09
(partial telerik removal) remove Telerik from Platform.

### LINKED ISSUE #2439: Telerik Removal
# Description of Problem
We still utilize the out of date telerik components in a few places throughout the platform. We need to finish up the complete removal so we can finally remove these components from new installations. 

Right now it appears that the digital assets module (dnn file manager) and the Html module are the major items that's holding us back. They both still reference and make heavy usage on the `DotNetNuke.Web.Deprecated` web controls. 

## Solution # 1
Remove usage and replace with asp.net components. 

## Solution # 2
Rewrite these as a SPA module and fix both the telerik and webforms issue at the same time. 

### Current Usage
A quick search on the solution found the following areas which should be address. This list isn't complete and will be updated as additional areas are identified. 

- DotNetNuke.Web.Deprecated
- DotNetNuke.Website.Deprecated
- DotNetNuke.Modules.DigitalAssets
- DotNetNuke.Modules.Html

### Host Settings

- EnableTelerikCDN
- TelerikCDNBasicUrl
- TelerikCDNSecureUrl

### Build Scripts
Module.build (spell checker, assembly copy, package creation)
CreateCommunityPackages.build (file copies, package creation)

### Other files
Tests\App.config
Config\SiteUrls.config
WebControlSkin (css styles)
admin.css (css styles)
dnn.jquery.js (RadComboBox type check)

zyhfish:
Fix #2439


mitchelsellers:
I believe that for this we will want to keep the web.deprecated assembly, at least until 10.x as removing this, and the associated sub-controls would be a breaking change.

kurtwilbies:
Remove Telerik. There is no way back. Communicating this change to potential users is enough (blogging: explain how to use web.deprecated assembly in 9.4.0 and that's enough; the only question is how to use module settings ???).

mitchelsellers:
We cannot accept this PR at this time, due to the breaking change and not matching the workflow of how we will remove Telerik.

PR #2683, opened 2019-04-09
Data Consent feature for DNN

donker:
## Summary

This PR implements a number of enhancements to the platform to allow administrators to require users' consent to storage and usage of data. This is an essential part of GDPR compliance in Europe.

In essence the login experience is enhanced with a new step "DataConsent" which kicks in once the admin has activated the feature for the site. A user's consent is recorded as well as the time it was done. The action is also logged. This allows basic forensics to take place. For more advanced scenarios this solution also allows the Data Consent step to be done elsewhere on the site through a "Consent Redirect" setting.

### Account removal

An integral part of consent is the option to opt out and even delete one's profile. Because it is hard to tell if all extensions will respond well to a user being eliminated from the database, the solution provides three mechanisms for user account deletion (by the user):

1. Off. Account deletion remains the responsability of the site owner.
2. Manual. The user is marked as wishing to be removed and the account is unauthorized.
3. Delayed Hard Delete (default). The user is marked as wishing to be removed and deleted. The user will be removed after a specified time. This allows admins to recuprate an account in case of a mishap during the grace period.
4. Hard Delete. The account is immediately removed.

Note that to achieve number 3, a scheduled task has been added.

### Discussion

Although care has been taken to provide a powerful mechanism that will address most needs for GDPR compliance in this area it is noted that always be more powerful and complete. The criteria with which I began this were that it would be as minimal an intervention in terms of code changes and additions and it would allow 3rd party developers to leverage this to create a more powerful system.

### Testing

Note what needs to be done is to check more complex integration scenarios With that I’m thinking:
- multi portal setups with portal groups
- users that have to go through various combo’s of: Data Consent, Update Password and Complete Profile. These items are now “on the same level” in the framework, so the risk of interference is greatest at that point.
- test the extension point. Can you create something simple to retrieve the user from the token and then do some fancy consent logic and mark the user as consented in the framework?


DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -2407,7 +2412,7 @@ public virtual void RestoreUser(int userId, int portalId)
                                       isApproved,
                                       refreshRoles,
                                       lastIpAddress,
-                                      passwordResetToken,
+                                      GetNull(passwordResetToken),
```
mitchelsellers: Was this a "bug fix" along the way?

DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -1794,6 +1809,46 @@ public static bool RemoveUser(UserInfo user)
             return retValue;
         }
 
+        /// <summary>
+        /// Delete the contents and folder that belongs to a user in a specific portal
+        /// </summary>
+        /// <param name="user">The user for whom to delete the folder. 
+        /// Note the PortalID is taken to specify which portal to delete the folder from.</param>
+        public static void DeleteUserFolder(UserInfo user)
```
mitchelsellers: For the purposes of future-proofing, can we make this private/internal to avoid adding new public API's (Unless we have to)

DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -2028,7 +2083,11 @@ internal static void UpdateUser(int portalId, UserInfo user, bool loggedAction,
                 DataCache.ClearUserCache(portalId, user.Username);
             }
 
-            if (!user.Membership.Approving) return;
+            if (!user.Membership.Approving)
```
mitchelsellers: Can this format change back to the inlined version for consistency with the rest of the platform

DNN Platform/Library/Security/Roles/RoleController.cs
```diff
@@ -654,6 +654,22 @@ public static bool DeleteUserRole(UserInfo objUser, RoleInfo role, PortalSetting
             return canDelete;
         }
 
+        /// <summary>
+        /// Completely remove all a user's roles for a specific portal. This method is used when 
+        /// anonymizing a user
+        /// </summary>
+        /// <param name="user">User for which all roles must be deleted. The PortalId property
+        /// is used to determine for which portal roles must be removed.</param>
+        public static void DeleteUserRoles(UserInfo user)
```
mitchelsellers: Same question on this one about possible internal/private

DNN Platform/Library/Services/Users/PurgeDeletedUsers.cs
```diff
@@ -0,0 +1,94 @@
...
+                                    break;
+                                case "d":
+                                    thresholdDate = DateTime.Now.AddDays(-1 * settings.DataConsentDelay);
+                                    break;
+                                case "w":
+                                    thresholdDate = DateTime.Now.AddDays(-7 * settings.DataConsentDelay);
+                                    break;
+                            }
+                            var deletedUsers = UserController.GetDeletedUsers(portal.PortalID);
+                            foreach (UserInfo user in deletedUsers)
```
mitchelsellers: Do we have any concerns on the amount of records that we will get with this?  Going forward it should be a small number, but not sure if this could get out of hand and limit the ability to process successfully with the 30 second query timeout.

DNN Platform/Library/Data/DataProvider.cs
donker: Looks like it to me. I tried to update using Null.NullGuid and the field was not NULL. So IMO that is not the behaviour one expects. From what I can tell no code actually tries to clear the reset token. There is no mechanism for this. All code just lets it expire. I wanted to NULL it so no open vectors remained to an account.

DNN Platform/Library/Services/Users/PurgeDeletedUsers.cs
donker: It runs as a scheduled task. I could not imagine a scenario where the amount of soft deleted accounts was too large to handle in memory or time. Isn't it OK for this to time out and pick up again after 30 mins in the exceptional case that the site has massive amounts of soft deleted users?

DNN Platform/Library/Entities/Users/UserController.cs
donker: ctrl-K ctrl-D does this for me in VS. I'll change it back.

DNN Platform/Library/Security/Roles/RoleController.cs
donker: Done

DNN Platform/Library/Entities/Users/UserController.cs
donker: Done

DNN Platform/Library/Services/Users/PurgeDeletedUsers.cs
sleupold: if a sproc timed out due to the amount of users, no records will be affected ad the same happen again on next run. However, I agree, a huge number should be an edge case.

DNN Platform/Library/Data/DataProvider.cs
sleupold: are you able to pass database Null instead?

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
```diff
@@ -0,0 +1,168 @@
...
+GO
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}ResetTermsAgreement]') AND type in (N'P', N'PC'))
+	DROP PROCEDURE {databaseOwner}[{objectQualifier}ResetTermsAgreement]
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}ResetTermsAgreement]
+	@PortalId	INT
+AS
+	UPDATE {databaseOwner}{objectQualifier}UserPortals
```
sleupold: Isn't it irritation/misleading, if HasAgreedToTermsOn has a value but HasAgreedToTerms not?

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
```diff
@@ -0,0 +1,168 @@
...
+INSERT INTO {databaseOwner}[{objectQualifier}EventLogTypes] ([LogTypeKey], [LogTypeFriendlyName], [LogTypeDescription], [LogTypeOwner], [LogTypeCSSClass]) 
+VALUES (N'DATA_CONSENT_SUCCESS', N'User Consented', N'', N'DotNetNuke.Logging.EventLogType', N'OperationSuccess')
+GO
+
+INSERT INTO {databaseOwner}[{objectQualifier}EventLogTypes] ([LogTypeKey], [LogTypeFriendlyName], [LogTypeDescription], [LogTypeOwner], [LogTypeCSSClass]) 
+VALUES (N'DATA_CONSENT_FAILURE', N'User Failed to Consent', N'', N'DotNetNuke.Logging.EventLogType', N'AdminAlert')
+GO
+
+IF NOT EXISTS (SELECT * FROM {databaseOwner}{objectQualifier}EventLogConfig WHERE LogTypeKey = 'DATA_CONSENT_SUCCESS')
+BEGIN
```
sleupold: What is the benefit of just keeping the past 10 records?
IMO success records should only be kept, if there is a concrete need - otherwise logging should turned off.

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
```diff
@@ -0,0 +1,168 @@
...
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}DeleteUserProfile]') AND type in (N'P', N'PC'))
+	DROP PROCEDURE {databaseOwner}[{objectQualifier}DeleteUserProfile]
+GO
+
+CREATE PROCEDURE {databaseOwner}[{objectQualifier}DeleteUserProfile]
+	@PortalId	INT,
+	@UserId INT
+AS
+    IF EXISTS (SELECT * FROM {databaseOwner}{objectQualifier}UserProfile up
```
sleupold: IMO this distinction is not correct, it doesn't cover superuser profiles properly.
You should distinguish upon @PortalID parameter, whether to delete portal specific properties (positive value) or all properties (PortalID is Null or -1). It depends on Userportals records, whether to call with portalID (other userportals record with other portalID exists) or without (otherwise).

DNN Platform/Library/Data/DataProvider.cs
donker: The default value of this column is NULL in SQL. Without this change I can never reset to that value. But with this change the column will be NULL again. So I guess this is actually exactly what you want Sebastian.

DNN Platform/Library/Services/Users/PurgeDeletedUsers.cs
donker: > 
> 
> This looks really great, and thank you for the contribution! It's going to make a lot of people quite happy to see this in the release notes for v9.4.0!
> 
> Can you easily move the sql changes into the main 940 file? This early in the cycle I'd rather keep them in a single file, but since we really don't have anything defined for sql changes it's really just my personal preference for release management and upgrade performance and not a required change at all.

Done

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
mitchelsellers: @sleupold I believe the reason to leave the value is that you can infer "last agreed" by this.

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
mitchelsellers: @sleupold This is a good idea.  @donker Possibly disabled by default?  Thoughts.  Users could turn this on, but the data is in fact recorded elsewhere

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
sleupold: IN this case, I would store "HasAgreedToTermsVersion" instead of "HasAgreedToTerms" and compare with PortalSetting "TermsVersion", which would be increased by the resetButton.

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
donker: OK. Done. It's gone.

Website/Providers/DataProviders/SqlDataProvider/09.04.00.01.SqlDataProvider
donker: I can't tell what this is about.

Issue #2691, opened 2019-04-12
IP is missing in the logs when user request comes with IPv6


mikebigun:
## Description of bug
Admin log in DNN is not displaying the IP address when user request is coming with IPv6. 

## Steps to reproduce
1. Login to the system using REMOTE_ADDR having IPv6 address
2. Go to Manage -> Admin Logs

## Current result
When user logged in, IP address is missing in the logs. The only way to see incoming IP is to check IIS logs in where it is captured.

## Expected result
When users are logging in, administrators would like to be able to check IPv6 addresses in the logs if there are some.

## Screenshot of missing IP

![image](https://user-images.githubusercontent.com/22524011/56043363-c756ec80-5d45-11e9-8594-fa249cfffcda.png)

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

PR #2692, opened 2019-04-12
IP is missing in the logs when user request comes with IPv6 

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2691

In this RP I extended existing mechanism to read IP address from request. Default IP remains IPv4, nothing is changed in terms of application functionality and public API's. Only thing is now it is possible to extract IPv6 from request. I used it just for the admin logs purposes when user is logging in. So administrators can now check logs and determine IPv6 addresses. 

See how does it look:

image

### DIFF
--- a/DNN Platform/Library/DotNetNuke.Library.csproj
+++ b/DNN Platform/Library/DotNetNuke.Library.csproj
@@ -1491,6 +1491,7 @@
       <SubType>Code</SubType>
     </Compile>
     <Compile Include="Services\UserProfile\UserProfilePicHandler.cs" />
+    <Compile Include="Services\UserRequest\IPAddressFamily.cs" />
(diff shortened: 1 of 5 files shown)


bdukes:
Do we think there's a problem if we start returning IPv6 addresses from `IUserRequestIPAddressController.GetUserRequestIPAddress`?  It seems like there's a lot going on here when we could just return whatever the IP address was, in whatever format is correct.

mikebigun:
Hi @bdukes @daguiler 
We had `IUserRequestIPAddressController.GetUserRequestIPAddress` dedicated to get IPv4 only. It is public API, so the way I'm thinking is that potentially it can be in use in other 3rd party features.
To avoid regressions, I decided to keep it unchanged for existing API consumers and make it more flexible if someone want to get IPv6.

PR #2709, opened 2019-04-19
DNN-29425 Reduce number of calls for ModuleSettings

nadeem-akhtar-nu:
## Summary

Number of calls for stored procedure GetModuleSettings was reported very high. It was identified that these calls are generating from CKEditorProvider from class: SettingsUtil functions: CheckExistsModuleInstanceSettings and CheckExistsModuleSettings

It could be seen that current version of CKEditorProvider has already been modified to use other function call which is subsequently using cache. 
These are probable reasons for clients using older version of CKEditorProvider 

- DNN up to 9.1.x is using older version of CKEditorProvider
- For custom builds, CKEditorProvider might not be updated and hence continue using older version

The provided solution is to use cache for GetModuleSettings to avoid heavy stress from database.

### DIFF
--- a/DNN Platform/Library/Obsolete/ModuleController.cs
+++ b/DNN Platform/Library/Obsolete/ModuleController.cs
@@ -105,6 +105,18 @@ namespace DotNetNuke.Entities.Modules
         [Obsolete("Deprecated in DNN 7.3.  Please use the ModuleSettings property of the ModuleInfo object")]
         public Hashtable GetModuleSettings(int ModuleId)
         {
+            var cacheKey = string.Format(DataCache.ModuleSettingsByModuleIdCacheKey, ModuleId);
+            var moduleSettings = ModuleId <= 0 ? GetModuleSettingsInternal(ModuleId) :
+             CBO.GetCachedObject<Hashtable>(
+                new CacheItemArgs(cacheKey, DataCache.ModuleCacheTimeOut, DataCache.ModuleCachePriority),
+                c => GetModuleSettingsInternal(ModuleId)
+             );
+
+            return moduleSettings;
+        }
+
+        private Hashtable GetModuleSettingsInternal(int ModuleId)
+        {
             var settings = new Hashtable();
             IDataReader dr = null;
             try
--- a/DNN Platform/Library/Common/Utilities/DataCache.cs
+++ b/DNN Platform/Library/Common/Utilities/DataCache.cs
@@ -164,6 +164,7 @@ namespace DotNetNuke.Common.Utilities
 
         public const string ModuleCacheKey = "Modules{0}";
         public const string ModuleSettingsCacheKey = "ModuleSettings{0}";
+        public static string ModuleSettingsByModuleIdCacheKey => "ModuleSettingsModuleId{0}";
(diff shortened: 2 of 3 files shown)


bdukes:
@nadeem-akhtar-nu thanks for the contribution, however I'm not sure if this change will end up solving your problem.  If someone is experiencing this issue with the CKEditor Provider, upgrading to this next version of DNN (your PR currently targets DNN 9.4.0) would resolve the problem via the included CKEditor Provider package, so this change won't be in use.  Am I missing a scenario?  Is there a reason a site would continue using an older version of the CKEditor Provider without its fixes?

mitchelsellers:
Additionally, if we add additional caching for this retrieval, the saving/updating of a setting would need to invalidate the cache.

nadeem-akhtar-nu:
@bdukes 
Thanks for your valuable feedback.
My finding is, CKEditor changes applied in release 9.2.0 from Mar 27, 2018 but it could be seen that couple of databases (pdnn_dnnowj8s8 and pdnn_dnnowj8s8_sandbox) showing version 9.2.1 last updated on August 14, 2018.
The pdnn_dnnowj8s8 logs from January 01, 2019 to April 01, 2019 shows 37,812,404 calls to the SP which is about 420,137 calls per day and for some days more than 1M calls. This number is alarming for Central DB team so they raised this ticket for optimization.
By considering that if some installation will upgrade the CKEditor library than current change would not impact functionality. But in my view, till then the number of calls to stored procedure problem requires some solution from core. Also please note that the module count in pdnn_dnnowj8s8 is 432 so there must be lot of repeating calls for same module settings.

@mitchelsellers 
That is correct and this point is already considered in PR by removing cache entry on module setting changes.

DNN Platform/Library/Obsolete/ModuleController.cs
```diff
@@ -104,6 +104,18 @@ public void DeleteTabModuleSettings(int tabModuleId)
         [EditorBrowsable(EditorBrowsableState.Never)]
         [Obsolete("Deprecated in DNN 7.3.  Please use the ModuleSettings property of the ModuleInfo object")]
         public Hashtable GetModuleSettings(int ModuleId)
+        {
+            var cacheKey = string.Format(DataCache.ModuleSettingsByModuleIdCacheKey, ModuleId);
+            var moduleSettings = ModuleId <= 0 ? GetModuleSettingsInternal(ModuleId) :
```
aaniscenko: Why there is a check for ModuleId <= 0  is there any difference with positive ModuleId?

sleupold: -1 usually equals Null in DNN - e.g. for a newly created module, which hasn't been saved yet.

mitchelsellers:
Based on the review feedback, and due to the change being introduced only to support an Obsolete method, which will be removed in 10.0, Sadly we cannot accept this change.

As @daguiler mentions, the possibility of other impacts, as well as the addition of technical debt for a feature that will be removed in the next subsequent release (As 9.4.0 will be followed by 10.0) at this time it does not make sense to adjust.

The CKEditor project, if the one incurring this hit, needs to be updated to use the proper API's as the true long-term fix for this.

PR #2715, opened 2019-04-19
Remove references to old modules

### LINKED ISSUE #2714: Old modules referenced in new sites
## Description of bug
When creating a new site, there is a User Accounts page under Admin, which has a module that doesn't work.  There are also references to many other modules which were replaced by the Persona Bar throughout the default templates and resources.

## Steps to reproduce
1. Create a new DNN site
2. Go to the Extensions page in the Persona Bar
3. Make sure Modules is the selected type
4. Scroll down to see User Accounts and view the page on which it is in use

## Current result
A page with an empty module, which shows an error message when attempting to view its settings

## Expected result
Page and module do not exist

## Screenshots
Module Usage for Users and Roles

Settings error for User Accounts module

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

bdukes:
Fixes #2714


bdukes:
I haven't tested these changes yet, so I'm sure there's some link that I'll have missed that'll break something.

I'd like to remove the Users and Roles package and/or desktop module, but I think some of those controls are used by auth providers, so I left them for the time being.  That is, from a SQL perspective, I only removed a `ModuleControls` row if the ascx file doesn't exist.

ohine:
This is awesome! Thank you for the contribution!

I'm curious, the `DotNetNuke.Data.SqlDataProvider` file. Are you manually editing it? Or is there a tool you're using to regenerate it?

bdukes:
I'm using a [LINQPad](https://www.linqpad.net/) query to read the file and remove rows, but it's mostly by hand.  I would be great if we had some way to auto-generate the file.

I'm also surprised how much is in that file, it seems like the installer and templates _should_ be able to handle most of this data, and avoid this situation where isn't out-of-sync.

ohine:
Bummer, I was really hoping you'd written that magic generator 😃 

It's been on my list to figure this piece out, it's one of those rainy day tasks and just haven't had a chance yet. 

I would love to bump things from [v08.00.04](https://github.com/dnnsoftware/Dnn.Platform/blob/development/Website/Install/DotNetNuke.install.config.resources#L7) to v09.04.00 for a clean install with the next release. It would allow us to document all the various pieces and possible come up with a better way to automatically bump this on each major release.

bdukes:
I ran through a test of these changes in a new install, and they all appeared to work correctly, no SQL errors logged, and the admin page is gone.  However, there's still a Host page for Superuser Accounts.  

Looking into it, it appears that the page was [removed in 9.1.0](https://github.com/dnnsoftware/Dnn.Platform/commit/6922058f006f7c67fc4877938f093da52f1431fc), but [added back in 9.2.0](https://github.com/dnnsoftware/Dnn.Platform/commit/908c55d742d894a7f19ddc990cdb06ef617562b6).  The related issue in JIRA was [DNN-10420](https://dnntracker.atlassian.net/browse/DNN-10420), which lead me to discover that the Persona Bar is using this module via an iframe to allow editing profile properties.  

So both this host page and the site-level admin page will need to remain in place until that functionality is migrated completely into the Persona Bar.  I've amended the PR to leave the `\\Admin\\UserAccounts` page.

Otherwise, my testing was successful, so this is ready to merge from my perspective.

mitchelsellers:
@bdukes Do we have a new issue to track the removal of those pages later by chance?  I can create one if not.

bdukes:
I'm not aware of any issue to remove those pages, feel free to create it.  Thanks!

PR #2722, opened 2019-04-21
update version labels for various build types / branch names

ohine:
## Summary
This pull request should change the build labels for pull requests to -pr from -PullRequest, release* to -rc, and other branch as -unstable.


valadas:
I am not sure if my comment makes sense, but what about stable/final releases ?

ohine:
These are the default tags, if we tag a commit it overrides these labels. I think...

PR #2732, opened 2019-04-23
Fixed bug in Azure folders

daguiler:
## Summary
Fixes a bug in ea26f8356df3efeb874653e9b7b3b92813c158de.
When the **Sync Batch Size** parameter is left blank, it's saved as an empty string instead of a null string. And empty strings throw an error when cast to integer.

## Steps to reproduce
1. Login as Host
2. Open Persona Bar > Manage > Global/Site Assets
3. Add New Folder Type > Set up Azure Folder Provider (**leave Sync Batch Size blank**)
4. Add New folder with Azure Folder type
5. Synchronize the folder and subfolders

## Current behavior
Following error occurs:

image


valadas:
I usually do it as a one line, but looks fine to me too that way.

```c#
int intValue = int.TryParse(stringValue, out intValue) ? intValue : defaultValue;
```

PR #2738, opened 2019-04-25
Update README to include links to archive repos

### LINKED ISSUE #2737: Update README.md to include links to Dnn.Releases.Archive repos
## Description of problem
Currently, there are no links in the README to the **DNN Releases Archive** repositories.  This would be super helpful in pointing people on how to find older releases of DNN. 

## Description of solution
Add new section specifically for **DNN Release Archives**.  This section would contain a clear and concise list of all DNN Releases Archive repositories.

n/a

## Screenshots
n/a

## Additional context
@dnnsoftware/approvers please assign this issue to me.  I'll submit a PR for it.  Thanks!

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

nvisionative:
## Summary
Update README.md to include links to Dnn.Releases.Archive repos

Resolves #2737

I wasn't sure which branch to target, so feel free to update to match the correct branch target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,18 @@ All the details on contributing with Git and on our working methods can be found
 
 In addition, since DNN is part of the .NET Foundation, we require our contributors to abide by their [Code of Conduct](https://www.dotnetfoundation.org/code-of-conduct) rules and requirements also.
 
+## DNN Releases Archives
+
+* [DNN Releases Archive 1.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.1x)
+* [DNN Releases Archive 2.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.2x)
+* [DNN Releases Archive 3.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.3x)
+* [DNN Releases Archive 4.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.4x)
+* [DNN Releases Archive 5.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.5x)
+* [DNN Releases Archive 6.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.6x)
+* [DNN Releases Archive 7.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.7x)
+* [DNN Releases Archive 8.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.8x)
+* [DNN Releases Archive 9.x](https://github.com/dnnsoftware/Dnn.Releases.Archive.9x)
+
 ## .NET Foundation
 
 This project is supported by the [.NET Foundation](https://dotnetfoundation.org).


mitchelsellers:
I would prefer that we find a better location to promote/showcase this information.  There has been historical concern regarding individuals accidentally gaining access to older versions thinking they are supported etc.  To this point, I wouldn't want to promote these old archive releases for the same reason.

I would offer two suggestions as an opportunity for this.

1.) Create a new document, such as upgrading.md, inside of the .github folder.  In there include the suggested upgrade path (https://www.dnnsoftware.com/wiki/suggested-upgrade-path) Then at the bottom of this it would be easy to reference these safely.  I would also include a note about all of those containing releases with known security vulnerabilities and that they should not be used for production sites.

2.) Add the desired links into the WIKI article, to the same effect.

I'm fine either way, but I don't want to promote this on the homepage.

nvisionative:
This is not about upgrades alone.  There are many use cases for needing an old release package that is nowhere else to be found.  This is about helping people find the files they don't know even exist but are trying to find here.  We have blogged about this, promoted it via DNN meetups, social post, etc.  However, It is still a **very common** occurrence to have to help point someone to these files for their needs (upgrades, testing backward compatibility, etc.).  I have personally had five recommendations to place this information specifically here (be that right, wrong or indifferent) and said I would be happy to create an issue and submit a PR.

I personally believe we should do everything we can to make this easier for people to find.  If there is a better place to put this much-needed information (here on GitHub), great.  I just don't believe the recommendations meet all the needs (use cases) for accessing these older releases.  

These are the current official repos for the archived official releases of DotNetNuke / DNN dating back to the very beginning.  

If we are trying to keep things leaner on the README, that is totally understood.  Perhaps a separate markdown file can be created to hold these links and simply linked to in the **Quick Links** section.  Would that work?

By the way, I will absolutely work on finding a home for this information in the new [DNN Docs](https://dnndocs.com).  That's a great idea!

nvisionative:
FYI...a new issue has now been created on the DNNDocs repo to add this documentation.
https://github.com/DNNCommunity/DNNDocs/issues/287

hismightiness:
I agree with Mitch and Oliver about this.  I don't think we should be making it easier for insecure software to be found and installed.  The documentation area is the best placement for this.  This makes it available to the people that really do need older versions for the right reasons, and makes the availability limited (so as to prevent accidental spread of insecure DNN installations).  :)

nvisionative:
That is a valid argument in my opinion @hismightiness - thanks for articulating it that way.  Nonetheless, we get A LOT of questions from community members that are looking for these older releases, for valid reasons.  Are you recommending we don't make this easier at all for them to find in the spot they are looking (this GitHub repo)?  To me, making it clear that any link(s) provided to them have a **(Not Recommend For Production Use)** or a similar message.

mitchelsellers:
@nvisionative I think that it would be an acceptable compromise to link a separate document, but I would prefer that it include detailed warnings & notices.  

I can work with you on a few ideas on this if absolute, but I'd rather reference the documentation as the sole way. And we could easily link to that from the readme

nvisionative:
Okay, we can chat offline about what absolutely needs to be in the documentation as it relates to warnings/notices.  I'll work on getting that into the new DNNDocs.  Then I'll circle back to this issue and submit a PR to add a simple link to the new documentation.  Thanks for working with me on this.  It should help alleviate some of the ongoing difficulties in finding these.

I'll also go ahead and close #2737

valadas:
I am all for having short links in the readme to popular "looked for" stuff and have that point to a more detailed section in the docs.

ohine:
Closing the PR as it sounds like they'll be another pull request later down the road which addresses the concerns. If that's not the case, please reopen the PR.

PR #2744, opened 2019-04-29
Unregister user to use same logic as in Data Consent

donker:
If Data Consent is enabled a user must be offered the same deletion choice as when they consent. So this PR covers the scenario when a user has already consented but wishes to delete their account.

### DIFF
--- a/Website/DesktopModules/Admin/Security/EditUser.ascx
+++ b/Website/DesktopModules/Admin/Security/EditUser.ascx
@@ -10,6 +10,25 @@
 
 <%@ Import Namespace="DotNetNuke.UI.Utilities" %>
 
+<%
+  var confirmText = ClientAPI.GetSafeJSString(LocalizeString("UnregisterUser"));
+  if (PortalSettings.DataConsentActive && User.UserID == UserInfo.UserID)
+  {
+    switch (PortalSettings.DataConsentUserDeleteAction)
+      {
+        case PortalSettings.UserDeleteAction.Manual:
+          confirmText = ClientAPI.GetSafeJSString(Localization.GetString("ManualDelete.Confirm", "~/DesktopModules/Admin/Security/App_LocalResources/DataConsent.ascx.resx"));
+          break;
+        case PortalSettings.UserDeleteAction.DelayedHardDelete:
+          confirmText = ClientAPI.GetSafeJSString(Localization.GetString("DelayedHardDelete.Confirm", "~/DesktopModules/Admin/Security/App_LocalResources/DataConsent.ascx.resx"));
+          break;
+        case PortalSettings.UserDeleteAction.HardDelete:
+          confirmText = ClientAPI.GetSafeJSString(Localization.GetString("HardDelete.Confirm", "~/DesktopModules/Admin/Security/App_LocalResources/DataConsent.ascx.resx"));
+          break;
+      }
+  }
+%>
+
 <script language="javascript" type="text/javascript">
     /*globals jQuery, window, Sys */
     (function ($, Sys) {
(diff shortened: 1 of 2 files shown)


BravoSierra:
GDPR, article 17 ;)
_The data subject shall have the right to obtain from the controller the erasure of personal data concerning him or her without undue delay and the controller shall have the obligation to erase personal data without undue delay_ 
Thanks Peter

Website/DesktopModules/Admin/Security/EditUser.ascx
```diff
@@ -10,14 +10,33 @@
 
 <%@ Import Namespace="DotNetNuke.UI.Utilities" %>
 
+<%
+  var confirmText = ClientAPI.GetSafeJSString(LocalizeString("UnregisterUser"));
+  if (PortalSettings.DataConsentActive && User.UserID == UserInfo.UserID)
```
daguiler: Hi @donker, can you please explain what the `User.UserID == UserInfo.UserID` condition is for? I see it down below as well and it seems that `PortalSettings.DataConsentActive` can be enabled but this other condition would defeat that setting and revert back to previous behavior

KenTombs:
Hi
just picked up on this discussion. I/we provide data protection/ISO27/compliance services etc. for our customers so I get to act as a Data Protection Officer for them.  If I can contribute to any of this then would be pleased to do so, as looking at using DNN for GDPR work. 

So for the example above. If a user asks to delete their account, that would take precedence over any consent - no argument and should be in a timely manner (i.e. not at the next system clean up five years later). 

However, if the users account includes any personal data needed for the businesses tax/accounting records, then you cannot just delete it until after the retention schedule for tax/accounting has been applied (7 years usually).  Those regulations take precedence locally over GDPR. So you'd need to build in a decision tree and you can see its it gets complicated.  

Realise jumped in half way here but hope that helps.

mitchelsellers:
@KenTombs Your input is helpful!  We do allow control over how the delete is handled by the administrator including a delayed hard-delete or simple notification.  This would allow those with more complex situations to be able to implement their own solution that would meet any retention requirements, etc.

KenTombs:
Its when you have 100,000 users and need to auto-administer, then schedules and processes become critically important.

Happy to contribute.

Website/DesktopModules/Admin/Security/EditUser.ascx
ohine: I'm guessing it's to only allow the User to request their account is removed. However, I'll wait to merge till @donker can give some more details.

ohine:
Looks good, just waiting to merge till the question from @daguiler is resolved.

BravoSierra:
@KenTombs then the admin just doesn't allow hard-deletes, it's a setting option. We've also discussed anonymizing or pseudonymizing rather than deleting but there is a challenge in multi-dnn instances so Peter's choice seems the best trade-off and offers better GDPR compliancy regarding at least 8 articles (5,6,7,12,13,15,16 and18). A good move forward.

KenTombs:
These are exactly the trade off issues we have to grapple with. The GDPR compartment suggests one thing and other regulations contradict it , no-win solutions. As in all regulatory regimes, 'Article Absolutism' can cause as much trouble as it solves, it is a simple way to do something just be aware that it may need changing later. Ideally I would suggest that where a development decision is reached such as seems to be the case here, the 'reasoning' behind it is explained maybe in a version specific help text for that item. This would help any auditor/court later to understand why something worked the way it does, and with regulators' fines going through the roof every little helps the customer. 

The reason I write all this is its exactly one of the problems we will have in adapting DNN to be a compliance platform. For example, the built in security analyser is amazing for legal purposes. Whats missing is a true workflow capability to let us focus on the good stuff customers want to see.  Lets see what the coming year brings.

Website/DesktopModules/Admin/Security/EditUser.ascx
donker: I wanted to be absolutely sure you cannot get here while editing someone else's profile. There is a lot of reuse of the profile editing controls and I was afraid it might not be the editing user who is deleting. In that case it shouldn't be marked as a request so the old logic is used.

daguiler: Cool, so using old logic in such case was the intended purpose.

PR #2746, opened 2019-04-29
Splits the Lists table population in 500 items batches

### LINKED ISSUE #2743: 9.3.2 RC0 timeout during clean installation
## Description of bug
During clean installation, sql scripts still timeout on less than basic servers.

## Steps to reproduce
Try to do a clean 9.3.2 RC0 install, notice timeout error

## Current result
I got a timeout in the DotNetNuke.Data.SqlDataProvider in the section where it tries to insert more than 5000 rows in the Lists table.

## Expected result
Installation should run fine

## Error log
Huge error log, but basically a timeout during creation of the data that goes into the Lists table for the countries and regions (more than 5000 records in one go).

## Additional context
This is on a not so low spec server, dual xeon physical server, 2 sockets x 8 cores, 16 GB RAM about half used. raid5 array of spinners, not near full at all. There was already a PR for adding more GO statetements on some sql scripts, maybe we also need more here, maybe run this part in two 2500 rows batches (cut it in half)...

## Affected version

* [x] 9.3.2
* [ ] 9.3.1
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
Not browser related

valadas:
Closes #2743 

## Summary
This PR splits the creation of rows in the Lists table into 500 item batches intead of a single bath with more than 5000 lines. This is to prevent a timeout issue during installation on some environments.


valadas:
I did not test the syntax change but went for the minimal change that fixed the issue. I am not sure on support for the other syntax on various SQL server versions so wanted to avoid introducing that change in a bug fix release that is already in RC stage. SQL server script generation looks more like the current syntax except it generates a go in each line and has hardcoded values instead of variables.

mitchelsellers:
Fair enough.  @ohine any thoughts

valadas:
Just let me know if you want me to test other solutions on that same machine since I have a good testbed to reproduce. Make scare with the other solution was that it may not work everywhere (azure SQL, old versions of SQL server, etc.) Which I don't have a good testbed for.

PR #2765, opened 2019-05-07
Rename "Azure" to "Azure Storage", and prefix settings with "Storage"

### LINKED ISSUE #2572: Azure connector needs some basic help text
## Description of problem
The azure connector needs some basic info in the UI to explain what it does and what keys to get where.

## Description of solution
The UI needs to be to updated to add some basic hints to set it up.
I do not use Azure so not sure about this, but I think the field names might not match exactly the names of the corresponding Azure fields to enter there???

## Additional context
This is partially to try and help with https://github.com/DNNCommunity/DNN.Blog/issues/130

## Affected version

* [x] 9.3.0
* [x] 9.2.2

ohine:
## Summary
Fixes #2572

### DIFF
--- a/DNN Platform/Connectors/Azure/App_LocalResources/SharedResources.resx
+++ b/DNN Platform/Connectors/Azure/App_LocalResources/SharedResources.resx
@@ -118,10 +118,10 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="AccountName.Text" xml:space="preserve">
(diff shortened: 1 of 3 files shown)


bdukes:
Does there need to be some migration since the `DefaultDisplayName` was changed, and that's used in the mapping names?

ohine:
@bdukes I didn't think so, I figured it was just used for display. I figured on upgrades it would just show the new value, but I haven't tested that yet.

ahoefling:
@ohine this is great! Thanks for submitting this much needed clarification on the Azure Storage Connector. Just about everyone I have talked to about the Azure Connector that didn't know better didn't realize it was just for Azure Storage

Issue #2761, opened 2019-05-06
Remove DotNetNuke.Library Circular Dependencies


ahoefling:
## Description of problem
The core library `DotNetNuke.Library` utilizes C# Reflection to circumvent a circular dependency problem in the platform. Since `DotNetNuke.Library` can't reference another library (such as `DotNetNuke.Web.Mvc`) that references itself `DotNetNuke.Library`. There is code scattered throughout the platform that uses C# Reflection which loads the assembly at runtime. While this works it creates side-effects in the platform that are prone to runtime crashes and performance slow-downs.

An example of how we are using reflection in the `ModuleControlFactory`
https://github.com/dnnsoftware/Dnn.Platform/blob/c1a352acb1ae7f52549d1e605822f56ae837ed45/DNN%20Platform/Library/UI/Modules/ModuleControlFactory.cs#L58-L64

## Description of solution
My plan/proposal is to extra shared contracts so the different Module Libraries don't need to reference the entire `DotNetNuke.Library` Project. A new project would be created called `DotNetNuke.Library.Contracts`. All shared interfaces and contracts required for the Module Libraries would be moved to the new `DotNetNuke.Library.Contracts` project. This would then be referenced by both `DotNetNuke.Library` and the Module LIbraries. This would then remove the circular dependency and allow us to re-write the code above to

```c#
case ".mvc":
    controlFactory = new MvcModuleControlFactory();
    break;
```

* New Project: `DotNetNuke.Library.Contracts`
* Module Library: References `DotNetNuke.Library.Contracts`
* `DotNetNuke.Library` References: `DotNetNuke.Library.Contracts`

## Reasoning for Change
Creating a contracts project is fundamental to creating the building blocks of a wrapper that **removes the dependency on System.Web**. After completing this feature we would be able to create a wrapper around `System.Web` which will be a big step towards a .NET Core migration.

## Description of alternatives considered
Leaving things the way they are makes a migration off of `System.Web` a very large undertaking. By making changes now that create abstractions around `System.Web` it will make the migration easier to manage in smaller chunks.

## Screenshots
N/A

## Additional context
N/A

I would like to include this change as part of the Razor Pages Feature. It is not required, but **highly** recommended.
Related: #2599

mitchelsellers:
[...]

PR #2774, opened 2019-05-10
Added Dependency Injection to all Module Patterns and Removed Circular Dependencies in Module Pipeline

ahoefling:
# UPDATED 5/13/2019 @ 09:54 EDT
Fixes: #2761 

First major improvement towards .NET Core

## Summary
Removed Reflection calls in the ModuleControlFactory (now the ModuleControlPipeline) by removing Circular Dependencies and adding Dependency Injection.

## Module Pipeline
The different module patterns were invoked in the `ModuleControlFactory` which has been moved to a new library called `DotNetNuke.ModulePipeline`. The root level `ModuleControlFactory` (Yes, there were more than 1) has been renamed to `ModuleControlPipeline` which contains an interface `IModuleControlPipeline` which makes it easy to register into the Dependency Injection Container.

Since all the independent module patterns have the exact same interface of `IModuleControlFactory` we were able to register each Module Pattern with the Dependency Injection container and then inject the different Factories into the `ModuleControlPipeline` where it determines which one to use.

Prior to this change the following libraries where instantiated by reflection with each module load which created a performance hit

* `DotNetNuke.Web.Mvc`
* `DotNetNuke.Web.Razor`

## Dependency Injection
Added the ASP.NET Core Dependency Injection libraries
* `Microsoft.Extensions.DependencyInjection`
* `Microsoft.Extensions.DependencyInjection.Abstractions`

New Dependency Injection Library specific for DNN
* `DotNetNuke.DependencyInjection`

New NUSPEC for 3rd-party libraries using dependency injection. The new library properly depends on the Microsoft NuGets so all the 3rd party developer has to do is include the following library.
* `DotNetNuke.DependencyInjection

Added IDnnStartupto the `DotNetNuke.Common.Globals` which provides easy access anywhere in DNN to use the Service Locator pattern to resolve dependencies. **It is always recommended to use constructor injection when possible, but the `ServiceProvider` provides a workaround if not possible**.

Added `IDnnStartup` interface that exists in the `DotNetNuke.DependencyInjection` library. Any file that implements this interface is invoked at App Start and registers all services in the implemented class. **(See Usage API for more details)**

The `IDnnStartup` is configured at the beginning of the Application Startup logic and is one of the first things that is invoked.

## Usage API
To use the new Dependency Injection Libraries you will add the `DotNetNuke.DependencyInjection` NuGet (which will be available after the next release).

**Example**
 Suppose you have a simple interface of `IMessageService` and an implementation of `MessageService`

**IMessageService**
```c#
public interface IMessageService
{
    string GetMessage();
}
```

Create a `Startup.cs` file like below. 
```c#
{
    {
        services.AddScoped<IMessageService, MessageService>();
    }
}
```

### MVC Modules
Create a `Startup` file like the above example and include it in the MVC Module assembly typically at the root level.

**HomeController**
```c#
public class HomeController : DnnController
{
    protected IMessageService MessageService { get; }
    public HomeController(IMessageService messageService)
    {
        MessageService = messageService;
    }

    public ActionResult Index()
    {
        var model = new HelloWorldModel
        {
            Message = MessageService.GetMessage()
        };
        return View(model);
    }
}
```

### SPA Modules using Web API
Create a `Startup` file like the above example and include it in the MVC Module assembly typically at the root level.

**HomeController**
```c#
public class HomeController : DnnApiController
{
    protected IMessageService MessageService { get; }
    public HomeController(IMessageService messageService)
    {
        MessageService = messageService;
    }

    public string Get()
    {
        return MessageService.GetMessage();
    }
}
```

### WebForms
Create a `Startup` file like the above example and include it in the Web Forms Module assembly typically at the root level.

Web forms doesn't support constructor injection but using the Service Locator pattern you can access the `IMessageService`

```c#
public partial class Index : PortalModuleBase
{
    protected IMessageService MessageService { get; }

    public Index()
    {
        MessageService = DotNetNuke.Common.Globals.DependencyProvider.GetService<IMessageService>();;
    }

    Message => MessageService.GetMessage();
}
```

### Razor3 Modules
Create a `Startup` file like the above example and include it in the Razor3 Module assembly typically at the root level. Make sure to register your model or constructor injection will not work. 

**Startup.cs**
```c#
{
    {
        services.AddScoped<IMessageService, MessageService>();
        services.AddScoped<IndexModel>();
    }
}
```

**IndexModel.cs**
```c#
public class IndexModel
{
    protected IMessageService MessageService { get; }
    public IndexModel(IMessageService messageService)
    {
        MessageService = messageService;
    }

    public string Message => MessageService.GetMessage();
}
```


ahoefling:
@ohine can you take a look at the build considerations with this PR. I added 2 new projects and want to make sure that the installer will include the new assemblies correctly. Here is a list of the new assemblies that need to be included in the bin directory
* `DotNetNuke.DependencyInjection`
* `DotNetNuke.ModulePipeline`
* `Microsoft.Extensions.DependencyInjection`
* `Microsoft.Extensions.DependencyINjection.Abstractions`

As well, I create a `nuspec` file for `DotNetNuke.DependencyInjection` and want to make sure that file gets published to the nuget feed correctly for 9.4

ohine:
For sure, I’ll take a look the build and also get a review going. However first, thank you for this amazing contribution!!

ahoefling:
Perfect! I'll try running the full build locally and verifying it with my different modules that have Dependency Injection built in

nvisionative:
True EPICness!!!  Thank you @ahoefling!

mitchelsellers:
As I mentioned offline, thank you for this Andrew. Amazing contribution. I’ll get a full review in shortly. But your collaboration and documentation on this is fantastic

Build/Tools/NuGet/DotNetNuke.DependencyInjection.nuspec
```diff
@@ -0,0 +1,25 @@
...
+    <license type="expression">MIT</license>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <iconUrl>http://www.dnnsoftware.com/favicon.ico</iconUrl>
+    <projectUrl>https://github.com/dnnsoftware/Dnn.Platform</projectUrl>
+    <description>
+      Provides API references needed to use Dependency Injection for the DNN Platform.
+    </description>
+    <copyright>DNN and DotNetNuke are copyright 2002-2019 by DNN Corp. All Rights Reserved.</copyright>
+    <dependencies>
+      <dependency id="Microsoft.Extensions.DependencyInjection" version="2.1.1" />
```
mitchelsellers: Do we need the abstractions package as well?

DNN Platform/DotNetNuke.DependencyInjection/IServiceRegistration.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+
+namespace DotNetNuke.DependencyInjection
+{
+    public interface IServiceRegistration
```
mitchelsellers: Can we get XML documentation for this so users will have intelligence detail.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Web.UI;
+#endif
+
+namespace DotNetNuke.ModulePipeline
+{
+
+    public class ModuleControlPipeline
+#if NET472
+        : IModuleControlPipeline
+#endif
```
mitchelsellers: For future support can we just add a comment here as to why this compiler directive was needed

DNN Platform/DotNetNuke.ModulePipeline/Startup.cs
```diff
@@ -0,0 +1,16 @@
+﻿using DotNetNuke.DependencyInjection;
+using DotNetNuke.UI.Modules;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace DotNetNuke.ModulePipeline
+{
+    public class Startup : IServiceRegistration
+    {
+        public void ConfigureServices(IServiceCollection services)
```
mitchelsellers: Similar to above. Note why needed

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcControllerFactory.cs
```diff
@@ -0,0 +1,16 @@
+﻿using DotNetNuke.Common;
+using System;
+using System.Web.Mvc;
+using System.Web.Routing;
+
+namespace DotNetNuke.Web.Mvc
+{
+    class DnnMvcControllerFactory : DefaultControllerFactory, IControllerFactory
```
mitchelsellers: This class is missing an access modifier

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using DotNetNuke.Web.Mvc.Framework.Controllers;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Linq;
+
+namespace DotNetNuke.Web.Mvc.Extensions
+{
+    internal static class StartupExtensions
+    {
+        public static void AddMvc(this IServiceCollection services)
```
mitchelsellers: If we can get some XML comments to document purpose here as well

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: I wasn't sure if we wanted to add comments in to explain the reasoning behind this or not. Considering the nature of OSS it is probably best to create the comment in code.

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcControllerFactory.cs
ahoefling: Good catch!

Build/Tools/NuGet/DotNetNuke.DependencyInjection.nuspec
ahoefling: I just confirmed that it is not necessary and I'll remove the `Microsoft.Extensions.DependencyInjection.Abstractions`

valadas:
This is awesome, but yes, I would love xml documentation.

DNN Platform/DotNetNuke.DependencyInjection/IServiceRegistration.cs
ahoefling: I am going to rename `IServiceRegistration` -> `IDnnStartup` to be more consistent with the `Microsoft.AspNetCore.Hosting.IStartup` API (which isn't included yet).  In AspNetCore you can create a `Startup` file by convention or use the `IStartup` interface. The `IDnnStartup` interface will have the convention method of `ConfigureServices` which will be invoked correctly as-is.

Using the `IStartup` interface really isn't an option right now and isn't a good idea as that interface returns an `IServicePrincipal` which won't work well with how many instances of `IDnnStartup` there are going to be when we really take advantage of Dependency Injection.

Here is the AspNetCore Startup logic for reference. 
```c#
public static IWebHostBuilder UseStartup(this IWebHostBuilder hostBuilder, Type startupType)
{
    var startupAssemblyName = startupType.GetTypeInfo().Assembly.GetName().Name;

    hostBuilder.UseSetting(WebHostDefaults.ApplicationKey, startupAssemblyName);

    // Light up the GenericWebHostBuilder implementation
    if (hostBuilder is ISupportsStartup supportsStartup)
    {
        return supportsStartup.UseStartup(startupType);
    }

    return hostBuilder
        .ConfigureServices(services =>
        {
            if (typeof(IStartup).GetTypeInfo().IsAssignableFrom(startupType.GetTypeInfo()))
            {
                services.AddSingleton(typeof(IStartup), startupType);
            }
            else
            {
                services.AddSingleton(typeof(IStartup), sp =>
                {
                    var hostingEnvironment = sp.GetRequiredService<IHostEnvironment>();
                    return new ConventionBasedStartup(StartupLoader.LoadMethods(sp, startupType, hostingEnvironment.EnvironmentName));
                });
            }
        });
}
```

The most common usage in AspNetCore is not using the `IStartup` interface, and I really think in an AspNetCore version of Dnn we should continue to use our own `IDnnStartup` to give us access to our customized startup code for both module pipelines and library code.

DNN Platform/DotNetNuke.ModulePipeline/Startup.cs
ahoefling: Added XML comments

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcControllerFactory.cs
ahoefling: I added XML comments to this class as wel

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
ahoefling: Added xml comments

DNN Platform/DotNetNuke.DependencyInjection/IServiceRegistration.cs
ahoefling: Added xml comments

ahoefling:
Added some new features that are worth noting.

## IDnnStartup
The `IServiceRegistration` has been renamed to `IDnnStartup` to follow AspNetCore standards. No other changes have been made for this file.

## Safely Load Assemblies
During the startup process every assembly in the bin directory is loaded into the App Domain and our Dependency Injection Startup routines attempt to retrieve All Types. This is used to find the implementations of `IDnnStartup` in each assembly and register services. 

If there is a bad 3rd party assembly that can't load we want it to safely be ignored, otherwise the entire website would crash. When the user tries to use their 3rd-party library they will see usage errors but it will be isolated to their module or similar area not the entire website.

## XML Comments
I went through all new code and added xml comments to properly document everything

ahoefling:
I created a samples repository which has sample modules using Dependency Injection. This sample code demonstrates the basics for registering services and using them. https://github.com/ahoefling/DNN.DependencyInjection.Samples

Currently I have samples for
* MVC
* Razor3
* WebForms

I plan to add a lot more documentation in this repo on using this new feature.

Repo Contains:
* DNN 9.4.0-alpha installer with DI Assemblies
* Private NuGet feed which includes 
  * DotNetNuke.DependencyInjection
  * DotNetNuke.Library
* Sample Modules
  * MVC
  * WebForms
  * Razor3
  * (SPA Coming Soon)

ahoefling:
## Update
Added Dependency Injection for Web Api Controllers in the `DnnApiController`. This is a controller that is commonly used in SPA Modules.

ahoefling:
@ohine can you look at the latest CI build. Looks like the failure is unrelated to my latest changes

ohine:
That's a weird failure for sure, it appears the devops build agent lost internet connectivity? Bizarre. Attempting to re-run that pipeline now...

CBPSC:
EPIC!

bdukes:
This is great!  For constructor injection in Web Forms, were you not able to turn that on for some reason?  Now that we're targetting .NET 4.7.2, you should just be able to set [`HttpRuntime.WebObjectActivator`](https://docs.microsoft.com/en-us/dotnet/api/system.web.httpruntime.webobjectactivator?view=netframework-4.7.2).  If there are issues with that, we don't need to hold up this PR, just wondering if that could be enabled, also.

ahoefling:
@bdukes thanks for bringing up the `HttpRuntime.WebObjectActivator` this is something I was trying to leverage especially for the WebForms modules. Considering the `HttpRuntime.WebObjectActivator` gives us access to the `IServicePincipal` this would be a great object to use for the service locator instead of our `DotNetNuke.Globals.DependencyProvider`. The `HttpRuntime.WebObjectActivator` also should provide dependency injection into WebForms pages, I'm not sure if it works for WebForms controls as many articles I ran across had issues with WebForms controls which is what our WebForms modules are built on top of.

When I tried using the `HttpRuntime.WebObjectActivator` I ran into **a lot** of problems with the website crashing. I had a theory that it was tied to the ClientDependency library but no real way to determine where the problem was because the Application Startup code wasn't running.

bdukes:
:+1: When I had played with it, I got it working somewhat, but only after lots of crashing, so I'm happy to leave that out for now.  Hopefully we can pinpoint the issue and get it working, but it's not a showstopper to get this in.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,194 @@
...
+        // This file multi-targets .NET Framework and .NET Standard,
+        // which is needed as DNN migrates to .NET Core. The 'NET472'
+        // pre-processor directives are to fully support Legacy DNN.
+        // As the Pipeline is upgraded to be more complaint with 
+        // .NET Standard 2.0 use the apprioprate pre-processor directives.
+#if NET472
+        : IModuleControlPipeline
+#endif
+    {
+        private static readonly ILog TracelLogger = LoggerSource.Instance.GetLogger("DNN.Trace");
```
bdukes: Is this a typo?  Should it be `TraceLogger` instead of `TracelLogger`?

bdukes: (I understand this was copied from other code, so the original (lack of) intention may be lost)

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,194 @@
...
+            _controlFactories.Add(".vbhtml", razor3);
+            _controlFactories.Add(".mvc", mvc);
+            _controlFactories.Add("default", fallthrough);
+        }
+
+#if NET472
+        private IModuleControlFactory GetModuleControlFactory(string controlSrc)
+        {
+            string extension = Path.GetExtension(controlSrc.ToLowerInvariant());
+            IModuleControlFactory factory = _controlFactories[extension];
```
bdukes: This should use `TryGetValue` instead of the indexer, since a `Dictionary<T,U>` will throw when a key isn't found via the indexer.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,194 @@
...
+            _controlFactories.Add(".cshtml", razor3);
+            _controlFactories.Add(".vbhtml", razor3);
+            _controlFactories.Add(".mvc", mvc);
+            _controlFactories.Add("default", fallthrough);
+        }
+
+#if NET472
+        private IModuleControlFactory GetModuleControlFactory(string controlSrc)
+        {
+            string extension = Path.GetExtension(controlSrc.ToLowerInvariant());
```
bdukes: Consider giving the dictionary a `StringComparer.OrdinalIgnoreCase` argument (on line 43), so that we don't need to worry about casing here

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,194 @@
...
+            }
+
+            // set the control ID to the resource file name ( ie. controlname.ascx = controlname )
+            // this is necessary for the Localization in PageBase
+            if (control != null)
+            {
+                var fileNameWithoutExtension = Path.GetFileNameWithoutExtension(controlSrc);
+                if (fileNameWithoutExtension != null)
+                {
+                    control.ID = fileNameWithoutExtension.Replace('.', '-');
```
bdukes: I'm noticing that we don't replace `.` with `-` in `LoadModuleControl` (line 110), anyone know it that's on purpose (or why we're replacing here)? @dnnsoftware/tag @dnnsoftware/tag-group @dnnsoftware/approvers

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        public Control CreateModuleControl(ModuleInfo moduleConfiguration)
+        {
+            string extension = Path.GetExtension(moduleConfiguration.ModuleControl.ControlSrc.ToLowerInvariant());
+            var moduleControl = new ModuleControlBase();
+            moduleControl.ModuleContext.Configuration = moduleConfiguration;
+
+            switch (extension)
+            {
+                case ".mvc":
```
bdukes: should `IModuleControlFactory` have a `GetLocalResourceFile` method?

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: I just copied and pasted this code it needs to be updated to `TraceLogger` thanks for pointing this out.

DNN Platform/DotNetNuke.Web.Razor/DotNetNukeWebPage.cs
```diff
@@ -59,6 +60,14 @@ public abstract class DotNetNukeWebPage<TModel> :DotNetNukeWebPage where TModel
     {
         private TModel _model;
 
+        public DotNetNukeWebPage()
+        {
+            var model = Globals.DependencyProvider.GetService<TModel>();
+            Model = model == null ?
+                Activator.CreateInstance<TModel>() :
+                model;
```
mtrutledge: This could be simplified to Model = model ?? Activator.CreateInstance<TModel>();

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: This code is the exact same code copied from the old `ModuleControlFactory`. I am hesitant to make any changes that modifies the original code that was unchanged in this pull request. *I am in support of working through follow up issues. *

If I am in the minority here, we can make changes as needed.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: We can create a `GetLocalResourceFile` to extract a method for this lookup, great idea!

DNN Platform/DotNetNuke.Web.Razor/DotNetNukeWebPage.cs
ahoefling: Yes, we should be able to simplify this. Thanks for the suggestion

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
bdukes: No, I completely agree that any issues raised with the copied code should be follow-up issues.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: I was thinking we want the exception to be thrown to display the module error to the user that something is wrong with their code. There is a bigger try-catch statement around this code so the website won't crash but just the module code and display a DNN Error

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: Good idea, I'll update this

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: Could you create an issue on GitHub for this so we don't lose sight of it? This is something that I am not as familiar with and want to make sure we get it documented correctly

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: I originally thought this was new code and not existing code. I think we should mark this as a follow-up item and create a new GitHub Issue. This is the exact same code that was in the original `ModuleControlFactory` that has been ported to this class.

DNN Platform/DotNetNuke.Web/Startup.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            var startupTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .Where(x => x != Assembly.GetAssembly(typeof(Startup)))
+                .SelectMany(x => x.SafeGetTypes())
+                .Where(x => typeof(IDnnStartup).IsAssignableFrom(x) &&
+                            x.IsClass &&
+                            !x.IsAbstract);
+
+            var startupInstances = startupTypes
+                .Select(x => (IDnnStartup)Activator.CreateInstance(x));
```
bdukes: We probably need to wrap this in a `try`/`catch`, in case there's a bug in someone's constructor code, we don't want to take the website down.

ahoefling:
@mtrutledge I have fixed your requested change, thanks for the feedback

@bdukes I have fixed some of your requested changes and have some follow up we need to finish in the review. Thanks for the feedback on these!

Website/release.config
```diff
@@ -124,7 +124,7 @@
     <trust level="Full" originUrl=".*" />
     -->
     <!-- set debugmode to false for running application -->
-    <compilation debug="false" strict="false" targetFramework="4.5">
+    <compilation debug="false" strict="false" targetFramework="4.7.2">
```
bdukes: Do we already have code to update the `web.config` on upgrades?

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
bdukes: Don't we want it to use the fallback if the extension isn't in the dictionary?  My reading of the code is that it's valid (though uncommon) for an entry to be missing.

DNN Platform/DotNetNuke.Web/Startup.cs
bdukes: @ahoefling this is my only remaining issue/question before I'm ready to approve (just wanted to make sure you didn't miss it since I added while you were reviewing previous comments)

ahoefling: In this scenario let's say someone creates a `Startup` file that implements `IDnnStartup` in their 3rd-party library. They decide to go against the best practices (which happens) and have some code in the constructor that causes an exception. Placing a try-catch around the activator code will prevent the website from crashing.

Yes this makes sense to me, I'll add it in. Great feedback!

Website/release.config
ahoefling: This was missed when I put in the 4.7.2 PR. Where does the upgrade code for the web.config exist? If it is not in there we need to make sure it is included.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: Follow-up item created #2780 marking this as resolved.

DNN Platform/DotNetNuke.Web/Startup.cs
bdukes: I added this comment on the `Select` call, but the `try` would need to be around the `foreach`, FYI.  Wanted to make sure I didn't throw you off track.

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: Follow-up item created #2780 marking this as resolved

bdukes: Resolved by opening #2780

DNN Platform/DotNetNuke.Web/Startup.cs
ahoefling: I updated the code and added a try-catch for both the foreach and the select

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
bdukes: This one was resolved by #2781

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
ahoefling: You are ABSOLUTELY correct, I completely missed this. I will go ahead and fix it. Thanks for pointing this out.

Website/release.config
bdukes: @ohine would this be `Upgrade.cs` code that calls an XML Merge template?

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Reflection;
+
+namespace DotNetNuke.DependencyInjection.Extensions
+{
+    /// <summary>
+    /// <see cref="Type"/> specific extneions to be used
```
mtrutledge: Minor typo in the comment. "extnesions" should be "extensions"

ahoefling: Good find, thanks for reporting the typo.

Issue #2788, opened 2019-05-14
Add .NET Framework 4.7.2 to DNN Upgrade Script


ahoefling:
## Description of problem
In DNN 9.4 we are upgrading the .NET Framework from 4.5 -> 4.7.2. The changes have been applied correctly to the web.config for new installations but not upgrades.

## Description of solution
There needs to be new code added to update the web.config for sites that are being upgraded.

## Description of alternatives considered
N/A

## Screenshots
N/A

## Additional context
N/A

## Affected version

* [X] 9.4

## Affected browser
N/A

PR #2789, opened 2019-05-14
Added DNN Upgrade script for 9.4: .NET 4.7.2

ahoefling:
fixes: #2788 

## Summary
Adds upgrade script to update the targetFramework elements in the web.config to be .NET 4.7.2. This is needed to properly upgrade websites to 9.4 as .NET 4.7.2 is introduced in the new version

### DIFF
--- a/Website/Install/Config/09.04.00.config
+++ b/Website/Install/Config/09.04.00.config
@@ -1,0 +1,6 @@
+<configuration>
+  <nodes configfile="web.config">
+    <node path="/configuration/system.web/compilation" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
+    <node path="/configuration/system.web/httpRuntime" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
+  </nodes>
+</configuration>
--- a/Website/Install/Config/09.04.00.config
+++ b/Website/Install/Config/09.04.00.config
@@ -1,6 +1,6 @@
 <configuration>
   <nodes configfile="web.config">
-    <node path="/configuration/system.web/compilation" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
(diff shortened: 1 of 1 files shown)


Website/Install/Config/09.04.00.config
```diff
@@ -0,0 +1,6 @@
+<configuration>
+  <nodes configfile="web.config">
+    <node path="/configuration/system.web/compilation" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
+    <node path="/configuration/system.web/httpRuntime" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
```
bdukes:
```suggestion
    <node path="/configuration/system.web/httpRuntime" action="updateattribute" name="targetFramework" value="4.7.2" />
```

Website/Install/Config/09.04.00.config
```diff
@@ -0,0 +1,6 @@
+<configuration>
+  <nodes configfile="web.config">
+    <node path="/configuration/system.web/compilation" action="update" key="targetFramework" value="4.7.2" collision="ignore" />
```
bdukes:
```suggestion
    <node path="/configuration/system.web/compilation" action="updateattribute" name="targetFramework" value="4.7.2" />
```

donker:
Should the web.config also be updated in the same PR? Or does this get run on a new install as well?

Website/Install/Config/09.04.00.config
ahoefling: Thanks for the advice, I wasn't really sure and grabbed the wrong example to base my change off of. Update on it's way

Website/Install/Config/09.04.00.config
ahoefling: See comment above, thanks for the feedback and update is on it's way

bdukes:
@donker the `web.config` change was made in a previous PR; we recently noticed that the upgrade logic was missing.

PR #2797, opened 2019-05-16
Addition of [Obsolete] Attribute to Users Online

mitchelsellers:
Back in DNN 8.x, or earlier, the UsersOnline HttpModule was removed from the default installations of DNN.  The configuration settings and actual source code for this feature were left inside the platform, however, they have not been tested or supported.

As such, the methods supporting this are legacy and no-longer properly managed.  To follow the published deprecation standards, since no prior notice was given.  I've flagged all implementation methods for the UserOnline feature for removal in v11.0.0.  

The DNN Community initiative, which has the supplemental UsersOnline module, could easily extract this code and install as an extension, which would be the preferred long-term solution as it.

One additional change was included, to update the configuration file on a UnitTest project to match the proper current configuration file.

### DIFF
--- a/DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
+++ b/DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
@@ -29,7 +29,6 @@ namespace DotNetNuke.Entities.Users
     /// -----------------------------------------------------------------------------
     /// Project:    DotNetNuke
     /// Namespace:  DotNetNuke.Entities.Users
-    /// Class:      PurgeUsersOnline
     /// -----------------------------------------------------------------------------
     /// <summary>
     /// The AnonymousUserInfo class provides an Entity for an anonymous user
--- a/DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
+++ b/DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
@@ -37,6 +37,7 @@ namespace DotNetNuke.Entities.Users
     /// </remarks>
     /// -----------------------------------------------------------------------------
     [Serializable]
+    [Obsolete("Support for users online was removed in 8.x, other solutions exist outside of the DNN Platform.  Scheduled removal in v11.0.0.")]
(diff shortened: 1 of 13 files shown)


sleupold:
has the http module been included in UsersOnline module?
We still have a number of clients using it and want to be able to get the online status, e.g. when fighting against spam users.

DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
```diff
@@ -29,7 +29,6 @@ namespace DotNetNuke.Entities.Users
     /// -----------------------------------------------------------------------------
     /// Project:    DotNetNuke
     /// Namespace:  DotNetNuke.Entities.Users
-    /// Class:      PurgeUsersOnline
     /// -----------------------------------------------------------------------------
     /// <summary>
     /// The AnonymousUserInfo class provides an Entity for an anonymous user
```
bdukes: This is in the `Users Online` directory, should it be marked as `Obsolete` as well?  It's only used from `DataProvider.UpdateUsersOnline`:
https://github.com/dnnsoftware/Dnn.Platform/blob/a1b217e038127f673fbfa82624e4d91f81993aab/DNN%20Platform/Library/Data/DataProvider.cs#L2498-L2519

valadas:
@sleupold Actually after looking at it, it appears the users online module was never even moved to github... If there is interest in updating the module, it is still on codeplex here https://archive.codeplex.com/?p=dnnusersonline I am willing to move it to github if we have interested contributors... Just let me know...

DNN Platform/Library/Entities/Users/Users Online/AnonymousUserInfo.cs
donker: To add to this: this is in a folder called "Users Online" containing code that is just meant for this part. All classes should be marked obsolete IMHO. It can easily be implemented as a 3rd party sln external to the core. Since we're axing it, remove all related parts IMHO.

mitchelsellers: I corrected the missing items

Issue #2762, opened 2019-05-07
Make Default Module Action Menu Configurable


mikesmeltzer:
Follow-up to RFC #2700 

## Description of problem

Currently, the path to the default Module Action menu control (/admin/Menus/ModuleActions/ModuleActions.ascx) is hardcoded within the C# in the Container base class and does not allow a developer to configure the site to use a different action menu control by default on a per site basis.

The default action menu can be overridden on a per module basis by using a container that uses it's own action menu control. However, it would be ideal if we could define the action menu at a portal level and override the default action menu regardless of the container being used across the site, unless the container supplied it's own action menu as you can today.

## Description of solution
Add a new Portal Setting called DefaultActionMenu, similar to how the default skin path, container path, etc. are stored.

Have the setting defaulted to the existing action menu path and allow users to swap this behaviour out by changing the setting value to point to an action menu control of their choosing.

This change would include the SQL to include the new portal setting in the DB, change the container base class to read from the portal setting instead of the hardcoded path, and changes to the Portal Settings and Portal Settings Controller to add the new setting into the API.

## Affected version

This is net new functionality targeted for DNN v9.4.

## Affected browser

This is a server side feature and is therefore browser agnostic.

ohine:
Closing issue as PR was merged.

PR #2804, opened 2019-05-19
Make Default Module Action Menu Configurable

mikesmeltzer:
Fixes issue #2762 

## Summary

Added a new property to the Host Setting entity that will allow solution developers to change what control get's loaded for the portal's default Module Action Menu.

Adjusted the the Container.cs class to read from the new HostSetting instead of a hard coded value.

The new behaviour will default to the core module action menu if a HostSetting entry for the new property is not found in the database,

I added a git ignore for SolutionInfo.cs as local build scripts were modifying that file to update the versioning.

Note:  I've tested that the SQL script works when running manually but it is not being ran (or appears that way) when I modify a nightly 9.4 build to include this script.  If someone can put another set of eyes on the script and see if they can spot the issue, if not I'd like to see if it works properly when we do a RC build (the script not running will not break the site/code but it would be nice to add the value into the DB).


sleupold:
Wouldn't it be easier, to provide the default value in the API instead of the database table? This way, no record would be needed, unless using a custom action menu.

mikesmeltzer:
@sleupold That is exactly how I have it currently. The default value is supplied in the API and if it exists in the database then it will use that one instead. You can find the defaulting in the PortalSettingsController.cs file change.

For including it in the DB during install, my thought was from a discover-ability perspective that it would be easier for developers to get started using this if they could see the setting in the DB instead of it being hidden away.

Since it works without the script, if the script isn't working properly then it's not a blocker to move ahead without it.

.gitignore
```diff
@@ -59,6 +59,7 @@ _ReSharper*
 
 # Others
 
+SolutionInfo.cs
```
bdukes: If `SolutionInfo.cs` is already a tracked file, adding it to the `.gitignore` won't stop it from being tracked.  The `.gitignore` only affects new files.  Is this ignoring a copy of `SolutionInfo.cs` that isn't committed in the repo?

Website/Providers/DataProviders/SqlDataProvider/09.04.00.SqlDataProvider
```diff
@@ -277,6 +277,30 @@ AS
 
 GO
 
+/* Issue #2762  - Make Default Module Action Menu Configurable */
+/***************************************************************/
+
+BEGIN
+
+	INSERT INTO {databaseOwner}[{objectQualifier}PortalSettings] (PortalID, SettingName, SettingValue, CreatedByUserID, CreatedOnDate, LastModifiedByUserID, LastModifiedOnDate)
```
bdukes: As @sleupold stated, I'd prefer we didn't have the SQL script and just allowed the default in code to be used in the majority of cases.

bdukes:
Thanks @mikesmeltzer!

Is the alternative action menu something open source that you can point to, or is it custom/proprietary?

.gitignore
mikesmeltzer: The solutioninfo.cs change was recommended by Oliver that I add it because one of the local build scripts was modifying the the shared file solutioninfo.cs file (to update the version) and then was getting picked as a change.  After adding this to the .gitignore file my Visual Studio is now no longer showing changes to the file, across the projects, and has the icon to be ignored.  Is there another way I should be doing this? Should this be removed?

bdukes: Unless I'm misunderstanding the scenario, adding the file to the `.gitignore` did not cause changes to the file to be ignored (try manually making a change to the file and see if it shows as a change).

Website/Providers/DataProviders/SqlDataProvider/09.04.00.SqlDataProvider
mikesmeltzer: Sure, I'll remove it and have some troubleshooting time on why it didn't fully work :)

mikesmeltzer:
@bdukes There aren't any alternatives at the moment. I am intending to do a blog post for the 9.4 release informing users as to what the feature is and how one would get started building their own menus.

.gitignore
mikesmeltzer: I just made a change on my end and it seems to be  being ignored. See attached pic.

image

mikesmeltzer: If there is confusion I can just remove this change from this PR and we can discuss the scenario/issue at tomorrows meeting and see if there is a better way forward. @ohine has a better idea of the issue and can probably explain better than I. I'm not sure why others haven't ran into the issue on their local env. - it had to do with the version on the DLL's being incorrect and not loading.

bdukes: Weird, I still see a change if I edit that file, in VS 2017 and on the command line.  Thanks for removing, we can look at what might need to change around that separately.  
SolutionInfo.cs listed under Changes in Visual Studio 2017's Team Explorer

ohine:
Thanks for the contribution @mikesmeltzer! Sorry for the confusion around the SolutionInfo file too.

ohine:
[Successful build](https://dev.azure.com/dotnet/DNN/_build/results?buildId=17434) queued manually.

PR #2817, opened 2019-05-25
Update PetaPoco to Nuget Reference and fix TableName attribute bug

ohine:
## Summary
In PR #2764, we upgraded PetaPoco which caused our custom mapper, `PetaPocoMapper`, to never register at runtime. This caused all of our custom attributes to stop functioning. 

```
namespace My.CustomModule.Data.Models
{
    [TableName("CustomModule_CustomEntries")]
    [PrimaryKey("Id", AutoIncrement = true)]
    [Scope("ModuleId")]
    public class CustomEntry : Entity
    {
        public string Title { get; set; }
        public DateTime Date { get; set; }
        public string Details { get; set; }
    }
}
```

Which would throw the following exception. 

```
DotNetNuke.Services.Exceptions.ModuleLoadException: Invalid object name CustomEntry
```

This PR fixes that, and it also swaps out the direct local reference to `PetaPoco` to the Nuget package `PetaPoco.Compiled` and updates us again from v6.0.404 to v6.0.415. I also checked the license and it hasn't changed so we do not need any updates to the licenses folder. 

TL;DR: GetMapper in v5.0.1 didn't have a default param and would return a `StandardMapper` instead of our custom `PetaPocoMapper` when the mapper wasn't registered yet. Now we are setting a new variable to be a `StandardMapper` and passing it into the `GetMapper` call.


DNN Platform/Library/Data/PetaPoco/PetaPocoMapper.cs
```diff
@@ -114,17 +116,15 @@ public static void SetMapper<T>(IMapper mapper)
             _lock.EnterWriteLock();
             try
             {
-                if (Mappers.GetMapper(typeof (T), mapper) is StandardMapper)
+                if (Mappers.GetMapper(typeof (T), _defaultMapper) is StandardMapper)
```
daguiler: I think this will always be true

DNN Platform/Tests/DotNetNuke.Tests.Data/DotNetNuke.Tests.Data.csproj
```diff
@@ -74,7 +74,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="PetaPoco">
-      <HintPath>..\..\Components\PetaPoco\bin\PetaPoco.dll</HintPath>
+      <HintPath>..\..\Library\bin\PetaPoco.dll</HintPath>
```
ahoefling: Instead of referencing the `Library\bin\PetaPoco.dll` could we update the test project to use the nuget file. This will limit a dependency on the compiled dll and force NuGet to resolve the dependency which I believe is the best practice we should be following

DNN Platform/Tests/DotNetNuke.Tests.Integration/DotNetNuke.Tests.Integration.csproj
```diff
@@ -91,7 +91,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="PetaPoco">
-      <HintPath>..\..\Components\PetaPoco\bin\PetaPoco.dll</HintPath>
+      <HintPath>..\..\Library\bin\PetaPoco.dll</HintPath>
```
ahoefling: (Same as above)

Instead of referencing the Library\bin\PetaPoco.dll could we update the test project to use the nuget file. This will limit a dependency on the compiled dll and force NuGet to resolve the dependency which I believe is the best practice we should be following

DNN Platform/Library/Data/PetaPoco/PetaPocoMapper.cs
ahoefling: I'm not sure if I understand this logic 100% percent, could you explain how this statement would ever by false. 

If I am understanding the code correctly it tries to get the mapper on the entity of `T` and if there is no mapper on entity of `T` it uses the default mapper variable `_defaultMapper`. In our case it is just the `StandardMapper`.

Can you confirm?

---
Some general thoughts I had about this change which I don't think we can easily implement but would like to get us thinking about. At first look of this code my first instinct was how can we update it to use Dependency Injection instead of instantiating the IMapper at line 45 and re-using it. Since this method is a static method I understand why we are instantiating it, but I would love to start thinking about using Dependency Injection more and how we can update this code to use it. **This last bit is just some thoughts and we don't need to add Dependency Injection**
--

DNN Platform/Tests/DotNetNuke.Tests.Data/DotNetNuke.Tests.Data.csproj
ohine: I had the same thought, however the reasoning behind my initial decision was to try and keep the dependabot pull requests down to a minimum. Since they only look at a single folder/configuration file and don't do a recursive search. We'd end up with 3 pull requests to update the reference in the Dnn.Platform project. 

I'm happy to make this change, if that's something everyone agrees on. However, I think we'd be OK with the single nuget reference and then just using a standard file based reference on the support projects.

Another thought (untested and not researched), we might be able to do something similar with the yarn-workspace monorepo configuration so we can decrease the PR workload when these references are updated.

DNN Platform/Library/Data/PetaPoco/PetaPocoMapper.cs
ohine: It's a bit of bizarre code for sure. It took me a little bit to fully wrap my head around it. Initially in #2764 I was incorrect in my assumptions. I think now it's at least working as it was before that PR. 

Nothing should be using the StandardMapper, every `T` I threw at the function while attached to the debugger returned our custom `PetaPocoMapper`. It's a very bad way (IMO) of checking if our mapper is registered yet or not for the requested type. 

Basically, On the first call for each type, `T`, our custom mapper isn't registered yet. The call to GetMapper wouldn't find that type, and it would then return the _defaultMapper. In #2764 I was using our custome mapper type, so the check for `is StandardMapper` was always false and our custom mapper was never registered as the result. 

Chatting offline with @daguiler earlier about this, as he had basically the same questions. The idea I came up with was to introduce a new class called `NullMapper` or `NotYetRegisteredMapper` (which would replace the `is StandardMapper` check), and this would make things a bit easier to understand. It would also allow the StandardMapper to be used, which I'm not even sure is possible in our implementation, but as it is right now (and in all previous version) I think the code would error if T actually used the StandardMapper. 

As for Dependency Injection. For sure! I'd love to  work together on improving the platform with DI. Lets plan to get some of that into v9.4.1 or v10 (depending on any breaking changes, etc). At least v6 of PetaPoco is netstandard so that should help us quite a bit on the journey to .netcore support.

ahoefling:
@ohine thanks for the detailed explanations here, I have no issues with holding up this PR at all and the information you provided is very useful.

DNN Platform/Tests/DotNetNuke.Tests.Data/DotNetNuke.Tests.Data.csproj
bdukes: I'm going to go ahead and merge, but I'd like to see us using NuGet for references where possible.  We'll need to investigate if there's a better way to manage that with Dependabot

PR #2822, opened 2019-05-28
Fixes PetaPoco lacking mapper

donker:
Without a mapper PetaPoco will default to using the object name as table name. DNN's petapoco mapper should be used when creating the DB to set the default mapper.

This fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/971

### DIFF
--- a/DNN Platform/Library/Data/PetaPoco/PetaPocoDataContext.cs
+++ b/DNN Platform/Library/Data/PetaPoco/PetaPocoDataContext.cs
@@ -61,8 +61,8 @@ namespace DotNetNuke.Data.PetaPoco
         {
             Requires.NotNullOrEmpty("connectionStringName", connectionStringName);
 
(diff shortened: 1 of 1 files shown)


bdukes:
It looks like this is a duplicate of #2817, but with a different (simpler?) resolution.  @ohine @daguiler @mitchelsellers do y'all have any thoughts on preference of this solution over the other, since you've looked into the issue more?  I think we want to keep the move to the NuGet package from #2817, but if this allows us to avoid the confusing default mapper check, this option may be preferred.

ohine:
I think we should merge #2817 and then work on cleaning up this bit of code for v9.4.1 or v10.0. The fix in #2817 is closer to the code we've been using since adding PetaPoco to the platform.

valadas:
Closing this as it duplicated #2817

PR #2833, opened 2019-06-04
Add GitHub security policy

bdukes:
This is just a copy of the content from https://www.dnnsoftware.com/community/security/security-center.  We may want to look at adjusting the security process to make use of GitHub's new functionality.

The template for the security policy also has a section about supported versions, but I wanted to start with what's there, and then we can add.  Here's a rough sketch of some content we might want to consider:

>
>## Supported Versions
>
>Only the latest release of DNN Platform receives security patches.  We do make an effort to provide workarounds or backports of patches when possible.
>The last release of DNN Platform 10.x is intended to be marked as a Long Term Support (LTS) version, which will receive official patch support while ongoing work continues on later versions.
>
>| Version           | Supported          |
>| ----------------- | ------------------ |
>| LTS (10.x)        | :white_check_mark: |
>| Current (9.4.x)   | :white_check_mark: |
>| 9.3.x             | :x:                |
>| 8.x               | :x:                |
>
>

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+We make every effort to ensure speedy analysis of reported issues and, where required, provide workarounds and updated application releases to fix them. If you see suspected issues/security scan results please report them by sending an email to:
+
+[security@dnnsoftware.com](mailto:security@dnnsoftware.com)
+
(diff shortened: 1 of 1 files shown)


valadas:
Just a note, when creating an issue, the list of issue templates now has 2 templates for reporting a security issue.

PR #2845, opened 2019-06-13
Wrong text in welcome email when users are added manually

mikebigun:
fixes: https://github.com/dnnsoftware/Dnn.AdminExperience/issues/367

Summary:
Recently we moved email sending functionality from UI to the Platform.
In this PR I'm checking whether user is a host or not to have a message with a right text to be sent to user while registration. In short, user gets approved automatically when performed by a host manually in a private mode, so we just want to send welcome email without any approval notifications.

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
+++ b/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
@@ -1,4 +1,5 @@
-﻿using DotNetNuke.Common;
+﻿using System;
+using DotNetNuke.Common;
 using DotNetNuke.Common.Utilities;
 using DotNetNuke.Entities.Portals;
 using DotNetNuke.Services.Mail;
@@ -8,6 +9,8 @@ namespace DotNetNuke.Entities.Users
 {
     public class UserRegistrationEmailNotifier
     {
+        private static Lazy<UserInfo> CurrentUser => new Lazy<UserInfo>(() => UserController.Instance.GetCurrentUserInfo());
+
         public UserRegistrationEmailNotifier()
         {
         }
@@ -15,8 +18,8 @@ namespace DotNetNuke.Entities.Users
         public static void NotifyAdministrator(UserInfo user)
         {
             // avoid self-notification (i.e. on site installation/super user creation)
-            var currentUser = UserController.Instance.GetCurrentUserInfo();
-            if (currentUser != null && (currentUser.UserID == Null.NullInteger || currentUser.UserID == user.UserID))
+            if (CurrentUser.Value != null && 
+                (CurrentUser.Value.UserID == Null.NullInteger || CurrentUser.Value.UserID == user.UserID))
             {
                 return;
             }
@@ -35,7 +38,9 @@ namespace DotNetNuke.Entities.Users
             switch (PortalSettings.Current.UserRegistration)
             {
                 case (int)PortalRegistrationType.PrivateRegistration:
-                    NotifyUser(user, MessageType.UserRegistrationPrivate);
+                    NotifyUser(user, CurrentUser.Value != null && CurrentUser.Value.IsSuperUser ?
+                        MessageType.UserRegistrationPrivateNoApprovalRequired :
+                        MessageType.UserRegistrationPrivate);
                     break;
                 case (int)PortalRegistrationType.PublicRegistration:
                     NotifyUser(user, MessageType.UserRegistrationPublic);
--- a/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
+++ b/DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
@@ -38,7 +38,7 @@ namespace DotNetNuke.Entities.Users
             switch (PortalSettings.Current.UserRegistration)
             {
                 case (int)PortalRegistrationType.PrivateRegistration:
-                    NotifyUser(user, CurrentUser != null ?
+                    NotifyUser(user, CurrentUser != null && CurrentUser.UserID != Null.NullInteger ?
                         MessageType.UserRegistrationPrivateNoApprovalRequired :
                         MessageType.UserRegistrationPrivate);
                     break;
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
```diff
@@ -8,15 +9,17 @@ namespace DotNetNuke.Entities.Users
 {
     public class UserRegistrationEmailNotifier
     {
+        private static Lazy<UserInfo> CurrentUser => new Lazy<UserInfo>(() => UserController.Instance.GetCurrentUserInfo());
```
bdukes: Since this is a `static Lazy<T>`, it will only resolve once for the entire life of the application.  `UserController.Instance.GetCurrentUserInfo` is properly optimized enough that I don't think there's any problem adding another call to it in `NotifyAdministrator` if necessary.

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
```diff
@@ -35,7 +38,9 @@ public static void NotifyUser(UserInfo user)
             switch (PortalSettings.Current.UserRegistration)
             {
                 case (int)PortalRegistrationType.PrivateRegistration:
-                    NotifyUser(user, MessageType.UserRegistrationPrivate);
+                    NotifyUser(user, CurrentUser.Value != null && CurrentUser.Value.IsSuperUser ?
```
bdukes: I don't think `IsSuperUser` is the right check here.  This should apply for _any_ user created by an admin (or, really, anyone with permission to create users), right?  In the case of private registration, isn't that everyone?  Our goal is to avoid sending this message:
>Your account details will be reviewed by the website Administrator and you will receive a notification upon account activation.

Is that message _ever_ true when using private registration?  If registration is private, users cannot register themselves, and so there is no review.

If there needs to be any check here, maybe it should be a check for whether the user is approved or in the unverified users role.

sleupold: IMO the superuser check is not sufficient, users might be added by a site admin OR superuser

bdukes: @daguiler since you originally reported the issue, am I missing anything? ☝️ 

It seems like we just need to change that message, rather than check anything.

mikebigun: @bdukes , I'm afraid you are wrong. 

1. Private registration means we require administrators to approve user accounts. So even in a Private mode anyone is able to register himself. 

image

2. Another thing is that we notify users on a post user creation even `OnUserCreated`. That means user is already created and db entry exists at the moment we send email. In such case, do we really need to check admin roles and other permission staff? Look like it is senseless. I even think `IsSuperUser` check is not needed here. If someone is logged, that means registration happened by someone with user creation right. If nobody logged, registration comes from outside, so we require approval. Am I right? It's enough to check if `CurrentUser` is not null.

https://github.com/dnnsoftware/Dnn.Platform/blob/1a867de6f65f1d5472eed5ca25dfef5cd504a77a/DNN%20Platform/Library/Entities/Users/UserEventHandlers.cs#L37-L45

cc: @sleupold

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
mikebigun: Fixed. Agree with you.

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
mikebigun: Thanks @sleupold.  I commented above.

Flow is:
Check permissions -> Create User -> If User Created successfully - > Notify User by Email
Notification is a last step in the process, means we already checked all permissions.
Here we only need to define if it was self-registration or not.

bdukes: Ah, you're right, I was thinking of `NoRegistration`.

I think your check for whether someone is logged in is simple enough, but you need to check whether `UserID != Null.NullInteger` because `GetCurrentUserInfo` returns an empty `UserInfo` instead of `null` if they aren't logged in.
https://github.com/dnnsoftware/Dnn.Platform/blob/1a867de6f65f1d5472eed5ca25dfef5cd504a77a/DNN%20Platform/Library/Entities/Users/UserController.cs#L209-L225

mikebigun: You are right. Just tested it. Works as expected now!

ohine:
If this is going to go into v9.4, it needs to be rebased onto that branch. Otherwise it will be merged into v10.

daguiler:
Hi @mikebigun 
I'm not getting the welcome email now for admin created users.
Are you getting the welcome email?

DNN Platform/Library/Entities/Users/UserRegistrationEmailNotifier.cs
sleupold: AFAIR there was a specific treatment of superusers before, to prevent passwords being sent to them. This is not necessary anymore, as we do not send passwords and the check might be removed.

daguiler:
Here's what I tried:
1) website's user registration: Private
2) logged out, navigate to /Register and register User One (using my email account)
3) I receive User One's welcome email
4) log in as host, open PB > Manage > Users and add User Two (authorized, my email account again)
5) no welcome email

daguiler:
Well, I just tested the same on 9.3.2 and I didn't get the email either, so it seems it's not a change in this PR.

daguiler:
Well, you can disregard my silly comments above, cause I just overlooked the rather obvious "Send An Email To New User" checkbox in the Users panel !

daguiler:
Hi @mikebigun , can you rebase to target release/9.4.x please? I'd do it directly from here but every time I tried it ended up adding unwanted stuff

daguiler:
Closing in favor of #2847

Issue #2800, opened 2019-05-16
Removal of methods marked for removal in v10.0.0


mitchelsellers:
This is a place holder for the removal of all methods marked for removal in version 10.0.0.

I will be submitting a Pull Request shortly after the "Development" branch is updated to build 10.0 releases that remove all of these items.

I will then coordinate with @ohine to publish an alpha build & blog post thereafter.

sleupold:
this is still valid

PR #2869, opened 2019-06-23
Removal of Obsolete Code

mitchelsellers:
## Summary
This the first commit for the resolution of #2800, this addresses solely the removal of methods that were flagged as obsolete for removal in 10.x.  Initial smoke testing has been completed against the core, however, due to the dependencies between this and the PB additional testing and future PR's might be necessary.

## Items for later consideration
There are a few other items to consider as well that might be follow-on work.

### Removal of Stored Procedures
Calls to the following Stored Procedures, were removed
DeleteContentWorkflowState
UpdateContentWorkflowState
GetContentWorkflowState
GetContentWorkflowStates
AddContentWorkflowLog
GetContentWorkflowLogs
DeleteContentWorkflowLogs
GetContentWorkflowSource
AddContentWorkflowSource

However, it appears that other areas are STILL using them, therefore I do not believe we SHOULD remove these but a more detailed review is necessary.

### jQuery Objects
The obsolete items were removed from the jQuery.cs file, however, further investigation might result in a complete removal of the class.


valadas:
@ohine Is the build failing because of changes in CI regarding tests or is there more to investigate here?

sleupold:
is this conflicting with effort from @mikesmeltzer ?

mikesmeltzer:
Thanks @sleupold.  Is this an effort to remove the Workflow API itself or just stale stored procedures within the API?

mikesmeltzer:
Actually, I just found the code with the obsolete folder at Dnn.Platform/DNN Platform/Library/Entities/Content/Workflow/.  I'm assuming the rest is staying intact.  No impact here that I can see but thanks @sleupold for the heads up.

mitchelsellers:
Correct the effort here is obsolete removal only.

sleupold:
Thanks for the reply - I jus waned to make sure, we are all in line.

mitchelsellers:
\azp run

mitchelsellers:
Re-running validation as the test failure on this seems to be a false positive

mitchelsellers:
@valadas Thought son this now?  Looks like the build for AdminExperience is failing, but not sure why, not due to this, as its a yarn issue?

valadas:
@mitchelsellers It appears to be some change on the development branch of AdminExperience unrelated to this, I am taking a better look at this...

valadas:
@mitchelsellers Oh, I think I know why, my build fixes (case sensitivity with c:\temp) are merged into release/9.4.x but not into development, we should update development with fixes from 9.4.x to fix the build process for that branch too.

valadas:
But Brian did it in https://github.com/dnnsoftware/Dnn.Platform/pull/2928 I am confused...

valadas:
Ok, so it was because the last pipeline build was done before the branch was updated with the build fixes, it now builds fine. Unit tests do not run though but there was also a fix for that, that we need to also merge into development

Issue #2819, opened 2019-05-26
hreflang tags


Tychodewaard:
## Description of problem
If you have multiple languages in a website, Google expects to have hreflang tags applied in the website.

## Description of solution
Every page that is multilingual should have a correct hreflang be present in the head as described in (https://support.google.com/webmasters/answer/189077?hl=en). 

## Description of alternatives considered
There is no alternative. 

## Affected version
all

## Affected browser
N/A

roman-yagodin:
Must have.

valadas:
Hey Stalebot, please do not close :)

jac69:
Hi

Why have you closed this issue? I believe we still have this SEO problem.

Best
Jose

mitchelsellers:
Due to branching changes, we will replay this against the release 10 branch once it is created, No action needed by submitter at this time.

Navilois:
Still relevant - very important for multilang sites.


### Temporary solution
Inject hreflang into the html header until this issue is resolved:
`<script runat="server">

    protected override void OnLoad(EventArgs e)
    {
        base.OnLoad(e);
        // Temporary solution: provides html hreflang until sitemap fixed in DNN 10 (?) -> https://github.com/dnnsoftware/Dnn.Platform/issues/2819
        var smProv = new DotNetNuke.Services.Sitemap.CoreSitemapProvider();
        var smUrls = new List<DotNetNuke.Services.Sitemap.SitemapUrl>();
        try
        {
            smUrls = smProv.GetUrls(PortalSettings.PortalId, PortalSettings, "0.9");
            foreach (DotNetNuke.Services.Sitemap.SitemapUrl url in smUrls)
            {
                if (url.Url == Request.Url.Scheme + "://" + Request.Url.Authority + Request.RawUrl)
                {
                    foreach (DotNetNuke.Services.Sitemap.AlternateUrl alt in url.AlternateUrls)
                    {
                        AttachCustomHeader("<link rel='alternate' hreflang='" + alt.Language.Substring(0,2) + "' href='" + alt.Url + "' />");
                    }
                }
            }
        }
        catch
        {
        }

    protected void AttachCustomHeader(string CustomHeader) { HtmlHead HtmlHead = (HtmlHead)Page.FindControl("Head"); if ((HtmlHead != null)) { HtmlHead.Controls.Add(new LiteralControl(CustomHeader)); } }

</script>`

jac69:
This bot is boring.

[...]

PR #2870, opened 2019-06-24
Sitemap.aspx does not produce alternate links in multilingual website

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2819

See demo: [DEMO](https://drive.google.com/file/d/1ewxm5j7-fvbOeey1CZpac9A3xcMrFBgY/view?usp=sharing)

To validate generated sitemap, there is a bunch of validators available. Validation requires sitemap.xml to be available on public. To pass xml from local environment to public, I generated sitemap locally and mock xml response using [mocky.io](https://www.mocky.io/). Passing public URL to the validator does the trick. No issues detected.

Please be noted, adding `xhtml:link` tags makes xml broken to be displayed in browser, however, document parses correctly in a sitemap validators and looks as per official google documentation, see [Sitemap](https://support.google.com/webmasters/answer/189077) chapter.

### DIFF
--- a/DNN Platform/Library/Services/Sitemap/SitemapBuilder.cs
+++ b/DNN Platform/Library/Services/Sitemap/SitemapBuilder.cs
@@ -362,17 +362,18 @@ namespace DotNetNuke.Services.Sitemap
             writer.WriteElementString("changefreq", sitemapUrl.ChangeFrequency.ToString().ToLowerInvariant());
             writer.WriteElementString("priority", sitemapUrl.Priority.ToString("F01", CultureInfo.InvariantCulture));
 
-            //if (sitemapUrl.AlternateUrls != null)
-            //{
-            //    foreach (AlternateUrl alternate in sitemapUrl.AlternateUrls)
-            //    {
-            //        writer.WriteStartElement("link", "http://www.w3.org/1999/xhtml");
-            //        writer.WriteAttributeString("rel", "alternate");
-            //        writer.WriteAttributeString("hreflang", alternate.Language);
-            //        writer.WriteAttributeString("href", alternate.Url);
-            //        writer.WriteEndElement();
-            //    }
-            //}
+            if (sitemapUrl.AlternateUrls != null)
+            {
+                foreach (var alternate in sitemapUrl.AlternateUrls)
+                {
+                    writer.WriteStartElement("link", "http://www.w3.org/1999/xhtml");
+                    writer.WriteAttributeString("rel", "alternate");
+                    writer.WriteAttributeString("hreflang", alternate.Language);
+                    writer.WriteAttributeString("href", alternate.Url);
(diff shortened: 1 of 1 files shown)


bdukes:
After reviewing this, there's some history to consider:
* This functionality was commented out in 40394e1b92bf2a5397b45ab28f5a7294286c93b7 (which, I believe, corresponds to the 7.4.2 release)
* There's a discussion in [Jira: DNN-4122](https://dnntracker.atlassian.net/browse/DNN-4122)
  * >A common case for multi-lingual sites is to have them running off different domain names so they are relevant to their locale. This in turn means they are separate sites to crawlers, thus they should not appear with mixed domains in a sitemap as this is incorrect. When locales in DNN are mapped to different portal aliases, as they are effectively separate sites at that point.
  * The final decision seems to be an agreement on a different sitemap per domain/language
* In the issue mentioned, #2819, @Tychodewaard asks for a `<link rel="alternate" hreflang="…" href="…" />` tag, not a sitemap change

That said, this seems like the simpler solution, but we'll need to have some discussion, since we're reversing a previous decision.

cc: @sleupold @daguiler @EPTamminga @mitchelsellers

daguiler:
This seems to be correctly implemented as per the standard but the obvious question is why this was commented out before.
This code was commented out 4 years ago, here: https://github.com/dnnsoftware/Dnn.Platform/pull/471/
Maybe someone remembers why?

daguiler:
This was attempted previously: https://dnntracker.atlassian.net/browse/DNN-9223

mikebigun:
Thanks to all for sharing the history of initial problem. Checking all the tickets reported previously I found community tried to resolve sitemap XML creation problem for the multilingual site.
Having multiple languages, i.e. `en-US` and `en-AU`, system was writing both to a single sitemap. That was causing XML structure problem due to multiple root elements and etc...

Finally, decision was to split sitemap per locale. 
If fact, with this change we do not break implemented behavior. We still generates one sitemap = one locale. I added both `en-US` and `en-AU` to the system, see how sitemaps are created.

`http://dnnhost.com/sitemap.aspx` - will result to `sitemap.en-us.xml` (default locale)
`http://dnnhost.com/en-US/sitemap.aspx` - will result to `sitemap.en-us.xml`
`http://dnnhost.com/en-AU/sitemap.aspx` - will result to `sitemap.en-au.xml`

image
Having this changes we just adding extra elements with an alternative links to help google properly crawl sites (see highlighted areas). This should not produce any regressions. 

image

daguiler:
Make sense @mikebigun.
Any concerns @dnnsoftware/tag with re-enabling this?

mikebigun:
Dear community, please let me know if there are any concerns.

valadas:
Just commenting this so stalebot does not close it.

PR #2877, opened 2019-07-02
Fixes an issue with the boolean values in the Google Analytics Connector

valadas:
Right or wrong the connectors infrastructure only supports passing strings as config values. Knockout has a strange behaviour where it treats any string as true and empty strings as false. However we get back from the connector UI either "true" or "false" for some reason. This PR handles this difference of behaviour when sending and receiving the settings for boolean values.

I have tested that it saves "false" or "true" in the database and that this PR restores the module UI problem, however I am not sure then where those database values are used, so testing that could be nice :)

It also adds a confirmation on save since that part was missing compared to the Azure Connector.

### DIFF
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -134,8 +134,8 @@ namespace DNN.Connectors.GoogleAnalytics
                 { "TrackAdministrators", trackForAdmin},
                 { "AnonymizeIp", anonymizeIp},
                 { "TrackUserId", trackUserId},
-                { "DataConsent", HandleCustomBoolean(portalSettings.DataConsentActive) },
-                { "isDeactivating", HandleCustomBoolean(false) }
+                { "DataConsent", HandleCustomBoolean(portalSettings.DataConsentActive.ToString()) },
+                { "isDeactivating", HandleCustomBoolean("false") }
             };
 
             return configItems;
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -134,7 +134,7 @@ namespace DNN.Connectors.GoogleAnalytics
                 { "TrackAdministrators", trackForAdmin},
                 { "AnonymizeIp", anonymizeIp},
                 { "TrackUserId", trackUserId},
-                { "DataConsent", portalSettings.DataConsentActive ? "true" : ""},
+                { "DataConsent", HandleCustomBoolean(portalSettings.DataConsentActive) },
                 { "isDeactivating", ""}
             };
 
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -149,7 +149,7 @@ namespace DNN.Connectors.GoogleAnalytics
         /// <returns></returns>
         private string HandleCustomBoolean(string value)
         {
-            if (value.ToLowerInvariant().Trim() == "true")
+            if (value.Trim().Equals("true", StringComparison.OrdinalIgnoreCase))
             {
                 return "true";
             }
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -87,7 +87,7 @@ namespace DNN.Connectors.GoogleAnalytics
             var analyticsConfig = AnalyticsConfiguration.GetConfig("GoogleAnalytics");
             var portalSettings = new PortalSettings(portalId);
 
-            // Important, knockout handles emptry strings as false and any other string as true
+            // Important, knockout handles empty strings as false and any other string as true
             // so we need to pass empty strings when we mean false, however it passes us back the string "false"
             // when saving the settings in the SaveConfig method, so we need to handle that case too
             var trackingId = string.Empty;
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -190,7 +190,7 @@ namespace DNN.Connectors.GoogleAnalytics
                 else
                 {
                     trackingID = values["TrackingID"] != null ? values["TrackingID"].ToLowerInvariant().Trim() : string.Empty;
-                    urlParameter = values["UrlParameter"] != null ? values["UrlParameter"].ToLowerInvariant().Trim() : string.Empty;
+                    urlParameter = values["UrlParameter"]?.Trim() ?? string.Empty;
                     trackForAdmin = values["TrackAdministrators"] != null ? values["TrackAdministrators"].ToLowerInvariant().Trim() : string.Empty;
                     anonymizeIp = values["AnonymizeIp"] != null ? values["AnonymizeIp"].ToLowerInvariant().Trim() : string.Empty;
                     trackUserId = values["TrackUserId"] != null ? values["TrackUserId"].ToLowerInvariant().Trim() : string.Empty;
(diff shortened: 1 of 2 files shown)


valadas:
@mitchelsellers @ohine I think this completes 9.4.0

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -88,11 +87,14 @@ public bool HasConfig(int portalId)
...
             var portalSettings = new PortalSettings(portalId);
 
-            var trackingId = String.Empty;
-            var urlParameter = String.Empty;
-            var trackForAdmin = false;
-            var anonymizeIp = false;
-            var trackUserId = false;
+            // Important, knockout handles emptry strings as false and any other string as true
+            // so we need to pass empty strings when we mean false, however it passes us back the string "false"
+            // when saving the settins in the SaveConfig method, so we need to handle that case too
```
bdukes:
```suggestion
            // when saving the settings in the SaveConfig method, so we need to handle that case too
```

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -88,11 +87,14 @@ public bool HasConfig(int portalId)
             var analyticsConfig = AnalyticsConfiguration.GetConfig("GoogleAnalytics");
             var portalSettings = new PortalSettings(portalId);
 
-            var trackingId = String.Empty;
-            var urlParameter = String.Empty;
-            var trackForAdmin = false;
-            var anonymizeIp = false;
-            var trackUserId = false;
+            // Important, knockout handles emptry strings as false and any other string as true
```
bdukes:
```suggestion
            // Important, knockout handles empty strings as false and any other string as true
```

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -108,46 +110,52 @@ public bool HasConfig(int portalId)
...
-                { "TrackAdministrators", trackForAdmin.ToString()},
-                { "AnonymizeIp", anonymizeIp.ToString()},
-                { "TrackUserId", trackUserId.ToString()},
-                { "DataConsent", portalSettings.DataConsentActive.ToString()},
-                { "isDeactivating", false.ToString()}
+                { "TrackAdministrators", trackForAdmin},
+                { "AnonymizeIp", anonymizeIp},
+                { "TrackUserId", trackUserId},
+                { "DataConsent", portalSettings.DataConsentActive ? "true" : ""},
+                { "isDeactivating", ""}
```
bdukes:
```suggestion
                { "isDeactivating", HandleCustomBoolean(false) }
```

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -108,46 +110,52 @@ public bool HasConfig(int portalId)
...
                 { "UrlParameter", urlParameter},
-                { "TrackAdministrators", trackForAdmin.ToString()},
-                { "AnonymizeIp", anonymizeIp.ToString()},
-                { "TrackUserId", trackUserId.ToString()},
-                { "DataConsent", portalSettings.DataConsentActive.ToString()},
-                { "isDeactivating", false.ToString()}
+                { "TrackAdministrators", trackForAdmin},
+                { "AnonymizeIp", anonymizeIp},
+                { "TrackUserId", trackUserId},
+                { "DataConsent", portalSettings.DataConsentActive ? "true" : ""},
```
bdukes:
```suggestion
                { "DataConsent", HandleCustomBoolean(portalSettings.DataConsentActive) },
```

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -165,28 +173,27 @@ public bool SaveConfig(int portalId, IDictionary<string, string> values, ref boo
...
                 }
                 else
                 {
-                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToString().Trim() : String.Empty;
-                    urlParameter = values["UrlParameter"] != null ? values["UrlParameter"].ToString().Trim() : String.Empty;
-
-                    bool.TryParse(values["TrackAdministrators"].ToLowerInvariant(), out trackForAdmin);
-                    bool.TryParse(values["AnonymizeIp"].ToLowerInvariant(), out anonymizeIp);
-                    bool.TryParse(values["TrackUserId"].ToLowerInvariant(), out trackUserId);
+                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToLowerInvariant().Trim() : string.Empty;
```
bdukes: should tracking ID and URL parameter be lower-cased?

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -108,46 +110,52 @@ public bool HasConfig(int portalId)
...
 
+        /// <summary>
+        /// Handles custom conversion from "true" => "true"
+        /// Anything else to "" to support the strange knockout handling of string as booleans
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        private string HandleCustomBoolean(string value)
+        {
+            if (value.ToLowerInvariant().Trim() == "true")
```
bdukes:
```suggestion
            if (value.Trim().Equals("true", StringComparison.OrdinalIgnoreCase))
```

mitchelsellers:
@valadas if you can address the above items I can get a review in.  I tried to commit a few of the changes, but it looks like your branch protection doesn't allow edits from us.

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
valadas: Nope, the object is lower case but the key is Upper Case

valadas:
@mitchelsellers @bdukes , changes done and had to modify one suggestion because HandleCustomBoolean takes a string, not a bool. Should be good to go now.

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
bdukes: @valadas I'm asking about the `ToLowerInvariant()` call on the value, that shouldn't be there, should it?
```suggestion
                    trackingID = values["TrackingID"]?.Trim() ?? string.Empty;
```

DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -165,28 +173,27 @@ public bool SaveConfig(int portalId, IDictionary<string, string> values, ref boo
...
                 else
                 {
-                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToString().Trim() : String.Empty;
-                    urlParameter = values["UrlParameter"] != null ? values["UrlParameter"].ToString().Trim() : String.Empty;
-
-                    bool.TryParse(values["TrackAdministrators"].ToLowerInvariant(), out trackForAdmin);
-                    bool.TryParse(values["AnonymizeIp"].ToLowerInvariant(), out anonymizeIp);
-                    bool.TryParse(values["TrackUserId"].ToLowerInvariant(), out trackUserId);
+                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToLowerInvariant().Trim() : string.Empty;
+                    urlParameter = values["UrlParameter"] != null ? values["UrlParameter"].ToLowerInvariant().Trim() : string.Empty;
```
bdukes:
```suggestion
                    urlParameter = values["UrlParameter"]?.Trim() ?? string.Empty;
```

valadas: Any reason for removing the .ToLowerInvariant or is that a mistake?

bdukes: It seemed that `ToLowerInvariant` was added to make sure that `True` was turned into `true` for the Boolean settings, but I don't see any reason to make the values of the two text settings always be lower case.  I assumed it was just copied from the Boolean case without realizing.

PR #2879, opened 2019-07-02
Filename illegal characters check

donker:
We should be checking all filenames for illegal characters. This is now done right after a check of illegal extension.


DNN Platform/Library/Common/Globals.cs
```diff
@@ -94,6 +94,7 @@ public sealed class Globals
         public static readonly Regex BaseTagRegex = new Regex("<base[^>]*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
         public static readonly Regex FileEscapingRegex = new Regex("[\\\\/]\\.\\.[\\\\/]", RegexOptions.Compiled);
         public static readonly Regex FileExtensionRegex = new Regex(@"\..+;", RegexOptions.Compiled);
+        public static readonly Regex FileValidNameRegex = new Regex(@"^(?!(?:PRN|AUX|CLOCK\$|NUL|CON|COM\d|LPT\d)(?:\..+)?$)[^\x00-\x1F\xA5\\?*:\"";|\/<>]+(?<![\s.])$", RegexOptions.Compiled);
```
Yabgu: Regex is not case-insenstive will this "com1" "con" "aux" "nul" "lpt" ....

v-karbovnichy: Fixed in #2915

PR #2882, opened 2019-07-03
DNN-30483: fix several bugs around export/import area.

### LINKED ISSUE #2880: Several bugs around Export/Import function
1. after import, the pages' order aren't same with source;
2. the page link will be wrong if the url type is tab and referenced to some other pages;
3. extensions still export even disable the extensions option;
4. sometimes matched to wrong module in target;
5. if module's order changed, update module will trigger workflow and cause page in wrong state;
6. when module matched multiple times in different pages, the module content will not update in other pages but only import for the first time;

zyhfish:
Fix #2880


daguiler:
Closing in favor of https://github.com/dnnsoftware/Dnn.Platform/pull/3001

PR #2886, opened 2019-07-08
Closes #2885: Allow multiple DesktopModules in a single package

### LINKED ISSUE #2885: Install multiple DesktopModules in a single Package
**Creating an issue so that I can submit a pull request for an existing patch**

## Description of problem
I am making functionality customizations to an intranet which all share the same database connection, entities and business logic, using MVC, and it makes no sense to create lots of Packages / ModuleDefinitions each with their own assemblies, data folders, etc when MVC can cleanly host large modular projects.

The obvious use is to be able to have a DesktopModule per Controller in a project

## Description of solution
When uploading a package which has multiple <component type="Module"> DesktopModule definitions allow the DesktopModule to have its own <friendlyName> and <description>.

These are **already set up** in the DNN DesktopModule database structure / entity structure, but at the moment each DesktopModule in a Package takes the FriendlyName and Description of the package. This results in a "Key already exists" dictionary error.

Note that the change will not break compatibility with any existing packages / modules; if a DesktopModule does not specify a FriendlyName / Description itself the Package FriendlyName / Description will be used, which is the current behavior.

Create a package for each controller by changing the build scripts, and try to fudge it so that both packages use the same assemblies and folders. But that would be messy and this is a very simple patch.

kestasjk:
## Summary
A very simple change to the DesktopModule installation logic allowing multiple DesktopModules in a single Package.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/DesktopModuleInfo.cs
+++ b/DNN Platform/Library/Entities/Modules/DesktopModuleInfo.cs
@@ -540,6 +540,9 @@ namespace DotNetNuke.Entities.Modules
                         case "moduleName":
                             ModuleName = reader.ReadElementContentAsString();
                             break;
+                        case "friendlyName":
+                            FriendlyName = reader.ReadElementContentAsString();
+                            break;
                         case "foldername":
                             FolderName = reader.ReadElementContentAsString();
                             break;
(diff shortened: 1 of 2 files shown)


kestasjk:
Note that the conflict here is trivial; this does not actually conflict with anything in the development branch.

valadas:
@kestasjk the conflict can be resolved by merging the latest changes from the development branch, as is, this PR would undo the recent changes for supporting version numbers with 4 blocks like 01.01.01.01

kestasjk:
Understood; I will close this and submit one with the change against the development branch.

valadas:
That works. 

Ideally for easier patching of PRs if you get requested some minor change:
1. In your local clone, switch to the target branch
```
git checkout development
```
2. If not done already, setup the upstream repository
```
git remote add upstream https://github.com/dnnsoftware/Dnn.Platform
```
3. Ensure your target branch is up to date
```
git fetch upstream
git pull upstream development
```
4. Create a branch specific to the subject you are working on
```
git checkout -b issue-2886
```
5. Do your thing as usual, commit, push submit PR, etc.

Then if changes are needed you can just return to that subject specific branch and add more commits and push, that will update the PR in place without any other steps.

PR #2891, opened 2019-07-12
Update GlobalResources.resx

Tychodewaard:
## Summary
Copyright in mails sent to friends, friend requesrts etc was 2013 hard coded. No [portal]


bdukes:
`[year]` isn't one of the tokens that's supported for emails:
https://github.com/dnnsoftware/Dnn.Platform/blob/a977d93170b6f5285fc6b753577b248fd88af522/DNN%20Platform/Library/Services/Social/Messaging/Scheduler/CoreMessagingScheduler.cs#L135-L143

Once you add year into that list, this change looks good to me, thanks!

Tychodewaard:
Ok, I need a little help here. Because, I figured I'd add : 

`template = template.Replace("[YEAR]", getFullYear);`

But there is nog such thing as getFullYear. And I don't find any documentation where/how I can get the correct class (or whatever it is called. Sorry, I am a marketeer)

bdukes:
@Tychodewaard I went ahead and added the C# change.  I've asked for a couple of reviews so we can get this integrated.  Thanks!

PR #2898, opened 2019-07-18
Fix ModuleSettings > Added to Pages grid, pager wasn't working Issue #2897

### LINKED ISSUE #2897: Module Settings > Added to Pages pager does not work
## Description of bug
In Module Settings > Added to Pages section, the pager does not work when there is a large number of pages.

## Steps to reproduce
1. Go a DNN Site that has > 20 pages
2. Add an HTML module, and choose the option to "Display Module on All Pages"
3. Click the pager to go to page 2
4. Page doesn't advance

## Current result
Grid does not advance to other rpages

## Expected result
Grid should advance to other pages

## Additional context
The PR does the following:
1. Wires up the PageIndexChange event
2. Updates the styling on the Pager

I also looked at other locations where GridView is used and Digital Assets Manager is the only other location currently that uses paging, and it has the proper events to do paging

## Affected version
* [x] 9.3.2

https://github.com/dnnsoftware/Dnn.Platform/pull/2898

msant7:
In Module Settings > Added to Pages section, the pager does not work when there is a large number of pages.

Closes #2897

### DIFF
--- a/DNN Platform/Website/Portals/_default/Skins/_default/WebControlSkin/Default/GridView.Default.css
+++ b/DNN Platform/Website/Portals/_default/Skins/_default/WebControlSkin/Default/GridView.Default.css
@@ -46,3 +46,18 @@
     color: #000;
     text-decoration: none;
 }
+
+.dnn-grid-pager tr td {
+    padding:0;
+    border:none;
+}
+.dnn-grid-pager tr td a, .dnn-grid-pager tr td span {
+    padding: 6px;
+}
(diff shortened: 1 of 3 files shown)


mitchelsellers:
@dnnsoftware/tag Before I merge this, should we rebase this to the Release 9.4.x branch for inclusion in a 9.4.1 release?

This fix, in reality will be needed for 9.4.x, if we remove Telerik in 10.x this grid will be updated.

PR #2903, opened 2019-07-22
CK Editor: Browser - Taking a long time to load [Performance Issue]

mikebigun:
partially fixes https://github.com/DNN-Connect/CKEditorProvider/issues/121

This is an improvement into the DNN.Platform that will help to avoid performance issues when loading list of files on a browsing using CKEditor.
Having a huge amount of files (tested 80K files in one folder), it was detected that application invokes `GetUrl` method on every iteration. Using performance analyzer I found it invokes and builds portal setting every time we want to get URL.

Suggested 2 simple but very important improvements:
1. Avoid regex. Instead of that, I used `Arrays` and `IndexOfAny` to check for an invalid characters. It is 5X times faster vs. regular expressions we used before.
2. No need to create portal settings every time. We are getting list of files linked to a particular portal, so just ensuring that we use current portal id we can re-user instance of current settings.

### DIFF
--- a/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
+++ b/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
@@ -23,7 +23,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text.RegularExpressions;
 using DotNetNuke.Common;
 using DotNetNuke.Common.Internal;
 using DotNetNuke.Common.Utilities;
@@ -39,7 +38,7 @@ namespace DotNetNuke.Services.FileSystem
     {
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(StandardFolderProvider));
 
-        private static readonly Regex InvalidFileUrlCharsRegex = new Regex(@"[%;?:@&=+$,]", RegexOptions.Compiled);
+        private static readonly char[] InvalidFileUrlChars = new char[] { '%', ';', '?', ':', '@', '&', '=', '+', '$', ',' };
 
         #region Public Properties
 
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
```diff
@@ -193,13 +192,16 @@ public override string GetFileUrl(IFileInfo file)
         {
             Requires.NotNull("file", file);
 
-            var portalSettings = GetPortalSettings(file.PortalId);
+            var portalSettings = file.PortalId == PortalSettings.Current.PortalId ?
```
bdukes:
```suggestion
            var portalSettings = file.PortalId == PortalSettings.Current?.PortalId ?
```

bdukes: It could be possible that this API is used from outside of an HTTP request, so that `PortalSettings.Current` might be `null`

bdukes:
Except for the one suggested change, LGTM, thanks!

DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
mikebigun: You are right @bdukes 
Fixed!

DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
```diff
@@ -193,13 +192,16 @@ public override string GetFileUrl(IFileInfo file)
         {
             Requires.NotNull("file", file);
 
-            var portalSettings = GetPortalSettings(file.PortalId);
+            var portalSettings = file.PortalId == PortalSettings.Current?.PortalId ?
```
Yabgu: Is this thread safe? If not please assign this to variable first than check:

```
var currentPortalSettings = PortalSettings.Current;
var portalSettings = file.PortalId == currentPortalSettings?.PortalId ?
    currentPortalSettings : 
    GetPortalSettings(file.PortalId);
```

mikebigun: Yes. It is thread safe. I do not see any problems using it without extra variable.

mikebigun:
Dear approver's, let me know please if you have more concerns on it.

mitchelsellers:
@dnnsoftware/tag SHould we merge this to Development for 10.0 or should this be rebased for 9.4.1

valadas:
If there is no big risk, I would merge in 9.4.1, it would be a very welcome performance improvement. I don't know enough about thread safety to have an opinion on the risk here but other than that comment I don't see any risk...

mikebigun:
Team, I created another PR to merge it to `release/9.4.*` 

See https://github.com/dnnsoftware/Dnn.Platform/pull/2918

Please, close this one if it is not valid.

mitchelsellers:
I'm closing this one as we are going to merge the one into the 9.4.x branch, which will get merged into development later

PR #2909, opened 2019-07-23
Bug template

valadas:
Adds information about targeting the correct branch when submitting pull requests


OllyHodgson:
I guess the only problem with specifying the version numbers is you'll have to update these after every point release. Is that actually a problem, or would something like `[ ] 9.4.x nightly build` be better?

valadas:
@OllyHodgson it was already that way, I think there is a release checklist item to update the issue templates after each release.

PR #2911, opened 2019-07-24
Fix missing modules after import site

### LINKED ISSUE #2910: Module missing after import data into site
## Description of bug
When exporting site with option"Include Deletions", TabModules.IsDeleted = 1 (true) is included as part of export data. Then, import the data into the site with existing Modules will cause Modules.IsDeleted updated to 1 (true). Therefore, subsequent export the site without "Include Deletions" will not contain the Modules and TabModules which cause the module missing after import into the new site. Refer to steps to reproduce to see the impact to the user. 

## Steps to reproduce
1. Create a new page, add HTML module A and HTML module B into the page; 
3. Delete module A; 
4. Go to Settings > Export/Import, export the website in Full Mode, make sure include deletion is enabled; 
5. Import the exported package into the same site; 
6. Go to Content > Recycle bin, restore the deleted module; 
7. Edit the new page, at this time the 2 modules should both display on the page, publish the page; 
8. export the website in full mode again, named it as "V2"; 
9. create a new child empty site, import with V2 package; 
10. navigate to the new page in child site.

## Current result
Module A is missing in the imported site.

## Expected result
Module A should appear in the imported site.

## Screenshots
The recording is based on Evoq but it is the same for DNN.Platform. Issue is reproducible for all modules. I have tested with normal HTML module. 
https://www.screencast.com/t/uM0NLi2W

## Error log

## Additional context
Found the root cause - https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs#L900

ModuleInfo.IsDeleted is shared by both TabModules and Modules table. Modules.IsDeleted should based on ExportModule.IsDeleted and TabModules.IsDeleted should based on ExportTabModule.isDeleted. However, ModuleInfo.IsDeleted is only set to ExportTabModule.isDeleted value.

When import data to the same site, Modules.IsDeleted value is updated to be same as ExportModule.IsDeleted. After restored the TabModules, TabModules.IsDeleted is changed to false but Modules.IsDeleted is remains as true. When exporting the site without included deletions, Modules is not exported because Modules.IsDeleted is remains as true.

To resolve the issue, we should update Modules.IsDeleted with ExportModule.IsDeleted value.
```
                                local.IsDeleted = otherModule.IsDeleted; // set as ExportModule.IsDeleted value
                                _moduleController.UpdateModule(local); // update Modules
                                local.IsDeleted = other.IsDeleted; // set back to ExportTabModule.IsDeleted value
```

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

tingung:
Fixes #2910

## Summary
 
Root cause: 
ModuleInfo.IsDeleted is shared by both TabModules and Modules table. Modules.IsDeleted should based on ExportModule.IsDeleted and TabModules.IsDeleted should based on ExportTabModule.isDeleted. However, ModuleInfo.IsDeleted is only set to ExportTabModule.isDeleted value.

When import data to the same site, Modules.IsDeleted value is updated to be same as 
ExportModule.IsDeleted. After restored the TabModules, TabModules.IsDeleted is changed to false but Modules.IsDeleted is remains as true. When exporting the site without included deletions, Modules is not exported because Modules.IsDeleted is remains as true.

Solution: 
Set ModuleInfo.IsDeleted to ExportModule.IsDeleted when updating Modules
- This is to handle scenario when ExportTabModule.IsDeleted and ExportModule.IsDeleted are having different value.

Confirmation video - https://drive.google.com/open?id=1uhmOT31yk9zKAqk7eULNS9Q8uUdLr5wA

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -897,7 +897,7 @@ namespace Dnn.ExportImport.Components.Services
                                 local.DisplayTitle = other.DisplayTitle;
                                 local.DisplayPrint = other.DisplayPrint;
                                 local.DisplaySyndicate = other.DisplaySyndicate;
-                                local.IsDeleted = other.IsDeleted;
+                                //local.IsDeleted = other.IsDeleted;
                                 local.IsShareable = otherModule.IsShareable;
                                 local.IsShareableViewOnly = otherModule.IsShareableViewOnly;
                                 local.IsWebSlice = other.IsWebSlice;
@@ -915,7 +915,7 @@ namespace Dnn.ExportImport.Components.Services
                                 }
 
                                 // updates both module and tab module db records
-                                _moduleController.UpdateModule(local);
+                                UpdateModuleWithIsDeletedHandling(other, otherModule, local);
                                 other.LocalId = local.TabModuleID;
                                 otherModule.LocalId = localExpModule.ModuleID;
                                 Repository.UpdateItem(otherModule);
@@ -978,6 +978,18 @@ namespace Dnn.ExportImport.Components.Services
             }
 
             return count;
+        }
+        /*
+            Update Modules.IsDeleted with ExportModule.IsDeleted and not ExportTabModule.IsDeleted. 
+            ExportTabModule.IsDeleted may different from ExportModule.IsDeleted when Module is deleted.
+            Change ModuleInfo.IsDeleted to ExportModule.IsDeleted and reverting to ExportMabModule.IsDeleted after 
+            updating Modules.
+        */
+        private void UpdateModuleWithIsDeletedHandling(ExportTabModule other, ExportModule otherModule, ModuleInfo local)
+        {
+            local.IsDeleted = otherModule.IsDeleted;
+            _moduleController.UpdateModule(local);
+            local.IsDeleted = other.IsDeleted;
         }
 
         private int ImportModuleSettings(ModuleInfo localModule, ExportModule otherModule, bool isNew)
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -897,7 +897,6 @@ namespace Dnn.ExportImport.Components.Services
                                 local.DisplayTitle = other.DisplayTitle;
                                 local.DisplayPrint = other.DisplayPrint;
                                 local.DisplaySyndicate = other.DisplaySyndicate;
-                                //local.IsDeleted = other.IsDeleted;
                                 local.IsShareable = otherModule.IsShareable;
                                 local.IsShareableViewOnly = otherModule.IsShareableViewOnly;
                                 local.IsWebSlice = other.IsWebSlice;
@@ -985,11 +984,11 @@ namespace Dnn.ExportImport.Components.Services
             Change ModuleInfo.IsDeleted to ExportModule.IsDeleted and reverting to ExportMabModule.IsDeleted after 
             updating Modules.
         */
-        private void UpdateModuleWithIsDeletedHandling(ExportTabModule other, ExportModule otherModule, ModuleInfo local)
+        private void UpdateModuleWithIsDeletedHandling(ExportTabModule exportTabModule, ExportModule exportModule, ModuleInfo importModule)
         {
-            local.IsDeleted = otherModule.IsDeleted;
-            _moduleController.UpdateModule(local);
-            local.IsDeleted = other.IsDeleted;
+            importModule.IsDeleted = exportModule.IsDeleted;
+            _moduleController.UpdateModule(importModule);
+            importModule.IsDeleted = exportTabModule.IsDeleted;
         }
 
         private int ImportModuleSettings(ModuleInfo localModule, ExportModule otherModule, bool isNew)


DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -897,7 +897,7 @@ private int ImportTabModulesAndRelatedItems(TabInfo localTab, ExportTab otherTab
                                 local.DisplayTitle = other.DisplayTitle;
                                 local.DisplayPrint = other.DisplayPrint;
                                 local.DisplaySyndicate = other.DisplaySyndicate;
-                                local.IsDeleted = other.IsDeleted;
+                                //local.IsDeleted = other.IsDeleted;
```
mitchelsellers: If a code element is no longer needed it should be removed, not commented out as we have history tracking available via Git.

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -979,6 +979,18 @@ private int ImportTabModulesAndRelatedItems(TabInfo localTab, ExportTab otherTab
 
             return count;
         }
+        /*
+            Update Modules.IsDeleted with ExportModule.IsDeleted and not ExportTabModule.IsDeleted. 
+            ExportTabModule.IsDeleted may different from ExportModule.IsDeleted when Module is deleted.
+            Change ModuleInfo.IsDeleted to ExportModule.IsDeleted and reverting to ExportMabModule.IsDeleted after 
+            updating Modules.
+        */
+        private void UpdateModuleWithIsDeletedHandling(ExportTabModule other, ExportModule otherModule, ModuleInfo local)
```
mitchelsellers: I would like to get better names for these parameters as it is a bit hard to follow what is being done.

Possible suggestion:
other = exportedTabModule
otherModule = exportedModule
local = importedModule

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
tingung: agree. my mistake to overlook.

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
tingung: Okay. these looks a lot better.

bdukes:
I agree with @mitchelsellers, there's no reason to include deleted modules in the export.

tingung:
Hi @mitchelsellers ,
AFAIK, there is Include Deletions  option which allowed deleted item to be export which I think is a feature instead of the 'bug'. 

Before I created this PR, I have some doubt too and here is my doubt when I am discussing with ESW team. 

both TabModules/Modules shared the ModuleInfo to perform update. Both TabModules and  Modules have IsDeleted attribute. When deleting module, TabModules.IsDeleted is  updated to true but Modules.IsDeleted remains false. Therefore, restoring module only change TabModules.IsDeleted to false but not updating Modules. I am trying to figure out the business logic of Modules.IsDeleted to have broader picture of  the system before finalizing fix for DNN-31029.  There are few things come to my mind. 

1. Both TabModules.IsDeleted and Modules.IsDeleted should be the same. (Deleting / restoring module should update both)
2. TabModules.IsDeleted can be different from Modules.IsDeleted can be different. ModuleInfo.IsDeleted should splits into Module.IsModuleDeleted and ModuleInfo.IsTabModuleDeleted (This is tricky) to represent different value
3. Modules.IsDeleted is never being used. (Remove it and update import / export process).
4. Workaround during import process - update Modules.IsDeleted with ModuleInfo.IsDeleted (set from ExportModule.IsDeleted) and later set it back to ExportTabModule.IsDeleted.

This PR is based on option no4 which I agree is not ideal but it is minimal change based on ModuleInfo design and far less risky.

What do you think? Appreciated if you or anyone from community can provide broader picture.

mitchelsellers:
Personally, I'd rather see a pull request come in that fixes the glitch, removing the ability to export deleted content.  It just doesn't make sense at all to do that, so I'd be more in favor of that rather than a hackish fix

tingung:
Hi @mitchelsellers , 

The hackish fix is actually caused by sharing of the ModuleInfo.IsDeleted in between Modules.IsDeleted and TabModules.IsDeleted when updating the DB. Both sharing the same field when updating but I am not sure the reason behind. If someone can explain, we can either refactor to have two fields for IsDeleted instead of one in ModuleInfo or if both field should be always the same then we need to change IsDeleted for Modules when restoring / deleting Module. This PR is more a fix to live with a single field with ModuleInfo.IsDeleted because @zyhfish actually told me Modules.IsDeleted is only used during import/export which lead to me decide to fix in this manner.

About removing the export deleted content, as I have said it is more like a feature with deletion included set as an option when exporting site. Unless there is agreement we don't need this feature, I didn't see it as a good solution to resolve a bug.

@amarjeet , @zyhfish , appreciated if you can provide your thoughts?

zyhfish:
HI @mitchelsellers , this could be a function design in Export/Import. when start an export, then will be an option "Include Deletions", if its selected, then the soft deleted content will be included in the package as well.
image

mitchelsellers:
I get that, but that feature, really, doesn’t make sense at all.

david-poindexter:
I agree with you @mitchelsellers - why would I ever want/need the ability to do that?

valadas:
I do not use that feature much, and the few times I used it, I never had a need to export deleted content. So in my very very own use case, I would also just remove that unless there is anybody really needing that option... @tingung did you ever need to include deletions ?

One use case I can think of (and I am not ever sure if that works that way) is if you have a production site that you export to a staging site, in the staging you delete some modules and you want to apply that change to the production change (the deletions) using an incremental export. Does it even work that way?

tingung:
Hi @valadas, 
I am maintaining the project on behalf of ESW. Internally, we have a customer defect reported. I agree it is an edge scenario and most people may not use  it. I  have checked through the history and the include deletions option is there since the beginning. I don't have enough product experience to give the real scenario of the include deletions. 
@zyhfish may shed the light here. He is the reporter of the customer defect.

tingung:
Hi everyone,

This is from @zyhfish - IMO this option is designed for continues integration, that user only need to make changes in staging site, and all changes will applied in production automatically.

There are two customers reported this issue but we are pending approval from PCA to reveal such information here. 

Thanks.

valadas:
So the goal is that if there is a module on both and you delete on one and import on the other using "include deletions" it would delete that module in the other. Right?

sleupold:
I am understanding it this way: if you delete it in source website, it is moved to recycle bin and should show up in recycle bin as well on the target site.

valadas:
I have not tested this but I think that is the goal of the "include deletions" can someone who uses that feature confirm?

sleupold:
I am not sure, but some of the behavior might aim at Evoq specific features.

zyhfish:
Hi @valadas , AFAIK we have some customers use this feature to do synchronization between staging and production environment.

mitchelsellers:
I think we should arrange a discussion on this. 

Moving deleted items just does not make sense and could result in a lot of additional clutter in the destination environment

sleupold:
@mitchelsellers 
it depends on the purpose, it might make sense, if the target environment is expected to be a "perfect" clone of the source env.

tingung:
> I think we should arrange a discussion on this.
> 
> Moving deleted items just does not make sense and could result in a lot of additional clutter in the destination environment

@mitchelsellers , user can always choose not to include deletions as it is by default disabled too. For user who want perfect synchronization in between staging and prod environment (quoting @zyhfish and @sleupold ), they need the include deletions option. In real client environment, they need this function and there are customers reported this issue. 

I am quite new to DNN community but please let me know the procedures to arrange such discussion. Is it a meeting only? If yes, I can arrange the schedule and host the meeting. 

Thanks.

v-karbovnichy:
I am now the guy who had taken role of @daguiler in ESW for DNN.

What others are saying of using this option by means of cloning the environment makes perfect sense here imo. Imagine a situation where a customer reports some bug, then you export/import his data to your place to play with it and debug, but the bug is not reproducible at your place. And it's because import/export did not copy those items that cause the bug.
Also perfect case where this can be an issue - with performance. For example, customer reports slowness, and it's because of a huge number of deleted items affects everything else. We would not be able to satisfy customer request of debugging and fixing it if we would not be able to copy his environment data as is.

I'm open for further discussions.

mitchelsellers:
If there is a community need for this feature, I guess I can support keeping this, but I'm guessing that there are a lot of unexpected consequences with the existing platform.

With regards to the question above.

TabModule.IsDeleted = This would be to delete a particular instance of a Module on a Particular Page
Module.IsDeleted = This would be to delete the actual module

It is possible that you can remove just 1 instance of a Module.  You would have 1 entry where a TabModule would be deleted, but others are not.

If a Module is deleted however, the behavior is to remove all (I believe)

At this point I would like to clean this up, if this means that we include more data, that is fine, but I would prefer to not have hack style solutions as they are hard to maintain.  If we need to discuss this at the next TAG meeting we can.

sleupold:
@mitchelsellers 
Module.IsDeleted: a module instance and all references to any page are deleted
TabModule.isDeleted: the module instance is deleted from a single page (i.e. the page reference is marked as deleted).

In the UI, the user deletes a page reference to a module (TabModule) and the module (reference) shows up in the recycle bin. If the module (reference) is purged from the recycle bin, the tabModule record is deleted and, if there are no more active or deleted-marked page references to a module instance, it is purged from the database as well.

bdukes:
I think we're come to an agreement that the functionality of retaining deleted modules in the export can have some value for customers and should be retained.

Regarding the behavior now in question, I think this PR ends up being a fairly "straightforward" workaround for DNN's API having mixed semantics around the `IsDeleted` property on the `ModuleInfo` class.  TL;DR that property updates the `Modules` table and is read from the `TabModules` table.

When calling `ModuleController.UpdateModule`, the `IsDeleted` column sets `Modules.IsDeleted` column (only):

https://github.com/dnnsoftware/Dnn.Platform/blob/30bd1a1bd9d2962a36f6e39403f00ba5877dcfab/DNN%20Platform/Library/Entities/Modules/ModuleController.cs#L1858-L1868

When filling a `ModuleInfo` instance, the `IsDeleted` column from the `vw_Modules` view is used, which comes from the `TabModules` table:

https://github.com/dnnsoftware/Dnn.Platform/blob/30bd1a1bd9d2962a36f6e39403f00ba5877dcfab/DNN%20Platform/Library/Entities/Modules/ModuleInfo.cs#L562

https://github.com/dnnsoftware/Dnn.Platform/blob/584b0f54bd6b6d237594dd9f96053f14db0b1a53/Website/Providers/DataProviders/SqlDataProvider/09.00.01.SqlDataProvider#L166-L168

Those other columns, `IsModuleDeleted` and `IsTabDeleted`, don't appear to be used anywhere in the core (except for [one `IsTabDeleted` check in `GetModuleByDefinition`](https://github.com/dnnsoftware/Dnn.Platform/blob/584b0f54bd6b6d237594dd9f96053f14db0b1a53/Website/Providers/DataProviders/SqlDataProvider/09.00.01.SqlDataProvider#L185-L203)).

So, given that this change is isolated to `PagesExportService`, I think it's appropriate.  Alternatively, if only the Tab Module is deleted, we could call [`ModuleController.DeleteTabModule(tabId, moduleId, softDelete: true)`](https://github.com/dnnsoftware/Dnn.Platform/blob/30bd1a1bd9d2962a36f6e39403f00ba5877dcfab/DNN%20Platform/Library/Entities/Modules/IModuleController.cs#L112), but if the row in `Modules` is deleted, there's no better way to set it while retaining the correct meaning of the "read" value, other than re-reading and busting the cache potentially many times during the import.

There are updates we can make to this API, but, for now, this service that consumes the API works just with this change, IMO.

valadas:
The only thing is since this targets development, it would go into Dnn10 right? Or do we need to ask @tingung to re-submit after 9.4.0 is released ?

daguiler:
We should include this 9.4.1, what do you think?

valadas:
Well it currently targets development which will got for Dnn10, if there is no breaking change and it is resubmitted targeting release/9.4.x soon, then sure.

daguiler:
Closing in favor of https://github.com/dnnsoftware/Dnn.Platform/pull/3022

PR #2912, opened 2019-07-27
Removes duplication of license files

### LINKED ISSUE #2904: Leftover license files when upgrading to 9.4.0
There was a cleanup of license files as part of 9.4.0 but some where forgotten, the following files are not cleaned-up during upgrade:

image

valadas:
This PR removes a duplication of license files in the Licenses directory on clean installs that are not duplicated on upgrade.

Closes #2904


valadas:
I have messed the rebase somehow and the selectize version bump is shown as a diff even though they are the same on my branch and the target branch, odd stuff. Anyway, I have replaced this with #2914 
@mitchelsellers @david-poindexter if you could re-review the new one please.

PR #2915, opened 2019-07-30
Ensure proper regex matching

donker:
When ensuring a file name is valid should also ensure we're getting the regex options right.


v-karbovnichy:
Looks good for me

PR #2917, opened 2019-07-31
Fix wrong navigation after creating the group

### LINKED ISSUE #2916: After creating a new group in the page, page redirect to Group activity page instead of current page
## Description of bug
Provide a clear and concise description of what the bug is.
After creating a new group in the page, page redirect to Group activity page instead of current page. 

## Steps to reproduce
1. Go to any page
2. Click on "Edit"  (pencil icon) at the bottom of the page. The page should in edit mode.
3. Click the "Add Module".
4. Add Module Pop Up will open - Scroll down and Add Social Groups 
5. Drag social groups in one of html div section 
6. Click on Auto Configure 
7. Click on Create New Group and fill in the details. 
8. Click on Create Group button.

## Current result
Explain what the current result is.
Page navigates to the Group-Activity/groupid URL. 
And in order to publish the page, user again need to click on respective page and click on Pencil icon and then click on publish button in order to publish the page. 

## Expected result
Provide a clear and concise description of what you expected to happen.
After clicking on create group button, the page should remain at the Edit Mode such that the user can publish the page.

## Screenshots
If applicable, provide screenshots to help explain your problem.

## Error log

## Additional context
Both edit mode and actual page are sharing the same module, make changes will affecting both unless we check for the condition for the edit mode. However, I think it is  acceptable to stay at current page after creating group in regardless in edit mode or not. 

## Affected version

* [ ] 10.0.0 nightly build
* [ ] 9.4.1 nightly build
* [ ] 9.4.0 release candidate
* [ ] 9.3.2 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

tingung:
Fixes #2916
## Summary

Fix wrong navigation after creating group
- Navigate to current page after creating group 
- Add overloading methods  to support closingUrl
- Update List.ascx to get current url from Request (System.URI) object and invoke GetCreateUrl with current url as closing url. 

Confirmation video - https://drive.google.com/open?id=1S5qmFuSL3UkuqyGPmd8yyk4W8Ek4IwUs

### DIFF
--- a/DNN Platform/Library/Common/Utilities/UrlUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/UrlUtils.cs
@@ -219,8 +219,13 @@ namespace DotNetNuke.Common.Utilities
 
         public static string PopUpUrl(string url, Control control, PortalSettings portalSettings, bool onClickEvent, bool responseRedirect)
         {
-            return PopUpUrl(url, control, portalSettings, onClickEvent, responseRedirect, 550, 950);
-        }       
+            return PopUpUrl(url, control, portalSettings, onClickEvent, responseRedirect, string.Empty);
+        }
+
+        public static string PopUpUrl(string url, Control control, PortalSettings portalSettings, bool onClickEvent, bool responseRedirect, String closingUrl)
+        {
+            return PopUpUrl(url, control, portalSettings, onClickEvent, responseRedirect, 550, 950, true, closingUrl);
+        }
 
         public static string PopUpUrl(string url, Control control, PortalSettings portalSettings, bool onClickEvent, bool responseRedirect, int windowHeight, int windowWidth)
         {
--- a/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
+++ b/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
@@ -768,11 +768,15 @@ namespace DotNetNuke.UI.Modules
 
         public string EditUrl(string keyName, string keyValue, string controlKey)
         {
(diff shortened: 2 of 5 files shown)


DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -219,8 +219,13 @@ public static void OpenNewWindow(Page page, Type type, string url)
 
         public static string PopUpUrl(string url, Control control, PortalSettings portalSettings, bool onClickEvent, bool responseRedirect)
         {
-            return PopUpUrl(url, control, portalSettings, onClickEvent, responseRedirect, 550, 950);
```
valadas: This could be a breaking change for other modules using this method

tingung: Should not be. There is another method overloading. 
` public static string PopUpUrl(string url, Control control, PortalSettings portalSettings, bool onClickEvent, bool responseRedirect, String closingUrl)
        {
            return PopUpUrl(url, control, portalSettings, onClickEvent, responseRedirect, 550, 950, true, closingUrl);
        }`
the exact same method with closingUrl = string.empty. The default behavior for closingURL is string.empty.

valadas: You are correct, sorry about that.

tingung:
PR targets 9.4.x is created - https://github.com/dnnsoftware/Dnn.Platform/pull/3213. closing this PR.

PR #2920, opened 2019-07-31
DNN-31366 - Deleted page can still be selected as parent page

### LINKED ISSUE #2919: Deleted page can still be selected as parent page
## Description of bug
There is a bug in the Pages screen when selecting a parent page. 

The deleted page is found inside the Recycle Bin. Behavior changes when the deleted page in Recycle Bin is deleted. Delete page is no longer visible in the Parent Page dropdown.

Note: Only deleted page that has a parent page can still be selected as a parent page. The deleted page that does not have parent page, once deleted is not visible in the Parent Page dropdown.

## Steps to reproduce
1. Create a page called "PAGETODELETE" and select a parent page for it. 
2. After creation, delete PAGETODELETE. 
3. Click Add Page, in the new page form, go to Parent Page dropdown 

## Current result
In Pages screen, you are still able to see PAGETODELETE page.

## Expected result
In Pages screen, you should not be able to see PAGETODELETE page.

## Screenshots
image

## Error log

## Additional context

## Affected version

* [x] 9.3.2 latest supported release
* [x] 9.2.2 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

berkarslan-xo:
Partially Fixes #2919 (There are total 2 PRs; for the other PR, you can check: https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1082 ).

## Summary

A new overload which accepts an additional parameter named as "includeDeletedChildren" is added.

Demo: https://drive.google.com/open?id=1yg6OuVuLMbwCbBlFHXDqQa8Abd3WPIn_

Note: This PR should be merged BEFORE https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1082

### DIFF
--- a/DNN Platform/Library/Entities/Tabs/TabController.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabController.cs
@@ -2473,7 +2473,8 @@ namespace DotNetNuke.Entities.Tabs
                                  false,
                                  false,
                                  false,
-                                 false);
+                                 false,
+                                 true);
         }
 
         /// <summary>
@@ -2497,7 +2498,8 @@ namespace DotNetNuke.Entities.Tabs
                                  includeDeleted,
                                  includeURL,
                                  false,
-                                 false);
+                                 false,
+                                 true);
         }
 
         /// <summary>
@@ -2525,7 +2527,8 @@ namespace DotNetNuke.Entities.Tabs
                                  includeDeleted,
                                  includeURL,
                                  checkViewPermisison,
-                                 checkEditPermission);
+                                 checkEditPermission,
+                                 true);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


sleupold:
IMO this could be handled well by an additional (computed) property in the API (where computation IMO belongs)

valadas:
Yes but active and not deleted does not have the same meaning, specially here, we do want to be able to add sub-pages to inactive pages, but not to deleted pages. The IsDeleted property already exists, the frontend just needs to ignore those pages that are deleted. Or adding an api that has a additional parameter to inclue or explude deleted pages could be ok. 

The problem here is that there is a modification of a current public api, adding a new overload that takes an additional boot to included deleted pages or not and then redirecting the other one to this new overload and mark it deprecated would be more acceptable. But as it is, this has the potential to break anything that expected the old signature without any previous warning.

berkarslan-xo:
The problem here was that I did this change because in the tree display (which appears when you click on the combobox of Parent Page), child pages are only loaded when you click on a parent page through AJAX call but that little arrows on the lefthand side of the pages appear only when HasChildren property is true (if there are no children, arrows don't appear as you might guess). I changed the part that I needed to change in AdminExperience repo about not listing the deleted children (I fixed it in the place that serves the AJAX call) but in order to not show that little arrows when there are no children, I made this change. 

As @sleupold suggested, another solution might be adding a new computed property into TabInfo class named as "HasNonDeletedChildren". Because I don't want to call a method for each of the pages just to check if a page has any child for performance reasons. Is this solution ok?

valadas:
Would have to dig into the previous PR and check what needs to be done here, but there are multiple ways to skin this cat. What is really important is to not change the behaviour of an existing api or a common component because it may be used elsewhere for different purposes and needs.

I did not take the time to go check the details of all this but we either:
:a: add an api that returns what we want without modifying existing ones (overload)
:b: we handle that special need in that specific UI part by skipping nodes that have isDeleted true

The problem with this PR is that it changes what is returned for people already expecting the current behaviour.

berkarslan-xo:
I have added a new overload which accepts an additional parameter named as "includeDeletedChildren".

mitchelsellers:
@berkarslan-xo Can we get you to submit this against the 9.4.x branch rather than Development?

berkarslan-xo:
> @berkarslan-xo Can we get you to submit this against the 9.4.x branch rather than Development?

I have created https://github.com/dnnsoftware/Dnn.Platform/pull/2925 now.

berkarslan-xo:
Dear approvers, I have created the requested PR for 9.4.x (#2925) but it's not picked up yet for a review, do you need anything on my side?

And could you please let me know what I need to do about https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1082 ? Should I retarget it to another branch or is it ok as it is?

mitchelsellers:
Closing this as it will be processed with the other PR against the release branch.

PR #2928, opened 2019-08-07
Update development with fixes from v9.4.0-rc

ohine:
The referenced changes have already been approved. Since this is a release management task we are self approving the merge without any reviews.


mitchelsellers:
@ohine Should this be merged?

bdukes:
@mitchelsellers the [build](https://dev.azure.com/dotnet/9defc7c7-d48d-4a8b-9f46-fd11ba3c5aad/_build/results?buildId=20161) failed because of a failure in the call to the Azure API, I think that's what @ohine was talking about being in the way of generating a new RC…

PR #2932, opened 2019-08-10
Enables and documents local builds

### LINKED ISSUE #2120: Producing installation package without CI
## Description

Here's the proposal I submitted on TAG Slack some time last month. It presents my idea on detaching the process of packaging Platform, Evoq or even custom installation bundles from the CI platform.

## Current state

To build a package of DNN Platform it is required to build its source code and combine with artifacts produced by Persona Bar and some other repositories.

Until recently, this was done on TeamCity, which created a lock-in and its configuration promoted a pipeline which was hardly replicable on local developer machine.

## Intended state

Entire build pipeline should be easily executable from the Dnn.Platform repository:

1. Building solution
2. Downloading all required packages
3. Packaging into the zip files

## Getting there

Instead of relying on TeamCity or other 3rd party, which cannot be controlled directly (ie. exists only on CI infrastructure/software), I propose to employ NuGet packages.

### Example pipeline

First, the module would be built and published

1. Build module dependency
2. Create zip packages as usual
    * eg. DDRMenu_09.02.01_Install.zip
4. Wrap packages in a nupkg file (new)
    * DDRMenu.9.2.1.nupkg
6. Publish the nupkg to a feed

Now that module can be consumed by Platform build

1. Build Platform solution
2. Get all required packages from the feed
3. Copy the zip installers to the required location
4. Run packaging script

## Benefits

This has a number of positive effects on what is possible:

* Repository will become self-contained
* Packaging will be repeatable
* Currently there is no way guarantee that packaging same revision produces identical packages!
* Each module will be allowed to be published and versioned independently
* Module can have dependencies which will actually be resolved
   * Albeit, only at build-time but it's a start
* One would clone Dnn.Platform, add their own modules and package a custom installation package
   * Could use a private feed or use open feed of 3rd party components

## Tooling proposal

I mention NuGet, but not the official client. Instead, there is wonderful open source replacement called [Paket](http://fsprojects.github.io/Paket/). It has a number of advantages:

* Can (easily) work outside of a solution/csproj, just downloading the necessary packages
* Comes with a simple text format, both for defining dependencies as well as creating nupkgs
* Guarantees stable folder structure, which is important for easy scripting
* Generates a lock file, like yarn, which guarantees repeatable builds

valadas:
Closes #2120 

## Summary

* Moves the temp dir locally to prevent case sensitivity issues with c:\temp vs c:\Temp on some systems
* Since GitVersion is used to dynamically set versions on manifests, the build script now backs-up the manifests before the build as restores them as they where before the build to prevent unwanted changes in Pull Requests
* Added some folders to gitignore to not commit build results
* Removes the source package from the release process
* Fixes a duplicate folder in Build/Cake and Build/cake (case sensitvity) which showed as two folders in github but a single one in Windows (case insensitive)
* Enables parallel MSBuild execution and shaves 20 seconds of build time for Platform.


david-poindexter:
Fantastic @valadas - I’ll try my best to review.

valadas:
Please hold before merging this, I have a few more tweeks

valadas:
@mitchelsellers This should be good to go now, but I noticed the script that should run the Unit Tests fails, not sure if related to one of these commits or something else...

PR #2953, opened 2019-08-18
Allows loading dependencies during startup even if one fails

### LINKED ISSUE #2952: Startup code can fail loading all dependencies if a single type fails
## Description of bug
The startup code tries to get types safely but if a single type fails for any reason, all the types in that assembly are ignored, this can cause very odd and hard to troubleshoot bugs.

## Steps to reproduce
1. Create a basic module with EF6
2. Forget to put the EF6 assemblies in the bin folder
3. Add a webapi to the module
4. Notice the webapi does not work, no errors are logged even with highest levels of logging.

* [ ] 10.0.0 nightly build
* [ ] 9.4.1 nightly build
* [x] 9.4.0 release candidate
* [ ] 9.3.2 latest supported release

valadas:
Closes #2952 

Previously if one type in an assembly was failing, all the types of that assembly where ignored by the startup code. This PRs fixes this and returns all the other types to prevent odd behaviours.

### DIFF
--- a/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
+++ b/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Reflection;
 
 namespace DotNetNuke.DependencyInjection.Extensions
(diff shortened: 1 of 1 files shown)


mitchelsellers:
And just for the record I believe this has been there for a while. I’m not 100% sure if new behavior but for sure something we need to monitor in future PRs

valadas:
I did not create an issue for it, but the //TODO: syntax shows up in the visual studio tasks panel. The reason we have not implemented it yet is because it would have created a new dependency for the logging here and we wanted to push that to later. So maybe we can do that when we start using more dependency injection..

PR #2968, opened 2019-09-02
DNN-32726: avoid null reference exception when call to Globals.LinkClick method.

### LINKED ISSUE #2967: Possible null reference exception in StandardFolderProvider.GetFileUrl method
https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs#L206
this code will call to Globals.LinkClick method:
https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Common/Globals.cs#L3359
here it doesn't check the portal settings value, if the code called in a schedule, it will throw out an null reference exception.

zyhfish:
Fix #2967


mitchelsellers:
Any chance we could get this retarged or submitted for 9.4.1?

valadas:
Closing this one in favor of #2988 
Thanks @zyhfish

Issue #2971, opened 2019-09-03
Empty namespace xmlns="" added to several dependentAssembly nodes in web.config during DNN upgrade


tvatavuk:
## Description of bug
Module installed on upgraded DNN 9.4.0 RC1 can stop working with exception like this:

`Could not load file or assembly 'System.Web.Http.WebHost, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)`

## Steps to reproduce
List the steps to reproduce the behavior:

### Clean DNN install scenario

1. Clean install DNN 9.4.0 RC1 using DNN_Platform_9.4.0-rc1.20742_Install.zip.
1. Install module that depend on eg. System.Web.Http.WebHost. 
   1. Here is 2sxc install package that works on DNN 7.4.2 - 9.4.0 RC1 [2sic_2SexyContent_09.43.02_dnn940comp_Install.zip](https://github.com/dnnsoftware/Dnn.Platform/files/3569384/2sic_2SexyContent_09.43.02_dnn940comp_Install.zip)
1. And module to page. Module is working.

### Upgrade DNN scenario

1. Clean install DNN 7.4.2.
1. Upgrade to DNN 9.4.0 RC1 using DNN_Platform_9.4.0-rc1.20742_Upgrade.zip.
1. Install module that depend on eg. System.Web.Http.WebHost. 
   1. Here is 2sxc install package that works on DNN 7.4.2 - 9.4.0 RC1 [2sic_2SexyContent_09.43.02_dnn940comp_Install.zip](https://github.com/dnnsoftware/Dnn.Platform/files/3569384/2sic_2SexyContent_09.43.02_dnn940comp_Install.zip)
1. And module to page. Module is not working. We get exception:

`Could not load file or assembly 'System.Web.Http.WebHost, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)`

## Current result

In upgraded web.config, several nodes for dependent assemblies have namespace attribute empty `xmlns=""`.

```xml
      <dependentAssembly xmlns="">
        <assemblyIdentity name="System.Web.Http" publicKeyToken="31BF3856AD364E35" />
[...]

PR #2979, opened 2019-09-06
Fix potential XML namespace bug

bdukes:
The background for this issue is documented in #2971.  This PR fixes a bug that can be introduced when using `collision="save"` with an incorrectly created XML Merge file.  This can result in an element with a namespace being replaced by an element that has the namespace reset (e.g. `<dependentAssembly xmlns="">`), resulting in that element being ignored.


valadas:
Since this is not regression, I have assigned it to 9.4.1, let me know if you want to to it differently...

iJungleboy:
Very cool, awesome & thanks!

PR #3001, opened 2019-09-17
fix several bugs around export/import area.

### LINKED ISSUE #2880: Several bugs around Export/Import function

zyhfish:
Fix #2880

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Common/CompressionUtil.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Common/CompressionUtil.cs
@@ -158,7 +158,7 @@ namespace Dnn.ExportImport.Components.Common
 
         #region Private Methods
 
-        private static bool AddFileToArchive(ZipArchive archive, string file, int folderOffset, string folder = null)
+        public static bool AddFileToArchive(ZipArchive archive, string file, int folderOffset, string folder = null)
         {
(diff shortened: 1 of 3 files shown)


DNN Platform/Modules/DnnExportImport/Components/Common/CompressionUtil.cs
```diff
@@ -158,7 +158,7 @@ public static bool AddFileToArchive(string file, string archivePath, int folderO
 
         #region Private Methods
 
-        private static bool AddFileToArchive(ZipArchive archive, string file, int folderOffset, string folder = null)
+        public static bool AddFileToArchive(ZipArchive archive, string file, int folderOffset, string folder = null)
```
Yabgu: At least this could be been moved outsize private methods region

valadas: I agree, missed that. However I would even prefer to remove all regions and replace that by stylecop rules for order of methods...

PR #3003, opened 2019-09-17
Use standalone version of Selectize script

### LINKED ISSUE #3002: Calendar on Module Start Date & End Date - Error
## Description of bug
After a clean install of 9.4, when I tried to edit a module start date and end date. The calendar popup does not show. Instead, there are multiple errors that showed up on the console of the developer tool.

## Steps to reproduce
1. Click on Start Date or End Date Field when editing a module.
2. Open up developer tools inspector.

## Current result
Nothing happens, but errors on console.

Error: 
- Selectize: Dependency MicroPlugin is missing
- Make sure you either: (1) are using the "standalone" version of Selectize, or (2) require MicroPlugin before you load Selectize.
- Uncaught TypeError: w.define is not a function
- Uncaught ReferenceError: Selectize is not defined
- Uncaught TypeError: $(...).dnnComboBox is not a function

## Affected version
* [ ] 10.0.0 nightly build
* [ ] 9.4.1 nightly build

## Affected browser
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [X] Edge

bdukes:
Fixes #3002 

I plan to add one more commit which adjusts the version of the Selectize JS Library package to 0.12.6.1, and see if that works for upgrading from 9.4.0 to 9.4.1.


bdukes:
😞 update with 0.12.6.1 package failed:

>DotNetNuke.Data.PetaPoco.PetaPocoHelper - [4] Error executing SQL: `;Exec dbo.AddPackage @0, @1, @2, @3, @4, @5, @6, @7, @8, @9, @10, @11, @12, @13, @14, @15, @16`
>Cannot insert duplicate key row in object 'dbo.Packages' with unique index 'IX_Packages'. 
>The duplicate key value is (DNN, Selectize, JavaScript_Library, `<NULL>`, 0.12.6).
>The statement has been terminated.

valadas:
Hmm, my thought was that this PR would have been for all extension types, but looks like only Modules...
https://github.com/dnnsoftware/Dnn.Platform/pull/2666/files

valadas:
@bdukes Did you try just using the same version, I have been digging into the installer code to try and find a solution to this and from the looks of it, it checks if the version is >= than the currently installed one and the stored procedure supports creating or updating. This is different from modules package type where the check is > and not >= ...
So I _think_ you could just remove the fourth digit and give a try ...

bdukes:
@valadas weird, that seemed to have worked.  I've removed the version bump from the PR.  If someone else can do a test upgrade (from an older version to 9.4.0, then 9.4.1), that would help verify that it truly fixes the issue.

I'm confused because the whole reason we upgraded the library in the first place in #2907 was because there was a file in the package that wasn't updated during the upgrade to 9.4.0.

valadas:
Strange, but yeah, I will try a couple different upgrades this weekend. Please do not merge until we get some positive reports on this.

thabaum:
image
 
after upgrading from version 9.4 to 9.4.1 I still have two selectize javascript libraries with no trash can to toss one out.  I was trying to dig up how to remove this to see if it cures another issue I am having with a vendor module, noticed the RC1 out so thought I would give it a try.  I have same modules on a seperate site with a fresh copy of 9.4 installed and the vendor module does not have the issue.

I can now view and edit module settings.  Trying to cleanup an older installation along with setup some new ones so I should get a lot of tests in here soon settings things up to give loads of feedback.

valadas:
Hmm, this is expected since libraries installation does not work like modules, they do not upgrade and you can have multiple versions installed. Also, when the manifest specifies it is a system package, then there is no UI to remove them... So technically, this is by design.

I agree though that this is not perfect since it will add up with each new version, on the other side there might be modules consuming that older version specifically and removing it would be a breaking change for them. I never found it to be a good idea to offer javascript libraries to module developers for this specific reason.

edisfera:
I have upgraded from 9.4.0 to 9.4.1. In the upgraded 9.4.0 I removed the duplicated selectize.js library days ago form Extensions as weell as the unnecessary PersonaBar modules (only one PersonaBar module is listed). When I enter any module settings I still get the following errors in the browser console windows: 
Selectize: Dependency MicroPlugin is missing selectize.min.js:3:6108
<nessuna etichetta di gruppo> selectize.min.js:3:6170
TypeError: w.define is not a function selectize.min.js:4:2404
Make sure you either: (1) are using the "standalone" version of Selectize, or (2) require MicroPlugin before you load Selectize. selectize.min.js:3:6186
ReferenceError: Selectize is not defined dnn.combobox.js:4:5
TypeError: $(...).dnnComboBox is not a function 427:3183:47
TypeError: $(...).dnnComboBox is not a function 427:3210:137
The error doesn't seem fixed... is there a way to fix manually?

edisfera:
Just for the record, out of 3 updated DNN to 9.4.1 only one had the selectize problem solved. The other two were still showing the dependency problem as written in my previous post.
I tried emptying the cache, restarting the application with no luck.
Finally I repaired the installation of selectize from the Extensions the selectize.zip taken from DNN_Platform_9.4.1_Upgrade/Install/JavaScriptLibrary. I emptied the cache, restarted the application and forced reload (CTRL+F5) and the problem went away from the other 2 updated DNN.
So it seems the problem as been actually fixed.

PR #3021, opened 2019-09-24
DNN-31029 - Set ModuleInfo.IsDeleted to ExportModule.IsDeleted when

### LINKED ISSUE #2910: Module missing after import data into site

tingung:
Fixes #2910

## Summary
 
Root cause: 
ModuleInfo.IsDeleted is shared by both TabModules and Modules table. Modules.IsDeleted should based on ExportModule.IsDeleted and TabModules.IsDeleted should based on ExportTabModule.isDeleted. However, ModuleInfo.IsDeleted is only set to ExportTabModule.isDeleted value.

When import data to the same site, Modules.IsDeleted value is updated to be same as 
ExportModule.IsDeleted. After restored the TabModules, TabModules.IsDeleted is changed to false but Modules.IsDeleted is remains as true. When exporting the site without included deletions, Modules is not exported because Modules.IsDeleted is remains as true.

Solution: 
Set ModuleInfo.IsDeleted to ExportModule.IsDeleted when updating Modules
- This is to handle scenario when ExportTabModule.IsDeleted and ExportModule.IsDeleted are having different value.

Confirmation video - https://drive.google.com/open?id=1uhmOT31yk9zKAqk7eULNS9Q8uUdLr5wA

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -897,7 +897,7 @@ namespace Dnn.ExportImport.Components.Services
                                 local.DisplayTitle = other.DisplayTitle;
                                 local.DisplayPrint = other.DisplayPrint;
                                 local.DisplaySyndicate = other.DisplaySyndicate;
-                                local.IsDeleted = other.IsDeleted;
+                                //local.IsDeleted = other.IsDeleted;
                                 local.IsShareable = otherModule.IsShareable;
                                 local.IsShareableViewOnly = otherModule.IsShareableViewOnly;
                                 local.IsWebSlice = other.IsWebSlice;
@@ -915,7 +915,7 @@ namespace Dnn.ExportImport.Components.Services
                                 }
 
                                 // updates both module and tab module db records
-                                _moduleController.UpdateModule(local);
+                                UpdateModuleWithIsDeletedHandling(other, otherModule, local);
                                 other.LocalId = local.TabModuleID;
                                 otherModule.LocalId = localExpModule.ModuleID;
                                 Repository.UpdateItem(otherModule);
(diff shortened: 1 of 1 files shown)


valadas:
Oops, there is a merge conflict though, can you resolve that.

tingung:
> Oops, there is a merge conflict though, can you resolve that.

yes. checking.

PR #3025, opened 2019-09-25
Warns before install/upgrade about needing .Net Framework 4.7.2

### LINKED ISSUE #2990: Upgrade to DNN 9.4 - Could not load file or assembly 'netstandard, Version=2.0.0.0
## Description of bug
While upgrading a site from DNN 9.1 to 9.4, encountered a .net error message 
"Could not load file or assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified."

## Details
Understanding that DNN 9.4 has a requirements change for ASP.NET 4.7.2, I checked with the host to make sure that 4.7.2 was installed/available. Confirmed that ASP.NET 4.5 was in place, so they updated and restarted the server. However the error message remains after.

The Error message, or asp.net error is encountered in the initial upgrade actions on the /Install/Install.aspx?mode=upgrade step

## Affected version

[X ] 9.4.0 latest supported release

valadas:
Closes #2990 

Adds a check in the existing IIS configuration step to make sure that .Net Framework 4.7.2 at least is installed. This is the first step for both installs and upgrades so that the upgrade does not continue if that condition is not met and users can just re-fire the upgrade after doing windows updates without needing to restore and all that...

### DIFF
--- a/Website/Install/App_LocalResources/Installwizard.aspx.de-DE.resx
+++ b/Website/Install/App_LocalResources/Installwizard.aspx.de-DE.resx
@@ -538,4 +538,7 @@ Gehen Sie im Windows Explorer zum Basisverzeichnis Ihrer Website ({0}). Klicken 
   <data name="AcceptTerms.Text" xml:space="preserve">
     <value>I accept the terms of the &lt;a href="../Licenses/Dnn_Corp_License.pdf" target="_blank"&gt;license agreement&lt;/a&gt;</value>
   </data>
+  <data name="DotNetVersion472Required.Text" xml:space="preserve">
+    <value>.Net Framework 4.7.2 ist erforderlich. Suchen Sie nach Windows-Updates, bevor Sie fortfahren.</value>
+  </data>
 </root>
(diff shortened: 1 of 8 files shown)


Website/Install/App_LocalResources/Installwizard.aspx.de-DE.resx
```diff
@@ -538,4 +538,7 @@ Gehen Sie im Windows Explorer zum Basisverzeichnis Ihrer Website ({0}). Klicken
   <data name="AcceptTerms.Text" xml:space="preserve">
     <value>I accept the terms of the &lt;a href="../Licenses/Dnn_Corp_License.pdf" target="_blank"&gt;license agreement&lt;/a&gt;</value>
   </data>
-</root>
+  <data name="DotNetVersion472Required.Text" xml:space="preserve">
+    <value>.Net Framework 4.7.2 ist erforderlich. Suchen Sie nach Windows-Updates, bevor Sie fortfahren.</value>
```
sleupold: DNN benötigt .Net Framework in der Version 4.7.2 oder neuer. Bitte prüfen Sie verfügbare Updates für Windows, bevor Sie mit der Installation fortfahren.

PR #3032, opened 2019-09-27
Updates .Net 4.7.2 warning for NL and DE languages

valadas:
Updates the warning messages for NL and DE languages as per @sleupold and @Timo-Breumelhof translations


iJungleboy:
German is ok for me too. But I believe my review-right have been revoked ;) - at least I don't see a button for it.

PR #3033, opened 2019-09-27
Portal settings setters should not persist to DB

donker:
Erroneously the earlier code copied the pattern from TimeZone and persisted to the database the value being set. But the controller that fills this object passes through here so the PortalSettings keep getting reset and not being cached. This also raises the question whether TimeZone should be refactored.


valadas:
Since this modifies a public property behaviour, I would vote to retarget this to Dnn10 to be extra-safe. Thoughts? Is this change mainly for cleanup/performance or does it cause a bug?

mitchelsellers:
@donker can you confirm this is actually a bug in the new code?

PR #3037, opened 2019-09-28
Make sure TimeZone doesn't store to database just by setting the prop

donker:
Currently when setting the TimeZone property on the PortalSettings object it will persist to the database and clears the cache. This is obviously not a good idea. The front end already sets the TimeZone when an admin changes the value in the site settings personabar module.

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -583,7 +583,7 @@ namespace DotNetNuke.Entities.Portals
             get { return PortalController.GetPortalSettingAsBoolean("IsLocked", PortalId, false); }
         }
 
-        public TimeZoneInfo TimeZone { get; set; }
+        public TimeZoneInfo TimeZone { get; set; } = TimeZoneInfo.Local;
 
         public string PageHeadText
         {
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -92,8 +92,6 @@ namespace DotNetNuke.Entities.Portals
         }
         #endregion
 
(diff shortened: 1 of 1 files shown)


donker:
This fails the LoadPortalSettings_Sets_TimeZone_Property_To_Local_TimeZone test. I have no idea how to adjust this test so it will pass. We'll need to discuss.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -590,16 +589,7 @@ public bool IsThisPortalLocked
...
-        {
-            get { return _timeZone; }
-            set
-            {
-                _timeZone = value;
-                PortalController.UpdatePortalSetting(PortalId, "TimeZone", value.Id, true);
-            }
-        }
-
+        public TimeZoneInfo TimeZone { get; set; }
```
bdukes: Should we continue defaulting this to `TimeZoneInfo.Local`?
```suggestion
        public TimeZoneInfo TimeZone { get; set; } = TuneZoneInfo.Local;
```

valadas: Watch out, you have TuneZone instead of TimeZone in your suggested change here...

valadas:
After thinking about this again, is this just a code cleanup effort or is it causing a bug? If this is just cleanup, I think I would prefer targetting this for 10 release just to make sure we don't break something along the way. Everything that touches times scares me :)

donker:
> 
> 
> After thinking about this again, is this just a code cleanup effort or is it causing a bug? If this is just cleanup, I think I would prefer targetting this for 10 release just to make sure we don't break something along the way. Everything that touches times scares me :)

From my point of view it's code cleanup

valadas:
@donker can you look at the conflict and the failing test...

PR #3039, opened 2019-09-29
Remove disposal of MemoryStream for email attachments sent using SendTokenizedBuilkEmail

### LINKED ISSUE #3036: Remove disposal of MemoryStream for email attachments sent using SendtokenizedBuilkEmail
## Description of bug
The `LoadAttachments` method in `SendTokenizedBulkEmail` (DotNetNuke.Services.Mail) contains a `using` statement that unfortunately disposes of the MemoryStream for attachments prior to the sending of emails.  This applies to **Synchronous** and **Asynchronous** methods of sending mail.

The culprit `using` statement is [HERE](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Services/Mail/SendTokenizedBulkEmail.cs/#L349-L366).

This was reported initially as a bug in the old DNN.Newsletter core module via DNNCommunity/DNN.Newsletter#10, but after diagnosis by multiple community members, including @valadas, @donker and @EPTamminga, it was determined to be a bug, as referenced above, in **DNN.Platform**.

## Steps to reproduce
1. Install DNN.Newsletter 08.00.00 in DNN 9.x
2. Add module to a page.
3. Send an email with attachment(s) using the module.
4. See error in Event Log.

## Current result
The email is sent with attachment(s), but the attachment(s) are unable to be opened by the email recipient(s).  Each attachment is actually empty.  

An error is thrown and can be viewed in the Event Log:
`Cannot access a closed stream.`

## Expected result
Attachment(s) should be sent fully intact with the email.  Recipient(s) should be able to open the attachments without error.

## Screenshots
n/a

## Error log
```
AbsoluteURL:/Default.aspx

ExceptionGUID:8d181b82-db66-423b-a30b-f8f6aaad2d83

AssemblyVersion:

PortalId:-1

UserId:-1

TabId:-1

RawUrl:

Referrer:

UserAgent:

ExceptionHash:jVBJAjOaPiNuV4Hsc/+ph/YcMtk=

Message:Cannot access a closed Stream.

StackTrace:

at System.IO.__Error.StreamIsClosed() at System.IO.MemoryStream.Read(Byte[] buffer, Int32 offset, Int32 count) at System.Net.Mime.MimePart.Send(BaseWriter writer, Boolean allowUnicode) at System.Net.Mime.MimeMultiPart.Send(BaseWriter writer, Boolean allowUnicode) at System.Net.Mime.MimeMultiPart.Send(BaseWriter writer, Boolean allowUnicode) at System.Net.Mail.SmtpClient.Send(MailMessage message)
InnerMessage:

InnerStackTrace:

Source:mscorlib

FileName:

FileLineNumber:0

FileColumnNumber:0

Method:

Server Name: xxxxxxx
```

## Additional context
n/a

## Affected version
* [x] 9.x

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

david-poindexter:
Resolves #3036 

## Summary
The `using` statement has been removed to avoid disposal of the MemoryStream prior to sending of email.

### DIFF
--- a/DNN Platform/Library/Services/Mail/SendTokenizedBulkEmail.cs
+++ b/DNN Platform/Library/Services/Mail/SendTokenizedBulkEmail.cs
@@ -347,24 +347,22 @@ namespace DotNetNuke.Services.Mail
 			foreach (var attachment in _attachments)
 			{
 				var buffer = new byte[4096];
-                using (var memoryStream = new MemoryStream())
+                var memoryStream = new MemoryStream();
+                while (true)
                 {
-                    while (true)
+                    var read = attachment.ContentStream.Read(buffer, 0, 4096);
+                    if (read <= 0)
                     {
-                        var read = attachment.ContentStream.Read(buffer, 0, 4096);
-                        if (read <= 0)
-                        {
-                            break;
-                        }
-                        memoryStream.Write(buffer, 0, read);
+                        break;
                     }
-
-                    var newAttachment = new Attachment(memoryStream, attachment.ContentType);
-                    newAttachment.ContentStream.Position = 0;
-                    attachments.Add(newAttachment);
-                    //reset original position
-                    attachment.ContentStream.Position = 0;
+                    memoryStream.Write(buffer, 0, read);
                 }
+
+                var newAttachment = new Attachment(memoryStream, attachment.ContentType);
+                newAttachment.ContentStream.Position = 0;
+                attachments.Add(newAttachment);
+                //reset original position
+                attachment.ContentStream.Position = 0;
 			}
 
 			return attachments;


DNN Platform/Library/Services/Mail/SendTokenizedBulkEmail.cs
```diff
@@ -347,24 +347,22 @@ private List<Attachment> LoadAttachments()
 			foreach (var attachment in _attachments)
 			{
 				var buffer = new byte[4096];
-                using (var memoryStream = new MemoryStream())
+                var memoryStream = new MemoryStream();
```
bdukes: I would prefer if there's some defensive code to ensure the stream gets disposed if there's an exception loading the data into it.  Something like this:

```
Attachment newAttachment;
MemoryStream memoryStream = null;
try
{
    memoryStream = new MemoryStream();
	// the while loop
    newAttachment = new Attachment(memoryStream, attachment.ContentType);
    memoryStream = null;
}
finally 
{
	memoryStream?.Dispose();
}
```

david-poindexter: @bdukes `try/finally` is the same as `using` when you dispose in the `finally`.  This will not work.  Did you mean a `try/catch` perhaps?

bdukes: @david-poindexter this pattern only disposes if there's an exception thrown.  After the `Attachment` is constructed, `memoryStream` is set to `null`, so that in the `fianlly` block nothing is disposed.

donker: The "LoadAttachments" method effectively copies the attachments of the email into new streams. This is then used in the email method call to actually send a bulk email. There is a potential for a lot of these copies being made. So there is a case to be made to properly clean this up. But IMHO that needs to be done by the code that sends the email. IMO the issue requires a bigger intervention in this class.

david-poindexter: @bdukes okay, now I see what you are saying.  I will test that pattern and push a new commit if it works.  Thanks!

david-poindexter: @donker are you in agreement with the suggested pattern made by @bdukes?  Or are you saying you believe this will be problematic too?  I have not had the ability to test this yet due to some serious computer issues I am dealing with.  Hopefully, I'll be back up and running soon.

mitchelsellers: I would agree with @bdukes on this one, we need at least some form of protection, otherwise, we will have a memory leak situation and given that things aren't exactly perfect in this area, I think it is a requirement.

I also agree with @donker that longer-term this should most likely be done in a more efficient manner 

Looking at the C# reference documentation for SMTP and email sending, there is a best practices recommendation for Using statement (Which would be expected)

david-poindexter:
It is probably worth mentioning that this is not used anywhere in DNN.Platform or DNN.AdminExperience.  😉

valadas:
I was requested as a reviewer on this, but I defer to @bdukes and @mitchelsellers opinions on this since I am no expert in this area... Ping me up if you need anything though.

DNN Platform/Library/Services/Mail/SendTokenizedBulkEmail.cs
david-poindexter: @bdukes and @mitchelsellers - just pushed changes, worked like a charm.  Thanks again for the review and recommendations for this fix.  In the bigger picture, this area needs some more TLC.

PR #3046, opened 2019-10-01
Fix issue with dynamic web api controllers #3045

### LINKED ISSUE #3045: Dynamically compiled Web APIs don't resolve since DNN 9.4
## Description of bug
When a system has WebApi controllers which are compiled dynamically - in our case 2sxc Apps can include .cs files which are compiled on the fly - they must be given to the .net framework for activation. This worked well till DNN 9.4, which accidentally broke this functionality because DNN 9.4 replaced the `IHttpControllerActivator`. 

I already found out how to fix this. 

## Steps to reproduce
1. Create a DNN 9.2 and a 9.4.x and install 2sxc on both
2. Add an app-module (2sxc) to a page and install the angular demo app from https://2sxc.org/en/apps/app/tutorial-angular-8
3. In the UI, click on the WebAPI demo
4. As you can see, it works in <9.4 and fails in 9.4.x

### Result in Pre 9.4
image

### Result in 9.4
image

### Result in 9.4 with fix applied
image

## Current result
In 9.4 the WebApi calls fail with this message:
```
{"Message":"No HTTP resource was found that matches the request URI 'http://dnn940rc1raw2.dnndev.me/API/2sxc/app/auto/live/api/simple/hello'.","MessageDetail":"No controller was created to handle this request."}
```

## Expected result
This should work

## Additional context
The solution is fairly simple, I believe @ahoefling should be able to confirm this.

## Affected version
* [x] 9.4.1 nightly build
* [ ] 9.2 (not affected)

## Affected browser
* [x] all

## Solution
The solution is to correct the code in the `DnnHttpControllerActivator` in `DotNetNuke.Web`. Here's the fix:
```c#
using DotNetNuke.Common;
using System;
using System.Net.Http;
using System.Web.Http.Controllers;
using System.Web.Http.Dispatcher;

namespace DotNetNuke.Web.Api
{
    public class DnnHttpControllerActivator : IHttpControllerActivator
    {
        public IHttpController Create(HttpRequestMessage request, HttpControllerDescriptor controllerDescriptor, Type controllerType)
        {
            // first try to just get it from the DI - if it's there
            return (IHttpController) Globals.DependencyProvider.GetService(controllerType) ??
                   // If it's not found (null), then it's probably a dynamically compiled type from a .cs file or similar
                   // Such types are never registered in the DI catalog, as they may change on-the-fly.
                   // In this case we must use ActivatorUtilities, which will create the object and if it expects 
                   // any DI parameters, they will come from the DependencyInjection as should be best practice
                   (IHttpController) ActivatorUtilities.CreateInstance(Globals.DependencyProvider, controllerType);
        }
    }
}

```

iJungleboy:
fixes #3045 

## Summary
Corrects the `IHttpControllerActivator` to ensure dynamically compiled controllers work as well.

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/DnnHttpControllerActivator.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/DnnHttpControllerActivator.cs
@@ -3,6 +3,7 @@ using System;
 using System.Net.Http;
 using System.Web.Http.Controllers;
 using System.Web.Http.Dispatcher;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace DotNetNuke.Web.Api
 {
@@ -10,7 +11,13 @@ namespace DotNetNuke.Web.Api
     {
         public IHttpController Create(HttpRequestMessage request, HttpControllerDescriptor controllerDescriptor, Type controllerType)
         {
-            return (IHttpController)Globals.DependencyProvider.GetService(controllerType);
+            // first try to just get it from the DI - if it's there
+            return (IHttpController) Globals.DependencyProvider.GetService(controllerType) ??
+                   // If it's not found (null), then it's probably a dynamically compiled type from a .cs file or similar
+                   // Such types are never registered in the DI catalog, as they may change on-the-fly.
+                   // In this case we must use ActivatorUtilities, which will create the object and if it expects 
+                   // any DI parameters, they will come from the DependencyInjection as should be best practice
+                   (IHttpController) ActivatorUtilities.CreateInstance(Globals.DependencyProvider, controllerType);
         }
     }
 }


iJungleboy:
I believe an ideal reviewer would be @ahoefling

mitchelsellers:
@ahoefling do you happen to have a reference to the behavior from the prior release before the changes?  (Thought it might save me some digging)

ahoefling:
The `DnnHttpControllerActivator` class is new in DNN 9.4 as before it was using the Microsoft default provided Activator. Here is a link to the [history](https://github.com/dnnsoftware/Dnn.Platform/commits/development/DNN%20Platform/DotNetNuke.Web/Api/DnnHttpControllerActivator.cs) for your reference:

DNN Platform/DotNetNuke.Web/Api/DnnHttpControllerActivator.cs
```diff
@@ -3,14 +3,21 @@
...
     {
         public IHttpController Create(HttpRequestMessage request, HttpControllerDescriptor controllerDescriptor, Type controllerType)
         {
-            return (IHttpController)Globals.DependencyProvider.GetService(controllerType);
+            // first try to just get it from the DI - if it's there
+            return (IHttpController) Globals.DependencyProvider.GetService(controllerType) ??
+                   // If it's not found (null), then it's probably a dynamically compiled type from a .cs file or similar
+                   // Such types are never registered in the DI catalog, as they may change on-the-fly.
+                   // In this case we must use ActivatorUtilities, which will create the object and if it expects 
+                   // any DI parameters, they will come from the DependencyInjection as should be best practice
```
ahoefling: @iJungleboy can you clarify this comment? 
- Does this mean the Controller will use the `IServiceProvider` to resolve any parameters? 
- Does this mean the controller is now added to the `IServiceProvider` and on subsequent calls the container will resolve it correctly?

mitchelsellers: Looking into this a bit more, I actually think the fix might be simplier, but it needs testing.

Per this documentation: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities?view=aspnetcore-3.0

The ActivatorUtilities will instantiate from the service provider, OR directly.  High-level theory, that this MIGHT actually do what we want internally, but would need to check.

mitchelsellers: Scratch that - ActivatorUtilities uses reflection and ServiceProvider is different

Activator Utilities - https://github.com/aspnet/DependencyInjection/blob/1.0.0-rc1/src/Microsoft.Extensions.DependencyInjection.Abstractions/ActivatorUtilities.cs
Service Provider - https://github.com/aspnet/DependencyInjection/blob/master/src/DI.Abstractions/ServiceProviderServiceExtensions.cs

iJungleboy: @ahoefling 

1. First of all it's really important that dynamic code works. These .cs files can change at runtime as people are modifying their apps, so IDnnStartup really won't work. It's like razor files, just for APIs. So that class is like a temporary DLL similar to a just compiled aspx file in the .net cache. 

2. This DI question is correct. `ActivatorUtilities` creates this dynamic class which was just compiled previously and will get recompiled on any changes. But it uses DI for any dependencies inside it - that's why I'm not using `Activator...` but `ActivatorUtilities` - very important. 

3. Such dynamically compiled files will never need to be injected into something - their purpose is the very top layer - just like a custom design which says what CSS to use, and needs DI, but won't be requested by other parts. They also don't have real namespaces or anything that's usefull - again just like compiled aspx files, just for webapi.

iJungleboy: also note that this change doesn't break anything - because if the DNN 9.4 way works, my additional code will never be hit. It's only a fallback in case the break in DNN 9.4 fails, + it re-activates the ability to use just-in-time compiled WebApis.

mitchelsellers: Actually @iJungleboy I don't believe that is the case.  If you call ActivatorUtilities without a default constructor or one that it can match dependencies on, it will throw an exception, whereby the existing code returns a null model.

We will need to validate this use case as well

iJungleboy: @mitchelsellers anything in a DI with a constructor that doesn't match dependencies will break. My code can't fix that. But without my code, that's also not fixed, so basically that case isn't resolved with this. But as of now I can't think of why someone would create a WebApi Controller with a constructor that needs dependencies which won't resolve.

iJungleboy: Just for clarification: With my change, previous dynamically compiled WebApi controllers work again. Here's the demo-controller that works again from the angular8 demo app
```c#
// In case you need to know more about WebApis, 
// visit the tutorials on https://2sxc.org/dnn-tutorials/
using DotNetNuke.Web.Api;
using ToSic.SexyContent.WebApi;
using System;
using System.Web.Http;

[AllowAnonymous]
[ValidateAntiForgeryToken]
public class SimpleController : SxcApiController
{
	[HttpGet]
	public string Hello(string name = null)
	{
		return name == null ? "Hello from the controller in web api" : "Hello " + name;
	}


	[HttpGet]
	public int[] Numbers()
	{
		return new int[] { 7, 12, 2972 };
	}


	[HttpGet]
	public dynamic Something(string name) 
	{
		return new {
			Name = name,
			Birthday = new DateTime(2012, 05, 12)
		};

	}

}
```
From https://github.com/2sic/app-template-angular8/blob/master/local/api/SimpleController.cs

Just a note: it's inheriting from the `SxcApiController`, which inherits from the default `DnnApiController`.

mitchelsellers: I guess I’m trying to figure out how/where you are injecting this today inside of DNN. As the route mapper etc is all from assemblies loaded in the /bin etc.

iJungleboy: @mitchelsellers that's fairly straightforward - just like DNN has a ControllerSelector, so does 2sxc. It automatically builds the .cs files and returns the type. Code is here: https://github.com/2sic/2sxc/blob/2513822f8a638439cda07cb5167fbc71610a3bc4/Sxc%20WebApi/AppApiControllerSelector.cs

iJungleboy: Basically these lines are the interesting ones: https://github.com/2sic/2sxc/blob/2513822f8a638439cda07cb5167fbc71610a3bc4/Sxc%20WebApi/AppApiControllerSelector.cs#L98-L107

iJungleboy: @ahoefling just noticed your second question. No, the dynamically compiled class is _not_ added to the catalog of types, so it's meant as an entry point but not for injecting into other things. 
That would actually be impossible, since the file can change at runtime and would need recompiling at certain times. It could also get removed at runtime - it's really a lot like a aspx, ascx or cshtml.

ahoefling: @iJungleboy - I hear your concerns about the new changes in 9.4.0+ and want to help, but I still politely disagree with this change.

@iJungleboy 
> ActivatorUtilities

Thanks for the clarification on the `ActivatorUtilities` I have not used that method before and it is really powerful.

@iJungleboy 
> also note that this change doesn't break anything

I understand that this change doesn't break anything and does support the generated code model that your module uses. Just because it doesn't break anything doesn't mean this is something that the platform should be supporting. It opens up many edge cases the platform would need to support.

@iJungleboy 
> AppApiControllerSelector

Thanks for providing the exact part of 2sxc where the ApiController is compiled and placed into memory. This is very helpful in understanding the problem.

@iJungleboy 
>  No, the dynamically compiled class is not added to the catalog of types, so it's meant as an entry point but not for injecting into other things.

If your module is creating a generated WebApi that goes outside the bounds of what DNN is doing, you need to register it via the `IDnnStartup`. **This will prevent the need for this change at all**.

To clarify on this point, you are registering the WebApi Controllers so DNN knows how to resolve them correctly. The way Dependency Injection works at the Controller level (yes, this includes .NET Core) is the controller is registered with the Container and when the Controller is instantiated it can then automatically pass in any dependencies into the Controllers Constructor with no need to manually use a helper method such as the `ActivatorUtilities.CreateInstance`.

I understand this is something that has broke your module , but AFAIK DNN doesn't officially support generated code in this manner. **This really isn't the same thing as a `.cshtml` or `.ascx` file, it is generating code and storing it in memory at runtime.**

If you want your module to continue to work with 9.4.0+ I would recommend adding some code to your `AppApiControllerSelector` to add the new controller to the container. Currently there is no reference to the  Container exposed so that will need further discussion with @mitchelsellers if that is something the platform could do. 

@mitchelsellers @iJungleboy 
If we decide to go ahead with this change I recommend we simplify this code to use 
```c#
public IHttpController Create(HttpRequestMessage request, HttpControllerDescriptor controllerDescriptor, Type controllerType)
{
    return ActivatorUtilities.GetServiceOrCreateInstance(Globals.DependencyProvider, controllerType);
}
```

Providing verbose comments are useful for our review but this would simplify the solution. See [Library Docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance?view=aspnetcore-3.0)

mitchelsellers:
Closing this request, as the change does not align with the patterns & practices recommendations for the long-term planning and growth of the DNN Platform.  All standard development practices are possible with the existing implementation.  For the specific impact to 2sxc, the vendor will need to work around this limitation that is in place for standards and security, a possible solution for this has been included earlier in the communication chain on this pull request.

@iJungleboy I'm emailing you directly on this as well.

PR #3059, opened 2019-10-02
Fix wrong non alphanumeric regex when validate password

### LINKED ISSUE #3058: Not able to use '_' (Underscore) while updating the password
## Description of bug
Changing the password with _ is not allowed when minRequiredNonalphanumericCharacters is set to "1"

## Steps to reproduce
1. Edit web.config. Set minRequiredNonalphanumericCharacters="1"
2. Create a new user.  
3. Change the password of the new user to password with underscore. 

## Current result
Change password failed even the password contains _ (underscore) which is non alphanumeric character and meets the length requirement. 

## Expected result
'_' underscore should consider as non alphanumeric. 

## Screenshots
If applicable, provide screenshots to help explain your problem.
image

## Error log

## Additional context
- The root cause of the issue is the regex expression to check non alphanumeric character is wrong. Glolabs::InvalidCharacters regex is  "[^A-Za-z0-9_-]" which consider both _ and - as alphanumeric. 
## Affected version

* [x] 10.0.0 nightly build
* [x] 9.4.1 nightly build

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

tingung:
Fixes #3058
## Summary

- Change the regex to match non alphanumeric character to [^A-Za-z0-9] . Alphanumeric are character from A to Z and 0 to 9 only. - _ are non alphanumeric characters.


valadas:
Does anyone see a reason whey - and _ would have been not accepted previously because of ActiveDirectory or other authentication providers or some such? I am approving the changes, just wondering why it was like that...

Issue #3122, opened 2019-10-07
Permissions are missing when copying modules between different localizations


mikebigun:
## Description of bug
When copying modules between different localization's it does not make permissions copied.  

## Steps to reproduce
1. Login as super user
2. Add multiple languages, mark all pages as translatable 
3. Add HTML Pro module to any page
4. Go to `Module Settings` -> `Permissions` -> add new role with a view permission (i.e. Community Manager role)
5. Go to current `Page Settings` -> `Localization Tab`
6. Mark `Copy Module` checkbox
7. Click on `Update Localization`

## Current result
Go to translated version of the page and go to module settings -> permissions. Role that was added for original version of module has not been copied to its translated version. 

## Expected result
Permissions should be also copied when copying modules between different localization's.

## Screenshots

![image](https://user-images.githubusercontent.com/22524011/66313199-81986c80-e901-11e9-86a4-19b2ffe90d57.png)

## Affected version

* [x] 10.0.0 nightly build
* [x] 9.4.1 nightly build
* [x] 9.4.0 latest supported release

mikebigun:
@mitchelsellers @valadas @bdukes 

Dear community, let me know please if you see any problems or risks to have permissions copied.

sleupold:
how do we handle permissions for the special "Translators" roles? A translator should b granted permission to his language only (unless granted by a differnt role or individually).

mikebigun:
Right, for the special "Translators" there is a host setting `DefaultTranslatorRoles-{culture}` with a list of roles which we want to add automatically when module is localized to a particular locale. By default we add `Administrators` and `TranslatorRole-{culture}`. 

But still it is not clear if I assigned i.e. `Moderators` and granted them view permissions for my original module. Then I copied module to another locale and found Moderators role is missing there.
So, is this expected behavior? Do we expect users to re-add missing roles manually to a newly copied modules? Or do we expect them to include roles to a `DefaultTranslatorRoles-{culture}` host setting to add them automatically on localization?

To me, this looks like we need to copy all roles from a source module, including `Moderators` (in my case). On top of that, "Translators" will be added.

sleupold:
I would expect, other permissions should be copied as well (except for any translators role) and respective translator role for new language being added explicitely.
Instead of the template, I would prefer the actual translator role being stored in PortalLanguages or PortalLocalization table (new column translatorRoleID int Null)

valadas:
Fixed by #3162

PR #3132, opened 2019-10-09
DNN-32978 Copying pro html module don't modify the permission

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/3122

Summary:
On copying modules between different locales, we also need to copy permissions that we granted for the original module. Having that we also need to make sure we are not explicitly adding inherited VIEW permissions from the tab and permissions for the source module translator.  Any other permission should be copied to the destination module to avoid users to compare and re-add missing roles. 

See demo: [DEMO](https://drive.google.com/file/d/1LDfsP0P0diRV-GJKLoe8Ax30kULWTIyv/view?usp=sharing)

### DIFF
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -296,9 +296,9 @@ namespace DotNetNuke.Entities.Modules
                 HostController.Instance.GetString(translatorSettingKey, null);
 
             var translatorRoles =
-                translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? Enumerable.Empty<string>();
+                translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
 
-            return translatorRoles.Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase));
+            return translatorRoles?.Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase)) == true;
         }
 
         /// <summary>
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -270,10 +270,8 @@ namespace DotNetNuke.Entities.Modules
             }
 
             var tabPermissions = TabPermissionController.GetTabPermissions(module.TabID, module.PortalID);
-
-            return tabPermissions
-                .ToList()
-                .Find(x => x.RoleID == permission.RoleID && x.PermissionKey == permissionViewKey) != null;
+            
+            return tabPermissions?.Where(x => x.RoleID == permission.RoleID && x.PermissionKey == permissionViewKey).Any() == true;
         }
 
         /// <summary>
@@ -300,8 +298,7 @@ namespace DotNetNuke.Entities.Modules
             var translatorRoles =
                 translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? Enumerable.Empty<string>();
 
-            return translatorRoles.Any() &&
-                    translatorRoles.Select(x => x.ToLower()).Contains(permission.RoleName.ToLower());
+            return translatorRoles.Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase));
         }
 
         /// <summary>
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -38,6 +38,7 @@ using DotNetNuke.Data;
 using DotNetNuke.Entities.Content;
 using DotNetNuke.Entities.Content.Common;
 using DotNetNuke.Entities.Content.Taxonomy;
+using DotNetNuke.Entities.Controllers;
 using DotNetNuke.Entities.Modules.Actions;
 using DotNetNuke.Entities.Modules.Definitions;
 using DotNetNuke.Entities.Portals;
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -250,6 +251,93 @@ private void CopyTabModuleSettingsInternal(ModuleInfo fromModule, ModuleInfo toM
...
+            if (!module.InheritViewPermissions || permission.PermissionKey != permissionViewKey)
+            {
+                return false;
+            }
+
+            var tabPermissions = TabPermissionController.GetTabPermissions(module.TabID, module.PortalID);
+
+            return tabPermissions
+                .ToList()
+                .Find(x => x.RoleID == permission.RoleID && x.PermissionKey == permissionViewKey) != null;
```
bdukes: Why does this need a `ToList`?  Couldn't we just use `Any()` to check the condition?

```
return tabPermissions.Any(p => p.RoleID == permission.RoleID && p.PermissionKey == permissionViewKey)
```

mikebigun: `tabPermissions` is a `TabPermissionCollection` 

See: https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Security/Permissions/TabPermissionCollection.cs 

It does not implement `Any()` method. However, it can be easily added.

bdukes: Gotcha.  How about this:
```
return tabPermissions.Where(p => p.RoleID == permission.RoleID && p.PermissionKey == permissionViewKey).Any();
```

DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -250,6 +251,93 @@ private void CopyTabModuleSettingsInternal(ModuleInfo fromModule, ModuleInfo toM
...
+
+            var translatorSettingValue = 
+                PortalController.GetPortalSetting(translatorSettingKey, portalId, null) ?? 
+                HostController.Instance.GetString(translatorSettingKey, null);
+
+            var translatorRoles =
+                translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? Enumerable.Empty<string>();
+
+            return translatorRoles.Any() &&
+                    translatorRoles.Select(x => x.ToLower()).Contains(permission.RoleName.ToLower());
```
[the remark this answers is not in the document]
mikebigun: Fixed.

DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -250,6 +251,90 @@ private void CopyTabModuleSettingsInternal(ModuleInfo fromModule, ModuleInfo toM
...
+            var translatorSettingKey = $"DefaultTranslatorRoles-{ culture }";
+
+            var translatorSettingValue = 
+                PortalController.GetPortalSetting(translatorSettingKey, portalId, null) ?? 
+                HostController.Instance.GetString(translatorSettingKey, null);
+
+            var translatorRoles =
+                translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? Enumerable.Empty<string>();
+
+            return translatorRoles.Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase));
```
Yabgu: You don't need to use "Enumerable.Empty()" below you can instead write all these like in the following

`return true == translatorRoles?.Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase));`

if it is null it will get resolved to `null == true` which will give the same result quickly. Or if you want you can do it in one-liner:

```
return true == translatorSettingValue?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
    .Any(r => r.Equals(permission.RoleName, StringComparison.OrdinalIgnoreCase));
```

You already did the very same thing on line 274 below

DNN Platform/Library/Entities/Modules/ModuleController.cs
```diff
@@ -250,6 +251,90 @@ private void CopyTabModuleSettingsInternal(ModuleInfo fromModule, ModuleInfo toM
...
+            var permissionViewKey = "VIEW";
+
+            if (!module.InheritViewPermissions || permission.PermissionKey != permissionViewKey)
+            {
+                return false;
+            }
+
+            var tabPermissions = TabPermissionController.GetTabPermissions(module.TabID, module.PortalID);
+            
+            return tabPermissions?.Where(x => x.RoleID == permission.RoleID && x.PermissionKey == permissionViewKey).Any() == true;
```
Yabgu: You can simplify this as in the following:

return tabPermissions?.Any(x => x.RoleID == permission.RoleID && x.PermissionKey == permissionViewKey) == true;

DNN Platform/Library/Entities/Modules/ModuleController.cs
mikebigun: I agree. But both variants are workable. Why suggested solution is better?

DNN Platform/Library/Entities/Modules/ModuleController.cs
mikebigun: No, we can't use `Any`. This is already mentioned above
https://github.com/dnnsoftware/Dnn.Platform/pull/3132#discussion_r333040908

DNN Platform/Library/Entities/Modules/ModuleController.cs
Yabgu: You are generating Enumerable.Empty<string>() just to work around for null value which is redundant

daguiler:
@dnnsoftware/tag Anyone against including this in 9.4.x?

bdukes:
I'm fine with including this in 9.4.x

mikebigun:
Team, you can close this one in favor of https://github.com/dnnsoftware/Dnn.Platform/pull/3162

PR #3137, opened 2019-10-10
Revision of build process part 1

donker:
This is the first in a series of revisions of the way we build DNN. It follows the merging on AE into the main repository. What this merge does is the following:

1. Move all content from ./Website to ./DNN Platform/Website and merge with what was there. The goal is to make it clear what is under source control and what is the website being built. ./DNN Platform/Website is source. ./Website is where the build process copies content to so it can start packaging.

2. Move the DNN packaging from MSBuild to Cake. Cake is better suited for this type of task and the old MSBuild tasks were very convoluted due to legacy code.


mitchelsellers:
@donker I'm not sure what the issue is here, but this showing lots of conflicts that restrict the merge?

.gitignore
```diff
@@ -229,34 +115,13 @@ DNN [Pp]latform/[Pp]roviders/*/[Bb]in/*
...
-Website/Portals/_default/Containers/Xcillion/
-DNN Platform/Dnn.AuthServices.Jwt/Package/
-Website/Portals/_default/Install/Dnn.CKEditorDefaultSettings.xml
-Website/Portals/_default/Install/Dnn.CKToolbarButtons.xml
-Website/Portals/_default/Install/Dnn.CKToolbarSets.xml
-Website/DesktopModules/Admin/SiteExportImport/App_LocalResources/ExportImport.resx
-/TestResult.xml
-/Website/DesktopModules/Admin/Dnn.EditBar
-/Website/DesktopModules/Admin/Dnn.PersonaBar
+D:\DNN\Dnn.AdminExperience\Extensions\Content\Dnn.PersonaBar.Extensions\admin\personaBar\Dnn.Roles\scripts\bundles\rw-widgets.svg
```
bdukes: This absolute path shouldn't be in the `.gitignore`, right?

Build/Cake/thirdparty.json
```diff
@@ -0,0 +1,50 @@
...
+    "destination": "Install/JavaScriptLibrary"
+  },
+  {
+    "name": "Microsoft.CodeDom.Providers.DotNetCompilerPlatform",
+    "folder": "DNN Platform/Components/Microsoft.CodeDom.Providers.DotNetCompilerPlatform/",
+    "destination": "Install/Library",
+    "version": "02.00.01"
+  },
+  {
+    "name": "Microsoft.CodeDom.Providers.DotNetCompilerPlatform_net64",
```
bdukes:
```suggestion
    "name": "Microsoft.CodeDom.Providers.DotNetCompilerPlatform_net46",
```

bdukes:
I've rebased on `release/9.4.x`, no more merge conflicts

.gitignore
donker: True ... I'll repair

donker: But if it's OK, I'll do that after merge as otherwise I have to do a lot of resolving locally given I'm out of sync.

SolutionInfo.cs
```diff
@@ -34,3 +34,6 @@
 [assembly: AssemblyProduct("http://www.dnnsoftware.com")]
 [assembly: AssemblyCopyright("DotNetNuke is copyright 2002-2018 by DNN Corporation. All Rights Reserved.")]
 [assembly: AssemblyTrademark("DNN")]
+[assembly: AssemblyVersion("9.4.2.0")]
+[assembly: AssemblyFileVersion("9.4.2.5782")]
+[assembly: AssemblyInformationalVersion("9.4.2-unstable.5782+Branch.reorg1.Sha.b37c5e45ed0dde233bee7c8572d70940e5bda93e")]
```
bdukes: Is this supposed to be added in here?  Or when the manifests are reverted, should this be reverted as well?

.gitignore
bdukes: I went ahead and pushed that change

SolutionInfo.cs
donker: Good catch. I'll repair after merge in a fresh PR.

donker:
> 
> 
> Ok, so from a read of the changes, I have nothing to say, this is a major improvement!
> 
> However I tried to install a clean site and the Persona Bar does not load because of this missing file
> image
> 
> I would like that we get to a state that works before merging. I need to catch a few zzz, but I will ping you up in the afternoon...

I do not get the same result here. Strange. OK, I guess we're going to have to wait on this for a few weeks as I'm about to leave. Pity.

valadas:
I will try to give it a more detailed look in a few hours, sorry.

david-poindexter:
I have tested this locally and am seeing the same as @valadas

valadas:
Ok, one difference here needs to be fixed:
image
This should be named .resources

valadas:
image

Here is one finding, the resources.zip on Dnn.PersonaBar.Extensions is missing the output from the yarn build and has other files unrelated, maybe this is getting zipped from the wrong source path or something?

donker:
> Here is one finding, the resources.zip on Dnn.PersonaBar.Extensions is missing the output from the yarn build and has other files unrelated, maybe this is getting zipped from the wrong source path or something?

I can't replicate that, Daniel. Here it builds this resource zip correctly from Yarn. Can you check your build output?

donker:
> 
> 
> Ok, one difference here needs to be fixed:
> image
> This should be named .resources

I just fixed this.

valadas:
@donker well, I see the yarn build running but then the packaging does not have the build result, I am thinking maybe the build script has the wrong source folder. David had the same results as me. Can you try re-cloning into a brand new folder and building to see, maybe you have some ignored files from past builds that makes it work for you but not for us...

I am still a bit swamped today but if you want to take a look at it together I should have a bit more time starting tomorrow... Just ping me up.

donker:
It is probably due to the various paths in the two MS build solutions running in parallel. It's a royal clusterf. We could try to solve this as well by collapsing those two into a single one. But it requires quite a bit of work to figure out all the dependencies. I just gave it a shot but no cigar yet. 

I propose to move AE's module.build to a file called AEModule.build and similar with Package.targets. These are the two files that are different from the main files and specifically tweaked for the AE.

valadas:
Cool, well ping me up if you want an extra pair of eyes, I got it working before this PR by referencing the main build script for everything but having 1 override for the AE Extensions project which was a bit different.

bdukes:
I can confirm that a clean install results in `/Install/Module/Dnn.PersonaBar.UI_09.02.02.Install.zip` with a `Resources.zip` which is missing its `/scripts/exports` folder.  When re-running the build a second time, that folder is in place.

It seems like there may be timing issues around when things are built/packaged/cleaned.

valadas:
Ok, so just updated with the changes that fixed previous build issue and it did not solve this one, it looks like we have missing stuff in PersonaBar.UI and EditBar.UI packages for some reason.

donker:
I went ahead and merged the build scripts from DNN and AE into one folder. Hopefully this resolves the quirks you were seeing.

david-poindexter:
I am still getting the same errors as before when testing the install package.

david-poindexter:
Okay, first build results in above-documented errors post-install load of the site.  Second build works with no errors post-install load of site (Persona Bar shows fine and no console errors).

valadas:
Ok, so I am having different results on first build and on next builds.

The first build is missing exports folder for PersonaBar.UI

image

If you rebuild a second time, then they are there. So I think there might be just an order of operations issue or a parallelism issue ?

bdukes:
The new build is still missing `/Install/Module/Dnn.PersonaBar.UI.09.04.02_Install.zip/Resources.zip/scripts/exports/export-bundle.js` (otherwise, it appears to be complete).

bdukes:
That `Resources.zip` is created on [line 7113](https://dev.azure.com/dotnet/DNN/_build/results?buildId=23724&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=7113) of the build output, and the `lerna build` command which creates the `export-bundle` finishes on [line 8998](https://dev.azure.com/dotnet/DNN/_build/results?buildId=23724&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=8998)

The `Resources.zip` is created in the `Minification` task:
https://github.com/dnnsoftware/Dnn.Platform/blob/087ee3678ff56d136bafa0839bfb0a8385204ead/Build/BuildScripts/AEPackage.targets#L49-L52

The `Minification` task is run by the `Package` task:
https://github.com/dnnsoftware/Dnn.Platform/blob/087ee3678ff56d136bafa0839bfb0a8385204ead/Build/BuildScripts/AEPackage.targets#L13-L25

The `Package` task is in a list of tasks to run after build:
https://github.com/dnnsoftware/Dnn.Platform/blob/087ee3678ff56d136bafa0839bfb0a8385204ead/Build/BuildScripts/AEModule.build#L11

According to [MSBuild docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/target-element-msbuild?view=vs-2019#remarks):
>The first target to execute is specified at run time. Targets can have dependencies on other targets. For example, a target for deployment depends on a target for compilation. The MSBuild engine executes dependencies in the order in which they appear in the `DependsOnTargets` attribute, from left to right.

Therefore, we'd expect `RunYarn` to execute before `Package`, but it's not.  So, maybe the docs are wrong, or maybe something is getting overwritten and the source of one of these instructions is actually coming from somewhere else.

Does anyone know of a good way to trace what MSBuild thinks it's doing?

donker:
> 
> 
> The [new build's artifacts](https://dev.azure.com/dotnet/DNN/_build/results?buildId=23724&view=artifacts) are still missing `/Install/Module/Dnn.PersonaBar.UI.09.04.02_Install.zip/Resources.zip/scripts/exports/export-bundle.js` (otherwise, it appears to be complete).

Oddly enough I can't replicate this. I just recloned and just ran the build script.

bdukes:
Definitely weird.  You're just getting a fresh clone and running `./build.ps1`, and nothing more?

bdukes:
It seems like it could be a timing issue, but it's weird that it seems consistent on DevOps and my machine.  Are you using VS 2019 or 2017?  I wonder if that might be a contributing factor.  I'm using 2019 (and I asssume DevOps is also).

donker:
> 
> 
> Definitely weird. You're just getting a fresh clone and running `./build.ps1`, and nothing more?

Completely fresh clone. Then:

.\Build.ps1 -Target BuildAll

So that should be equivalent, yes. I have both 2017 and 2019 installed on my machine. But build is called from the scripts so cannot see how that is a factor.

valadas:
Yes and then the next builds are fine. It only happens on the first build. Same happens in CI.

donker:
So in answer to @bdukes : I compared my successful build log with the one in DevOps and in my case the zip was created on line 8587 and the export-bundle on 8331. So it seems they are indeed reversed.

But ... but. On a rebuild it's reversed. Hmm.

bdukes:
Alright, I think I've tracked this down.  Not quite sure what the right answer is.

While the `Package` task depends on the `RunYarn` task, `RunYarn` has a condition `Condition="$(YarnWorkingDirectory.Length) > 0 AND '$(Configuration)|$(Platform)' == 'Release|AnyCPU'"`.  The variable `YarnWorkingDirectory` is only set in [`Dnn.PersonaBar.Extensions/Module.build`](https://github.com/dnnsoftware/Dnn.Platform/blob/087ee3678ff56d136bafa0839bfb0a8385204ead/Dnn.AdminExperience/Extensions/Content/Dnn.PersonaBar.Extensions/Module.build#L19).  So, there is nothing indicating that `yarn build` needs to run before the packaging of `Dnn.PersonaBar.UI`.

The easiest solution is probably to move the [Yarn commands](https://github.com/dnnsoftware/Dnn.Platform/blob/087ee3678ff56d136bafa0839bfb0a8385204ead/Build/BuildScripts/AEModule.build#L48-L49) into Cake, and make sure they're called before `CompileSource`.  Another (unsatisfying) solution would be to run `yarn build` for both projects.  There's probably a way within MSBuild to get these two pieces to connect, but I don't know what it is.

donker:
Together with Daniel we worked at this and added the PersonaBar.Extensions project as dependency for the PersonaBar.UI project.

bdukes:
Conflict has been addressed, once it's done building let's go ahead and approve this PR with a squash (I'm heading out of town, so don't wait for me to push the squash button)

david-poindexter:
Thanks @bdukes !

And thanks @donker and @valadas for working on this.  It is great!!!

PR #1999, opened 2018-01-31
DNN-7818: add send verification mail link in unverified message.

zyhfish:

PR #3144, opened 2019-10-14
DNN-7818: add send verification mail link in unverified message.

zyhfish:
new PR on behalf of #1999 to set target to 9.4.x branch


thabaum:
Thank you @zyhfish

PR #3188, opened 2019-10-22
Updates development with latest changes from 9.4.x

valadas:
This is a release management task and we are self approving it for that reason


valadas:
Too many conflicts, will resubmit

PR #3191, opened 2019-10-22
Merge 9.4.x to development

bdukes:
@mitchelsellers used a squash on #3189, but we need a merge commit to connect the commits into both branches.  This PR includes a merge commit (and still needs to be approved/integrated via merging), and also happens to fix a merge conflict we missed in `.github/ISSUE_TEMPLATE/bug-report.md`


mitchelsellers:
My Bad!

mitchelsellers:
@bdukes You ok with me "Merge Committing" this as an admin function to fix this

bdukes:
yup! thanks, @mitchelsellers

valadas:
Do I need to do something or is this good to go now?

valadas:
@bdukes in general we should not use "Squash and merge", only if the commits are really too granular for a single thing we would use that right?

bdukes:
In general, squash or rebase are both fine (and, yes, the difference between them would be how important the individual commits are).  But when we're copying commits from one long-lived branch to another long-lived branch, that's when a merge commit is important.

Issue #3186, opened 2019-10-21
Possible memory leak (dependency injection)?


allanedk:
## Description of bug
I'm sorry if this is not the right place to post this. But I suspect that Dependency Injection implemented in DNN 9.4 is causing memory leaks and if so it might be important for others too.

After upgrading from DNN 8.0.4 to 9.4.1 the memory consumption of my website (the w3wp process) is increasing indefinitely until all server resources is consumed. The site is **not** slowing down or affected in any way though, i.e. no errors in the admin logs.

The site is not using any 3rd part modules. But only "thin" frontend modules pulling data from custom build web api services (a project with controllers that inherits from DnnApiController).

No source code was changed when upgrading other than changing target .NET Framework version of the project to 4.7.2. No compiler errors and no errors in the DNN log.

I've been "debugging" in production by deactivating the use of different parts of the service, but without luck. 

When analyzing a memory dump of the IIS w3wp process (using dotMemory from Jetbrains and DebugDiag from MS) I see the the garbage collector heap usage (generation 2) is quite large with about 2/3 of the total memory usage. Also there are a lot of "objects ready for finalization". (see [https://ibb.co/Yt1GVkG](https://ibb.co/Yt1GVkG) )

And at the top of memory usage list is _Microsoft.Extensions.DependencyInjection.ServiceProvider_ (see [https://ibb.co/vjK5rsc](https://ibb.co/vjK5rsc) ).

I've run out of ideas. If this is caused by Dependency Injection, then others might experience the same. Perhaps without knowing it yet. I have a lot of traffic on the site (20-25k user session per day) and the memory consumption hits 2 GB in about 6 hours at which point the app pool recycles.

Is it possible for this to be related to DI, perhaps in combination with using DnnApiController (my code was not changed during upgrade)? Could DI prevent the garbage collector from eliminating dead objects?

If needed I can add source code examples (even though no exceptions are thrown) or more info from the memory profiler.

Thanks a lot.

## Steps to reproduce
I have not been able to reproduce this locally. Perhaps a different stress testing tool is needed. Or maybe it's related to the production server environment (Windows Server 2008 R2 + IIS 7.5)

## Current result
Ever increasing memory consumption.

## Expected result
Memory usage should be stable. It was stable at 600 MB in DNN8 using the same code.

## Screenshots
This is screenshots of some of the memory profiler reports:
[https://ibb.co/Yt1GVkG](https://ibb.co/Yt1GVkG)
[...]

PR #3192, opened 2019-10-23
Fixed Memory Leak in Web API Modules

ahoefling:
Fixes: #3186

## Summary
Fixes the significant Memory Leak discussed in #3186 in Web API Modules. I duplicated the test I performed in https://github.com/dnnsoftware/Dnn.Platform/issues/3186#issuecomment-544974596

### Test Result
* 506,787 requests in 360 seconds: 257.6 MB of Memory
image

This is a marked improvement from 11,553.2 MB of Memory

### Description of Changes
I removed the `DnnHttpControllerActivator` which is not the recommended approach for Dependency Injection in Web API 2 by Microsoft. This was an oversight when Dependency Injection was originally implemented in v9.4.0.

I created a `DnnDependencyResolver` which handles communication with the `Globals.DependencyProvider` and applies the implementation of `IDependencyResolver` to the Web API 2 Configuration. This allows the controller life-cycle to be handled by ASP.NET Web API 2

If you want more information on this, check out the [Docs](https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/dependency-injection) from Microsoft


david-poindexter:
Fantastic @ahoefling!

mitchelsellers:
\azp run

ahoefling:
Some unit tests failed that I need to fix, if you want to test this against the happy path please clone the repo and run the create install command
```
./build.ps1 -Target CreateInstall
```

valadas:
Awesome @ahoefling I have a test bed here ready to upgrade and compare, I created a test that calls 5 endpoints 20 times with different parameters in a module that uses the Dependency Injection that was added in 9.4.0 and also Entity Framework 6.

With Dnn 9.4.1, I rand that test for 5 minutes and memory usage kept growing up to 506MB after 5 minutes. I let the site cool down for 15 minutes and that value never went down significantly from there.

With the build from this PR and the same test, it stopped growing at 217MB, then I let the site cool down and it went back to about 170MB.

Also I see nothing broken on the site so far.

So this looks good to me.

valadas:
Just the build is failing, I think because of a test, other than that, looks good to me.

ahoefling:
I fixed the failing unit tests and updated the `IServiceProvider` mock to build a full Service Provider. This is much easier than trying to mock it out. Doing so allowed me to add more unit tests to the new `DnnDependencyResolver` so we now have 100% code coverage over the new class that this PR adds 🥂 .

After the successful PR build I grabbed the installer and ran the same test provided in the issue.

Memory prior to the load test
image

Memory after the load test
image

Interestingly enough I had less memory allocation in the IIS Worker Process than before. It appears this does solve the memory leak

bdukes:
🎉 Thanks @ahoefling!

PR #3204, opened 2019-10-26
Move email above username field.

thabaum:
## Summary
Email address is at the bottom below password.  when you have to enter in the captcha code you have to scroll up to get to enter password again.  would be nice not to have to re-enter passwords as this is an issue as well but at least the email address is below the username field with this pull request if accepted.

image

NOTE:  I wish I could move "Display Name"  below User Name but this is where the bio registration fields go so no way to really move this easily.

If you have Email as Username set the email address is at the bottom making a user enter passwords first and I can't find any other website that does this.  However email address under Username is more industry standard I believe for a registration form.

### DIFF
--- a/Website/controls/user.ascx
+++ b/Website/controls/user.ascx
@@ -3,9 +3,9 @@
 <%@ Register TagPrefix="dnn" TagName="label" Src="~/controls/labelControl.ascx" %>
 <table cellSpacing="0" cellPadding="1" border="0" summary="User Design Table">
 	<tr>
-		<td class="SubHead" width="175">
+	    <td class="SubHead" width="175">
 			<dnn:label id="plFirstName" runat="server" controlname="txtFirstName" text="First Name:"></dnn:label></td>
-		<td class="NormalBold" noWrap>
+	    <td class="NormalBold" noWrap>
 			<asp:textbox id="txtFirstName" tabIndex="1" runat="server" cssclass="NormalTextBox" size="25"
 				maxlength="50"></asp:textbox>&nbsp;*
 			<asp:requiredfieldvalidator id="valFirstName" runat="server" cssclass="NormalRed" display="Dynamic" errormessage="<br>First Name Is Required."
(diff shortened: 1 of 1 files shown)


sleupold:
IMO it would be best, if the order is confgurable by admins without limit

thabaum:
image

I didnt change "Username" position.  I changed Email Address

thabaum:
> 
> 
> IMO it would be best, if the order is confgurable by admins without limit

Yes, I totally agree and it looks like it would take breaking the BIO fields down as well to allow adding/removing those and putting them in an order that works best for the website purpose.

thabaum:
https://www.template.net/web-templates/htmlcss-templates/signup-form/

Here are 20+ HTML5 examples that I believe would be nice to get the login to look like one of these two IMO

image

image

thabaum:
I can make this change.  I only seen one other form that worked like this, but really it is to satisfy keeping the email and username above the password so I really do not care which way it goes.  I was just looking at the way most did it but I do like email on top as that was my preference originally.

Website/controls/user.ascx
```diff
@@ -30,10 +41,10 @@
 			<asp:label id="lblUsername" runat="server"></asp:label>
 			<asp:requiredfieldvalidator id="valUsername" runat="server" cssclass="NormalRed" display="Dynamic" errormessage="<br>Username Is Required."
 				controltovalidate="txtUsername" resourcekey="valUsername"></asp:requiredfieldvalidator></td>
-	</tr>
+
```
valadas: I think there is a missing </tr> here

valadas: arrg, stripped it...
I thinkg there is a missing `</tr>` here

thabaum: odd thanks for catching that dont understand how anything could of been missing.  It was pretty straight forward operation :)

thabaum:
> 
> 
> IMO it would be best, if the order is confgurable by admins without limit

To discuss this a little further can anyone give me a heads up on what to expect to make this happen if you coded it and I will give it a shot... it would be a good way to get my feet a little wetter here and build some confidence.  I figure a few things need to be thought about which is the db schema, sp's and admin ui plus back end code.  I am still getting used to github and the new tools along with the latest best practices and future but I am ready to fly soon I believe.  I should be able to do this in github overtime with help :)

Website/controls/user.ascx
```diff
@@ -20,6 +20,17 @@
 			<asp:requiredfieldvalidator id="valLastName" runat="server" cssclass="NormalRed" display="Dynamic" errormessage="<br>Last Name Is Required."
 				controltovalidate="txtLastName" resourcekey="valLastName"></asp:requiredfieldvalidator></td>
 	</tr>
+	</tr>
```
bdukes:
```suggestion
```

valadas:
Thanks @thabaum , nothing major here, but next time, we like to have an issue before a pull request and a link back to it in the issue comments (like closes #8948484) (using a big number here to not actually link to a real issue... And in a perfect world it is best to create a branch specific to your change instead of doing it in the development branch because if you do other work before we merge this, you have a better separation. But nothing huge, we are approving this. Thanks for your efforts

thabaum:
Thank you for the pointers @valadas @bdukes @david-poindexter @mitchelsellers and everyone helping me get my first PR through GitHub and DNN CE.  I had to figure out to re-request reviews and it seemed like I had to hit a refresh icon for david to re-review which I spent time looking for.  

I am also trying to understand the branches better so I can keep fixes and issues separate.  I will play with this more takes time which I am short on but really wanting to dig in here and help.  I will watch your video tutorials again refresh things as I want to +1 this PR soon hopefully.  Thanks again for accepting my first PR! 👍

thabaum:
> nothing major here, but next time, we like to have an issue before a pull request and a link back to it in the issue comments

https://github.com/isaacs/github/issues/1125

this current issue in GitHub could use some support... this is what I am trying to figure out and find easily how to create an issue and then create a branch.  And I wish to understand what is done on my fork as far as development version and how everyone gets to play with it.

PR #3230, opened 2019-11-02
Reduction of build warnings

### LINKED ISSUE #2648: Resolve Build Warnings on DNN.Platform Solution

eng-aomar:
[warnings_soultions.xlsx](https://github.com/dnnsoftware/Dnn.Platform/files/3800834/warnings_soultions.xlsx)

## Summary

Fixes  #2648 
Please note that the warning solutions are listed in the excel file attached in detail.
(https://github.com/dnnsoftware/Dnn.Platform/files/3800837/warnings_soultions.xlsx)

Only one warning is left by now :

	"C:\Dnn.Platform\packages\Yarn.MSBuild.1.16.0\build\Yarn.MSBuild.targets" cannot be imported again. It was already imported at "C:\Dnn.Platform\Build\BuildScripts\Module.Build (2,4)". This is most likely a build authoring error. This subsequent import will be ignored.	Dnn.PersonaBar.Extensions	11	C:\Dnn.Platform\Dnn.AdminExperience\Extensions\Content\Dnn.PersonaBar.Extensions	C:\Dnn.Platform\Dnn.AdminExperience\Extensions\Content\Dnn.PersonaBar.Extensions\Dnn.PersonaBar.Extensions.csproj	743


valadas:
Also, if this is no breaking change and you can target the release/9.4.x branch, it could be merged earlier, if not, this will only get into Dnn 10. We are planning on changing the branching strategy to make this easier on contributors but unfortunately cannot make that happen before Dnn10.

eng-aomar:
> The manifest versions need to stay blank because they are dynamically set by GitVersion, can you undo that change please.

Do you mean removing the version number 10.00.00 from version tag, replacing it with version="" instead of version="10.00.00" .
image

valadas:
It may works but I am not 100% sure, to be save I would remove the whole version="" thing on those files

DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
```diff
@@ -33,6 +33,8 @@ public class ModuleControlPipeline
     {
         private static readonly ILog TraceLogger = LoggerSource.Instance.GetLogger("DNN.Trace");
         private Dictionary<string, IModuleControlFactory> _controlFactories;
+
+        [Obsolete]
```
valadas: Also, this should not be marked obsolete, what is the reason for this change?

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcHandler.cs
```diff
@@ -57,7 +57,9 @@ bool IHttpHandler.IsReusable
 
         void IHttpHandler.ProcessRequest(HttpContext httpContext)
         {
+#pragma warning disable CS0612 // Type or member is obsolete
```
valadas: I would not override this warning, we should replace it with the new way or remove it or leave the warning in if not possible.

valadas:
I had some time to look a bit more into this pull request and there are too many changes that do not make sense... A lot of methods where marked deprecated without discussing deprecating them and without a deprecation message, also this resolves most warning by just disabling them, we need to instead switch usage to the new way of doing things and not just silence those warnings. I am sorry be we cannot accept this pull request the way it is

eng-aomar:
> I had some time to look a bit more into this pull request and there are too many changes that do not make sense... A lot of methods where marked deprecated without discussing deprecating them and without a deprecation message, also this resolves most warning by just disabling them, we need to instead switch usage to the new way of doing things and not just silence those warnings. I am sorry be we cannot accept this pull request the way it is

Thank you for your time reviewing this pull request.
I am managing to start from scratch with a different approach for solving the warnings, I will categorize the warnings according to their types, and make smaller pull requests, so we can reduce the number of warnings at least.

valadas:
Awesome, thanks

PR #3236, opened 2019-11-04
Further reorganization of build process

donker:
This PR will do the following:

1. Move the node stuff from the AE folder to the root folder. This allows us to leverage node in the future in pther parts of the project without creating confusion about its location and avoid creation of multiple node_modules folders.
2. New build script to create a local website using the "ResetDevSite" target in Cake
3. Add local settings for the above
4. Reorganize all MSBuild scripts to use the same variables (in Dnn_Platform.build) and the ability to override the value of the website path there using a local file. This allows debug building to the debug website.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "git.ignoreLimitWarning": true
```
bdukes: Are you intending to check in the `.vscode` folder?  Just wanting to make sure this was intentional, no big objection.

donker: Unintentional. I'll add to gitignore

bdukes:
I did a build of this branch and `release/9.4.x` and compared them.  The only difference that sticks out is that the `reorg2` branch includes an extra package in the `Install/Library` folder, `DNNCE_Website.Deprecated_09.04.02_install.zip`.  Does anyone know anything about that package.

It also appears that the `/Install/Module/DNNCE_01_Web.Deporecated_09.02.04_Install.zip` package was moved to `/Install/Library/DNNCE_Web.Deporecated_09.02.04_Install.zip`, but I wouldn't expect that to cause an issue.

bdukes:
After installing, the `reorg2` site has `DotNetNuke.Website.Deprecated.dll` in its `bin` and `release/9.4.x` did not have that DLL in the `bin`.  So it looks like this either fixed an oversight, or missed a special case exemption.

Also, `reorg2` includes `DotNetNuke.DependencyInjection.deps.json` in the `bin`, which `release/9.4.x` doesn't have (but they _both_ have `DotNetNuke.Abstractions.deps.json`).  These JSON files being included in the `bin` isn't going to cause a problem, but it also doesn't do anything AFAIK, and should probably be excluded.

donker:
> 
> 
> I did a build of this branch and `release/9.4.x` and compared them. The only difference that sticks out is that the `reorg2` branch includes an extra package in the `Install/Library` folder, `DNNCE_Website.Deprecated_09.04.02_install.zip`. Does anyone know anything about that package?
> 
> It also appears that the `/Install/Module/DNNCE_01_Web.Deporecated_09.02.04_Install.zip` package was moved to `/Install/Library/DNNCE_Web.Deporecated_09.02.04_Install.zip`, but I wouldn't expect that to cause an issue.

It used to be in "Modules" despite being a library because the logic that ensures a proper install order was faulty. I fixed this bug and moved this package back to the Library folder where it belongs.

valadas:
As for the DNNCE_Website.Deprecated_09.04.02_install.zip package, if I am not mistaken this was code to be removed but still needed in order to be able to migrate from the control panel to the persona bar.  This was done in order to still have access to the old codebehind logic on the older webforms based controls because the upgrade removes those files but the upgrade logic still needed some methods from there to perform it's thing.

If I am not mistaken this can be removed in any version as long as people follow the suggested upgrade path that has a step at 9.1.1 before moving forward to the latest version. This being said I don't know if we should keep it there "in case" for people not reading the suggested upgrade path and would try upgrading directly from 8.0.4 (or older) directly to 9.4.x or some such.

See https://dnntracker.atlassian.net/browse/DNN-9552 also...

I think we could keep this in the upgrade package only and get rid of it in 10 ??? Opinions? We could also include a cleanup in 10 to remove it ?

mitchelsellers:
Deprecated is 100% required currently

bdukes:
@mitchelsellers `Deprecated.Web` is currently included, not issue there.  `Deprecated.Website` is not currently getting installed (just checked it's not present in 9.3.2), so it can't be 100% required.

mitchelsellers:
Weird, I was 99.9% sure that was required for the FileManager, I'll have to look deeper in my notes.  

I did an NDepend Chart on all of this a while back

thabaum:
For things that have been removed or need changed or fixed up due to changes is there a place anywhere where we can look at these.   It would also be nice to categorize if it is a developer change needed or an administrative change.  Just a thought to have this list possibly a file in the upgrade package or a referenced webpage for things to double check have been done.  It is hard to read release notes to find these all the time for every single release it would be convenient to get these all in one location.  dnndocs maybe...

Issue #2140, opened 2018-06-29
Region Dropdown doesn't appear when registration page loads and autofills country field


yog-it:
## Description 
When the site is set to "Require a Valid Profile for Registration" on the registration settings tab and the Region and Country fields are set to "required" on the user profiles tab. The user can not register when the Country is auto-filled by the IP lookup mechanism because the Region dropdown never appears. This doesn't occur on the localhost.

## Steps to reproduce
1. Install a new DNN site on a server or app service (this doesn't occur on localhost).
2. Set the Country and Region fields to required in the Site Settings > Site Behavior > User Profiles tab.
3. Turn on "Require a Valid Profile for Registration" on the Security > Members Accounts > Registration Settings tab.
4. Logout and attempt to register. 

## Current result
Using a PC on a US IP address (only country IP I've tested), the user can't register because the State/Region field never fills with a dropdown when United State is auto entered by the IP lookup. The region field will continue to blank out every time the form is submitted.

## Expected result
The dropdown for the region appears, is populated and the user can submit the form.  

## Affected version

* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0
all 8 versions
Does not effect upgraded sites from earlier versions if the Country field remains DataType of DotNetNuke.UI.WebControls.DNNCountryEditControl instead of DotNetNuke.UI.WebControls.DnnCountryAutocompleteControl from a new installation.
  
## Affected browser

* [ x ] Chrome
* [ x ] Firefox
* [ x ] Safari
* [ x ] Internet Explorer
* [ x ] Edge

RFHavens:
This is one of our biggest issues as we are a membership based use case for our club. I get nothing but complaints for this issue. Even clearly writing instructions to "make it work" most do not read the instructions I have put up. This should be a top issue that should be able to be fixed in a timely manner. Of course a custom solution could be built but that would also require the developer to maintain it thru future updates to the core DNN code.

valadas:
This should be fixed in 9.3.0, basically two issues where fixed.

1. Some browsers autocomplete hid the built-in drop-down this not firing up the selection.
2. For the region list to populate, it needs to know the country first. On new installations the order is country then region. But to not disturb existing sites, if you upgrade then you should only need to move the country before the region.

If I remember correctly, the is also still an open issue about that order not getting applied everywhere, that I hope to get to for 9.3.1 or 9.3.2

yog-it:
[...]

PR #3244, opened 2019-11-05
Move Country Above Region in UserProfile.cs

thabaum:
Moved country above region to help resolve any issues from trying to set the region prior to setting the country might be concerned.

https://github.com/dnnsoftware/Dnn.Platform/issues/2140 I can reference this issue but I will not say it will fix it.  But it makes more sense to have it show up here in this file so I propose the change.

### DIFF
--- a/DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
+++ b/DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
@@ -63,7 +63,7 @@ namespace DotNetNuke.Entities.Users
         public const string USERPROFILE_City = "City";
         public const string USERPROFILE_Country = "Country";
         public const string USERPROFILE_Region = "Region";
-	public const string USERPROFILE_PostalCode = "PostalCode";
+		public const string USERPROFILE_PostalCode = "PostalCode";
 
         //Phone contact
         public const string USERPROFILE_Telephone = "Telephone";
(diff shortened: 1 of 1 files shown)


thabaum:
alright no additional commits!

thabaum:
I was not changing the order in the UI, just in the code to match the UI and make the code get the country list set first I was hoping to help cure any issues regarding loading the UI automatically.

thabaum:
@valadas I thought you had changed this already for newer installations?

valadas:
> I was not changing the order in the UI, just in the code to match the UI

Oh I am sorry, I misunderstood this PR then, reopening...

DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
```diff
@@ -61,9 +61,9 @@ public class UserProfile : IIndexable
         public const string USERPROFILE_Unit = "Unit";
         public const string USERPROFILE_Street = "Street";
         public const string USERPROFILE_City = "City";
-        public const string USERPROFILE_Region = "Region";
         public const string USERPROFILE_Country = "Country";
-        public const string USERPROFILE_PostalCode = "PostalCode";
+        public const string USERPROFILE_Region = "Region";
+	public const string USERPROFILE_PostalCode = "PostalCode";
```
valadas: Can you just adjust the spacing here, other than that, I am fine with this. For other reviewers this does not change behaviour just the order of declarations so it matches in code what we have in the UI for most situations.

thabaum:
Does it look right now?  When I saved it in GitHub it was moving all over the place but I think it settled in the right place this last time.

valadas:
> Does it look right now? When I saved it in GitHub it was moving all over the place but I think it settled in the right place this last time.

Hmm, looks like in 5dd0331 it got pushed more than necessary, then in 2fb8727 it got back too much left.

Not a huge deal and we are planning to add an editor config to avoid those inconsistencies anyway, just a remark.

valadas:
Might also be just a github UI glitch who knows, I will checkout the actual file after 9.4.2 is out...

Issue #3250, opened 2019-11-06
[Site Assets] Select All checkbox is not working


mikebigun:
## Description of bug
Select All checkbox on toolbar is not working properly. When user clicks on `Select All` and then navigates to different folder, checkbox remains checked, but no one item is selected. 

## Steps to reproduce
List the steps to reproduce the behavior:
1. Login as admin
2. Go to Persona Bar -> Site Assets
3. Navigate to `Root` folder
4. Click on `Select All`
5. Navigate to `Images` folder

## Current result
When user navigates to different folder, checkbox remains checked. You can observer that clicking on this checkbox it does not reflect its state correctly. 

## Expected result
`Select All` checkbox should work properly. It should reflect its state correctly. When `Select All` is checked, all items should be selected, otherwise deselected. 

## Screenshots

![image](https://user-images.githubusercontent.com/22524011/68304046-a96c2300-00ad-11ea-8a1b-7d2444bcaac1.png)

![image](https://user-images.githubusercontent.com/22524011/68304096-c3a60100-00ad-11ea-8f5d-6f8e66ef5df3.png)

## Affected version

* [x] 10.0.0 alpha build
* [x] 9.4.2 alpha build
* [x] 9.4.1 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

valadas:
I am not able to reproduce this in 9.4.2 RC1, is there anything else special in order to reproduce this issue? Does it happen with the default theme? Do you have any browser console errors?

valadas:
Just tried it on 9.4.0 and 9.4.1 too and not able to reproduce

thabaum:
I could not reproduce either
![image](https://user-images.githubusercontent.com/13577556/68537359-159d8f80-0317-11ea-810b-3479efd1842c.png)
![image](https://user-images.githubusercontent.com/13577556/68537365-24844200-0317-11ea-924d-186b1ba6bee5.png)

I do notice your view is different then stock.  You have large thumbnails when I have detailed view going.

thabaum:
reproduced
![image](https://user-images.githubusercontent.com/13577556/68537384-7331dc00-0317-11ea-8c38-56d210390081.png)

thabaum:
reproduced.  

Icon View is what sets off the bug.  I just had to click on images, select all images and click on the images folder again it would refresh looking like this:

[...]

PR #3251, opened 2019-11-06
DNN-29110 Site Assets > Select All is not working

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/3250

Summary: 

There is a dnnCheckbox control that needs to be updated according to the selected items. If no one item is selected, we should remove `dnnCheckbox` class, otherwise , if all items selected, add this class. 

Confirmation video: [DEMO](https://drive.google.com/file/d/1qXSI2pNX0mQNrlT2j6B5wX9Oaf7TXTUO/view?usp=sharing)

### DIFF
--- a/DNN Platform/Modules/DigitalAssets/ClientScripts/dnn.DigitalAssets.js
+++ b/DNN Platform/Modules/DigitalAssets/ClientScripts/dnn.DigitalAssets.js
@@ -2193,6 +2193,7 @@ dnnModule.digitalAssets = function ($, $find, $telerik, dnnModal) {
             if (totalItems == totalSelectedItems) {
                 $("#dnnModuleDigitalAssetsListViewToolbar input[type=checkbox]", "#" + controls.scopeWrapperId).attr('checked', true);
                 $('#dnnModuleDigitalAssetsListViewToolbar>span.dnnModuleDigitalAssetsListViewToolbarTitle', "#" + controls.scopeWrapperId).text(resources.unselectAll);
+                $('#dnnModuleDigitalAssetsListViewToolbar>span.dnnCheckbox').addClass('dnnCheckbox-checke');
             }
             updateSelectionToolBar();
         }, 2);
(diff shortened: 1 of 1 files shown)


DNN Platform/Modules/DigitalAssets/ClientScripts/dnn.DigitalAssets.js
```diff
@@ -2193,6 +2193,7 @@ dnnModule.digitalAssets = function ($, $find, $telerik, dnnModal) {
             if (totalItems == totalSelectedItems) {
                 $("#dnnModuleDigitalAssetsListViewToolbar input[type=checkbox]", "#" + controls.scopeWrapperId).attr('checked', true);
                 $('#dnnModuleDigitalAssetsListViewToolbar>span.dnnModuleDigitalAssetsListViewToolbarTitle', "#" + controls.scopeWrapperId).text(resources.unselectAll);
+                $('#dnnModuleDigitalAssetsListViewToolbar>span.dnnCheckbox').addClass('dnnCheckbox-checke');
```
Yabgu: a typo here shouldn't it be "dnnCheckbox-checked" not "dnnCheckbox-checke" you are missing last letter "d" here,

mikebigun: You are right! It is fixed now.

valadas:
I was not able to reproduce this in the platform, could this issue be Evoq specific?

mikebigun:
@valadas , it is reproducible using DNN Platform 9.4.1.22

See video I recorded to show the problem: https://drive.google.com/file/d/1ebRIUK_sK1URT3EITAOfsCNbIEiYMIXD/view?usp=sharing

valadas:
Yeah, saw the video after commenting, sorry :) But yeah, oddly it works fine here for me, so just wondering if the fix should be something different if not able to always reproduce it... Can anyone else reproduce this issue ?

thabaum:
I just tested this on two separate portals one running current beta and the other running upgraded to version 9.4.1 and this feature seems to be working selecting as needed..  Maybe a .dll file is out of date or you could use a fresh copy of file management aspx files?

What browser are you using?

Issue #3239, opened 2019-11-05
Dnn modules should follow Dnn version


valadas:
We have had issues in the past where a version number was not bumped and it caused upgrade issues since packages with the same version as the installed one are skipped. We use GitVersion to set the versions in manifest files for most modules but some are not auto-managed and get the last manually set version.

In 9.4.2-rc1, the following modules have version 9.3.2 instead of 9.4.2, which could cause issues if they have changes when upgrading from 9.3.2, 9.4.0, 9.4.1, I suggest we fix this before 9.4.2 official release to prevent issues with those upgrades.

## Modules
* Authentication
* Registration
* Search Results
* Users and Roles
* ViewProfile

## Authentication Systems
* Default Authentication

## Providers
* DotNetNuke Client Capability Provider

## Skin Objects
* All the skin objects

## Others
Those have never historically been always matching the Dnn version but I think they should (open to opinions though)
* Xcillion Skin and Containers
* Default theme and containers

## Unkowns
I am not sure on the purpose of delivering these Libraries, can someone explain their purpose and why we deliver 3 versions when we require .Net Framework 4.7.2 ?
* Microsoft CodeDom .Net Compiler (1.0.1)
* Microsoft CodeDom .Net Compiler for .NET 4.5 (2.0.1)
* Microsoft CodeDom .Net Compiler for .NET 4.6+ (2.0.1)

EPTamminga:
I support the suggestion that all standard components follow the main version #.

valadas:
@bdukes Those packages allow the usage of new C# features in ASCX files

But they are not installed, so if someone would want to use those new features, they would have to install it and then use that correct?

And it is safe to get rid of the older ones and just keep the latest right?

bdukes:
Yes, it's an optional install because it comes with `csc.exe`, and there are some hosting environments (I think GoDaddy was the main one) which don't allow deploying `exe` files.  So a package which used those new C# features would need to take a dependency on that package to ensure it was installed first.

bdukes:
And yes, it is safe to get rid of older ones.

valadas:
Cool, @donker already has most of this as part of #3254 and we discussed this is not a showstopper for 9.4.2 so moving the milestone for 9.4.3

valadas:
Merged

PR #3254, opened 2019-11-06
Set core object versions to core version

donker:
Addresses Issue #3239

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.04.02.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.04.02.SqlDataProvider
@@ -8,7 +8,126 @@
 /*****                                                  *****/
 /************************************************************/
 
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}{objectQualifier}SetCorePackageVersions') AND type in (N'P', N'PC'))
+DROP PROCEDURE {databaseOwner}{objectQualifier}SetCorePackageVersions
+GO
 
+SET ANSI_NULLS ON
+GO
+SET QUOTED_IDENTIFIER ON
+GO
+CREATE PROCEDURE {databaseOwner}{objectQualifier}SetCorePackageVersions
+AS
+DECLARE @Version VARCHAR(10)
+SET @Version = (SELECT TOP 1
+ CAST(v.Major AS VARCHAR) + '.'
+ + CAST(v.Minor AS VARCHAR) + '.'
+ + CAST(v.Build AS VARCHAR)
+FROM {databaseOwner}{objectQualifier}Version v
+ORDER BY v.Major DESC, v.Minor DESC, v.Build DESC);
+UPDATE {databaseOwner}{objectQualifier}Packages
+SET Version=@Version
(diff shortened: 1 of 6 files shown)


DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.04.02.SqlDataProvider
```diff
@@ -8,7 +8,126 @@
...
+'DotNetNuke.NAVSkinObject',
+'DotNetNuke.PRINTMODULESkinObject',
+'DotNetNuke.PRIVACYSkinObject',
+'DotNetNuke.SEARCHSkinObject',
+'DotNetNuke.SIGNINSkinObject',
+'DotNetNuke.TERMSSkinObject',
+'DotNetNuke.TITLESkinObject',
+'DotNetNuke.TREEVIEWSkinObject',
+'DotNetNuke.USERSkinObject',
+'DotNetNuke.VISIBILITYSkinObject',
```
bdukes: This seems like a maintenance issue waiting to happen… Can we not use something like `WHERE IsSystem=1` for, at least, most of these?

valadas: If I understand correctly, the packages listed here have no manifest, some others do have a minifest and actually get installed as an extension, so if we do WHERE IsSystem=1 we could hide upgrade failures on other extensions and make it hard to troubleshoot. Does that make sense?

bdukes: Yeah, so under what conditions does this list grow and we need to maintain it?  Does this only apply to "legacy" built-in components, and we'll never have a new one like this?

valadas: They are not legacy yet, but could become, not sure how skins will work after the move to .Net Core... But I haven't seen a new skinobject in ages so I doubt this list would grow. This being said since they are not packages, it would not be dramatic if one is forgotton, the main issue with packages versioning is that that won't upgrade if there was a change between versions and the version of the package is the same version as already installed.

donker: Yes, this is prone to maintenance. But the alternative is to include SQL like in the 9.3.2 script for every single release which I find worse. Besides: I don't think we will continue doing things this way.

I did look for alternatives to find these automatically, but it was not possible without introducing some level of risk of false positives and fals negatives. IMO we should go with this until we are weening ourselves off webforms and rewriting these.

bdukes: :+1: works for me, thanks for the discussion

donker:
> 
> 
> @donker Can you update this for inclusion in 9.4.3?

NP. I'll get to it.

valadas:
Just for context, my main problem with #3239 was for packages for the installer, since it causes an upgrade problem if there was code changes in a package that does not install due to it's version not being higher than the version the user is upgrading from (we had that problem in the past). This sql change (I don't disagree with it at all) but it would be a one time thing or a thing we need to do on each version or some such. What I mean is that those are not problematic, it would be just to look consistent...

Issue #3269, opened 2019-11-10
Add ability to style login in more detail


thabaum:
In response to the following issue among others along with my own interests the login form needs more ability apply more style.

https://github.com/dnnsoftware/Dnn.Platform/issues/2471

valadas:
Closing this one in favour of #3278 where this overall css discussion is being held. This issue has no details on what the problem is currently, as far as I can tell every single element can already be selected using css specificity.

Issue #2471, opened 2018-11-11
RFC: Improve login


donker:
## Description of problem
The login experience has remained untouched for years and shows its age:

1. It takes forever to load which is especially noticeable when you use the popup
2. When clicking login the form sits there while the user is being logged in - the user should see some sort of spinner
3. It's not responsive (issue #2469)
4. The "keep me logged in" checkbox is *below* the login/cancel buttons - this should be above
5. It is difficult to style/adapt given the nr of fixed classes - it should be templateable

## Description of solution
To attack problems nr 1 and nr 2 we could look at preloading a login screen. The complexity is that your page may be http and you want the login to be https. But we could consider having an improved experience only for sites that implement https and then go through a web api call to log in. This would allow us to do all the user feedback at the front end which makes things simpler.

To attack 3-5 we need to look at making this templateable. The complexity is that this control is extensible and auth providers can hook in. But IMHO this shouldn't trump us moving forward to improve this. One solution I've had some luck with is the use of a razor template inside the skin folder which describes how the login panel looks including the optional components.

Accuraty:
Please also - 
 
1. we need to have the option to unhook this 'login thing' from targeting the current page's Skin and ContentPane. Maybe there should be some settings at the Site (Portal) level that provide some choices about how the Event (user suddenly needs to authenticate) is handled and routed. What I am trying to say is it needs to uncouple from the current skin/pane logic and become its own thing with an OPTION (host and site level) to still happen the old way (which could/should be the natural default for the next few versions). The new, default end result needs to be simple and generic and work with the existing providers. 

2. please do not forget being able to login without a mouse. The tab order moving between fields needs to be logical and pressing enter needs to submit.

moorecreative:
++ on the comment for having 2-part authentication as an option, that would be a great feature!

kurtwilbies:
2-part auth is nowadays a must. My vote.

MaiklT:
Please please please add 2-factor authentication (e.g. using Google Authenticator)

thabaum:
Hello stale bot, I would like to bring this back to life.  2 Factor Authentication would be a nice touch to have soon for security reasons.

thabaum:
> IMO, points 1 to 5 are nobrainers and should be fixed

I agree.

thabaum:
[...]

PR #3270, opened 2019-11-10
Adds more control over login module form style

thabaum:
In reference to these issues:
https://github.com/dnnsoftware/Dnn.Platform/issues/3269
#2471 

## Summary
Creates additional css class's to allow more independent styles to be applied to login module page.

### DIFF
--- a/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
+++ b/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
@@ -3,7 +3,7 @@
 <%@ Register TagPrefix="dnn" Namespace="DotNetNuke.Web.UI.WebControls.Internal" Assembly="DotNetNuke.Web" %>
 <div class="dnnForm dnnLoginService dnnLoginForm dnnClear">
     <div class="dnnFormItem dnnLoginFormItemUsername">
-		<div class="dnnLoginLabelUsername">
+		<div class="dnnLabel dnnLoginLabelUsername">
 			<asp:label id="plUsername" AssociatedControlID="txtUsername" runat="server" CssClass="dnnFormLabel dnnLoginFormLabelUsername" />
 		</div>        
         <asp:textbox id="txtUsername" runat="server" />
@@ -27,7 +27,7 @@
         
     </div>
 	<div class="dnnFormItem dnnLoginFormItemRemember">
-		<asp:label id="lblLoginRememberMe" runat="server" CssClass="dnnLoginFormLabelRemember" />
+		<asp:label id="lblLoginRememberMe" runat="server" CssClass="dnnFormLabel dnnLoginFormLabelRemember" />
 		<span class="dnnLoginRememberMe"><asp:checkbox id="chkCookie" resourcekey="Remember" runat="server" /></span>
 	</div>
     <div class="dnnFormItem dnnLoginFormItemActions">
--- a/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
+++ b/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx
@@ -1,41 +1,41 @@
 <%@ Control Language="C#" Inherits="DotNetNuke.Modules.Admin.Authentication.DNN.Login" AutoEventWireup="false" CodeBehind="Login.ascx.cs" %>
 <%@ Register TagPrefix="dnn" Assembly="DotNetNuke" Namespace="DotNetNuke.UI.WebControls"%>
 <%@ Register TagPrefix="dnn" Namespace="DotNetNuke.Web.UI.WebControls.Internal" Assembly="DotNetNuke.Web" %>
-<div class="dnnForm dnnLoginService dnnClear">
-    <div class="dnnFormItem">
-		<div class="dnnLabel">
-			<asp:label id="plUsername" AssociatedControlID="txtUsername" runat="server" CssClass="dnnFormLabel" />
+<div class="dnnForm dnnLoginService dnnLoginForm dnnClear">
+    <div class="dnnFormItem dnnLoginFormItemUsername">
+		<div class="dnnLoginLabelUsername">
+			<asp:label id="plUsername" AssociatedControlID="txtUsername" runat="server" CssClass="dnnFormLabel dnnLoginFormLabelUsername" />
 		</div>        
         <asp:textbox id="txtUsername" runat="server" />
     </div>
-    <div class="dnnFormItem">
-		<div class="dnnLabel">
-			<asp:label id="plPassword" AssociatedControlID="txtPassword" runat="server" resourcekey="Password" CssClass="dnnFormLabel" ViewStateMode="Disabled" />
+    <div class="dnnFormItem dnnLoginFormItemPassword">
+		<div class="dnnLabel dnnLoginLabelPassword">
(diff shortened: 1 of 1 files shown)


thabaum:
I will see about doing this for the Registration page as well if everyone is ok with how it was done.  I studied a few things from the previous CSS classes and figured this approach does not break anything existing and people can then add CSS to the Custom CSS or Skin.css.  And if we want to get a more grip on the Default.css... this would potentially help however might be a breaking change so before playing with the CSS in default I think just adding this so it can get played with is great start for 9.

thabaum:
> Please do not remove dnnLabel and dnnFormLabel as it may unnecessarily break existing themes.

I added, not removed.  Only line I didn't add to was dnnLoginActions as it was unique enough.  The others I added a second class copying and pasting but I will have a look at the changes you requested.

I left all original classes, this should only add a second class that is more specific to each object.

thabaum:
Added CSS Classes to document (I hope I got them all so I can help with the dnndocs as needed):

`
	dnnLoginForm 
	dnnLoginFormItemUsername
	dnnLoginLabelUsername
	dnnLoginFormLabelUsername
	dnnLoginFormItemPassword
	dnnLoginFormLabelPassword
	dnnLoginFormItemCaptcha
	dnnLoginFormLabelCaptcha
	dnnLoginFormItemMessageCaptcha
	dnnLoginFormMessageCaptcha
	dnnLoginFormItemLogin
	dnnLoginFormLabelLogin
	dnnLoginPrimaryAction
	dnnLoginActionLogin
	dnnLoginSecondaryAction 
	dnnLoginActionCancel
	dnnLoginFormItemRemember
	dnnLoginFormLabelRemember
	dnnLoginFormItemActions
	dnnLoginFormLabelHelp
	dnnLoginActionsList
	dnnLoginActionRegister
	dnnLoginActionRecover
`

Original Existing CSS Classes:

`
dnnLoginRememberMe
dnnSecondaryAction
dnnActions
dnnLoginActions
dnnFormLabel
dnnFormItem
dnnSecondaryAction
dnnPrimaryAction
dnnFormLabel
dnnFormMessage 
dnnFormError 
dnnCaptcha
dnnForm dnnLoginService
dnnClear
`

Anymore changes speak now :)

valadas:
I still think we do not need that many classes, if an element has a single label child and a single field child, we can just access it from the parent class and the exisint .dnnLabel or what have you on the child using specificity instead of adding a class name on every element on top of the already existing ones. I am saying this for 2 reasons.

1. css files are static and better cached, etc. So I prefer having more css into the css and less in the html (just the minimum required to uniquely select something).
2. Although this looks pretty minor, every time we add a class, it needs to stay there forever in order to not break existing themes, so this is one of the things very easy to add but very hard to remove and we should keep that at the minimum required to solve a problem.

valadas:
And from a quick look, I don't even know if we actually have a problem to resolve here, what can't we uniquely target with css specificity currently on the login form ?

thabaum:
going forward should we use a certain naming convention for CSS that everyone agrees?  I can go through and get everything ready for styling for enhancing the user experience.

Should I use 

1. class="dnn-login-form-item-username"
2. class="dnnLoginFormItemUsername"  (Current naming convention not generally recommended or used but I can work with it)
3. class="dnn_login__form_item_username"
4. class="login-form-item-username" (removes the dnn which I believe is best practice and most common way CSS developers are used to seeing.)

We can also shorten the common form-item to type-f-i-object/action such as login-f-i-cancel.

Please advise.

I will go through all my PR's and change this to whatever works for everyone.  Trying to make them short as well so removing the - helps but if we can agree on a naming convention for the CSS classes I believe that needs to be set before going forward.  A general rule that can be broke if needed.

I believe everyone is in desperate need of being able to really manipulate everything in DNN as far as styles go.

These changes will have to stay around for a while I believe so please give me some feedback here.  You can see what I am trying to accomplish I wish to do it in the best way for everyone.  Thank you.

mitchelsellers:
Although you are leaving the existing styles in place, this to me seems to be really pushing the bloat on the login form, without a true tangible benefit to the platform.

I think trying to get an understanding of what cannot be styled, etc would be good, otherwise, at a minimum a more detailed review, as I think the goal could be addressed with quite a few less classes.

valadas:
I totally agree, a good start is to find what is unstylable uniquely due to a lacking class. From a quick look I think we can already access any single html element with css specificity with the existing classes.

thabaum:
In a way everything can be styled, it was more for making it easier for newbies to style each piece of the login.  Just seemed like a good idea at the time.  Reality is you can do it already... just takes more thought and a higher level of skills.

I was having second thoughts as well here but I still feel classes can be added to simply things so you can just use one class to style.

And I believe a few are missing.

I would really love some time to learn more and share my results and hear a discussion on these.

I did the Journal Module it was missing a lot in my opinion to make it easy to style the entire thing.  But I don't want to bloat anything for beginners or ease of use.   

Everything with an ID can be uniquely accessed just not easily.  You have to know more in depth how to edit those areas and for a lot of beginners this is pretty deep stuff.

Get rid of the cancel button... but when you remove it the page module fails to load.  I was going to play with this line a little more so that the sign in can move all the way to the right.

Personally I would just like to see the login boxes and all text and buttons center with the name of the field labels inside the text boxes with a login on the right.  I think all of these controls should pretty much be tokens for the login form and allow you to create your own login form using the HTML editor.  That would allow me to create all my own CSS and everything and change when I like.

image

this page has 

image

I think the DNN login should be similar in respect to login-form , login-input, login-submit and change some of theres to login-recover or something to that effect.  

So what is truely broken here is the naming convention, it is todays worst practice.

A part of these efforts I figured could go toward renaming classes for DNN to follow the best practices of today.  And put our DNN community twist to it.  

If I can get down exactly how we can approach this I can take on the chore of making it happen.

thabaum:
I have a feeling that navigating from old ways to new ways by version 11 is the goal, with hopefully version 11 or 12 introducing .net core?  So this CSS class naming convention and any other one should be reviewed for possible solutions to be best practices industry standard for each class.  This means using a dash - between words instead of lowercaseUppercase.  Otherwise we are not modernizing to todays standards at least.

One goal I really think should be focused on is Mobile first approach, but never forget about the power users and give some added features to them for using bigger screens such as expanding menus.

The administration panel can probably use another tweak to use less screen on mobiles.

Moving all form field labels inside of the form to get use more of the limited screens in mobile...

This is a bit of the same topic and maybe just doing some CSS tricks along with standard naming convention that is easy-to-read throughout DNN for classes could be something we can direct some time towards?

It is like an ever expanding universe, the more you know the more there is to know...

mitchelsellers:
I really appreciate the enthusiasm here, but this is approaching a much larger architecture discussion and something that if we are going to do, needs to be done across the board, not just on the login forms.

We have to consider the migration path, upgrade risks, and breaking changes of anything.  Part of this continued discussion is the future of the default.css file that is distributed with the platform.  

My $0.02 on this is that we close this for now, and work on a full plan of attack, that coordinates with longer-terms goals etc.  And focuses on solving issues

thabaum:
How do we work on it if we close it?

thabaum:
Maybe Request for Comments?

mitchelsellers:
We can do this as a RFC, but really it needs a bit of a working group discussion before it gets there.  I would want to get input from the designers, as well as developers.

We can discuss during the next TAG meeting as appropriate to try and create a team for it.

thabaum:
Thank you, I will start looking at other things I can help with until I hear word.  I dont have to do it if someone else can... just trying to help if needed where I can to make DNN the best solution for my future development efforts.

I am really looking 5 years down the road with DNN.  I almost had to turn away but I see there is hope so I am going to hang with it.

PR #3287, opened 2019-11-13
Set core object versions to core version

donker:
Addresses Issue #3239


DNN Platform/Providers/HtmlEditorProviders/RadEditorProvider/DotNetNuke.RadEditorProvider.dnn
```diff
@@ -1,6 +1,6 @@
 ﻿<dotnetnuke type="Package" version="5.0">
   <packages>
-    <package name="DotNetNuke.RadEditorProvider" type="Provider" version="09.01.00">
+    <package name="DotNetNuke.RadEditorProvider" type="Provider" version="09.04.02">
```
bdukes: Do these need to be set to 9.4.3, or are they updated automatically now?

donker: They will be updated after release from now on. This will happen in an upcoming PR and we'll make an automated task to do this in the future.

PR #3297, opened 2019-11-15
Fixes an upgrade loop issue with 9.4.3

### LINKED ISSUE #3296: 9.4.3 upgrade fails due to missing 09.04.03.SqlDataProvider
## Description of bug
The version set in the database versions table is set when the sql data providers run. When not present, the upgrade works but after installation users are redirected to the upgrade wizard because the dll has a more recent version than the database.

## Steps to reproduce
1. Install Dnn before 9.4.3
2. Upgrade to 9.4.3 using a nightly build

## Current result
Notice after the install wizard is finished, you get re-directed again to the install wizard.

## Expected result
The site should load

## Additional context
I gonna submit a PR to fix this now, but it would be nice to automate that in some way as part of the build scripts improvements, we could use GitVersion to see if that file version exists and if not create it, a blank file does the trick, we can just add actual sql in it if needed.

## Affected version

* [ ] 10.0.0 alpha build
* [x] 9.4.3 alpha build 
* [ ] 9.4.2 alpha build
* [ ] 9.4.1 latest supported release

valadas:
Fixes #3296 

## Summary
Adds 09.04.03.SqlDataProvider file 
The database versions table is updated by this file and this prevents Dnn from thinking it is due to an upgrade after the actual upgrade because the dll version is higher than the database version.


donker:
If you check in PR#3287 this one is obsolete. If you check in this one first the other PR will have a merge conflict.

valadas:
Oh, correct, ok, I will try to get some time to review #3287 then...

Issue #3116, opened 2019-10-05
Export to PDF Issue on SQL Console


kkadir:
## Description of bug
When a query result on the SQL Console has long text cells, which are wrapped in the result table; then the first exported PDF is normal but next exported PDF files are distorted as wrapped texts are not wrapped anymore on PDF files.

## Steps to reproduce
List the steps to reproduce the behavior:
1. Log in as a Super User
2. Go to Persona Bar > Settings > SQL Console
3. Write a query like `SELECT * FROM Tabs` and click the "Run Script" button
4. Observe that the result table is generated and some columns like IconFile, VersionGuid, UniqueId and etc. have wrapped long texts.
5. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded.
6. Open the PDF file and see that it is normal; all long texts are wrapped as expected.
7. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded for a second time.
8. Open the PDF file and see that, this time long texts are not wrapped and the PDF layout is distorted.

## Current result
PDF files containing long text values in any table cell are distorted after 2nd export.

## Expected result
PDF files should be consistently generated as the same without any distortion.

### 1st Export
![image](https://user-images.githubusercontent.com/8139167/62689138-e6534b80-b9d2-11e9-87b5-14e74646cd81.png)

### 2nd Export
![image](https://user-images.githubusercontent.com/8139167/62689550-afca0080-b9d3-11e9-9f22-a809759c09f8.png)

## Error log
N/A

## Additional context
Note that non-wrapped texts do not have any separators or spaces.

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

thabaum:
![image](https://user-images.githubusercontent.com/13577556/68997637-7cfa9880-085d-11ea-8678-eaba925f8c84.png)
![image](https://user-images.githubusercontent.com/13577556/68997669-dbc01200-085d-11ea-85cf-b78ff368b7fe.png)

sure takes a while and this spinner or hourglass replacement is hard to find, should try to center on the screen view if it can or be at the top of the page next closer to the button.

other than that I actually could not reproduce on a second time.

[...]

PR #3301, opened 2019-11-15
Update jsPDF to version 1.5.3

thabaum:
updates version jsPDF to version 1.3.2 help resolve issue #3116 

## Summary

Needs testing


bdukes:
Is there a reason you're updating to 1.3.2 instead of a newer version?  Is there a reason to believe this update fixes the issue in #3116?

thabaum:
i looked for a newer version I thought this was it... I must of got it from the wrong source

https://github.com/MrRio/jsPDF/releases

i see now it is here I will make adjustments thank you @bdukes

thabaum:
OK I updated the js to the latest version.  I notice at the bottom of the script it moves a line from inside to outside, maybe this helps cure the second go around creating a pdf.

I noticed it on the 1.3.2 version

`}(typeof self !== "undefined" && self || typeof window !== "undefined" && window || typeof global !== "undefined" && global || Function('return typeof this === "object" && this.content')() || Function('return this')()));`

Dnn.AdminExperience/Extensions/Content/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SqlConsole/scripts/jspdf.js
```diff
@@ -10,51 +10,67 @@
...
-    var jsPDF = function (global) {
-        'use strict';
-
-        var pdfVersion = '1.3',
-            pageFormats = { // Size in pt of various paper formats
+/** @preserve
+ * jsPDF - PDF Document creation from JavaScript
+ * Version ${versionID}
+ *                           CommitID ${commitID}
+ *
```
bdukes: I'm concerned that these placeholders indicate you're pulling a source version instead of a distribution version.   Can you pleases switch to https://cdn.jsdelivr.net/npm/jspdf@1.5.3/dist/jspdf.min.js

thabaum: OK will do thanks

thabaum:
OK i see the cdn min.js file is a bit different from the source version and from the original in dnn.

Please let me know how I did.

valadas:
I have 3 concerns.

1. Was this tested and does it resolve the issue ?
2. If we **only** have the minified version it makes debuggin hard, but on the other hand, this specific module is not setup to have a non-minified and a minified version, sooo I could live with that.
3. My major concern is that there was breaking changes in 1.5.0 and I do believe this affects us since we use one of those deprecated methods

image

Also, just gave it a quick try and it does not even initialize the library.
image

thabaum:
```
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
    typeof define === 'function' && define.amd ? define(factory) :
    (global.jspdf = factory());
}(this, function () { 'use strict';

    var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) {
      return typeof obj;
    } : function (obj) {
      return obj && typeof Symbol === "function" && obj.constructor === Symbol ? "symbol" : typeof obj;
    };
```

I imagine an adjustment here to connect the new to the old is needed?

thabaum:
`If you still want to use addHTML or fromHTML as they are still in jspdf. But if some changes break those plugins, you can still use jsPDF 1.4.1, which has a stable version for both plugins.`

jsPDF 1.4.1 the one that might work for DNN for now?

valadas:
Well, I did not take the time to dig into that part works, but yeah, we cannot just merge this PR. Changes need to be tested as working, fixing #3116 needs a confirmation that it does fix it and upgrading a library needs at least some testing that it does not break the whole feature all together, for now I am closing this PR, please resubmit if you take the time to find the proper way and test it.

thabaum:
I am working on it now...

thabaum:
nvm SELECT * FROM Tabs

Issue #3305, opened 2019-11-17
Email should be above username in User.ascx


thabaum:
## Description of bug
The email address is below password in User.asxc

## Expected result
Email is usually above username on a form.

valadas:
Merged

PR #3306, opened 2019-11-17
Move Email Above Username in User.ascx #3305

thabaum:
For issue #3305

## Summary

This will put email address field above username on the User.ascx form

### DIFF
--- a/DNN Platform/Website/controls/user.ascx
+++ b/DNN Platform/Website/controls/user.ascx
@@ -19,6 +19,16 @@
 				maxlength="50"></asp:textbox>&nbsp;*
 			<asp:requiredfieldvalidator id="valLastName" runat="server" cssclass="NormalRed" display="Dynamic" errormessage="<br>Last Name Is Required."
 				controltovalidate="txtLastName" resourcekey="valLastName"></asp:requiredfieldvalidator></td>
+	</tr>	
+	<tr>
+		<td class="SubHead" width="175">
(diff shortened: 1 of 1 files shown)


thabaum:
https://github.com/dnnsoftware/Dnn.Platform/pull/3204

already approved here for development release,  redo PR for current version 9.

thabaum:
tabindex... interesting and done.

Would this be something the login form would have or should have as well?

valadas:
Thank you!

Ok, so tabIndex can be omitted on inherently focusable elements like buttons and inputs which means that the tab order will follow the document declaration order. If it is assigned a value lower than 0 (like -1) it makes the element non-focusable by keyboard, when 0 it is the normal document flow behaviour and when higher than 0 then it will be in numerical order.

Most of the time the recommendation is to not specify it, this would be best practice in most situations. Here it is used I believe because people may re-style the form to let's say show the fields in 1 or two columns on mobile vs desktop... This would keep a logical tab order even though it looks different than the declaration order.

Ex: Mobile
```
FirstName(1)
LastName(2)
Email(3)
Password(4)
```

Could be presented on desktop as:
```
FirstName(1)     Email(3)
LastName(2)     Password(4)
```

This would ensure proper logical tab order to go from FirstName to LastName and so on. To be perfectly honest if this is styled with css, it does not change a thing since the declaration order stays the same, but let's suppose some theme would use javascript to change the order, then it would keep that logical entry order...

Issue #2225, opened 2018-08-11
DNN Journal Word Wrap turn off/on


thabaum:
## Description 
I am reading half of a word and having to scroll to next line to finish reading it.
## Steps to reproduce
Put a journal post long enough to make a word go to next line
## Current result
half the word is on one line the other half is on the other
## Expected result
Easier to read if you can see the entire word, otherwise you feel like your in elementary school learning how to read.
## Affected version

* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

Tychodewaard:
If you add to the css:

.journalitem .dnnClear {
word-break: initial;
}

It is fixed. 
But I don't know if this is the right approach.

PR #3307, opened 2019-11-18
Word-Break added to Journal P

thabaum:
Issue #2225 

## Summary

Keeps words from breaking in journal posts.


thabaum:
I am glad my dnn site is not the only one that pumps out an error page... this Github site does it all the time and just did it.

I did some looking around I am hoping this is an OK solution as it seems to make me happy.  I just need to work on the comment space, links and images in journal now see if I can doctor those up.  Let me know any concerns here.

valadas:
> ... pumps out an error page... this Github site does it all the time

Yeah it has been slow/intermittent all day for me too. Not an usual thing though...

Issue #3118, opened 2019-09-11
Close Person Bar when using restart App or clear cache


Timo-Breumelhof:
## Description of problem
When restarting DNN or clearing the main cache through Settings > Servers, DNN does reload the Persona Bar (opened) but not the page. I think the most common use case is that you want to clear the cache because of a rendering issue on a page or after some SQL statements that have been executed.
In that case you now have to close the persona bar and refresh the page after you restarted DNN.
(I have been working with  a module that only refreshed it's templates after a DNN restart and I can assure you this gets annoying)

## Description of solution
IMO, both "clear cache" and "restart DNN" should reload the page and not have the PB open afterwards.

## Description of alternatives considered
This is the result of the discussion here: dnnsoftware/Dnn.AdminExperience#1084

## Additional context
Provide a clear and concise description of what you expected to happen.

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

valadas:
merged

PR #3315, opened 2019-11-19
Reload page when the code tells you to

donker:
Solves issue #3118


Dnn.AdminExperience/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Servers.Web/src/components/App.jsx
```diff
@@ -31,6 +31,10 @@ class App extends Component {
     render() {
         const {props} = this;
         const buttonVisible = utils.isHostUser();
+        if (this.props.reloadPage) {
+            window.top.location = window.top.location.protocol + '//' + window.top.location.host + window.top.location.pathname + window.top.location.search;
```
bdukes: I think `window.top.location.reload()` should work in all supported browsers

PR #3319, opened 2019-11-20
Fix bug #3318

coretoco:
Closes #3318

## Summary
The TrackingId should be uppercase to work with Google Analytics.

PR #3322, opened 2019-11-20
Google Analytics TrackingID is stored in lowercase #3318

### LINKED ISSUE #3318: Google Analytics TrackingID is stored in lowercase
## Description of bug
When adding the analytics tracking code in analytics connector, it will be stored in lowercase. This makes analytics not tracking anything.

## Steps to reproduce
1. Go to 'Settings' --> 'Connectors'
2. Click on 'Google Analytics' --> Edit
3. Enter Google TrackingID (eg UA-123456789-0)
4. Click 'Save'

## Current result
The 'SiteAnalytics.config' in website root directory will contain the TrackingID in lowercase: (ua-123456789-0)

## Expected result
The 'SiteAnalytics.config' in website root directory should contain the TrackingID in an unchanged form: (UA-123456789-0)

## Affected version
* [x] 9.4.2 latest supported release

## Affected browser
* [x] All Browsers

coretoco:
Closes #3318 
Supersedes #3319 

## Summary
The TrackingId should be uppercase to work with Google Analytics.

### DIFF
--- a/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
+++ b/DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
@@ -189,7 +189,7 @@ namespace DNN.Connectors.GoogleAnalytics
                 }
                 else
                 {
-                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToLowerInvariant().Trim() : string.Empty;
+                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToUpperInvariant().Trim() : string.Empty;
                     urlParameter = values["UrlParameter"]?.Trim() ?? string.Empty;
                     trackForAdmin = values["TrackAdministrators"] != null ? values["TrackAdministrators"].ToLowerInvariant().Trim() : string.Empty;
                     anonymizeIp = values["AnonymizeIp"] != null ? values["AnonymizeIp"].ToLowerInvariant().Trim() : string.Empty;


DNN Platform/Connectors/GoogleAnalytics/GoogleAnalyticsConnector.cs
```diff
@@ -189,7 +189,7 @@ public bool SaveConfig(int portalId, IDictionary<string, string> values, ref boo
                 }
                 else
                 {
-                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToLowerInvariant().Trim() : string.Empty;
+                    trackingID = values["TrackingID"] != null ? values["TrackingID"].ToUpperInvariant().Trim() : string.Empty;
```
bdukes:
```suggestion
                    trackingID = values["TrackingID"] != null ? values["TrackingID"].Trim() : string.Empty;
```

Should this just allow for whatever casing is used?  Is there a reason to upper-case it?

bdukes: @david-poindexter I see your comment on #3318, which seems to say that the code _should_ allow any case to be saved, rather than forcing uppercase.  Am I missing something?

david-poindexter: Sorry for my clunky wording @bdukes - that's what I getting for rushing.  😆 

It is CURRENTLY respecting case and saving the TrackingID value just as entered by the admin.  However, this is causing a problem with Google Analytics functionality.  It needs to be UPPERCASE in order to work as expected.

PR #3326, opened 2019-11-21
Fix casing mismatch errors in security analyzer file compare

donker:
Fixes issue where security analyzer would not recognize default.aspx on 9.4 installations and report an error.


Dnn.AdminExperience/Extensions/Content/Dnn.PersonaBar.Extensions/Components/Security/Checks/CheckDefaultPage.cs
```diff
@@ -73,7 +73,7 @@ private bool CheckDefaultPageModified(out IList<string> modifiedFiles)
                 var fileName = node.Attributes["name"].Value;
                 var sum = node.Attributes["sum"].Value;
                 var file = Path.Combine(Globals.ApplicationMapPath, fileName);
-                if (!File.Exists(file) || Utility.GetFileCheckSum(file) != sum)
+                if (!File.Exists(file) || Utility.GetFileCheckSum(file).ToLower() != sum.ToLower())
```
bdukes:
```suggestion
                if (!File.Exists(file) || Utility.GetFileCheckSum(file).Equals(sum, StringComparison.OrdinalIgnoreCase))
```

PR #3329, opened 2019-11-21
Registered IPersonaBarContainer with DI

### LINKED ISSUE #3328: [Dependency Injection] Unable to create instance of PersonaBarContainer
## Description of bug
Evoq calls `PersonaBarContainer.SetInstace(IPersonaBarContainer, bool)` to override the default Persona Bar at application startup. The passed in `IPersonaBarContainer` argument is an instance of a class that derives from `PersonaBarContainer`.
Since 697a54e36808850e13c9149fde79985ae4646293, `PersonaBarContainer`'s constructor requires a `INavigationManager` argument (which is a breaking change for Evoq).
Creating an instance of `INavigationManager` during application start up is not possible, because:
1) The app startup event is not a DI friendly context
2) `Globals.DependencyProvider` is internal to core DotNetNuke assemblies only, so I can't call `Globals.DependencyProvider.GetRequiredService<INavigationMananger>()` from there.
The alternative is to create and register our derived PB container instance using the recommended `IDnnStartup` method, but then it would be ignored by `PersonaBarContainer.Instance` property.

## Steps to reproduce
1) Create a project with references to core DotNetNuke assemblies
2) Add a class that implements `IAppEvents`
3) Implement `ApplicationBegin()` method
4) Invoke PersonaBarContainer.SetInstance(IPersonaBarContainer, bool) from there

## Current result
In order to call above method, an instance of `INavigatorManager` is required, which cannot be obtained from DI.

## Expected result
There should be a way to derive from `PersonaBarContainer` and call the base constructor, or `PersonaBarContainer.Instance` should get the object from DI.

## Affected version

* [X] 10.0.0 alpha build
* [X] 9.4.3 beta build
* [X] 9.4.1 latest supported release

daguiler:
Fixes #3328 

## Summary
This makes `PersonaBarContainer.Instance` property use DI to build the returned object instead of simply newing up a `PersonaBarContainer` instance.
This allows third-party apps to register PB container derived classes with DI and override the default PB container.

### DIFF
--- a/Dnn.AdminExperience/Library/Dnn.PersonaBar.Library/Containers/PersonaBarContainer.cs
+++ b/Dnn.AdminExperience/Library/Dnn.PersonaBar.Library/Containers/PersonaBarContainer.cs
@@ -58,7 +58,7 @@ namespace Dnn.PersonaBar.Library.Containers
             {
                 if (_instance == null)
                 {
-                    _instance = new PersonaBarContainer(Globals.DependencyProvider.GetRequiredService<INavigationManager>());
+                    _instance = Globals.DependencyProvider.GetRequiredService<IPersonaBarContainer>();
                 }
 
                 return _instance;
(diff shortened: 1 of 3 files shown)


ahoefling:
This has really big impact to the DNN community, thanks @daguiler for discussing with everyone about the problem you were having in Evoq and using DI to solve the problem including this PR.

If you are unaware the `IPersonaBarContainer` gives you the ability to customize the Persona Bar as a vendor. This is something I was playing with awhile back on how I could add my own top level settings to the Persona Bar. Now if you create your own implementation of `IPersonaBarContainer` and register it you will be able to do something similar to Evoq with the Persona Bar.

This is something that I don't believe is for every single vendor, but for a vendor that really wants to customize the Persona Bar this is going to be a great addition. Just remember there is potential for conflicting registrations say if a vendor uses this technique and is also on Evoq

PR #3333, opened 2019-11-22
Search Results: Advanced Search Scopes - inaccurate behavior

### LINKED ISSUE #3332: Search Results: Advanced Search Scopes - inaccurate behavior
## Description of bug
Advanced Search > List of available Filters returns other options than the selected ones from the Settings > Results Scope for Content Type(s) and
This is due to how the code matches the scopes,
Currently, it will look for the selected scope, and will check if the name of the other scopes is contained within it. In this example it checks if "Blog Subscribe" contains "Blog" which returns true and therefore it ends up showing that scope as well which is not the expected behavior.

The line with doing the 'contains' match is
https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Website/DesktopModules/Admin/SearchResults/SearchResults.ascx.cs#L232
## Steps to reproduce
1. Go to the Search Results page of your site.
2. Go into Edit mode, and go to the settings of the Search Results Module.
3. Go to the 'Search Results Settings' Tab
4. For the 'Results Scope for Content Type(s), uncheck all of the entries and only leave "Blog Roster"
5. Save the Settings and Refresh the Page.
6. Click on the Advanced Drop-down located at the right of the search bar
7. Review the List of available Filters

## Current result
It is showing 2 different Scopes "Blog" and "Blog Roster" even though only one scope has been selected.

## Expected result
It should only show the scopes that were selected in the settings.

## Screenshots
https://drive.google.com/file/d/1QUWOku7j7fjJMusI24WS9BKNbe0GoGvL/view

## Affected version

* [ ] 10.0.0 alpha build
* [ ] 9.4.2 alpha build
* [x] 9.4.1 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

aelibyary:
Fixes #3332 

## Summary

The fix is as simple as the code should look for a exact match of the scope so we replaced Contains with Equals
Fix Confirmation Video: https://drive.google.com/file/d/1ge7l5F9W37mp_GizncDFIUCzATeSByJi/view


mitchelsellers:
@aelibyary Please note, as discussed with ESW leadership in the past, all PR's and Issues should only contain human-friendly descriptions.  If you need to track an internal ID, please add it to the comments, rather than the title.  I've updated this PR, but your cooperation is greatly appreciated.

aelibyary:
> @aelibyary Please note, as discussed with ESW leadership in the past, all PR's and Issues should only contain human-friendly descriptions. If you need to track an internal ID, please add it to the comments, rather than the title. I've updated this PR, but your cooperation is greatly appreciated.

Noted!, thank you.

v-karbovnichy:
This PR milestone should be changed.
This change was released in 9.6.1
Check the tags for this commit: f41c2ad0c72847370bf3cbde80bc0c0754fcc9f9
image

valadas:
Done

PR #3350, opened 2019-11-28
do not replace the protocol in query string.

### LINKED ISSUE #3349: login page will go into infinite redirect under ssl offloading environment
## Description of bug
login page will go into infinite redirect under ssl offloading environment

## Steps to reproduce
1. install Evoq latest release under an ssl offloading environment;
2. browse to the login page with returnurl value set to a full https URL;

## Current result
the login page will go into an infinite redirect.

## Expected result
The login page should open correctly.

zyhfish:
Fix #3349

### DIFF
--- a/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
@@ -2777,15 +2777,7 @@ namespace DotNetNuke.Entities.Urls
                                 //DNN-9158: prevent SSL Offloading infinite redirects
                                 if (!result.IsSecureConnection && result.IsSSLOffloaded && bestFriendlyNoScheme.StartsWith("https"))
                                 {
-                                    if (bestFriendlyNoScheme.Contains("?"))
-                                    {
-                                        var urlParts = bestFriendlyNoScheme.Split('?');
-                                        bestFriendlyNoScheme = $"{urlParts[0].Replace("https://", "http://")}?{urlParts[1]}";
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
```diff
@@ -2777,7 +2777,15 @@ private void SecurityCheck(HttpApplication app)
                                 //DNN-9158: prevent SSL Offloading infinite redirects
                                 if (!result.IsSecureConnection && result.IsSSLOffloaded && bestFriendlyNoScheme.StartsWith("https"))
                                 {
-                                    bestFriendlyNoScheme = bestFriendlyNoScheme.Replace("https://", "http://");
+                                    if (bestFriendlyNoScheme.Contains("?"))
+                                    {
+                                        var urlParts = bestFriendlyNoScheme.Split('?');
```
bdukes:
```suggestion
                                        var urlParts = bestFriendlyNoScheme.Split(new[] { '?' }, count: 2);
```

We might as well guard against multiple `?` characters, right?

Or maybe it should be something like this:
```
if (!result.IsSecureConnection && result.IsSSLOffloaded && bestFriendlyNoScheme.StartsWith("https://"))
{
    bestFriendlyNoScheme = "http://" + bestFriendlyNoScheme.Substring("https://".Length);
}
```

zyhfish: Thanks @bdukes , code updated.

Issue #2868, opened 2019-06-23
ModuleController.Instance.UpdateModuleSetting Not Updating


mitchelsellers:
## Description of the bug
When working with the ModuleController object and attempting to update a module setting the setting can be initially set, but subsequent calls to UpdateModuleSetting does not result in the value being changed

Example Code:

ModuleController.Instance.UpdateModuleSetting(moduleId, "MyName", "MyValue");

## Steps to reproduce
1.  Have a module settings action, or otherwise, that will set a value for a Module Setting using the above snippet.
2. After updating validate that the database has the proper value in the ModuleSettings table
3. Try to update the setting value with a new value

## Current result
The value in the DNN Database contains the older value, with the older timestamp.  It is not possible to overwrite the setting value.

## Expected result
The value in the database, and exposed to the code, should be the new value set.  With the timestamp of when the last action was completed.

## Additional context
This appears to only impact ModuleSettings, TabModuleSettings appear to be working as desired.

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1

Earlier versions were not verified

valadas:
Is this always the case or only under some conditions, I am surprised this was not noticed earlier...

mitchelsellers:
I have tested both an MVC module and a WebForms module and the behavior is consistent.

david-poindexter:
Closing per #3360

PR #3360, opened 2019-12-02
Fix UpdateModuleSetting in ModuleController

engineering87:
Fix the 'UpdateModuleSetting' method in ModuleController refactoring the update module conditions.
Issue https://github.com/dnnsoftware/Dnn.Platform/issues/2868

## Summary
Modified the update conditions, removed the 'settingExist' variable and check directly the settings value.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/ModuleController.cs
+++ b/DNN Platform/Library/Entities/Modules/ModuleController.cs
@@ -814,30 +814,28 @@ namespace DotNetNuke.Entities.Modules
                 var currentUser = UserController.Instance.GetCurrentUserInfo();
                 dr = dataProvider.GetModuleSetting(moduleId, settingName);
 
-	            var settingExist = false;
 	            string existValue = null;
                 if (dr.Read())
                 {
-					settingExist = true;
 	                existValue = dr.GetString(1);
                 }
 
 				dr.Close();
 
-				if (existValue != settingValue)
-	            {
-					dataProvider.UpdateModuleSetting(moduleId, settingName, settingValue, currentUser.UserID);
-					EventLogController.AddSettingLog(EventLogController.EventLogType.MODULE_SETTING_UPDATED,
-														"ModuleId", moduleId, settingName, settingValue,
-														currentUser.UserID);
-				}
-				else if (!settingExist)
-				{
-					dataProvider.UpdateModuleSetting(moduleId, settingName, settingValue, currentUser.UserID);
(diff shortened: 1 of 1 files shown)


engineering87:
Hi Brian,
I tested the change with a simple test module defined in this way:

```
protected void Page_Load(object sender, EventArgs e)
        {
            try
            {
                const string key = "TestProperty";
                ModuleController.Instance.UpdateModuleSetting(ModuleContext.ModuleId, key, "False");
                var moduleInfo = ModuleController.Instance.GetModule(ModuleContext.ModuleId, ModuleContext.TabId, true);
                var value = moduleInfo.ModuleSettings[key];
                if (value != null && value.ToString() == "False")
                {
                    ModuleController.Instance.UpdateModuleSetting(ModuleContext.ModuleId, key, "True");
                    moduleInfo = ModuleController.Instance.GetModule(ModuleContext.ModuleId, ModuleContext.TabId, true);
                    value = moduleInfo.ModuleSettings[key];
                    if (value == null || value.ToString() == "False")
                    {
                        throw new ModuleLoadException("Second update failed");
                    }
                }
                else
                {
                    throw new ModuleLoadException("First insert/update failed");
                }
            }
            catch (Exception exc) //Module failed to load
            {
                Exceptions.ProcessModuleLoadException(this, exc);
            }
        }
```

the TestProperty is correctly set in the ModuleSettings table.

Actually I haven't tried the same code with the unpatched version yet but reading the issue it seems that the problem is consistent. Regarding the original code, the second condition in 'else if' was executed only when the 'settingValue' was null.

bdukes:
I tried your test code and it appears to work correctly on 9.4.3 (without this patch).  @mitchelsellers were you testing a different scenario in #2868?

mitchelsellers:
@bdukes This appears to have fixed itself from when I initially posted it as well, as I just validated that I have modules working with 9.4.1 and 9.4.2 that do not experience the issue in #2868

bdukes:
@engineering87 had you run into this issue in practice, or were you just tackling it based on @mitchelsellers' issue #2868?  It would be great if we could isolate and reproduce the real issue to make sure any change we make is truly helpful.

engineering87:
Actually just tackling it based on the #2868 issue but I will make further checks and tests to try to replicate what has been reported.

mitchelsellers:
From what I can tell.  It appears that this is no longer an issue.  It was recreatable when I submitted the issue, but I cannot now.

PR #3361, opened 2019-12-03
Merges latest changes from 9.4.x into development

valadas:


valadas:
Closing in favor of #3366

Issue #3344, opened 2019-11-26
Possible memory leak in MVC modules


jsbsantos:
## Description of bug
Hello.

I have a project running in production, which I upgraded to the 9.4.1 version of DNN. After the upgrade I noticed that there was an unusual memory usage, closing in on max, resulting in website restarts.

Searching in the reported issues, I found one issue [related to a memory leak in the API](https://github.com/dnnsoftware/Dnn.Platform/issues/3186), which [was fixed in DNN 9.4.2](https://github.com/dnnsoftware/Dnn.Platform/pull/3192) version release. 

I went on to upgrade to the 9.4.2 on my dev environment to test things out. Turns out there was still a high memory usage. 

For my next test, I installed a fresh DNN site, with no modules and ran a load test (20 concurrent users for 5 minutes) and the results were good, with normal memory usage. After that I installed a simple module, created with a DNN template based on [Chris Hammond's template](https://marketplace.visualstudio.com/items?itemName=Chris-Hammond.DotNetNukeDNNDevelopmentProjectTemplates). This next test was done by adding this module to an empty page and running the same load test, which, turned out to be spot on: memory usage increased rapidly and was close to max.

I am attaching the module I used to test this behavior.

## Steps to reproduce

1. Install the attached TestModule
2. Add TestModule to any page
3. Run a load test on the page containing the module
4. Observe memory usage for IIS process

## Current result
Memory usage increases rapidly until it's maxed out.

## Expected result
Memory usage should remain stable.

## Screenshots
1) Memory Usage for IIS process running the clean test installation with a test module 

![image](https://user-images.githubusercontent.com/729094/69651148-8c50c180-1067-11ea-8b95-5a8e32157056.png)

## Additional context
I took a process dump from my production website and after analysis I theorized that the leak was related to the new DI. Digging into the dump file showed that the ServiceProvider is not disposing of module controllers, as you can see by the image bellow.

![image](https://user-images.githubusercontent.com/729094/69651437-fb2e1a80-1067-11ea-8e71-890b7c0c155b.png)

## Affected version
9.4.0, 9.4.1, 9.4.2 

* [TestModule_00.00.01.00_Install.zip](https://github.com/dnnsoftware/Dnn.Platform/files/3892796/TestModule_00.00.01.00_Install.zip)
* [TestModule.zip](https://github.com/dnnsoftware/Dnn.Platform/files/3892810/TestModule.zip)

Bladixx:
Hello 

[...]

PR #3369, opened 2019-12-04
Fixes Memory Leak for MVC Modules

ahoefling:
Fixes: #3344 

## Summary
Fixes memory leak in MVC modules by utilizing the Microsoft recommendations for Dependency Injection in ASP.NET MVC. The Docs were very useful and while they use the Unity Container, this was adapted for `IServiceProvider`
- [ASP.NET MVC Dependency Injection](https://docs.microsoft.com/en-us/aspnet/mvc/overview/older-versions/hands-on-labs/aspnet-mvc-4-dependency-injection)

### Notable Changes
- Removed `UseMvc()` method which registers all MVC controllers to the container. It will now use the default controller activation technique in ASP.NET
- Deleted Custom `IControllerFactory` as we are now using default ASP.NET controller factory
- Added new `IDependencyResolver` for DNN

### Notes
In my initial tests, this appears to have fixed the memory leak, but I would like some others to take the install package and test it out prior to us accepting this change


thabaum:
Nice work @ahoefling thank you very much for your effort!

PR #3374, opened 2019-12-05
Merges changes from 9.4.4 into develop

valadas:
Merges changes from 9.4.4 (here just the MVC memory leak fix) into current development. This is a release management task which only includes pre-approved pull-requests, for that reason we are self-approving it.


bdukes:
@valadas is there a reason this hasn't been merged yet?

valadas:
@bdukes because first I don't have permission :)

Second, it could stay here until wednesday and we would have a single one to merge instead of many if there are last minute fixes to 9.4.4... So no rush either.

PR #3372, opened 2019-12-05
Email Above Username in User.ascx.cs & Registration Form Issue #3371

thabaum:
Closes issue #3371
## Summary
Changed code to create email address field above Username on registration form and the User.ascx.cs file to match the User.ascx email order.

Issue #3371, opened 2019-12-05
Email is below password fields on registration form


thabaum:
## Description of bug
Provide a clear and concise description of what the bug is.

## Steps to reproduce
List the steps to reproduce the behavior:
1. Go to registration form
2. view order and see email is below the password fields.

## Current result
If you have to retype a captcha you might need to scroll up to re-enter passwords each time the code does not work.

## Expected result
Fields you may have to re-enter are close to the submit button.

## Screenshots
If applicable, provide screenshots to help explain your problem.

## Error log
Paste the error log that is related to this issue.

## Additional context
It would be nice to group the help icon and label together so they stick together if it is put on top of the text area.

## Affected version

* [ ] 10.0.0 alpha build
* [ ] 9.4.3 alpha build
* [ ] 9.4.2 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

valadas:
PR merged

PR #3386, opened 2019-12-06
Moves Email + Display Name on Registration Form & Email in User.ascx file above Username/Password on 

thabaum:
#3371 #3372
## Summary

Moves Email above Username in User.ascx.cs to match order of Users.aspx
Moves Registration Email Field above Username on the Registration Form.

### DIFF
--- a/DNN Platform/Library/UI/UserControls/User.cs
+++ b/DNN Platform/Library/UI/UserControls/User.cs
@@ -279,21 +279,21 @@ namespace DotNetNuke.UI.UserControls
                     txtIM.TabIndex = Convert.ToInt16(_StartTabIndex + 7);
                     txtFirstName.Text = _FirstName;
                     txtLastName.Text = _LastName;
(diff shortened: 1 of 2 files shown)


donker:
Hey @thabaum, does this change anything? It looks like it just moved stuff in code, not in UI, right?

thabaum:
@donker The registration form is created in code making it hard to see.  This is pretty much two related PR's in one.

The first one just moves in logical order email above username in the User.aspx with the last PR that was related "Chore".  It really does nothing but make sense when you look the order in the code.

"&"

"Enhancement" or "Chore" as well secondly this actually moves the email+display name on the registration form.  I dug all over trying to find it please correct me if I am wrong but the Display Name and email Registration HTML will be parsed prior to username and password fields on the registration form if this PR is accepted.  I am hoping I got this part 100% right as this is my main intention with this PR.

valadas:
Yep that should do it @thabaum

Issue #3387, opened 2019-12-06
Form Messages are not using 100% of screen


thabaum:
## Description of bug
form messages can only expand to 980px

## Steps to reproduce
List the precise steps to reproduce the bug:
1. Go to a page that has administrator only form message

## Current behavior
for high resolution screens this message is not expanding across the window.

## Expected behavior
The message area uses all of the allowed space.

## Screenshots
If applicable, provide screenshots to help explain the bug.
![image](https://user-images.githubusercontent.com/13577556/70355667-8f963b00-1827-11ea-89d0-c60f1ac4a50b.png)

## Error information
Provide any error information (console errors, error logs, etc.) related to this bug.

## Additional context
Provide any additional context that may be helpful in understanding and/or resolving the bug.

## Affected version

* [ ] 10.0.0 alpha build
* [ ] 9.5.0 alpha build
* [ ] 9.4.3 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

bdukes:
I think this is an intentional design decision to restrict the form messages from growing too large.  If you want them to be 100% in your site, you can override the styles via your theme or the site's stylesheet (i.e. `portal.css`).  My opinion would be that this is _not_ a bug/defect and should remain as-is.

thabaum:
ok no worries just looked odd on the form.  Felt like it should use the full screen or space allowed in the skin template instead of cutting itself off.

valadas:
Well, this is certainly an opiniated thing but I think the default of 100% and people can override it if they want is better than a fixed default and people can make it otherwise. 100% is certainly more consistent IMO Very ofter with css frameworks, the forms measure like around 1050px wide with the default 980px it looks like it is just missing a little bit... I marked is as an enhancements and approved the PR but if there is traction to not merge that, please comment...

valadas:
PR merged

PR #3388, opened 2019-12-06
dnnFormMessage 100% width fix

thabaum:
#3387
## Summary

allows a form message to use 100% of the area it is presented in.

image

will now look like 
image

if accepted as a solution.


valadas:
For reviewers, just in case you don't look at the issue linked, this is by design limited to a fixed pixel max-width. It is not a bug per se and it is an opiniated decision, one the theme can easily override. I for one think 100% is a better default than any fixed width, but yeah, please comment your opinions :)

valadas:
Cool, @thabaum not a big issue but when you mention an issue just with the pound sign, it only creates a link to it. If you add the words Fixes, Fix, Closes or Close and a few others, it will also auto-close the issue when the PR is merged. Like I said, no big deal just a little thing to know about to automate the process :) See https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords for details.

Issue #3390, opened 2019-12-08
register buttons are on the right side of the form instead of below captcha


thabaum:
## Description of bug
Provide a clear and concise description of the bug.

## Steps to reproduce
List the precise steps to reproduce the bug:
1. Go to a page with the user registration module
2. view screenshot

## Current behavior
list of registration buttons are aligned to the right of the captcha

## Expected behavior
buttons below captcha and other fields

## Screenshots
If applicable, provide screenshots to help explain the bug.
![image](https://user-images.githubusercontent.com/13577556/70397195-fc930780-19c4-11ea-9809-ee146a44623a.png)

## Error information
Provide any error information (console errors, error logs, etc.) related to this bug.

## Additional context
Provide any additional context that may be helpful in understanding and/or resolving the bug.

## Affected version

* [ ] 10.0.0 alpha build
* [ ] 9.5.0 alpha build
* [ ] 9.4.3 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

PR #3391, opened 2019-12-08
<DIV> Wrap Registration Buttons with css class

thabaum:
adds a div around the register button list along with a css class "registerButtonGroup"

issue #3390

## Summary

adds a missing DIV around the registration button group to allow the buttons to go below the other form fields instead of to the right of the security code div.


thabaum:
never mind i think it was a browser cache issue of some type... sorry

valadas:
Oh ok...

PR #3392, opened 2019-12-09
avoid create folder with invalid path.

zyhfish:


DNN Platform/Library/Services/FileSystem/FolderManager.cs
```diff
@@ -515,6 +520,12 @@ public virtual IFolderInfo AddFolder(FolderMappingInfo folderMapping, string fol
             return folder;
         }
 
+        internal virtual bool IsValidFolderPath(string folderPath)
+        {
+            var illegalInFolderPath = new Regex(string.Format("[{0}]", Regex.Escape(new string(Path.GetInvalidPathChars()))), RegexOptions.Compiled);
+            return !illegalInFolderPath.IsMatch(folderPath) && !folderPath.TrimEnd('/', '\\').EndsWith(".");
```
Yabgu: Your fix will not work for path like this: "../../../Dummy/File.txt"
You should improve your regex like this: "[{0}]|[/\\]{1,}\.\.[/\\]{1,}|^\.\.[/\\]{1,}|[/\\]{1,}\.\.$|^\.\.$"

PR #3396, opened 2019-12-10
Fixed Canonical and None mapping redirects

### LINKED ISSUE #2841: PortalAlias Settings Not Correctly Observed
## Description of bug
We just experienced this on the dnncommunity.org website.  Basically, we installed under one URL, upgraded a couple of times, then added a new URL as primary and the original URL was still being used.  From what I could tell, the first URL in the database was being used, despite a different domain name being specified as `IsPrimary` and Site Alias Mapping Mode being set to Redirect.

**IMPORTANT:**  If this is indeed a bug, it could have massive impacts on the ecosystem.  Websites may be incorrectly serving out the wrong domain names and the website owners may not have any clue that it's happening.

## Steps to reproduce
1. Install DNN 9.1.1 and use a specific URL (e.g., domain1.org, and www.domain1.org)
2. Upgrade to DNN 9.2.2
3. Upgrade to DNN 9.3.2
4. Add another domain, and set it to be the primary site alias now (e.g., domain2.org (primary), and www.domain2.org)
5. View the website using the new URL/portal alias

## Current result
Originally, it was redirecting to the correct primary URL, but a lot of time, it doesn't.

## Expected result
The primary URL should be used when the setting exists and a HTTP 301 redirect should be performed is Redirect is the mapping mode.

## Additional context
We managed to fix this by deleting the original primary alias (that was no longer primary) and adding it to the end.

## Screenshots
image

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

### LINKED ISSUE #3017: Portal alias redirecting to Primary alias when adding new language and enabling content localization
Portal alias redirecting to Primary alias when adding new language and enabling content localization. Before new language and content localization alias are working fine for canonical or none mode

## Description of bug
Portal alias redirecting to Primary alias when adding new language and enabling content localization. Before new language and content localization alias are working fine for canonical or none mode

## Steps to reproduce
1/ create a new portal with at least 1 published page
2/ Site Settings -> Languages ... add at least one additional language and enable it
3/ Enable localized content (with all pages translatable) - this should create localized copies of all pages and also create language-specific aliases
4/ translate and publish pages you want to test
5/ add an additional alias for the portal and also for each language
6/ set "Site Alias Mapping Mode" to None or to Canonical
7/ try to access a page using the additional alias - you will be always redirected to the primary alias

## Current result
When you try to access any of the secondary pages via the additional alias, you get redirected to the primary one. Eg.
https://aliasen931.dnndev.me/en-gb/what-we-do gets redirected to https://en931.morganphilips.com/en-gb/what-we-do

## Expected result
When access https://aliasen931.dnndev.me/en-gb/what-we-do it should not redirect

## Affected version
* [X ] 9.1.1
* [X ] 9.2.0
* [X ] 9.3.0
* [X ] 9.4.0 latest supported release

### LINKED ISSUE #3395: Canonical and None URL mappings don't work correctly on localized sites
## Description of bug
Canonical and None URL mappings don't work correctly on localized sites. This is pretty much what #3017 is about, but steps to reproduce are more precise.

## Steps to reproduce
1. Set up a new IIS web site for DNN 9.4.3. Name it `platform-9-4-3.dnndev.me`.
2. Install DNN 9.4.3 with default settings (as in `/install.aspx`).
3. Add non primary alias `www-platform-9-4-3.dnndev.me` (and corresponding IIS binding)
4. Ensure URL Mapping is set to `Canonical`
5. Add French as a second language
6. Enable content localization, create localized pages, mark all as translated, activate French pages, publish all French pages.
7. Create additional non-primary aliases:
7.a) `www-platform-9-4-3.dnndev.me/en-us` for English
7.b) `www-platform-9-4-3.dnndev.me/fr-fr` for French
7. Navigate to `www-platform-9-4-3.dnndev.me/fr-fr/`

## Current behavior
DNN redirects to `platform-9-4-3.dnndev.me/fr-fr` (the primary alias)

## Expected behavior
Given URL mapping is set to canonical, no redirection should be performed.

## Additional context
Some pitfalls with Chrome:
* Chrome hides "www." prefix from the URL, which can be very confusing. That's why I used "www-" (with a dash) to avoid this.
* Keep the Dev Tools panel open at all times with the cache disabled. This will get rid of confusing caching issues.

## Affected version
* [X] 10.0.0 alpha build
* [X] 9.5.0 alpha build
* [X] 9.4.x latest supported release
* [X] 9.3.x latest supported release
(not sure about older versions)

## Affected browser
N/A

daguiler:
Fixes #3395
Fixes #3017 
Fixes #2841 

## Summary
This pull request fixes wrong redirections that occur on Canonical and None URL mappings.
I have tested this as much as I could. I created following spreadsheet to track my tests and make sure this change does not introduce new regressions:
https://docs.google.com/spreadsheets/d/1c9DyWfLAG5qrAnPDZsvAg5VBJkSRWBciWtPxHARxWUA/edit?usp=sharing
Above spreadsheet contains:

Tab | Description
-- | --
DNN943 | A set of test URLs along with the expected and the actual responses when executed on DNN 9.4.3. Red colors on the right show cases where the response is not correct, as per my understanding. In such cases the Severity column shows 1 (minor issue) or 3 (big issue).
Patched | Same set of test URLs but with responses after applying this fix. Green colors in Severity column show how each test case improved with respect to original.
Bindings | IIS bindings I used in the test environment
Portals | How I set up the portals in the test environment. Main portal is localized, child portal is not localized
Tabs | Which pages I created in the test environment
Aliases | How aliases were configured.

I used following Powershell script to automate requesting each test URL:
```powershell
$domain = "platform-9-4-3.dnndev.me"
@(
	"en-$domain"
	"en.$domain"
	"fr-$domain"
	"fr.$domain"
	"$domain"
	"$domain/en-us"
	"$domain/fr-fr"
	"www-$domain"
	"www-$domain/en-us"
	"www-$domain/fr-fr"
	"www.$domain"
	"www.$domain/en-us"
	"www.$domain/fr-fr"
    "$domain/child"
    "www.$domain/child"
    "www-$domain/child"
) | % { @( $_, "$_/" ) | % {
    $response = Invoke-WebRequest "http://$_" -MaximumRedirection 0 -ErrorAction Ignore -UseBasicParsing
    $output = $_, $response.StatusCode
    If ($response.StatusCode -eq 301) {
        $output += $response.Headers['Location'], $response.Headers['X-Redirect-Reason']
    }
    $output -join "`t"
}}
```

This PR is currently targetting 9.4.4 instead of develop, but I know it needs to be re-targeted. I did this on purpose to test if it's easier to re-target from 9.4.4 than from develop.

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalAliasExtensions.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalAliasExtensions.cs
@@ -95,10 +95,6 @@ namespace DotNetNuke.Entities.Portals
         {
             var aliasList = aliases.ToList();
 
-            var defaultAlias = aliasList.ToList().Where(a => a.PortalID == portalId)
-                                        .OrderByDescending(a => a.IsPrimary)
-                                        .FirstOrDefault();
-
             //First check if our current alias is already a perfect match.
             PortalAliasInfo foundAlias = null;
             if (result != null && !string.IsNullOrEmpty(result.HttpAlias))
@@ -107,7 +103,6 @@ namespace DotNetNuke.Entities.Portals
                 foundAlias = aliasList.FirstOrDefault(a => a.BrowserType == browserType &&
                                                          (String.Compare(a.CultureCode, cultureCode,
                                                              StringComparison.OrdinalIgnoreCase) == 0)
-                                                         && a.IsPrimary
                                                          && a.PortalID == portalId
                                                          && a.HTTPAlias == result.HttpAlias);
                 if (foundAlias == null) //let us try again using Startswith() to find matching Hosts
@@ -115,7 +110,6 @@ namespace DotNetNuke.Entities.Portals
                     foundAlias = aliasList.FirstOrDefault(a => a.BrowserType == browserType &&
                                                          (String.Compare(a.CultureCode, cultureCode,
                                                              StringComparison.OrdinalIgnoreCase) == 0)
-                                                         && a.IsPrimary
                                                          && a.PortalID == portalId
                                                          && a.HTTPAlias.StartsWith(result.HttpAlias.Split('/')[0]));
                 }
@@ -164,10 +158,15 @@ namespace DotNetNuke.Entities.Portals
             }
             else
             {
+                // if we didn't find a specific match, return the default, which is the closest match
+                var defaultAlias = aliasList
+                    .Where(a => a.PortalID == portalId)
+                    .OrderByDescending(a => a.IsPrimary)
+                    .FirstOrDefault();
+
                 foundAlias = defaultAlias;
             }
 
-            //if we didn't find a specific match, return the default, which is the closest match
             return foundAlias;
         }
 
--- a/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
@@ -1700,7 +1700,9 @@ namespace DotNetNuke.Entities.Urls
                                     //var cpa = portalAliases.GetAliasByPortalIdAndSettings(result);
                                     string url = requestUri.ToString();
                                     RewriteController.CheckLanguageMatch(ref url, result);
-                                    var cpa = portalAliases.GetAliasByPortalIdAndSettings(result.PortalId, result, result.CultureCode, result.BrowserType);
+                                    var cpa = portalAliases
+                                        .Where(a => a.IsPrimary || result.PortalAliasMapping != PortalSettings.PortalAliasMapping.Redirect)
+                                        .GetAliasByPortalIdAndSettings(result.PortalId, result, result.CultureCode, result.BrowserType);
 
                                     if (cpa != null)
                                     {


valadas:
If I am not mistaken, this would Fix #3395, #3017 and #2841 Correct?

daguiler:
I'm not sure. Maybe we can add that specific scenario to the test sheet and check it out?

valadas:
Yeah, let me know if you test it before me, the changes look good to me but I did not review yet because I would like to make sure how many issues this closes :) If you have time before me to test those scenarios, please report back...

daguiler:
If I'm reading 2841 case correctly, it looks pretty much like my child site case, where there are several competing aliases, none of them having a culture code. The key difference is that in my case I have the primary alias in the first record of the PortalAlias table, whereas in 2841 the problem begun when the second row was selected as primary. So I changed the primary alias to be the one in the second row (www-* in my case):
dnn-2841-alias
I also set the mapping mode to Redirect (I think that was also the case in 2841)
With all this in place, I was not able to reproduce the issue. Every time I requested any of the 3 variants, I got correctly redirected to the selected primary.
Maybe I'm not testing 2841 correctly.

valadas:
Yeah, it was a test also to retarget, it worked without any conflict. Just need to let it rebuild and we are good to merge...

daguiler:
Thanks guys.
About 2841, there was a misunderstanding. My comment above should have been more precise. What I tried to say is that I was not able to reproduce 2841, even before applying this fix. So we can't really say this fixes 2841.

daguiler:
Maybe @hismightiness can chime in?

hismightiness:
#2841 only occurs under very specific conditions and it's not consistently reproducible, as noted in the Issue.

valadas:
Ok, do we leave #2841 closed until someone reports it back with maybe some steps to make it consistently happen?

hismightiness:
> Ok, do we leave #2841 closed until someone reports it back with maybe some steps to make it consistently happen?

I'd vote yes.  :)

PR #3397, opened 2019-12-10
Merges latest changes from 9.4.4 into develop

valadas:

Oddly even though #3374 was merged, it does not show those commits, not sure why, but let's try this again from master so we also get the merge commit from the actual 9.4.4 release


bdukes:
It looks like #3374 was squashed, rather than creating a merge commit, which is why this PR still showed a difference in commits

valadas:
Oh, this makes a lot of sense, thanks @bdukes yeah we need to not squash those updating one branch from another @dnnsoftware/approvers please.

PR #3401, opened 2019-12-12
Move all JS projects from the AE to their own folder

donker:
I realize this is opinionated, but I feel "hiding" the JS projects nested deeply inside the VS projects is a mistake as it gets hard to find and properly manage with gitignore. It also exposes the webpack stuff better and more consistently.


donker:
Closing and resubmitting after a rebase to resolve merge conflicts.

PR #3412, opened 2019-12-17
Fixed NuGet package specification

### LINKED ISSUE #3411: NuGet Package Manifests Not Correct
## Description of bug
Current NuGet packages include both an icon and iconUrl paths,  This is redundant and not per the specifications of NuGet

These should be updated to match the proper specifications as defined here: https://docs.microsoft.com/en-us/nuget/reference/nuspec.

mitchelsellers:
Fixes #3411 by removing the <iconUrl> node from all NuGet specifications to match documentation

### DIFF
--- a/Build/Tools/NuGet/Dnn.PersonaBar.Library.nuspec
+++ b/Build/Tools/NuGet/Dnn.PersonaBar.Library.nuspec
@@ -9,7 +9,6 @@
     <license type="expression">MIT</license>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <icon>images\logo-for-nuget.png</icon>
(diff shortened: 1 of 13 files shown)


bdukes:
The _Tip_ in the [`icon`](https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon) section of those docs says
>You can specify both `icon` and `iconUrl` to maintain backward compatibility with sources that do not support `icon`. Visual Studio will support `icon` for packages coming from a folder-based source in a future release.

It looks to me like we're still in a situation where it would be good to have both.

mitchelsellers:
But the "iconUrl" portion says

> iconUrl is deprecated. Use icon instead.

And it gives warnings on every publish of the packages

PR #3416, opened 2019-12-18
Added Configuration Merge Script for 9.5.0

mitchelsellers:
This PR includes a configuration merge script to be used as part of the installation/upgrade process for 9.4.0 to cleanup the installation from old data.

PR #3440, opened 2019-12-31
Moves 09.05.00.config into proper location

valadas:
#3416 was submitted in the wrong folder path. The new Website path was moved from the root into the `Dnn Platform` folder and #3416 was submitted into the now absent root Website folder.

This PR just moves the file into the proper location without changes.


mitchelsellers:
My bad....

PR #3456, opened 2020-01-08
Issues/3455 lerna build

### LINKED ISSUE #3455: Build: When building locally any changes made to dnn-react-common are not reflected and the library is always pulled from nom.
## Description of bug
If you make any changes to a module in the dnn-react-common components those changes will never be reflected upon subsequent builds. 

## Steps to reproduce
1. Go to Dnn.AdminExperience/Dnn.ReactCommon/src and update one of the components in some way. The easiest is to add a new icon or update an existing icon
2. Do the build and launch the site.
3. Navigate to the component with the changes and see they have not been put in place.

## Current behavior
The dnn-react-common package is never hoisted to the node_modules folder so the changes are never reflected. Instead, it leaves the package that was retrieved from npm in place.

## Expected behavior
The changes in dnn-react-common should have been reflected in the build

## Affected version
* [ ] 10.0.0 alpha build
* [X] 9.5.0 alpha build
* [ ] 9.4.3 latest supported release

mtrutledge:
Fixes #3455 

## Summary
Made sure that lerna knew about the workspace packages configured for yarn. Had to move React Common to the top to make sure it was built first. Also added scripts to the main package.json for a cleaner interface instead of calling lerna directly from msbuild.

### DIFF
--- a/Build/BuildScripts/AEModule.build
+++ b/Build/BuildScripts/AEModule.build
@@ -46,6 +46,6 @@
     <Message Importance="high" Text="Running Yarn for $(WorkingDirectory)" />
 
     <Yarn Command="install" WorkingDirectory="$(WorkingDirectory)" IgnoreExitCode="false" Condition="$(WorkingDirectory.Length) > 0" />
-    <Yarn Command="lerna run build --parallel" WorkingDirectory="$(WorkingDirectory)" IgnoreExitCode="false" Condition="$(WorkingDirectory.Length) > 0" />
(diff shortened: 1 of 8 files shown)


ohine:
Thanks for taking a look at this @mtrutledge! These changes make sense, but it looks like the CI build is still failing. If you need an extra set of eyes, I'm always happy to help!

mtrutledge:
Taking a look to see why the CI build is failing. It almost looks like the export bundle is trying to compile the dnn-react-common that is already compiled instead of just packaging it.

mtrutledge:
If this CI build works will @donker and @valadas please run a couple more tests on their end.

PR #3471, opened 2020-01-12
Prevents tracked versioned files to change when build script runs

valadas:
As part of CI we want to maintain the automation of versioning, but this causes a large number of tracked files to change when we run ./build.ps1 this is a pain point on developers submitting PRs.

On the other hand we also want to keep the ability to set a specific version manually for multiple reasons including debugging previous releases, etc.

This PR creates a temporary commit before all tasks run, then reverts the tree state to exactly how it was before build.ps1 ran. Also, the previous solution to this same problem was brittle since it left-over files during failed builds. This solution uses Cake Teardown method which runs even if any task in the middle fail. So it is a much more robust solution.

Also, automatic versioning will now only happen only if the user does not have a local settings file to not obstruct custom versioning.

Finally we check-in the last released version so that out of the box for developers to build towards a working site of the latest release, it produces no change at all.


valadas:
Ok, so CI chokes on not having a git user, I will adjust something up to this backup/restore does not run in CI.

PR #3482, opened 2020-01-14
Reinstate an old key that was deleted

donker:
This key was accidentally deleted during earlier work on GDPR compliance.


Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SiteSettings/App_LocalResources/SiteSettings.resx
```diff
@@ -1072,7 +1072,7 @@
     <value>Enable Localized Content</value>
   </data>
   <data name="EnableLocalizedContentHelpText.Text" xml:space="preserve">
-    <value />
+    <value>Enabling localized content allows you to provide translated module content in addition to displaying translated static text. Once localized contetnt is enabled the default site culture will be permanently set and <strong>cannot be changed</strong>.</value>
```
bdukes:
```suggestion
    <value>Enabling localized content allows you to provide translated module content in addition to displaying translated static text. Once localized content is enabled the default site culture will be permanently set and &lt;strong&gt;cannot be changed&lt;/strong&gt;.</value>
```

PR #3487, opened 2020-01-15
Use new Ban icon to indicate unauthorized users

### LINKED ISSUE #3114: Unauthorized icon is the same as the roles icon in users list
## Description of bug
gdpr features added icons to the left of the user name to indicate some statuses, for instance if they accepted the new terms, etc. (I am not 100% sure of all since I do not use that feature). But since then when looking at Unauthorized users, there is a grey shield icon to the left of the user names that is exactly like the icon on the right to edit roles for the user. This is confusing.

## Steps to reproduce
1. Ensure you have some unauthorized users
2. Go to Manager -> Users
3. In the show dropdown, select Unauthorized
4. Notice the duplicate icons

## Current result
The icons are the same which is confusing

## Expected result
We should find a different icon for indicating unauthorized users.
Maybe something like 
image

 or similar that is not used and is not a duplicate of another one.

## Screenshots
image

## Affected version

* [x] 9.4.0
* [?] 9.3.x

## Affected browser

* [x] Chrome
* [?] Firefox
* [?] Safari
* [?] Internet Explorer
* [?] Edge

donker:
Resolves #3114


david-poindexter:
@donker I edited your comment on this PR to ensure it will auto close the referenced issue for which this addresses.  For further reference, take a look at [this link for more information on the syntax requirements](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords).  Thanks.  :)

valadas:
Why not use cross_out.svg ? which is already in there and looks the same?

sleupold:
I would prefer an even more specific icon - e.g. greyed user with a question mark

valadas:
I don't know I think the chosen one is appropriate to indicate banned and that was what was in the issue... It's like a no entry singe, which I think makes sense to me. Exclamation means important to me or warning or some such, grayed out means disabled or some such, banned I like the no-entry one...

donker:
> 
> 
> Why not use cross_out.svg ? which is already in there and looks the same?

Didn't know it was there. Why didn't you mention the existence of this icon in the issue? It would have saved me a ton of time.

valadas:
>Didn't know it was there. Why didn't you mention the existence of this icon in the issue? It would have saved me a ton of time.

Sorry :)

PR #3516, opened 2020-01-22
Implemented cleanup glob support

### LINKED ISSUE #3227: Add support for wildcards in manifest cleanup component type
## Description of problem
The manifest has a cleanup component type to delete old files upon module upgrades, this is useful to get rid of unused files after some architectural change in modules. It currently supports specifying each file one by one in multiple "file" nodes or specifying a cleanup text file that lists all files one per line. Directories can also be deleted by specifying the directory name.

Problem 1: You cannot delete a directory unless it is empty, so you need to delete all files it contains before you can delete the directory. Some times, you may not know all the possible files that this folder can contain, for instance some javascript build systems use checksums in filenames for multiple reasons and that can be hard to track and keep a list of for all versions.

Problem 2: You may want to delete all files of a certain file type (example license files), so for old modules (and even for Dnn itself) it is very hard to know all the previous files that could be there but a wildcard could do the trick, such as /DesktopModules/MyModule/Licenses/*.txt then the manifest would put just the files for this version in there.

## Description of solution
Adding support for wildcards in filenames would solve both problems.
Problem 1:
```
DesktopModules\MyModule\Resources\*.*
DesktopModules\MyModule\Resources\
```
Would make sure, no mater what, that we can actually delete the folder.

Problem 2:
```
DesktopModules\MyModule\Resources\Scripts\*.js
```
Would get rid of any old javascript to be replaced by the new scripts during the module install.

Alternative 1: We could add a new package type just to support wildcards, but I think the proposed solution is simpler.

Alternative 2: We could add a property to the cleanup called let's say "ForceDelete" that would force the deletion of a folder even if it contains files, but that would only solve one problem and not the ability to delete all files matching a certain pattern only.

## Additional context
My idea would be to rewrite the file deletion logic, instead of looping through one filepath at a time, that logic would get the folder path of the specified file, then do a Directory.EnumerateFiles on it and then loop through that result and delete each found file that matches. This would not break a normal file path and name but would simply add support for wildcards with not much code change.

Also for this to work (deleting a folder and then replacing it with a new one), we need to ensure that the cleanup task always runs before the files installation. I _think_ this is already the case, but if not, that would have to be done together with this change for it to work.

Any thoughts on this? I would be willing to submit a PR as part of 9.4.x for this...

valadas:
Closes #3227

Adds support for globbing patterns in the manifest cleanup component.

Will soon link a documentation link to this PR.


valadas:
Documentation submitted for this feature at https://github.com/DNNCommunity/DNNDocs/pull/423

PR #3520, opened 2020-01-24
Fixed MVC dependency injection.

### LINKED ISSUE #3513: Dependency Injection in MVC is broken.
## Description of bug
I've identified a couple of issues with how DI was implemented for MVC in DNN.

### First issue
The first one basically fails to initialize the MVC DependencyResolver. The code that sets up the `DnnMvcDependencyResolver` is using the `Globals.DependencyProvider` property which is not yet initialized at that time. This code is part of the `Startup` class in DotNetNuke.Web.Mvc.dll (https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web.Mvc/Startup.cs#L19). Relevant snippet:

```csharp
services.AddSingleton(serviceProvider => ControllerBuilder.Current.GetControllerFactory());
services.AddSingleton<MvcModuleControlFactory>();

IDependencyResolver resolver = new DnnMvcDependencyResolver(Globals.DependencyProvider, DependencyResolver.Current);
DependencyResolver.SetResolver(resolver);
```

The call stack of the initialization is

```
DotNetNuke.Web.Mvc.dll!DotNetNuke.Web.Mvc.Startup.ConfigureServices(Microsoft.Extensions.DependencyInjection.IServiceCollection services)
DotNetNuke.Web.dll!DotNetNuke.Web.Startup.ConfigureServices(Microsoft.Extensions.DependencyInjection.IServiceCollection services)
DotNetNuke.Web.dll!DotNetNuke.Web.Startup.Configure()
DotNetNuke.Web.dll!DotNetNuke.Web.Common.Internal.DotNetNukeHttpApplication.Application_Start(object sender, System.EventArgs eventArgs)
```

The important snippet in `Application_Start` is:

```csharp
var startup = new Startup();
Globals.DependencyProvider = startup.DependencyProvider;
```

So from the first line in the `Application_Start` snippet we arrive to the `DotNetNuke.Web.Mvc.Startup.ConfigureServices` which uses the `Globals.DependencyProvider` before the second line of the `Application_Start` snippet has a chance to assign it.

### Follow up issue
After a quick dirty fix (using a Func<T> as constructor parameter of `DnnMvcDependencyResolver`, to delay the access to `Globals.DependencyProvider`) another issue occurs. The first line in the MVC `Startup` class snippet above triggers a recursive access to the Value property of a Lazy<T> that is internal to the `ControllerBuilder`. The call path is as follows:

-> `DotNetNuke.Web.Mvc.Framework.Modules.ModuleApplication.ctor`
-> `Globals.DependencyProvider.GetRequiredService<IControllerFactory>()`
-> calls the factory defined by `services.AddSingleton(serviceProvider => ControllerBuilder.Current.GetControllerFactory());`
-> `ControllerBuilder.Current.GetControllerFactory()`
-> internally calls a `SingleServiceResolver<IControllerFactory>.Current`
-> calls `DependencyResolver.Current.GetService<TService>()` (https://github.com/aspnet/AspNetWebStack/blob/master/src/System.Web.Mvc/SingleServiceResolver.cs#L52)
-> which arrives at the factory in the container again
-> which calls `ControllerBuilder.Current.GetControllerFactory()` again.

Basically what the `ControllerBuilder` (more exactly the `SingleServiceResolver`) does is to first try resolving the ControllerFactory from the current DependencyResolver, then from the instance provided by the `ControllerBuilder.Current.SetControllerFactory()` then from the default value of `DefaultControllerFactory`.

### Second follow up issue
Since the PR that fixed the MVC memory leak (https://github.com/dnnsoftware/Dnn.Platform/pull/3369/), the MVC Controllers are not registered in the container anymore. Because of that, when the `DnnMvcDependencyResolver` tries to resolve the controller in `GetService` (https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs#L42) it will return a null object. To be able to resolve controllers that have constructor dependencies (which is the whole purpose of this) they have to be registered in the container. In all integrations between a 3rd party DI Container and the MVC/WebApi pipeline that I could find they always register the controllers in the container.

### Second issue
The container request scope is not created in MVC, so scoped services are not supported. The WebApi `IDependecyResolver` interface has an `IDependencyScope BeginScope()` method defined, which is called by the platform to create the request scope. The MVC `IDependecyResolver` unfortunately does not have it (it might have been overlooked by the AspNet team). To create a request scope for MVC as well, we can use the example provided by David Fowler in this issue: https://github.com/dotnet/aspnetcore/issues/3172 (https://gist.github.com/davidfowl/563a602936426a18f67cd77088574e61). This was the only example that I could find.

Also looking through the changes in the MVC memory leak PR I've noticed that the controllers were registered as `transient` services. Which, I think, combined with the scope issue, was the actual issue for the memory leak. Basically what the dotnet core implementation of the IServiceProvider does when a scoped/transient service is resolved from the root (unscoped) container is to store them in the singleton scope. This makes the scoped services singletons and the transient services to pile up in the memory.

Because of the remark above, if we do not implement request scoping in MVC, the container will return the same controller instance (that was cached in the singleton scope even if it was registered as scoped) for every request which will make the `ControllerBase` throw the exception below from the second request onward:
singleton-controller

```
Error: Dependency Injection Sample - MVC Module is currently unavailable. DotNetNuke.Services.Exceptions.ModuleLoadException: A single instance of controller 'Dnn.DependencyInjection.Samples.Mvc.Controllers.HomeController' cannot be used to handle multiple requests. If a custom controller factory is in use, make sure that it creates a new instance of the controller for each request. ---> System.InvalidOperationException: A single instance of controller 'Dnn.DependencyInjection.Samples.Mvc.Controllers.HomeController' cannot be used to handle multiple requests. If a custom controller factory is in use, make sure that it creates a new instance of the controller for each request. at System.Web.Mvc.ControllerBase.VerifyExecuteCalledOnce() at System.Web.Mvc.ControllerBase.Execute(RequestContext requestContext) at DotNetNuke.Web.Mvc.Framework.Modules.ModuleApplication.ExecuteRequest(ModuleRequestContext context) in C:\wwwroot\dnn944source\DNN Platform\DotNetNuke.Web.Mvc\Framework\Modules\ModuleApplication.cs:line 126 at DotNetNuke.Web.Mvc.Framework.Modules.ModuleExecutionEngine.ExecuteModule(ModuleRequestContext moduleRequestContext) in C:\wwwroot\dnn944source\DNN Platform\DotNetNuke.Web.Mvc\Framework\Modules\ModuleExecutionEngine.cs:line 38 at DotNetNuke.Web.Mvc.MvcHostControl.ExecuteModule() in C:\wwwroot\dnn944source\DNN Platform\DotNetNuke.Web.Mvc\MvcHostControl.cs:line 200 --- End of inner exception stack trace ---
```

## Steps to reproduce
1. Install clean DNN 9.4.4
2. Download source code for the MVC DI sample: https://github.com/ahoefling/DNN.DependencyInjection.Samples/tree/mvc
3. Build in release and install the package generated in: `DNN.DependencyInjection.Samples-mvc\src\Dnn.DependencyInjection.Samples.Mvc\packages`
4. Add the module on a page
5. The "Make sure the controller has a parameterless constructor" exception is thrown.

## Affected version

* [ ] 10.0.0 alpha build
* [ ] 9.5.0 alpha build

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

dimarobert:
Fixes #3513 

## Summary
* Added a HttpModule (ServiceRequestScopeModule) that handles DI's request scope lifetime.
  * This enables scoped service resolutions for WebForms, MVC and WebApi.
  * It also stores the created scope inside the HttpContext.Items for easy access. This can be useful for vendors to access the request scope.

* Setup the ServiceRequestScopeModule as part of the Application_Start initialization.
* Register the WebApi controllers in the DI container.

Used example from https://github.com/dotnet/aspnetcore/issues/3172 as inspiration.

TODOs: 
* Properly include the `ServiceRequestScopeModule` HttpModule through the web.config, get rid of `PreApplicationStartMethod` assembly attribute in `ServiceRequestScopeModule` and the `Microsoft.Web.Infrastructure` package. The module should be the first in the pipeline for BeginRequest and the last for EndRequest, so that as much of the request as possible is covered.
* Improve naming and placement for some of the classes created.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
@@ -5,7 +5,9 @@
 using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
+using System.Web;
 using System.Web.Mvc;
+using DotNetNuke.Common.Extensions;
 
 namespace DotNetNuke.Web.Mvc
 {
@@ -15,20 +17,6 @@ namespace DotNetNuke.Web.Mvc
     /// </summary>
     internal class DnnMvcDependencyResolver : IDependencyResolver
     {
-        private readonly IServiceProvider _serviceProvider;
-        private readonly IDependencyResolver _resolver;
-
-        /// <summary>
-        /// Instantiate a new instance of the <see cref="DnnDependencyResolver"/>.
-        /// </summary>
-        /// <param name="serviceProvider">
-        /// The <see cref="IServiceProvider"/> to be used in the <see cref="DnnDependencyResolver"/>
-        /// </param>
-        public DnnMvcDependencyResolver(IServiceProvider serviceProvider, IDependencyResolver resolver)
-        {
-            _serviceProvider = serviceProvider;
-            _resolver = resolver;
-        }
 
         /// <summary>
         /// Returns the specified service from the scope
@@ -41,14 +29,11 @@ namespace DotNetNuke.Web.Mvc
         /// </returns>
         public object GetService(Type serviceType)
         {
-            try
-            {
-                return _serviceProvider.GetService(serviceType);
-            }
-            catch
-            {
-                return _resolver.GetService(serviceType);
-            }
+            var scope = HttpContext.Current?.GetScope();
+            if (scope != null)
+                return scope.ServiceProvider.GetService(serviceType);
+
+            throw new InvalidOperationException("IServiceScope not provided");
         }
 
         /// <summary>
@@ -62,14 +47,11 @@ namespace DotNetNuke.Web.Mvc
         /// </returns>
         public IEnumerable<object> GetServices(Type serviceType)
         {
-            try
-            {
-                return _serviceProvider.GetServices(serviceType);
-            }
-            catch
-            {
-                return _resolver.GetServices(serviceType);
-            }
+            var scope = HttpContext.Current?.GetScope();
+            if (scope != null)
+                return scope.ServiceProvider.GetServices(serviceType);
+
+            throw new InvalidOperationException("IServiceScope not provided");
         }
     }
 }
--- a/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
@@ -5,9 +5,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
-using System.Web;
 using System.Web.Mvc;
-using DotNetNuke.Common.Extensions;
+using DotNetNuke.Services.DependencyInjection;
 
 namespace DotNetNuke.Web.Mvc
 {
@@ -17,6 +16,12 @@ namespace DotNetNuke.Web.Mvc
     /// </summary>
     internal class DnnMvcDependencyResolver : IDependencyResolver
     {
+        private readonly IServiceProvider _serviceProvider;
+
+        public DnnMvcDependencyResolver(IServiceProvider serviceProvider)
+        {
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+        }
 
         /// <summary>
         /// Returns the specified service from the scope
@@ -29,7 +34,8 @@ namespace DotNetNuke.Web.Mvc
         /// </returns>
         public object GetService(Type serviceType)
         {
-            var scope = HttpContext.Current?.GetScope();
+            var accessor = _serviceProvider.GetRequiredService<IScopeAccessor>();
+            var scope = accessor.GetScope();
             if (scope != null)
                 return scope.ServiceProvider.GetService(serviceType);
 
@@ -47,7 +53,8 @@ namespace DotNetNuke.Web.Mvc
         /// </returns>
         public IEnumerable<object> GetServices(Type serviceType)
         {
-            var scope = HttpContext.Current?.GetScope();
+            var accessor = _serviceProvider.GetRequiredService<IScopeAccessor>();
+            var scope = accessor.GetScope();
             if (scope != null)
                 return scope.ServiceProvider.GetServices(serviceType);
 
--- a/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
@@ -1,0 +1,28 @@
+﻿using DotNetNuke.DependencyInjection.Extensions;
+using DotNetNuke.Web.Mvc.Framework.Controllers;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotNetNuke.Web.Mvc.Extensions
+{
+    public static class StartupExtensions
+    {
+        public static void AddWebApiControllers(this IServiceCollection services)
+        {
+            var controllerTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .SelectMany(TypeExtensions.SafeGetTypes)
+                .Where(x => typeof(IDnnController).IsAssignableFrom(x)
+                    && x.IsClass
+                    && !x.IsAbstract
+                );
+            foreach (var controller in controllerTypes)
+            {
+                services.AddScoped(controller);
+            }
+        }
+    }
+}
--- a/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
@@ -2,9 +2,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE file in the project root for full license information.
 // 
-using DotNetNuke.Common;
 using DotNetNuke.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection;
+using DotNetNuke.Web.Mvc.Extensions;
 using System.Web.Mvc;
 
 namespace DotNetNuke.Web.Mvc
@@ -13,11 +13,12 @@ namespace DotNetNuke.Web.Mvc
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddSingleton(serviceProvider => ControllerBuilder.Current.GetControllerFactory());
+            services.AddSingleton<IControllerFactory, DefaultControllerFactory>();
             services.AddSingleton<MvcModuleControlFactory>();
 
-            IDependencyResolver resolver = new DnnMvcDependencyResolver(Globals.DependencyProvider, DependencyResolver.Current);
-            DependencyResolver.SetResolver(resolver);
+            services.AddWebApiControllers();
+
+            DependencyResolver.SetResolver(new DnnMvcDependencyResolver());
         }
     }
 }
--- a/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
@@ -6,6 +6,7 @@ using DotNetNuke.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection;
 using DotNetNuke.Web.Mvc.Extensions;
 using System.Web.Mvc;
+using DotNetNuke.Common;
 
 namespace DotNetNuke.Web.Mvc
 {
@@ -18,7 +19,7 @@ namespace DotNetNuke.Web.Mvc
 
             services.AddWebApiControllers();
 
-            DependencyResolver.SetResolver(new DnnMvcDependencyResolver());
+            DependencyResolver.SetResolver(new DnnMvcDependencyResolver(Globals.DependencyProvider));
         }
     }
 }
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Web.Http.Dependencies;
+using DotNetNuke.Common.Extensions;
 
 namespace DotNetNuke.Web.Api.Internal
 {
@@ -25,7 +26,7 @@ namespace DotNetNuke.Web.Api.Internal
         /// </param>
         public DnnDependencyResolver(IServiceProvider serviceProvider)
         {
-            _serviceProvider = serviceProvider;
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
         }
 
         /// <summary>
@@ -36,7 +37,8 @@ namespace DotNetNuke.Web.Api.Internal
         /// </returns>
         public IDependencyScope BeginScope()
         {
-            return new DnnDependencyResolver(_serviceProvider.CreateScope().ServiceProvider);
+            var scope = System.Web.HttpContext.Current.GetScope();
+            return new DnnDependencyResolver(scope.ServiceProvider);
         }
 
         /// <summary>
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Web.Http.Dependencies;
 using DotNetNuke.Common.Extensions;
+using DotNetNuke.Services.DependencyInjection;
 
 namespace DotNetNuke.Web.Api.Internal
 {
@@ -37,7 +38,8 @@ namespace DotNetNuke.Web.Api.Internal
         /// </returns>
         public IDependencyScope BeginScope()
         {
-            var scope = System.Web.HttpContext.Current.GetScope();
+            var accessor = _serviceProvider.GetRequiredService<IScopeAccessor>();
+            var scope = accessor.GetScope();
             return new DnnDependencyResolver(scope.ServiceProvider);
         }
 
--- a/DNN Platform/HttpModules/DependencyInjection/ServiceRequestScopeModule.cs
+++ b/DNN Platform/HttpModules/DependencyInjection/ServiceRequestScopeModule.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Web;
+using DotNetNuke.Common.Extensions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Web.Infrastructure.DynamicModuleHelper;
+
+[assembly: PreApplicationStartMethod(typeof(DotNetNuke.HttpModules.DependencyInjection.ServiceRequestScopeModule), nameof(DotNetNuke.HttpModules.DependencyInjection.ServiceRequestScopeModule.InitModule))]
+
+namespace DotNetNuke.HttpModules.DependencyInjection
+{
+    public class ServiceRequestScopeModule : IHttpModule
+    {
+        public static void InitModule()
+        {
+            DynamicModuleUtility.RegisterModule(typeof(ServiceRequestScopeModule));
+        }
+
+        private static IServiceProvider _serviceProvider;
+
+        public void Init(HttpApplication context)
+        {
+            context.BeginRequest += Context_BeginRequest;
+            context.EndRequest += Context_EndRequest;
+        }
+
+        public static void SetServiceProvider(IServiceProvider serviceProvider)
+        {
+            _serviceProvider = serviceProvider;
+        }
+
+        private void Context_BeginRequest(object sender, EventArgs e)
+        {
+            var context = ((HttpApplication)sender).Context;
+            context.SetScope(_serviceProvider.CreateScope());
+        }
+
+        private void Context_EndRequest(object sender, EventArgs e)
+        {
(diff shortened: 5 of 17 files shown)


DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            var controllerTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .SelectMany(TypeExtensions.SafeGetTypes)
+                .Where(x => typeof(IDnnController).IsAssignableFrom(x)
+                    && x.IsClass
+                    && !x.IsAbstract
+                );
+            foreach (var controller in controllerTypes)
+            {
+                services.AddScoped(controller);
```
ahoefling: Controllers were explicitly removed from the Container because of memory leak problems. In earlier versions of Dependency Injection the controllers that got registered in the container would never be collected by the Garbage Collector. This means if you have 100,000 requests to a specific controller, you would have 100,000 instances of the controller in memory.

I am hesitant to add controllers back in unless there is a very good reason for it

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
```diff
@@ -41,14 +29,11 @@ public DnnMvcDependencyResolver(IServiceProvider serviceProvider, IDependencyRes
...
-            {
-                return _serviceProvider.GetService(serviceType);
-            }
-            catch
-            {
-                return _resolver.GetService(serviceType);
-            }
+            var scope = HttpContext.Current?.GetScope();
+            if (scope != null)
+                return scope.ServiceProvider.GetService(serviceType);
```
ahoefling: This looks right to me, I will need review some of the docs on scoped services before I could give any type of approval for adding scoped resolutions.

The only question that jumps out at me is if the scope is null instead of throwing an exception maybe we can have fallback logic. I think if we are all in on scoped containers that is unnecessary. (more to come later)

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
```diff
@@ -13,11 +13,12 @@ public class Startup : IDnnStartup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddSingleton(serviceProvider => ControllerBuilder.Current.GetControllerFactory());
+            services.AddSingleton<IControllerFactory, DefaultControllerFactory>();
```
ahoefling: Can you explain the intent of this change?

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
```diff
@@ -13,11 +13,12 @@ public class Startup : IDnnStartup
...
             services.AddSingleton<MvcModuleControlFactory>();
 
-            IDependencyResolver resolver = new DnnMvcDependencyResolver(Globals.DependencyProvider, DependencyResolver.Current);
-            DependencyResolver.SetResolver(resolver);
+            services.AddWebApiControllers();
```
ahoefling: Since the container is now stored on the Scope does this mean the controller lifecycle is managed correctly? In other words does the GC properly clean it up?

DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
```diff
@@ -36,7 +37,8 @@ public DnnDependencyResolver(IServiceProvider serviceProvider)
         /// </returns>
         public IDependencyScope BeginScope()
         {
-            return new DnnDependencyResolver(_serviceProvider.CreateScope().ServiceProvider);
+            var scope = System.Web.HttpContext.Current.GetScope();
```
ahoefling: This actually looks like it is the correct way to handle the DI Scope in System.Web

DNN Platform/HttpModules/DependencyInjection/ServiceRequestScopeModule.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Web;
+using DotNetNuke.Common.Extensions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Web.Infrastructure.DynamicModuleHelper;
+
+[assembly: PreApplicationStartMethod(typeof(DotNetNuke.HttpModules.DependencyInjection.ServiceRequestScopeModule), nameof(DotNetNuke.HttpModules.DependencyInjection.ServiceRequestScopeModule.InitModule))]
+
+namespace DotNetNuke.HttpModules.DependencyInjection
+{
+    public class ServiceRequestScopeModule : IHttpModule
```
ahoefling: Does this mean you got Dependency Injection working in HTTP Modules or is this needed to access the System.Web Scope?

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
dimarobert: This is explained in the linked issue. It triggers a recursive call inside the `ControllerBuilder` and throws an error.

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
dimarobert: yes, this SO https://stackoverflow.com/a/40845639/499220 explains how it works.

The memory leak was actually because the controllers where registered as transient and resolved from the root (un-scoped) container that was never disposed (apart from application shutdown).

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
dimarobert: see https://github.com/dnnsoftware/Dnn.Platform/pull/3520#discussion_r370655221 for the reasoning behind the memory leak.

DNN Platform/HttpModules/DependencyInjection/ServiceRequestScopeModule.cs
dimarobert: This HttpModule is the one that handles the entire request scope lifetime.
As stated in the TODOs, this should be the first HttpModule in the pipeline for BeginRequest so that the scope is opened as soon as possible in the request lifetime (all HttpModules after it can access the opened scope through the instance stored in HttpContext.Items - they will not have constructor DI though). The same HttpModule should be the last that executes it's EndRequest so that the scope is disposed as late as possible.

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
ahoefling: This makes a lot of sense, thanks for explaining 👍

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
ahoefling: perfect, thanks!

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
ahoefling: This makes sense, thanks for the explanation

DNN Platform/HttpModules/DependencyInjection/ServiceRequestScopeModule.cs
ahoefling: Perfect, thanks for the explanation on what is going on with this code. It helps me a lot and I'm sure it will help others that review this change later.

This sounds like the correct way and I would argue the way I should have implemented this in the first place. I'm actually fine keeping this as an attribute on the class. Something like this shouldn't be configured in the web.config, creates another point of failure on a part of the platform that isn't designed to be plug and play

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
mitchelsellers: I would like to re-do the same testing that was done to validate the fix that was put in place in 9.4.x to ensure that we don't have a leak, but this looks to be the better way regardless

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
mitchelsellers: I'm with @ahoefling on this one, I think this is ok, but looking to confirm

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
mitchelsellers: Awesome, this is a great discovery!

DNN Platform/DotNetNuke.Web/Api/Internal/DnnDependencyResolver.cs
mitchelsellers: @ahoefling I assume this is the note of the dependence on System.Web.  I'm ok with this as we are using System.Web now, so it makes sense, and this is internal to our stuff

dimarobert:
Here is an attempt at a complete explanation on how the service resolution works in the dotnet core DI system.

You register the services in one container (IServiceCollection). The services lifetime can be Singleton, Scoped or Transient. You then build a service provider from that container using `BuildServiceProvider`. This provider is considered the **root service provider**. It will be created at the startup of the application and disposed when the app is shutdown.

The service provider will hold a cache with all the services that it materialized throughout it's lifetime, no mater their lifetime.
When you resolve a service from the provider, based on the service lifetime, it will:
* create the service each time if it is Transient and cache it.
* try to get the service from the cache if it's there, otherwise create and cache the service for later resolutions. This applies to both Scoped and Singletons.

When the service provider is disposed it will go through the cache and dispose all services that implement IDisposable.

**From the root service provider you can create a scope.** What that means is that an IServiceScope instance is created from the root. It will contain a newly created service provider that can access the 
 root provider's cache, but has it's own cache as well. When you resolve a service from the scope it will:
* create the service each time if it is Transient and cache it.
* try to get the service from it's cache if it's there, then from the parent (root) provider's cache, otherwise it will create the service and cache it. The caching is a bit different here, if the service lifetime is Scoped, then the scope cache is used, otherwise if it is Singleton, the parent cache is used.

When the scope is disposed all the services that were stored in it's cache and implement IDisposable will be disposed.

Here is a gotcha with this: **a Scoped service that is resolved from the root provider will become Singleton**. If you think about it, when you resolve a service with Scoped lifetime from the root provider it will create and store it in it's cache, which means that for further resolutions, both the root and any scope providers that are created from it, will always return that cached instance, instead of creating a new one when needed.

Another one would be that Transient services resolved from the root provider will pile up and create a memory leak.

Because of the first one, the Root Service Provider can be created in two ways:
* `services.BuildServiceProvider(true)` - this is considered development mode and scope validations will be done before resolving a service. What it means is that if you try to resolve a Scoped service from the root provider an exception will be thrown so that you know something is wrong and the service changed it's lifetime.
* `services.BuildServiceProvider()` or `services.BuildServiceProvider(false)` - this is considered production mode, no checks are done, you are on your own.


## Conclusion
Request scope is kind of mandatory for a healthy DI architecture.
Also there should be a way to call `services.BuildServiceProvider(true)` in DNN, for development purposes and for the vendors to easily see if they make a mistake. It should be some kind of switch in `web.config`.

---
_Disclaimer: I am not completely sure if the Transient services resolved, **that are NOT implementing IDisposable**, are actually cached by the service provider, but the ones that do are definitely stored so that they are properly disposed._

DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
dimarobert: Please note that if the controllers are not registered in the container it will not be able to resolve them. Which means that in 9.4.4 (apart from the provider being always null) the controllers were always resolved from the asp net DefaultControllerFactory `_resolver` that was passed in, which only supports parameterless constructors, so no DI for them.

DNN Platform/DotNetNuke.Web.Mvc/DnnMvcDependencyResolver.cs
dimarobert: see https://github.com/dnnsoftware/Dnn.Platform/pull/3520#issuecomment-578186734 for my attempt to explain the need of a request scope.

dimarobert:
I'd rather have someone from the core team understand how everything here works in depth before we go forward.

There are a couple other things that we can try. For example for WebApi we could create a custom IHttpControllerActivator instead of the IDependencyProvider. This will give us access to the HttpRequestMessage and in turn to the HttpContextBase when resolving controllers (decouple from HttpContext.Current). But then, we would not be able to replace services in the asp net pipeline ([examples of services](https://exceptionnotfound.net/the-asp-net-web-api-2-http-message-lifecycle-in-43-easy-steps-2/)) from the container, if that is desired.
Other variant would be to create some sort of singleton service (e.g. `IScopeAccessor`) that will be updated by the HttpModule on each request with the current scope. Then in the `IDependencyProvider`s we could use the root provider to resolve the `IScopeAccessor` and resolve services from the current scope. 
We can also improve on the `IScopeAccessor` idea and add some events for scope creation/disposal. This way vendors can hook in and create their own logic. For example we (DnnSharp) have our own container for a while now. To have support in all areas (WebForms, MVC and WebApi) we had to override things in the asp pipeline and use the `PreApplicationStartMethod` assembly attribute to have a good (not ideal) composition root. Now, looking at a way to be compatible with DNN 9.4.x+, we need to still keep our internal container. That's in part also because we rely on some features that the dotnet core container does not provide, like keyed service registrations and child containers. If we'd have those events we could have a consistent way of managing our internal controller's scope.

mitchelsellers:
@ahoefling @bdukes or others, might we setup a quick conference call

ahoefling:
I'm going to be unavailable on vacation for a week. I'm more than happy to do a more thorough review when I get back the week of 2/3/2020

bogdan-litescu:
To add a bit of business context, DNN Sharp users can't upgrade to DNN 9.4 because of this issue. It would help us if we can tell users an approximate time of release and also if it will be version 9.5 or 9.4.5.

mitchelsellers:
@bogdan-litescu Can you shed any additional light on the impact here?  Were you using dependency injection PRIOR to 9.4?  This was a new feature in 9.4.x.  I just want to make sure that we aren't chasing a red-herring in your case.

dimarobert:
@mitchelsellers yes we are using our own DI for about 3 years now.

bdukes:
@bogdan-litescu @dimarobert so is DNN Sharp planning to update their DI to hook into this request scope?

bogdan-litescu:
@bdukes Robert is implementing a layer of abstraction so on DNN <9.4 we use our previous DI library (DryIoc it's called by the way - very powerful, very lean) and on DNN 9.4+ it will use the one that ships with DNN. With DryIoc we already hook into request scope, MVC, Web API, etc. and actually, that's what's causing the conflicts right now. Our goal is to get similar capabilities with the DNN implementation, otherwise, we'll have to downgrade a lot of code.

By the way, if there's a chance to adopt DryIoc as the DI provider in DNN we would like to have that conversation. The one in .Net core is pretty basic.

bdukes:
I've run a few verification tests, this is look good to me :+1:
* I've installed the MVC module from https://github.com/ahoefling/DNN.DependencyInjection.Samples/pull/3 and verified there's no error with this patch (and there _is_ an error without this patch)
* I've installed the test MVC module from https://github.com/dnnsoftware/Dnn.Platform/pull/3520 and the test Web API from https://github.com/dnnsoftware/Dnn.Platform/issues/3344 and verified that hitting those URLs repeatedly does not cause an increase in memory usage.

mitchelsellers:
@bdukes I too have tested this in this manner.

There is a TODO: Note from the user about moving the inclusion to the web.config, however, given the nature of how this is done etc, I actually think that the safer route is to leave this implementation as-is and use the attribute.

Thoughts?  Do we merge this in since it has been validated and we know that 9.4.4 is broken?

bdukes:
Yes, I agree, leaving it as is would be the best approach, and I'd be good with merging it for 9.5

bdukes:
Confused dog GIF

mitchelsellers:
@bdukes Same here....but it appears that a build is in progress......

dimarobert:
@bdukes @mitchelsellers @ahoefling I've made some changes/improvements by adding the IScopeAccessor that I was talking about https://github.com/dnnsoftware/Dnn.Platform/pull/3520#issuecomment-578204654. It gets rid of the System.Web dependency in the MVC and WebApi projects.
Take a look https://github.com/dimarobert/Dnn.Platform/pull/1

If that is something that you think is worth it, merge that PR so that the changes reflect here.

bdukes:
Thanks @dimarobert, that change looks good to me.  I'm going to build it and re-run my tests, but I think we can go ahead and include that change.

bdukes:
@dimarobert the build is failing with three scope related null reference exceptions.  Can you look into those? https://dev.azure.com/dotnet/DNN/_build/results?buildId=27962&view=ms.vss-test-web.build-test-results-tab&runId=1085634&resultId=102042&paneView=debug

bdukes:
Not sure if adding your IScopeAccessor changes will have any effect on those tests, either…

dimarobert:
@bdukes I've made an update to the https://github.com/dimarobert/Dnn.Platform/pull/1 PR to fix the unit tests. The fix was simple due to the IScopeAccessor.

bdukes:
Thanks @dimarobert! I've updated this PR with those changes.  They also passed my manual testing (no errors, no memory leaks).  I'm good to merge this @mitchelsellers if you are.

david-poindexter:
Awesomeness @dimarobert - great teamwork!

mitchelsellers:
Thanks everyone!  This is merged!

PR #3525, opened 2020-01-27
Update Default Website template links

### LINKED ISSUE #3163: Default home page out-of-the-box install gives wrong impression
## Description of problem
When you install DNN with a default template, you get a home page that moves you in a wrong direction: all links point to DNNSoftware.com, there is no clear instruction what to do next, there is no link to the extension overview, there is no link to the DNN Docs site, etc.

## Description of solution
A new default home page.

Change links; a completely new template; multiple templates

## Screenshots
image
.

## Additional context
A new template should be made available in other languages as well

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

## Affected DNN Versions

* [x] DNN941 and earlier

david-poindexter:
## Summary
Resolves #3163 

Default Website Template 3


valadas:
If you could at the same time add this:
image

it would make the cursor look like a hand on the actual link (which is a label for some reason)

david-poindexter:
> If you could at the same time add this:
> image
> 
> it would make the cursor look like a hand on the actual link (which is a label for some reason)

The cursor styling has now been pushed via new commit.

PR #3528, opened 2020-01-29
Don't expose file last modifed date/time via cachebusted URLs

### LINKED ISSUE #2850: RFC: Don't expose file last modifed date/time via cachebusted URLs
## Description of problem

Currently raw file last modifed date/time is used to produce cachebusted file URLs, like `http://mysite.com/myfile.pdf?ver=2018-04-20-144725-343`. I thinks there are cases then we do not want the information about file modification date/time to be exposed in public. Just a sample case - the site owner delayed publishing some important document, but still want it to look just like it is published in time.

## Description of solution

Just use some hash string instead of raw file last modifed date/time, something like this: `http://mysite.com/myfile.pdf?ver=MjAxOC0wNC0yMC0x`.

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

roman-yagodin:
Closes #2850

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -634,12 +634,9 @@ namespace DotNetNuke.Entities.Portals
             }
         }
 
-        /*
-         * add a cachebuster parameter to generated file URI's
-         * 
-         * of the form ver=[file timestame] ie ver=2015-02-17-162255-735
-         * 
-         */
+        /// <summary>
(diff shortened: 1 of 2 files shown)


bdukes:
❓ I don't see any new portal setting when I look at the diff.  Am I missing something @mitchelsellers?

mitchelsellers:
@bdukes GitHub's diff display was being stupid

bdukes:
Thanks @roman-yagodin!

PR #3544, opened 2020-02-05
DI: Register MVC/WebApi Controllers using TryAddScoped.

### LINKED ISSUE #3543: DI: Allow overriding MVC/WebApi Controller registrations.
## Description of problem
With the current way of registering the services in the DI Container it is a bit difficult (even impossible in some cases) to override Controller registrations.
In the case of WebApi Controllers, they are registered after all IDnnStartup implementations, at the end of [`DotNetNuke.Web.Startup.ConfigureServices()`](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web/Startup.cs#L61).
As for the MVC Controllers, they are registered in the MVC IDnnStartup. The iteration over the IDnnStartup implementations does not have a specific ordering, which means that most likely they will be called based on the name of the containing assembly. Because of this, the MVC Startup might be executed after some other vendors Startup's.
This ordering issue might affect other areas as well.

## Description of solution
The direct solution for the issue at hand would be to register the controllers using the [`TryAddScoped(controllerType)`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddscoped?view=dotnet-plat-ext-3.1#Microsoft_Extensions_DependencyInjection_Extensions_ServiceCollectionDescriptorExtensions_TryAddScoped_Microsoft_Extensions_DependencyInjection_IServiceCollection_System_Type_) IServiceCollection extension method, which will add the controller to the collection only if the controller hasn't already been registered.

A solution to the broader issue, would be to either define an ordering/priority scheme, that could span multiple levels, either just calling the DNN IDnnStartup implementations first and then the rest.

## Affected browser
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

dimarobert:
Fixes #3543 

## Summary
This is a quick fix for the assigned issue.
This will allow vendors to add/replace controller registrations with their own.
The DefaultControllerFactory is also added to the ServiceCollection using the `TryAddSingleton` for the same reasons.

Fix AddWebApiControllers typo for AddMvcControllers.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using DotNetNuke.DependencyInjection.Extensions;
 using DotNetNuke.Web.Mvc.Framework.Controllers;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -11,7 +12,7 @@ namespace DotNetNuke.Web.Mvc.Extensions
 {
     public static class StartupExtensions
     {
-        public static void AddWebApiControllers(this IServiceCollection services)
+        public static void AddMvcControllers(this IServiceCollection services)
         {
             var controllerTypes = AppDomain.CurrentDomain.GetAssemblies()
                 .SelectMany(TypeExtensions.SafeGetTypes)
@@ -21,7 +22,7 @@ namespace DotNetNuke.Web.Mvc.Extensions
                 );
             foreach (var controller in controllerTypes)
             {
-                services.AddScoped(controller);
+                services.TryAddScoped(controller);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


dimarobert:
@bdukes I see that this is part of the 9.5.0 milestone, but what do you think are the chances that it will get merged by then?

bdukes:
@valadas @ahoefling @mitchelsellers @donker any opinions about getting this change in 9.5.0?  I believe this is blocking DNN Sharp modules from working on 9.5.0.

valadas:
I have no objection merging this in 9.5.0 providing we redo some tests on MVC and SPA modules, but I am no expert in the DI area, you those who know more than me agree, I have no issue retargetting to 9.5.0 personally... Can anyone test DnnSharp modules with this change?

dimarobert:
> I have no objection merging this in 9.5.0 providing we redo some tests on MVC and SPA modules, but I am no expert in the DI area, you those who know more than me agree, I have no issue retargetting to 9.5.0 personally... Can anyone test DnnSharp modules with this change?

We'll have new builds in the next couple of days that should work with 9.5.0. With this PR it should be easier for us to maintain both DNN versions (pre 940 and 950+) without separate install packages for our products.

bogdan-litescu:
Please let us know if this can be merged in DNN 9.5 Thanks!

valadas:
Since 9.5.0 was code frozen for about 2 weeks and would get officially released like tomorrow, this pr would not have not went through the testing team. My vote would be to release 9.5.0 as it is ASAP and do a quick 9.5.1 Rc1 with this and other bug fixes that are ready this week with just a 1 week RC stage... Opinions?

bdukes:
Thanks @dimarobert for the contribution, unfortunately we dropped the ball getting this reviewed for 9.5.0, but this will be included in 9.5.1 (which we're planning to fast-track, RC should be released soon).

PR #3550, opened 2020-02-07
Fix for DisplayUserName

### LINKED ISSUE #3157: Registration: Users are unable to register when Display Name Format setting is set
## Description of bug
When user tries to register, they receive error: "The Display Name is invalid."

## Steps to reproduce
1. In the Registration Settings (PersonaBar: Settings -> Security -> Member Accounts -> Registration Settings) set **Display Name Format** field to [FIRSTNAME] [LASTNAME]
2. Log out
3. Click on Register link.
3. Try to register as different user
4. You will receive error: "The Display Name is invalid."

## Current result
So problem is, that users can not register themselves on the web, when Display Name Format setting is set. No matter what setting is Registration Form Type (Standard or Custom).
Workaround is to show Display name in registration form and clear Display Name Format setting, but it is a bit user aggressive.

## Expected result
Provide a clear and concise description of what you expected to happen.

## Screenshots
image

## Additional context
I also noticed that when the registration form type is set to Custom the setting Use Email Address as Username disappear, but affects the system. I would expect Email Address as Username  setting to be visible no matter what registration form type.

## Affected version
This problem occurs after upgrade to 9.4.1 release version.

SCullman:
Fixes  #3157
Change is minor - if no displayname can be set directly, is must not be checked if it is valid.

### DIFF
--- a/DNN Platform/Website/DesktopModules/Admin/Security/Register.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/Admin/Security/Register.ascx.cs
@@ -591,12 +591,14 @@ namespace DotNetNuke.Modules.Admin.Users
                 {
                     CreateStatus = UserCreateStatus.InvalidLastName;
                 }
-
-                var cleanDisplayName = PortalSecurity.Instance.InputFilter(User.DisplayName, filterFlags);
-                if (!cleanDisplayName.Equals(User.DisplayName))
-                {
-                    CreateStatus = UserCreateStatus.InvalidDisplayName;
-                }
+				if (string.IsNullOrEmpty(PortalSettings.Registration.DisplayNameFormat))
+				{
+					var cleanDisplayName = PortalSecurity.Instance.InputFilter(User.DisplayName, filterFlags);
+					if (!cleanDisplayName.Equals(User.DisplayName))
+					{
+						CreateStatus = UserCreateStatus.InvalidDisplayName;
+					}
+				}
             }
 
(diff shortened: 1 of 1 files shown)


DNN Platform/Website/DesktopModules/Admin/Security/Register.ascx.cs
```diff
@@ -591,12 +591,14 @@ private bool Validate()
                 {
                     CreateStatus = UserCreateStatus.InvalidLastName;
                 }
-
-                var cleanDisplayName = PortalSecurity.Instance.InputFilter(User.DisplayName, filterFlags);
-                if (!cleanDisplayName.Equals(User.DisplayName))
-                {
-                    CreateStatus = UserCreateStatus.InvalidDisplayName;
-                }
+				if (string.IsNullOrEmpty(PortalSettings.Registration.DisplayNameFormat))
```
bdukes: This should be validated if `DisplayNameFormat` is _not_ empty, right?

```suggestion
				if (!string.IsNullOrEmpty(PortalSettings.Registration.DisplayNameFormat))

```

SCullman: No. Only if DisplayNameFormat is empty, DisplayName can be set and has to be validated.
If DisplayNameFormat is set, DisplayName is NULL at that time and set later during user creation.

SCullman:
@bdukes , thanks for your quick review!

Issue #3563, opened 2020-02-13
Registration Button DIV tag and class missing


thabaum:
## Description of bug
DIV tag and class missing for registration buttons

## Steps to reproduce
List the precise steps to reproduce the bug:
1. Go to registration page
2. the registration buttons are in a UL tag

## Current behavior
Registration buttons are not able to be styled and are outside of a div tag

## Expected behavior
list of buttons should be inside the DIV tag with the appropriate class

## Screenshots
If applicable, provide screenshots to help explain the bug.
![image](https://user-images.githubusercontent.com/13577556/74486649-522adb00-4e72-11ea-865d-623c26a3a337.png)

## Additional context
I looked at my PR and these were present version, some reason in the 9.5 RC this DIV tag and class ended up missing after the merge.

## Affected version

* [ ] 10.0.0 alpha build
* [ x] 9.5.0 RC1
* [ ] 9.4.4 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge (Classic)
* [ ] Microsoft Edge Chromium

valadas:
This is what I see using the default theme:
![image](https://user-images.githubusercontent.com/6371568/74490852-dd6a9780-4e97-11ea-9d22-fa95495f2aaa.png)

The .dnnActions class makes that ul float:none; and display:block thus clearing any previous controls. Are you using a different theme on that site?

valadas:
> some reason in the 9.5 RC this DIV tag and class ended up missing after the merge

So far as I can see, there was no Div tag wrapping those controls even if I go as far back as 7.4.2
https://github.com/dnnsoftware/Dnn.Platform/blob/release/7.4.2/Website/DesktopModules/Admin/Security/Register.ascx

Can you explain a bit more what you mean by "after the merge", what merge?

thabaum:
You are looking at the popup modal.

Maybe one step is to put the module on a page

thabaum:
That was the issue, in the popup modal it works fine like your screenshot however if you put the module on a page it looks like my screenshot.

I can't remember why I had that DIV Tag in my fork maybe I added it while I was working on moving the user registration fields.

Maybe that PR will still work out that issue for modules 

If on a small window the register and cancel buttons go off the screen.  No flex...

valadas:
Shows up fine to me even with popups disabled here:

[...]

PR #3564, opened 2020-02-13
Added missing registerButtonGroup div and class

thabaum:
Fixes https://github.com/dnnsoftware/Dnn.Platform/issues/3563
## Summary

added the missing div tag around the buttons.


valadas:
Actually, I cannot reproduce the original issue, let me comment there...

valadas:
After digging around a bit:

This behaviour comes from admin.css which overrides the display:block; in favour of display: inline-block; for that ul

I would prefer a solution where `inline-block` gets replaced by `block` in that file. On a related note, admin.css was for modules placed directly in the ControlBar way back right old-timers?

thabaum:
I will try to work with creating a CSS Fix for this type of stuff.  I now remember what I had going on I thought I had forgot that line of code but it was only an issue while logged in as administrator.

Thanks @valadas

PR #3569, opened 2020-02-16
Persona bar: Show page title on hover, show more of long page titles

### LINKED ISSUE #3090: Long page titles are cropped in the Pages editor (UX enahncement)
When working with nested pages, or pages with long names, the page name gets cropped after only 10 or 12 characters:

image

There's no obvious way to see the full page name without clicking on and loading it in the the right pane.

Would it be better if the full page name could be shown when the user hovers over/taps/tabs to the link  (depending on their input device)?

OllyHodgson:
Fixes #3090.

1. No longer truncates the page name. This means screen readers should be able to read the entire page name.
1. Handles the text-overflow using CSS.
1. Adds the page name to the `title` attribute of the list item. Most desktop browsers will show this when hovering over the page name.
1. Tightens up the spacing a little, so slightly longer page titles fit. I'm pretty sure there's more we can do here, but I think this is a decent start.

I've tested in latest Chrome, Edge (Chromium) and Firefox. Could do with testing in a wider variety of browsers.


valadas:
Nice one, thanks!

PR #3570, opened 2020-02-16
Removes obsolete admin.css styling

thabaum:
Resolves https://github.com/dnnsoftware/Dnn.Platform/issues/3563

## Summary

Removes all current admin.css styling which includes old ribbon bar, edit bar and module CSS styling that takes effect when logged in as an administrator.


thabaum:
I tried testing out everything on the file manager and all admin actions in persona bar I have not ran into any issues as of yet. We can put some things back as needed as it is a clean slate to start with.

I noticed it helped align the settings/edit/move boxes that I always had problems with to where they need to be along with fixing the register button issue while logged in as admin.

I think the styles are all being applied by the new theme as needed.  Only concern I had was if anyone still used ribbon bar or edit CP but I figured in 9.5 they should consider moving to persona bar if they have not yet.

So this actually fixes 2 issues I was having 👍

valadas:
I agree, there are some z-indexes for the action menus and stuff. Still good but has a small potential for a breaking change on 3rd party themes maybe. I would keep this but merge it for 10 to be safe. Thoughs?

valadas:
Sorry, hit close by mistake, just reopened

thabaum:
IMO I think this cleanup should be safe for version 10 for sure. I can address that in another PR after this one has been completed for version 9.  I can use this PR and make minimal adjustments to fix the icons for modules aligning and the register button issue that I noticed it had cured and remove the edit and ribbon bar CSS.  While I am in here I will see if there is anything else I can help with.

That way I can make changes to version 9 for version 10 or does that matter?  I am not sure how the changed files in 9 might effect version 10 files.   So to play it safe I am thinking I need to make CSS changes here first in 9 then make a 10 version.  If it does not matter you can approve this for 10 and I will make a new version 9 PR.

I wish all the different CSS files in DNN could be edited in the CSS editor.  It would be nice to have easy access to files like these in DNN 10.

I am also very concerned with the CSS file load order which I will bring up that issue soon.  I think this has a bit of an issue with it that has been causing a lot of grief for CSS designers and module developers.  

For version 9 maybe I can release another PR with minimal adjustments to fix the icons for modules aligning and the register button issue that I noticed it had cured and remove the edit and ribbon bar CSS.

I can make another PR with this version for version 10.

We need to create one fix for version 9 to sweeten up this version as far as it can go.  The inline block and module icon CSS bugs DNN 9, I can leave all other css intact CSS and adjust those.

That and the CSS for the issue recently for the admin logs drop down which seems pretty buried in the code I spent a long time reading all the logic there trying to locate it.

I will be poking around here for the next couple weeks it looks like to dial these three issues in and move on.

thabaum:
I just went through all of the CSS trying to see if there was anything that relates to DNN 9.5

Do we use a Control Bar? Edit Bar? or any of the old CP's?  That is all the Z stuff relates to... I dont see that anything here effects my current installation of DNN 9.5

In fact things seem perfectly fine.  Can anyone get an example of when any of these things effect version 9.5 of DNN?  Persona Bar controls everything except couple pages I have had no issues with.  I think it is a bunch of unnecessary code that could lighten up DNN in my opinion here now unless someone can show an example of where any of this matters in the current version of DNN 9.

sleupold:
Due to the fact that all admin modules are PersonaBar extensions since DNN 9.2, any old control panel cannot be used any longer.

valadas:
After discussing this in the approvers meeting, due to the potential breaking nature of these changes we will be targeting to merge this as part of Dnn10 for now.

thabaum:
Thanks everyone for the help here!  Thank you @bdukes for the rebase.

PR #3575, opened 2020-02-18
Don't expose file last modifed date/time via cachebusted URLs

valadas:
Last merge of development into future/10 to delete the development branch


valadas:
This is a release management task and we are self-approving it for that reason

PR #3576, opened 2020-02-18
Merges latest changes from release/9.5.0 into future/10

valadas:
## Summary
Merges latest changes from release/9.5.0 into future/10


valadas:
Meh, will have to submit another PR with a manual merge...

Issue #3578, opened 2020-02-19
Portal.CSS is not loading last overriding all other stylesheets


thabaum:
## Description of bug
When editing Portal.CSS you have to pretty much use !important to get any CSS changes to take effect.  This is due to the load order of the css stylesheets

## Steps to reproduce
ONE EXAMPLE:

try to change logo size using Page.css.

  1  create a page
  2  upload a pagename.css to the root site directory
  3  include the following in the pagename.css file

` .navbar .brand {max-width:50px;}`

  4  Add pagename.css as Page Stylesheet in PersonaBar Page Settings editor
  5  Refresh the page notice nothing happens
  6   put the same code in Portal.css and it works however other CSS stylesheets load after Portal.css taking priority. I had to really dig to find where "CustomStylesheet" or Page Stylesheet was located. Funny part is Portal.css is called "Custom CSS" while Page.css files are "CustomStylesheet" and load way before default CSS loads with NO priority it just loads.

## Current behavior
Portal.CSS styles are loaded prior to module extensions CSS

## Expected behavior
Portal.css is loaded last and overrides all other styles as needed

## Additional context
Changing the load order of style sheets should fix the issue

## Affected version

* [x ] 10.0.0 alpha build
* [x ] 9.5.0 alpha build
* [x ] 9.4.4 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge (Classic)
* [ ] Microsoft Edge Chromium

sleupold:
I noticed the wrong order of CSS files loaded in client projects as well. 
AFAIR the proper order should be:
- default.css
- module.css of any module on the page
- skin.css (or skinname.css)
- container.css for any module container used on the page
- portal.css
- page.css (css file specified in page settings)

sleupold:
In addition, we might need to provide an option for skin developers, not to load just skin.css and container.css, but additionally dependent css files (menu.css, slider.css) which would need to be included in combined css file as well - same for module developers for module.css.

thabaum:
My current PR changes only DefaultPriority to 99 and Portal.css to 100.

I will look for when Page.css is loaded and what priority which I am guessing it is "DefaultPriority" so we need to possibly give it a priority of it's own as it is not listed below.

I can see a potential issue here.

[...]

PR #3579, opened 2020-02-19
Load DefaultPriority Css prior to PortalCss Priority Load Order

thabaum:
Portal CSS seems to load after modules and other things load causing issues when modifying a portal CSS.   Loading portalcss last will ensure these styles override all other previous css stylesheets loaded.

Fixes https://github.com/dnnsoftware/Dnn.Platform/issues/3578

## Summary

Moves DefaultPriority CSS (not to be confused with default.css which loads priority first 5) to 99 and PortalCss to 100 so it loads last no matter how many stylesheets get loaded first.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client/FileOrder.cs
+++ b/DNN Platform/DotNetNuke.Web.Client/FileOrder.cs
@@ -92,9 +92,9 @@ namespace DotNetNuke.Web.Client
         public enum Css
         {
             /// <summary>
-            /// The default priority (100) indicates that the ordering will be done based on the order in which the registrations are made
+            /// The default priority (99) indicates that the ordering will be done based on the order in which the registrations are made
             /// </summary>
-            DefaultPriority = 100,
+            DefaultPriority = 99,
 
             /// <summary>
             /// The default.css file has a priority of 5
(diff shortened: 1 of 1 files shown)


thabaum:
This maybe for version 10 but thought I would add it for a fix for 9 if OK, this should fix the load order of CSS styles making it easier to style and not have to use !important anytime you wish to style a vendor module.

bdukes:
Modules _should_ be using `module.css`, which would load with a priority of 10.  If they're using the default priority, that's a bug that the vendor should fix.  Making a change like this can have major impacts upon existing themes, so we definitely couldn't do anything until DNN 10, if we decided this change was worth it.

mitchelsellers:
I agree with @bdukes this would be a huge breaking change and I believe it is more indicative of a module vendor issue than a platform issue.

thabaum:
OK I can target this change for version 10 as I felt this was correct just didnt want to rebase but I did want some feedback on the change.

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request  looks like I have some schooling to do here to rebase a PR.

I feel Portal.css should load last no matter how many CSS files are added by anyone including the platform.

A lot of things in DNN hard code I believe uses "DefaultPriority" which would load last over Portal.css which is incorrect.  Portal.css should load last.  This is what everyone is expecting to happen and what I have been told since the beginning of my DNN years.  As you can see here in this PR that in fact is not the case.

I probably will just create a new PR and close this one with the changes.   Any other adjustments now is the time please let me know.

Thank you.
Cody

thabaum:
I am going to close this for now... I will present another PR for version 10 once I inspect Page.css load time to ensure it loads last.  And address any other issues presented in the issue discussion mentioned in this PR.

PR #3586, opened 2020-02-19
Data Consent JS

DNNMonster:
Js to enable the data consent flow on custom logins.  Not sure why I'm seeing other verified commits with the PR.  I'll resubmit if required.


DNNMonster:
I'll close it the PR.  I branched off master from a fresh clone of the repo.  Figured that would go in clean but it does not.  Grab the code if it works. I'll git better.

PR #3588, opened 2020-02-20
get the real client ip address.

### LINKED ISSUE #3587: ip filter won't work under load balance environment
## Description of bug
if DNN running under load balance environment, the ip filter of restrict admin user login won't work as the ip address will always get the load balance server's ip.

## Steps to reproduce
1. install the DNN installation under load balance environment;
2. set up the ip filter to only allow your internet IP to login;
3. browse to the website and try to login host user.

## Current behavior
you host user won't able to login.

## Expected behavior
the host user should able login via your internet network.

## Error information
the page will display "you are not allowed to login from the ip address".

## Affected version
* [ ] all versions

zyhfish:
Fix #3587

### DIFF
--- a/DNN Platform/Library/Common/Utilities/NetworkUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/NetworkUtils.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Net;
 using System.Net.Sockets;
+using System.Web;
 
 #endregion
 
@@ -241,6 +242,24 @@ namespace DotNetNuke.Common.Utils
                 return false;
             }
             return false;
+        }
+
+        /// <summary>
+        /// Get the Client IP Address.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static string GetClientIpAddress(HttpRequest request)
+        {
+            var ipAddress = request.ServerVariables["HTTP_X_FORWARDED_FOR"];
+
+            // If there is no proxy, get the standard remote address
+            if ((string.IsNullOrWhiteSpace(ipAddress)) || ipAddress.Equals("unknown", StringComparison.OrdinalIgnoreCase))
+            {
+                ipAddress = request.UserHostAddress;
+            }
+
+            return ipAddress;
         }
     }
 
--- a/DNN Platform/Library/Common/Utilities/NetworkUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/NetworkUtils.cs
@@ -245,10 +245,10 @@ namespace DotNetNuke.Common.Utils
         }
 
         /// <summary>
-        /// Get the Client IP Address.
+        /// Gets the Client IP address of the current request, from server variables if available, otherwise returns Request.UserHostAddress.
         /// </summary>
-        /// <param name="request"></param>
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Common/Utilities/NetworkUtils.cs
```diff
@@ -242,6 +243,24 @@ public static bool IsIPInRange(string currentIP, string startIP, string subnetma
             }
             return false;
         }
+
+        /// <summary>
+        /// Get the Client IP Address.
```
mitchelsellers: Since this is a new addition I'd like to get a bit more detail on this.  Something like

Gets the Client IP address of the current request, from server variables if available, otherwise returns Request.UserHostAddress

DNN Platform/Library/Common/Utilities/NetworkUtils.cs
```diff
@@ -242,6 +243,24 @@ public static bool IsIPInRange(string currentIP, string startIP, string subnetma
...
+        /// </summary>
+        /// <param name="request"></param>
```
mitchelsellers: Add

The current HttpRequest to the param documentation

DNN Platform/Library/Common/Utilities/NetworkUtils.cs
```diff
@@ -242,6 +243,24 @@ public static bool IsIPInRange(string currentIP, string startIP, string subnetma
...
+        /// <returns></returns>
```
mitchelsellers: Add 

The current ClientIpAddress

CMarius94:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) the specified header can return a list of IPs due to proxies. In our development environment we had this bug because we already have an internal method for getting the UserIP (MyTokens geolocation token, for example). 

Some load balancers/proxies do add their IP to this header, others don't, others use another header field. It all depends on their configuration, but in order to prevent unwanted behavior, i think you should split by comma and use the first IP found in the header.

DNN Platform/Library/Common/Utilities/NetworkUtils.cs
```diff
@@ -242,6 +243,24 @@ public static bool IsIPInRange(string currentIP, string startIP, string subnetma
...
         }
+
+        /// <summary>
+        /// Gets the Client IP address of the current request, from server variables if available, otherwise returns Request.UserHostAddress.
+        /// </summary>
+        /// <param name="request">The current http request.</param>
+        /// <returns>The current client ip address.</returns>
+        public static string GetClientIpAddress(HttpRequest request)
+        {
+            var ipAddress = request.ServerVariables["HTTP_X_FORWARDED_FOR"];
```
mitchelsellers: Per the comment by @CMarius94 and confirmed with details linked here https://stackoverflow.com/questions/753645/how-do-i-get-the-correct-ip-from-http-x-forwarded-for-if-it-contains-multiple-ip it looks like we should split on a comma here and use the first value

zyhfish: thanks for the review, updated the code

PR #3591, opened 2020-02-20
Search preview overflow

DNNMonster:
Add js to set preview display height.


DNNMonster:
Hmm..  Seem to be working for the devices I've tested.  But you're right it can be optimized a bit more.  We'll close the PR since it's also got the polluted git stuff going on.

PR #3593, opened 2020-02-21
Updates future/10 with latest changes from 9.5.0

valadas:
Merges latest changes from 9.5.0 into future/10


valadas:
@mitchelsellers and @bdukes as discussed, the two last commits are the merge issues we where having. One about the new default portal template and the other about the command that fires up the frontend build. I built this and did not spot anything wrong but yeah because of the folder moves, it looks like 6000 files but each move counts as 1 deleting and one creation, soooo. yeah, fun times, but it finally builds fine.

bdukes:
@valadas I just pushed a commit that resolved a bunch of differences between my merge and this merge.  I haven't yet done any deep comparison of other files that moved (in the `Build` and `Dnn.AdminExperience` folders).

valadas:
Wow, I have very bad internet right now, but yeah, I will try another build when able and take a look

valadas:
If I look at your commit, I see like 6000 changes, is that right?

bdukes:
If you look at it in the context of the PR, it still shows the number of changed files for the whole PR.  If you look at it as [an individual commit](https://github.com/valadas/Dnn.Platform/commit/bd15db184b4487daf3ad9cdc266f9a9c1b84d3a8), it's 208 files changed.  Most of those are copyright headers.

valadas:
@bdukes cool, got it, makes sense to me and I ran a build locally and tested around, looks fine 2 me...

bdukes:
@valadas yeah, if it's passing checks, I'm okay to merge it.  I'll try to find some time next week to do a couple more comparisons to make sure we didn't miss anything, and can submit any fixes as separate PRs

valadas:
Would you be fine if we merge this before the conference and fix later if we find any issues with it though?

PR #3600, opened 2020-02-28
Change Css Priority File Order + Add PageCss Priority #3578

thabaum:
https://github.com/dnnsoftware/Dnn.Platform/issues/3578

## Summary

What was done:

2 files changed
FileOrder.cs
modifies the DefaultPriority for all other CSS to 90
Sets PortalCss priority to 95
Creates and Sets PageCss priority to 100

Default.aspx.cs
Moves CustomStylesheet (Page.css) to load after all other stylesheets both programmatically in page source and using the new stylesheet file load order PageCss  priority.

This should be labeled "Fixing Change" but it is in fact probably going to break someones CSS.  I almost dont thing it will because the files are custom page and portal css files getting priority like everyone expects.  Most CSS Styles on these files will have !important.

Everytime module vendors support using Portal.css file to fix anything they always say use !Important.

I think this discussion may have started in the DNN Events module github issues and after talking to you and getting the same from others in the past I had to do some due diligence because I didn't believe things loaded in proper order.

Do you want this in version 9 for 9.6?

Steps to reproduce on of the issues a developer would deal with:

> try to change logo size using Page.css.
> 
>     create a page
>     upload a pagename.css to the root site directory
>     include the following in the pagename.css file
> 
>     .navbar .brand {max-width:50px;}
> 
>     Add pagename.css as Page Stylesheet in PersonaBar Page Settings editor
>     Refresh the page notice nothing happens
>     put the same code in Portal.css and it works however other CSS stylesheets load after Portal.css taking priority. I had to really dig to find where "CustomStylesheet" or Page Stylesheet was located. Funny part is Portal.css is called "Custom CSS" while Page.css files are "CustomStylesheet" and load way before default CSS loads with NO priority it just loads.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client/FileOrder.cs
+++ b/DNN Platform/DotNetNuke.Web.Client/FileOrder.cs
@@ -92,9 +92,10 @@ namespace DotNetNuke.Web.Client
         public enum Css
         {
             /// <summary>
-            /// The default priority (100) indicates that the ordering will be done based on the order in which the registrations are made
+            /// The default priority (90) indicates that the ordering will be done based on the order in which the registrations are made.  
+			/// PortalCss then PageCss load after DefaultPriority has loaded all assigned css pages.
             /// </summary>
-            DefaultPriority = 100,
+            DefaultPriority = 90,
 
             /// <summary>
             /// The default.css file has a priority of 5
@@ -147,9 +148,14 @@ namespace DotNetNuke.Web.Client
             SpecificContainerCss = 30,
 
             /// <summary>
-            /// The portal.css file has a priority of 35
+            /// The portal.css file has a priority of 95
             /// </summary>
-            PortalCss = 35,
+            PortalCss = 95,
+			
+			/// <summary>
+            /// The Page.css file has a priority of 100 and will load last.
+            /// </summary>
+            PageCss = 100,
         }
     }
 }
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -411,13 +411,6 @@ namespace DotNetNuke.Framework
                 Title += versionString;
             }
 
-			//register the custom stylesheet of current page
-			if (PortalSettings.ActiveTab.TabSettings.ContainsKey("CustomStylesheet") && !string.IsNullOrEmpty(PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString()))
-			{
-				var customStylesheet = Path.Combine(PortalSettings.HomeDirectory, PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString());
-				ClientResourceManager.RegisterStyleSheet(this, customStylesheet);
-			}
-
             // Cookie Consent
             if (PortalSettings.ShowCookieConsent)
             {
@@ -674,6 +667,13 @@ namespace DotNetNuke.Framework
             SkinPlaceHolder.Controls.Add(ctlSkin);
 
             ClientResourceManager.RegisterStyleSheet(this, string.Concat(PortalSettings.HomeDirectory, "portal.css"), FileOrder.Css.PortalCss);
+
+			//register the custom stylesheet of current page
+			if (PortalSettings.ActiveTab.TabSettings.ContainsKey("CustomStylesheet") && !string.IsNullOrEmpty(PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString()))
+			{
+				var customStylesheet = Path.Combine(PortalSettings.HomeDirectory, PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString());
+				ClientResourceManager.RegisterStyleSheet(this, customStylesheet, FileOrder.Css.PageCss);
+			}
 
             //add Favicon
             ManageFavicon();


valadas:
@thabaum 

1. The PR includes a lot of unrelated changes, can you please rebase this
2. I would really like to have a step by step of what css styling is not possible to do now so that we can review if we as a community do want to change this, it has a potential to be a huge breaking change. We try to avoid breaking changes as much as possible unless they are really required. Load order is only one thing considered for css rules, usually the best way to style something without causing a breaking change is to just be a little more specific thatn the rule to override. Also there is a default.css cleanup coming up for 10 which maybe can solve the issue at hand without it being a breaking change...

thabaum:
I was trying to push this to version 10.  What am I suppose to base this on to do so?  I was not going to put in 9 as it would be breaking possibly.

try to change logo size using Page.css without

- create  a page
- upload a pagename.css to the root site directory
- include the following in the pagename.css file  

> .navbar .brand {max-width:50px;}
- Add pagename.css as Page Stylesheet in PersonaBar Page Settings editor
- Refresh the page notice nothing happens
- put the same code in Portal.css and it works however other CSS stylesheets load after Portal.css taking priority

Pretty much only way to style things is to use !important a lot.

But then you find a problem when !important is used as there are times I cannot edit styles of modules

There is NO end CSS file a DNN administrator can use that makes sense to ensure the CSS is loaded last and takes priority over other css.  I think only people that have used Community Edition a lot for projects notice this issue but it is surely an issue.

Let me know what to base this PR on I will redo it as I don't know how to rebase quickly...

As discussed in my version 9 PR this is a version 10 solution so I rebased to 10.   Now what?

thabaum:
Curious, is EVOQ default.aspx file different with a change similar to what I have made or does it act the same? It would show the neglect by corporate to the community IMO if it did as this should of been fixed for everyone.  And if not they should toss me a bone ;)

david-poindexter:
> 1. The PR includes a lot of unrelated changes, can you please rebase this
> 2. I would really like to have a step by step of what css styling is not possible to do now so that we can review if we as a community do want to change this, it has a potential to be a huge breaking change. We try to avoid breaking changes as much as possible unless they are really required. Load order is only one thing considered for css rules, usually the best way to style something without causing a breaking change is to just be a little more specific thatn the rule to override. Also there is a default.css cleanup coming up for 10 which maybe can solve the issue at hand without it being a breaking change...

Agreed 100%

valadas:
Ok, for the unrelated changes, you had it right but the PR was submitted targetting the `develop` branch. There was a branch reorganization and `develop` is the branch to use for most things that will make it into the very next release like 9.5.1 or 9.6.0 depending on what comes in. For the potential breaking changes or major new stuff, the branch is `future/10` for version 10. After changing that target we now have only 2 changed files which is good.

Now for the actual css, !important is really a last resort thing and changing file load order is just the last condition css uses to define what wins, just being a tiny little bit more specific will make your rule win over a competing one. For instance, using your example, you could just do:
```
header .navbar .brand {max-width:50px;}
```

And this rule would win. Load order only affects what wins if the competing rules have exactly the same specificity. I am not saying by any means that the current load order is right or wrong, just that it is not a major problem and changing how it is currently would cause a breaking change on any unkown cases without a big need for it. So certainly worth discussing to see if we want to get into that. It is now clearer with the PR fix to see the changes involved...

thabaum:
OK, but I have been submitting issues for 10 years... should i dig them up?  I think some are on Jira

thabaum:
I thought I had based this on future/10 

This what was recommended my last PR which I targeted develop 9.x... so I did a new one as Mitch and Brian recommended, AND added the fix for putting Page.css to load last as it should as requested by Sebastian.

So not sure what to do here I am getting bounced around like a ping pong ball as far as where to base it.

This load order needs fixed IMO badly.  I will fight for it and others notice the issue as well. 

Why load Page.css before default.css?  It currently loads first when it should load last as my directions if you follow point out.  Is this correct?

thabaum:
One last note before i got to go.  This issue i would be creating many issues for.... over and over and wait another 10 years.  OR it was much faster although it took some days of studying the code source to resolve this, much less time than banging my head trying to get some simple stupid CSS style to take effect.

sleupold:
from my experience, there is an issue with loading sequence of CSS files, which doesn't follow documented order. There are options as well for skins and modules to override the documented (i.e. intended) order of css files (e.g. using includes), which interfere with combined css files, i.e. we never use this option at our clients (while the option of including files is helpful in complex scenarios with hierarical skin files supporting multiple variations, splitting up the skin into multiple hierarchical include files for headers, logos, login popups, link menues, footer lists etc..

sleupold:
the original ideal of combining files in DNN CDF system was reduction of files to be transferred, due to HTTP 1.x limitations. IMO we need to investigate, whether this still makes sense in HTTP 2.x, which supports a larger number of parallel downloads. This might improve performance using a larger number of smaller files. especially, if some of the files are alread transferred by previos page. We need to investigate, whether combining CSS (and JS files) might have become obsolete from a performance perspective.

thabaum:
I feel what you are talking about Sebastian.  I would love any solution as well if it increases performance and things are functioning as intended.  I can rebase this to 9 for a fix and maybe look for something new for 10?  Just let me know I am happy to help as I can.  There are a couple of ways of handling things here.  This just put things in logical order for my mind and what is currently being used in DNN 9.  I had even thought of removing PortalCss priority and use DefaultPriority for both Portal and page Custom stylesheets and set them up at the end of the page in order after any other CSS is called upon.   I am just trying to tune things in a bit here to make things work as expected, nothing for performance as DNN 11 should have many things changed by then.  

If the current way stays and something like this PR works ok I might even suggest adding a Module Custom Stylesheet that can get set similar to page.  Upload a css file and set it in the module settings to use custom CSS and be the very last to load after Page.css.

I liked that idea Sebastian also mentioned. And I also like the idea of being able to setup priorities in the database for all CSS files so you can set it to the old ways, use the new way OR create your own way for CSS or JS files to load.  A bit beyond the scope of what I was address as that is adding new features when I am trying to "Fix" what I consider a bug.

Maybe this is obsolete that I am working  on?  As long as I know something is in the works to make this stuff load in order I expect and everyone talks it should load in I am good to go.

Currently it is the wild west as far as knowing what will load last.

thabaum:
What kind of change would anyone be in support of if any?  Skin.css and Portal.css loads after Page.css, can we at least put Page.css set to default priority or have it load after all other CSS files load in default priority?

Portal.css I still believe should load after default priority.  Portal.css does currently and still will load after Skin css container css and module css files.  This order DOES NOT change in my PR nor am I proposing the fundamental order change.  So please let me know you understand this.  Thank you.

mitchelsellers:
Portal after Default = would mean that you cannot override portal, which would be the reverse issue of your current issue.

If anything, can you attend a future TAG meeting to discuss this?

thabaum:
So you wish to override portal.css?  Something that is written by an administrator to make a portal have styles?

I can try to attend a meeting and schedule it in if it can help here.  

Page.css should still load after portal.css.  DefaultPriority overrides Portal.css by design I take it.  So the fact is Portal.css does not load last as everyone is under the impression it does.  And page.css loads before skin.css.

Those are my two findings which came up from an added discussion with Sebastien when he brought up the page.css load order and I looked all over thinking it was default priority, just to find it with no priority loading above other css includes.

valadas:
>Page.css should still load after portal.css

I don't disagree and would be totally fine with this if we where just building something new, but we need to consider upgrades and 3rd parties that are used to this order. Also overriding a rule should not be done only by the file load order, and it should not be done with !important, it should be done by providing a more specific rule, that will win no matter the file load order. 

>And if anything here we can at least clean up the documentation as needed so we can all understand this load order.

@thabaum Can you provide a link to which documentation is wrong, I think this would be where to fix this other than just creating a breaking change. If documentation about load order is wrong, we need to fix that documentation. If people just don't understand CSS, unfortunately documenting that is out of our scope... Load order is really the last thing to consider for overriding styles, see https://www.w3.org/TR/CSS2/cascade.html 
image



Like @david-poindexter I am against this change. It would not fix all cases (only rules with same specificity), it would be a breaking change (swapping all existing 3rd party rules with same specificity), and it would encourage CSS bad practices (people will incorrectly fix the breaking change with !important and be under the wrong path thinking load order is more important than specificity). The correct way to not get into troubles is to know how css specificity works and understanding the Dnn load order that never changed as far as I know).

The correct best practice is to scope themes with a wrapper id or class, scope all theme rules to it. Same for modules, wrap the module with a class (not an id because there could be more than one on the page) and scope the module specific rules to that class. No conflicts, you know what wins and no !important needed anywhere. Like @mitchelsellers said on the issue, the admin accessible css UI options are a last resort and I guess if all else fails, this would be the only place where !important could be used as a very last resort.

valadas:
>This what was recommended my last PR which I targeted develop 9.x... so I did a new one as Mitch and Brian recommended, AND added the fix for putting Page.css to load last as it should as requested by Sebastian. So not sure what to do here I am getting bounced around like a ping pong ball as far as where to base it.

Sorry for that and I understand the frustration, some of us only manage to get time to reply a few days later than issues or PRs are submitted. Also many of us where super busy due to Dnn Summit. I for one am just catching up to put my comments now :) And well different people will have different opinions about changes, I for one am against breaking changes where they are not absolutely needed. You will get different comments and opinions for sure and when there are diverging opinions we need to take it into a TAG meeting to take a final decision. For sure at a minimum if the consensus ends up being to take it in, the move to Dnn10 is still valid at the least.

valadas:
As per todays TAG approvers meeting, we looked at a couple of module vendors and according to the usages found, this would be a breaking change for many 3rd parties, for that reason and because there are better ways to do this properly in CSS, we are closing this PR. We think this is an unnecessary breaking change that would not even solve the issue in all cases.

thabaum:
Could this be added to DNN Platform Version 11 possibly?

If not should we document to use` !important` in css for page and portal css to take priority over all other css files that are loaded. 

A module or skin loads top priority overriding Portal.css and Page.css which everyone expected before to be the last files to load.  Also if a developer loads a css file without declaring a load priority it takes priority over all others.  Instead they actually fight with all other modules and skins that have css files loading with default priority so you never get a true css file load order, each page generally varies depending on which modules or skins are loaded and how the developer declared the css files.  That module or skin then overrides Page.css and Portal.css which should load after default priority.   So for a developer you get best results for what you are trying to do however if someone tries to modify after getting a vendor module you generally MUST use !important in order for any CSS changes to take affect.

valadas:
>Could this be added to DNN Platform Version 11 possibly?
My vote would be no, we do not want to intruduce unnecessary breaking changes no matter the version, if there is wrong documentation about how it works, we should correct the documentation, not introduce a breaking change that is not required.

>If not should we document to use !important in css for page and portal css to take priority over all other css files that are loaded.
It cannot be on our responsability to document html/css/javascript, there are core web technologies that are not Dnn specific. Also !important is to avoid as much as possible, instead rules should be as specific as they need to be in order to do their job. Themes and Modules should use an enclosing scope so their css does not leak out of their intent and are still able to override more general rules. It is really rare that !important is required and many css frameworks make extensive use of it, which I hate.

>To prepare for the big change everyone can simply put !important on everything in the Portal.css and Page.css files to prepare for the switch.
Hmm, no, big no, this would be such an anti-pattern and would not replicate just a load order change!!! !important breaks css cascading, it is to avoid at all costs unless there is absolutely no other way.

PR #3596, opened 2020-02-24
Add debug build to lerna

donker:
This was overlooked in the last Lerna/Yarn overhaul. The user must be able to target a local dev site with the Lerna build. It is currently missing as it only runs in production mode.

PR #3601, opened 2020-02-28
Fix image-actions configuration

bdukes:
See https://github.com/calibreapp/image-actions/blob/08437f2fb2e1548570f2e24cf8c902191c2f5432/README.md#installation for "expected" config

Should fix failure of this check in #3596

### DIFF
--- a/.github/workflows/image-actions.yml
+++ b/.github/workflows/image-actions.yml
@@ -2,9 +2,13 @@ name: Compress images
 on: pull_request
 jobs:
   build:
+    name: calibreapp/image-actions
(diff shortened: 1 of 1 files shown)


valadas:
Looks good to me, but we will only know once merged...

bdukes:
Fortunately, it runs the new workflow on the PR, so it did catch an issue and failed on my first commit, but I fixed that one issue and it's green now.  So unless there's something else going on with #3596 we should be good.

valadas:
Ah, awesome

PR #3609, opened 2020-03-03
Data Consent JS 

### LINKED ISSUE #3193: Data Consent checkbox click handler fails on custom logon page
## Description of bug
Data consent check box does not trigger the removal of the disabled attribute from the submit button when the login control is placed on custom logon page.  

## Steps to reproduce
1. Create a new page called logon.
2. Add the Login Control
3. Set the sites default login to the the new page
4. Enable the Data Consent feature in the site privacy settings.
5. Log in as any regular user & you cannot proceed past the agree checkbox stage.

## Current result
User logs in but can't enable the submit button after agreeing to the terms.

## Expected result
User logs in and checks the box to agree to the Data Consent terms which should enable the submit.  Log in proceeds & user gets access.

## Error log
No js error shows in the developer console

## Additional context
It seems the browser is not executing this script found at the bottom of the  DesktopModules\Admin\Security\DataConsent.ascx control.  Tested placing an alert after the opening js script tag and before the click handlers but it is not executing.  Not seeing why..??  Tested a few different themes which should eliminate that as a possible cause.

## Affected version
9.4.1

## Affected browser
* [ ] Chrome

DNNMonster:
Possible option to Fix #3193

## Summary

Migrate the event handler js to login.ascx so that custom login pages will work with the Data Consent flow.   Setting the checked var may be optional.


donker:
Thank you for your contribution @DNNMonster. I have two issues with this approach:

1. It violates the idea of keeping stuff isolated. I.e. in this proposal there would be data consent specific code at a higher level. This risks breaks when someone intervenes in the DataConsent code and is no aware of something higher up
2. There are hard coded ClientID paths which also are brittle.

I will submit a separate PR which will address the issue by adding the onclick handlers using the code behind. It's also not perfect (I hate having JS in code) but it is the lesser of two evils IMHO.

DNNMonster:
Sounds fair to me.  I'll merge your fix when it's ready & run our installation with this for now since it allows us to use the Data Consent feature.

PR #3633, opened 2020-03-18
File extension whitelists for end users

donker:
Currently DNN uses one single file extension whitelist that governs what administrators and end users can upload (AE: Security > More > More Security Settings). This list includes files that may be needed by administrators to upload content to manage the site but that end users have no business uploading. This PR introduces a new whitelist mechanism that allows site maintainers (i.e. admins) to control what end users of a site can upload (within the limits of their own whitelist). Specifically this solution does the following:

1. It introduces a second whitelist setting (AE: Security > More > More Security Settings) for the host called "Default End User Extension Whitelist". If an administrator does *not* define an end user whitelist, this list will be used.

2. It allows the administrator to define a whitelist (AE: Site Settings > Site Behavior > More) called "Permitted File Extension List". The UI includes a dropdown which allows the admin to specify "default" (i.e. what is defined in 1 above), "custom", or "images only".

3. Upon update of any of the above settings and the global whitelist, there is a cascading check to ensure no end user whitelist allows uploading of a file not allowed by the global whitelist.


DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -585,6 +585,20 @@ public virtual IDataReader GetPortalSettings(int PortalId, string CultureCode)
...
 
+        internal virtual IDictionary<int, string> GetPortalSettingsBySetting(string settingName, string cultureCode)
+        {
+            var result = new Dictionary<int, string>();
+            var sql = "SELECT ps.PortalID, ps.SettingValue FROM " + DatabaseOwner + ObjectQualifier + "PortalSettings ps WHERE ps.SettingName=@0 AND COALESCE(CultureCode, @1, N'') = IsNull(@1, N'')";
+            using (var reader = PetaPocoHelper.ExecuteReader(ConnectionString, CommandType.Text, sql, settingName, cultureCode))
+            {
+                while (reader.Read())
+                {
+                    result[(int)reader.GetValue(0)] = (string)reader.GetValue(1);
```
bdukes:
```suggestion
                    result[reader.GetInt32(0)] = reader.GetString(1);
```

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -585,6 +585,20 @@ public virtual IDataReader GetPortalSettings(int PortalId, string CultureCode)
             return ExecuteReader("GetPortalSettings", PortalId, CultureCode);
         }
 
+        internal virtual IDictionary<int, string> GetPortalSettingsBySetting(string settingName, string cultureCode)
+        {
+            var result = new Dictionary<int, string>();
+            var sql = "SELECT ps.PortalID, ps.SettingValue FROM " + DatabaseOwner + ObjectQualifier + "PortalSettings ps WHERE ps.SettingName=@0 AND COALESCE(CultureCode, @1, N'') = IsNull(@1, N'')";
```
bdukes: Do we have anywhere else in the `DataProvider` that we generate SQL like this?  Are we intentionally moving away from stored procedures?

DNN Platform/Library/Common/Utilities/FileExtensionWhitelist.cs
```diff
@@ -141,5 +141,19 @@ private IEnumerable<string> NormalizeExtensions(IEnumerable<string> additionalEx
         {
             return additionalExtensions.Select(ext => (ext.StartsWith(".") ? ext : "." + ext).ToLowerInvariant());
         }
+
+        public FileExtensionWhitelist RestrictBy(FileExtensionWhitelist parentList)
+        {
+            var filter = parentList.AllowedExtensions.ToList();
```
bdukes: You can just use `_extensions` directly, since we're inside `FileExtensionWhitelist`:
```suggestion
            var filter = parentList._extensions;
```

bdukes: Alternatively, since we're just using contains, using a set may be more performant:
```suggestion
            var filter = parentList.AllowedExtensions.ToHashSet();
```

DNN Platform/Library/Common/Utilities/FileExtensionWhitelist.cs
```diff
@@ -141,5 +141,19 @@ private IEnumerable<string> NormalizeExtensions(IEnumerable<string> additionalEx
...
+            var filteredList = new List<string>();
+            foreach (var ext in _extensions)
+            {
+                if (filter.Contains(ext))
+                {
+                    filteredList.Add(ext.Substring(1));
+                }
+            }
+            return new FileExtensionWhitelist(string.Join(",", filteredList));
```
bdukes: If this gets called a lot, we may want to switch to a `StringBuilder` (or, at least, do a little performance testing)

mitchelsellers: This could also possibly be done more succintly with something like
var allowedExtensions = _extensions.Where(x => filter.Contains(x)).Select(s => s.Substring(1));

Just could be abit more condensed.readable.

DNN Platform/Library/Data/DataProvider.cs
mitchelsellers: Also, don't we have an API method already that gets a single setting?

DNN Platform/Library/Entities/Users/UserInfo.cs
```diff
@@ -114,6 +114,36 @@ public string FirstName
         [Browsable(false)]
         public bool IsSuperUser { get; set; }
 
+
+        /// <summary>
+        /// Gets whether the user is in the portal's administrators role
+        /// </summary>
+        private bool? isAdmin;
```
mitchelsellers: Given that the calculated value is never stored/used, no need for this to be a private backer

DNN Platform/Library/Entities/Users/UserInfo.cs
```diff
@@ -114,6 +114,36 @@ public string FirstName
...
+        public bool IsAdmin
```
mitchelsellers: I need to check a few things on the processes, but I believe that IsInRole already handles the IsSuperUser check as a superuser is always in every role.

DNN Platform/Library/Security/PortalSecurity.cs
```diff
@@ -809,19 +809,43 @@ public bool ValidateInput(string userInput, FilterFlag filterType)
...
-		
-		#region Public Shared/Static Methods
+
+        /// <summary>
+        /// This function loops through every portal that has set its own AllowedExtensionWhitelist
+        /// and checks that there are no extensions there that are restriced by the host
+        /// 
+        /// The only time we should call this is if the host allowed extensions list has changed
+        /// </summary>
+        public void CheckAllPortalFileExtensionWhitelists()
```
mitchelsellers: If this is internal, can we mark this internal to prevent external callers?

mitchelsellers: Additionally I think we might want a bit stronger name on this, as this isn't a check, it actually updates/changes data if necessary

DNN Platform/Website/Install/DotNetNuke.install.config.resources
```diff
@@ -43,7 +43,8 @@
     <EnableStrengthMeter>Y</EnableStrengthMeter>
     <EncryptionKey Secure="True"></EncryptionKey>
     <EventLogBuffer>N</EventLogBuffer>
-    <FileExtensions>jpg,jpeg,jpe,gif,bmp,png,svg,ttf,eot,woff,doc,docx,xls,xlsx,ppt,pptx,pdf,txt,xml,xsl,xsd,css,zip,rar,template,htmtemplate,ico,avi,mpg,mpeg,mp3,wmv,mov,wav,mp4,webm,ogv</FileExtensions>
+    <FileExtensions>jpg,jpeg,jpe,gif,bmp,png,svg,ttf,eot,woff,doc,docx,xls,xlsx,ppt,pptx,pdf,txt,css,zip,rar,template,htmtemplate,ico,avi,mpg,mpeg,mp3,wmv,mov,wav,mp4,webm,ogv</FileExtensions>
+    <DefaultEndUserExtensionWhitelist>jpg,jpeg,jpe,gif,bmp,png,svg,doc,docx,xls,xlsx,ppt,pptx,pdf,txt,zip,rar,ico,avi,mpg,mpeg,mp3,wmv,mov,wav,mp4,webm,ogv</DefaultEndUserExtensionWhitelist>
```
mitchelsellers: Upon further reflection, should the default end user allow zip & rar?

DNN Platform/Library/Data/DataProvider.cs
donker: My 2cts:

1. I felt we were encouraging everyone to move on from the archaic stored procedures for everything. So I kicked it off here.

2. Given that the use case for this method is not huge, I felt comfortable doing this in a way that was shortest to goal.

3. Yes, we have for a single setting. But given that an installation may have hundreds of portals, I felt it prudent to make this (which runs during a setting update) as efficient as possible.

DNN Platform/Library/Entities/Users/UserInfo.cs
donker: Not sure I am on the same page, but I am pretty confident that once we have this value then extension developers will begin using this value more and more instead of coding their own IsAdmin logic. And that might require accessing the value multiple times in your code.

On the other hand one could argue that maybe an extension developer might change the user's roles and then recalculation would be necessary.

On the fence about it. Given that the calculation is not trivial I persisted the value.

DNN Platform/Website/Install/DotNetNuke.install.config.resources
donker: rar probably not. But sharing of files is often done in zip files.

This raises another issue we may need to discuss, BTW.

DNN Platform/Library/Data/DataProvider.cs
sleupold: Using stored procedures still has a benefit, as SQL Server is able to compile it and maintain optimized query plans. This is nt important for this small query, but for many other queries in the code, i.e. imo it is not desirable to avoid sprocs and merge SQL code in other code files.

bdukes:
Looks like tests are failing.  I only glanced at the error, but you may need to add a mock value for this new setting or something like that.

neal1991:
Is this fix for CVE-2020-5186. Why there is no release description for security fix?

mitchelsellers:
@neal1991 Security issues ARE NOT DISCUSSED here per the terms of our security policy.

If this is applicable to a security notice it will be published as part of the Security Bulletins as part of the release process.

We DO NOT publish security bulletins for RC releases.

PR #3635, opened 2020-03-20
Fix error when MVC module is on a page multiple times

### LINKED ISSUE #3634: Exception on DNN 9.5.1 when same MVC module inserted several times on a page
## Description of bug

MVC Module when inserted multiple times on a page results in exception:

```
Error: Contact List Mvc is currently unavailable. DotNetNuke.Services.Exceptions.ModuleLoadException: A single instance of controller 'Dnn.ContactList.Mvc.Controllers.ContactController' cannot be used to handle multiple requests. If a custom controller factory is in use, make sure that it creates a new instance of the controller for each request. ---> System.InvalidOperationException: A single instance of controller 'Dnn.ContactList.Mvc.Controllers.ContactController' cannot be used to handle multiple requests. If a custom controller factory is in use, make sure that it creates a new instance of the controller for each request. at System.Web.Mvc.ControllerBase.VerifyExecuteCalledOnce() at System.Web.Mvc.ControllerBase.Execute(RequestContext requestContext) at DotNetNuke.Web.Mvc.Framework.Modules.ModuleApplication.ExecuteRequest(ModuleRequestContext context) at DotNetNuke.Web.Mvc.MvcHostControl.ExecuteModule() --- End of inner exception stack trace ---
```

Module used to test:
https://github.com/dnnsoftware/Dnn.Platform.Samples.Mvc/blob/master/Website/Install/Module/DNN_ContactList_Mvc_01.00.01_Install.zip

Related Issue Previously Reported: https://github.com/dnnsoftware/Dnn.Platform/issues/3608

## Steps to reproduce
1. Place MVC Module on the page twice
2. Display page
3. Exception shows in the second module. First module loads correctly.

## Current behavior
First module loads correctly. Second fails to load due to exception.

## Expected behavior
Both modules should load without exception.

## Screenshots
image

## Error information
Looks related to changes in 9.5.0 and continued in 9.5.1 to use Scoped instancing for IDNNControllers. 

Exceptions occurs here:

https://github.com/dnnsoftware/Dnn.Platform/blob/release/9.5.1/DNN%20Platform/DotNetNuke.Web.Mvc/Framework/Modules/ModuleApplication.cs#L127

Current workaround is to change TryAddScoped to TryAddTransient in the mvc load extension here:

https://github.com/dnnsoftware/Dnn.Platform/blob/release/9.5.1/DNN%20Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs#L25

## Affected version
* [ ] 10.0.0 alpha build
* [X]  9.5.1 alpha build
* [X]  9.5.0 latest supported release
* [ ]  9.4.4 

## Affected browser
Browser independent

bdukes:
Fixes #3634, thanks @DaveMartinSE!

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs
@@ -22,7 +22,7 @@ namespace DotNetNuke.Web.Mvc.Extensions
                 );
             foreach (var controller in controllerTypes)
             {
-                services.TryAddScoped(controller);
+                services.TryAddTransient(controller);
             }
         }
     }


mitchelsellers:
@bdukes Did you run a performance test on this?

bdukes:
The performance was broken before and not broken after 😉 

In reality, scoped would be one per request, and transient would be one per module, so the only time there would be a difference would be when there are multiple modules on one page, and that's the case where it's broken anyway.  But, to answer the question, no, I didn't run any performance comparison.

mitchelsellers:
I was thinking of the memory leak side of things, but this appears to be no impact.

PR #3643, opened 2020-03-24
Updates develop with latest changes from release/9.5.1

valadas:

Due to COVID-19 we are postponing the 9.5.1 release due to multiple factors, due to community slowdown regarding testing, ability to install by companies, etc.


valadas:
This also means 9.5.1 is no longer in code-freeze for the time being

PR #3655, opened 2020-03-30
validate file upload request.

### LINKED ISSUE #3654: validate file upload request

zyhfish:
Fix #3654.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/ValidationUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/ValidationUtils.cs
@@ -17,28 +17,31 @@ namespace DotNetNuke.Common.Utilities
             return FIPSCompliant.EncryptAES(key, key, Host.GUID);
         }
 
-        internal static string ComputeValidationCode(IList<string> extensions)
+        internal static string ComputeValidationCode(IList<object> parameters)
         {
-            if (extensions != null)
+            if (parameters != null && parameters.Any())
             {
-                var checkString = extensions.Select(i => i.ToLowerInvariant())
-                    .OrderBy(i => i)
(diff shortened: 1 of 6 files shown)


DNN Platform/Library/Common/Utilities/ValidationUtils.cs
```diff
@@ -0,0 +1,44 @@
...
+    public sealed class ValidationUtils
+    {
+        internal static string GetDecryptionKey()
+        {
+            var machineKey = Config.GetDecryptionkey();
+            var key = $"{machineKey ?? ""}{Host.GUID.Replace("-", string.Empty)}";
+            return FIPSCompliant.EncryptAES(key, key, Host.GUID);
+        }
+
+        internal static string ComputeValidationCode(IList<string> extensions)
```
bdukes: Can we add something more dynamic to this validation code?  Maybe a module ID?  Otherwise, if there are two uploads (e.g. journal upload and profile picture upload), I could copy the validation code from one to the other and bypass the restriction that way.

bdukes: I guess even just a random salt that we sent to the client alongside the validation code would work

bdukes: After discussion, it sounds like these sorts of changes make sense in this context.

bdukes: After further discussion, we've talked about adding portal ID (or GUID?) and user ID, which would provide some benefit.  The remaining issue would be around expiration the validation code, but we can accept this before that's handled.

zyhfish: Thanks @bdukes , added the portal id and user id in the check.

PR #3663, opened 2020-04-05
Ensure the portal ID being used for consent is the parent portal

### LINKED ISSUE #3662: Data consent will not store consent in parent/child portal scenario
A user's consent is stored in the UserPortal record. It is sent there from a portal with the current portal's ID. If this is a child portal there is no userportal record for that portal. Instead it should be recorded with the parent portal.

## Steps to reproduce
Create parent/child portal setup for DNN
1. Activate Data Consent
2. User logs in and consents
3. Log out
4. User connects again and sees the consent screen again

## Expected behavior
Should not see consent again

## Affected version
* [x] 10.0.0 alpha build
* [x]  9.5.1 alpha build
* [x]  9.5.0 latest supported release
* [x]  9.4.4 

## Affected browser
N/A

donker:
Fixes #3662

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -692,7 +692,7 @@ namespace DotNetNuke.Entities.Users
         public static void UserAgreedToTerms(UserInfo user)
         {
             Requires.NotNull("user", user);
-            MembershipProvider.Instance().UserAgreedToTerms(user.UserID, PortalController.GetEffectivePortalId(user.PortalID));
+            MembershipProvider.Instance().UserAgreedToTerms(user);
         }
 
         /// <summary>
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -692,9 +692,7 @@ namespace DotNetNuke.Entities.Users
         public static void UserAgreedToTerms(UserInfo user)
         {
             Requires.NotNull("user", user);
-            var u = user;
-            u.PortalID = PortalController.GetEffectivePortalId(u.PortalID);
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -692,7 +692,9 @@ public static void ApproveUser(UserInfo user)
         public static void UserAgreedToTerms(UserInfo user)
         {
             Requires.NotNull("user", user);
-            MembershipProvider.Instance().UserAgreedToTerms(user);
+            var u = user;
+            u.PortalID = PortalController.GetEffectivePortalId(u.PortalID);
+            MembershipProvider.Instance().UserAgreedToTerms(u);
```
bdukes, line 697: `var u = user;` doesn't have any effect here.  Since `UserInfo` is a class, you're still pointing to the shared instance, and this `PortalID` assignment will affect the `UserInfo` object that's passed in.  I think we'd want to avoid that, so it's probably better to either pass `portalId` as a parameter to `UserAgreedToTerms` or check the effective portal ID inside `UserAgreedToTerms`.

donker: OK, I'll need to rewrite that bit of the API though, which I wanted to avoid.

bdukes: Is it possible to move the `GetEffectivePortalId` call into `UserAgreedToTerms`?  That seems like the least intrusive change, if there's nothing standing in the way…

donker: I'll give it a shot. Hang on.

donker: Like it is now?

bdukes: yup, looks good to me!

PR #3674, opened 2020-04-10
Fixes #3671

### LINKED ISSUE #3671: Droplists showing value not text
## Description of bug
When using a control that is using type-a-ahead (like Country or Region) if you start to type the first three letters and use an arrow key to select from the list, the droplist displays the items value and not the text (i.e. enter "Uni" for united states and then hit the down arrow to select from list and it will show a number (I assume the items value from the database)

## Steps to reproduce
1. Open a page like registration that is including the Country lookup/listbox.
2. Start typing a country (i.e. "Uni" for United States)
3. A list will appear below the entered text with a list of countries.
4. Attempt to use the down arrow to select a country. The cursor stays in the text box which I assume is ok, but as you hit the down arrow, it cycles through item values (177, 242, etc) not the names of the countries you are looking for.

## Current behavior
See above

## Expected behavior
Would cycle through names of countries as you scroll through the list of names, not their value in the database.

## Screenshots
bug-before
bug-after

## Error information
No errors displayed in any browser

## Additional context

## Affected version

* [ ] 10.0.0 alpha build
* [ ]  9.5.1 alpha build
* [X ]  9.5.0 latest supported release
* [ ]  9.4.4 

## Affected browser
Only ones I tried it in.
* [ x] Chrome
* [x ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [x ] Microsoft Edge (Classic)
* [ ] Microsoft Edge Chromium

mitchelsellers:
This PR fixes an incomplete push from #3671 disables additional Telerik features that are not used by DNN Platform.

This is a potential breaking change for anyone with third-party modules that are using Telerik functionality, however, it is NOT a breaking change for a base DNN Platform installation.

### DIFF
--- a/DNN Platform/Website/Install/Config/09.06.00.config
+++ b/DNN Platform/Website/Install/Config/09.06.00.config
@@ -1,0 +1,12 @@
+<configuration>
+  <nodes configfile="Web.config">
+    <node path="/configuration/system.web/httpHandlers/add[@name='Telerik.Web.UI.ChartHttpHandler']" action="remove" />
+    <node path="/configuration/system.webServer/handlers/add[@name='Telerik.Web.UI.ChartHttpHandler']" action="remove" />
+    <node path="/configuration/system.web/httpModules/add[@name='RadUploadModule']" action="remove" />
+    <node path="/configuration/system.webServer/modules/add[@name='RadUploadModule']" action="remove" />
+    <node path="/configuration/appSettings" action="update" key="key" collision="overwrite">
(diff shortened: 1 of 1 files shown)


valadas:
@mitchelsellers Can you resolve the conflict, should be easy in the browser I believe.

mitchelsellers:
Argh it’s telling me to use the command line and I’m away from a PC....

mitchelsellers:
Ok, conflict fixed

PR #3675, opened 2020-04-12
Bring back Export Portal Template

donker:
Since the introduction of the PersonaBar we have lost the ability to export a portal template. This PR brings back a button on the site list that will take the user to a page from where the portal template can be created.


schotman:
Did some testing on this enhancement.
Exporting template works. Creating portal based on a 9.6 template works. 

Some additional notes.

Trying to create a site based on an 804 template, works after manually changing the template. An 804 template includes the admin pages. 
The following goes wrong:
-  one or more admin pages throw an error (module creation error), but since we don't need them, they can be excluded.
-  if I remove all the admin pages from the site template, site creation works, except that the site asset manager is now missing
-  if I keep file manager page in the template, it becomes a regular page
But if I keep the admin page and the file manager page in the template, the file manager is again part of the persona bar

So using a 804 template is possible but requires most of the admin pages to be removed.

mitchelsellers:
That’s would be expected behavior for a template

Issue #3660, opened 2020-04-02
Users can't log in to child portals when username is email


donker:
##  Description of bug
This concerns the master/child portal structure that can exist in DNN. If you have a master portal for a child portal, then during login using email the user is looked up in the child portal and not the master portal as it should.

## Steps to reproduce
In an installation with a master/child portal structure
1. Switch child portal to use email as username
2. Log out
3. Try to log in as any user

## Current behavior
Can't log in

## Expected behavior
Log in

## Affected version
* [? ] 10.0.0 alpha build
* [ x]  9.5.1 alpha build
* [ x]  9.5.0 latest supported release
* [ x]  9.4.4 

## Affected browser
n/a

thabaum:
Is this a "Site Group"  I have not yet played with that feature much, and "child portals" features vanished version 9.  Although I may not know where they are in the PersonaBar.
It is good to see you going through the "email as username" bugs.  I know it has me in the back of my mind saying "stay away" although each time I look at the feature it is very tempting to turn on.

donker:
Fixed by #3659

Issue #3658, opened 2020-04-02
Host can't login after portal switched to use email instead of usernames


donker:
## Description of bug
It looks like our logic to log in a user depends on that user being a member of the portal if the portal uses "email as username". This impedes hosts to log in as they are not typically a member of every child portal.

## Steps to reproduce
1. Use an existing site with various accounts
2. Switch the site to use emails instead of usernames (Security)
3. Log out
4. Try to log in as host

## Current behavior
Will not log in host. Neither email not username will work.

## Expected behavior
Logs in host.

## Affected version
* [ ?] 10.0.0 alpha build
* [ ?]  9.5.1 alpha build
* [ ?]  9.5.0 latest supported release
* [x ]  9.4.4 

## Affected browser
n/a

## Analysis
This happens in this bit of code:

https://github.com/dnnsoftware/Dnn.Platform/blob/f6f6ef6dcdf43d185cac2d3bf4dc55bb2ec5e7a8/DNN%20Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs#L255

As you can see we do ``` userByEmail = UserController.GetUserByEmail(PortalId, userName); ``` to find the user. This SPROC is somewhat lacking as it will only find a user if in that portal. IMHO a superuser should be found as well.

thabaum:
I always thought I could I just needed to use the host email address instead of username.  does this allow username for host instead?

Is there a way we can get it so we can use allow both Email or Username for option to login instead of just email or only username.

I am debating using email as username as it seemed to have some odd behaviors in the past.

PR #3677, opened 2020-04-13
Allow superusers login using email

### LINKED ISSUE #3676: Use Email Address as Username option doesn't handle super users
## Description of bug
When "Use Email Address as Username" option is enabled for a portal, super users can't log in.

## Steps to reproduce

1. Log in as a super user
2. Create a new portal. Don't set the super user as the admin of the new portal. Instead, create a new admin, for example Subsite Admin, subsiteadmin@change.me
3. Logout
4. Navigate to newly created portal and log in as subsiteadmin@change.me
5. Open SETTINGS > Security > Member Accounts > REGISTRATION SETTINGS and enable the "Use Email Address as Username" option.
6. Save the settings and log out.
7. Log back in using the super user's email address.

## Current behavior
"Login Failed. Please remember that passwords are case sensitive." error message is displayed

## Expected behavior
The super user should be able to login without issues.

## Affected version
* [X] 10.0.0 alpha build
* [X]  9.5.1 alpha build
* [X]  9.5.0 latest supported release
* [X]  9.4.4

daguiler:
Fixes #3676

## Summary
The logic in Login.ascx.cs was only authenticating users by email when the users existed in current portal, which is not the case when super users attempt to log in.
I changed this logic to search not only for portal users but also for super users as well. This is accomplished by calling `GetUsersByEmail` instead of `GetUserByEmail`.

### DIFF
--- a/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/AuthenticationServices/DNN/Login.ascx.cs
@@ -5,6 +5,7 @@
 #region Usings
 
 using System;
+using System.Linq;
 using System.Web;
 using Microsoft.Extensions.DependencyInjection;
 using DotNetNuke.Abstractions;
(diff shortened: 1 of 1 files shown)


bdukes:
Duplicate of #3659

bdukes:
I think @donker's fix in #3659 of updating `GetUserByEmail` to work with super users is better than switching to `GetUsersByEmail`.  Let me know if I'm missing anything, but looks like this issue will be addressed in 9.6.0 either way.  Thanks!

daguiler:
Oh man... I can't believe this. What a waste of time.
Neither #3658 nor #3660 popped up as potential duplicates when I created #3676.
And I even checked out latest code, including @donker 's fix, except I did not install the very latest 9.5.2 CI build, causing the stored procedure not to get updated. So I was able to reproduce the issue.

donker:
Sorry to hear about time wasted, Daniel. It seems we were both struggling with the same issues at the same time.

PR #3673, opened 2020-04-10
Adjusted Telerik Bindings

mitchelsellers:
This PR disables additional Telerik features that are not used by DNN Platform.

This is a potential breaking change for anyone with third-party modules that are using Telerik functionality, however, it is NOT a breaking change for a base DNN Platform installation.

PR #3683, opened 2020-04-15
Remove items from default web.config

donker:
This completes work from a PR #3673 where we adjusted bindings


mitchelsellers:
@donker Actually they need to be removed from here - https://github.com/dnnsoftware/Dnn.Platform/blob/c35fdc7fb75db0438f3b872ce4e279e3ea73e7c2/DNN%20Platform/Components/Telerik/DotNetNuke.Telerik.Web.dnn

PR #3696, opened 2020-04-21
Add DI support to scheduler

### LINKED ISSUE #3602: Support Dependency Injection from Scheduler
## Description of problem
There is no way to access services in the Dependency Injection container from a scheduler client.

## Description of solution
When DNN creates a scheduler client, it needs to use the DI container.

## Additional context
Based in a discussion at https://dnncommunity.org/forums/aft/890 and via https://dnntag.slack.com/

bdukes:
Fixes #3602, thanks for the direction @dimarobert!

### DIFF
--- a/DNN Platform/Library/Services/Scheduling/ProcessGroup.cs
+++ b/DNN Platform/Library/Services/Scheduling/ProcessGroup.cs
@@ -8,8 +8,9 @@ using System;
 using System.Reflection;
 using System.Threading;
 using System.Web.Compilation;
-
+using DotNetNuke.Common;
 using DotNetNuke.Instrumentation;
+using Microsoft.Extensions.DependencyInjection;
 
 #endregion
 
@@ -61,12 +62,14 @@ namespace DotNetNuke.Services.Scheduling
 
         public void Run(ScheduleHistoryItem objScheduleHistoryItem)
         {
+            IServiceScope serviceScope = null;
             SchedulerClient Process = null;
             try
             {
                 //This is called from RunPooledThread()
                 ticksElapsed = Environment.TickCount - ticksElapsed;
-                Process = GetSchedulerClient(objScheduleHistoryItem.TypeFullName, objScheduleHistoryItem);
+                serviceScope = Globals.DependencyProvider.CreateScope();
+                Process = GetSchedulerClient(serviceScope.ServiceProvider, objScheduleHistoryItem.TypeFullName, objScheduleHistoryItem);
                 Process.ScheduleHistoryItem = objScheduleHistoryItem;
                 
 				//Set up the handlers for the CoreScheduler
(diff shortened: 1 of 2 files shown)


bdukes:
@mitchelsellers I've tested that this works with existing scheduler clients and a new one that takes an `INavigationManager`.  I have not done any perf testing on it.

dimarobert:
@bdukes I think that `ActivatorUtilities.CreateInstance` can get by without the SchedulerClient registrations in the `AddSchedulerClients`. It uses the `IServiceProvider` to resolve the dependencies of the instantiated type not the type itself.
If that is the only reason that the clients are registered you should give it a try without the registrations and see if it still works.

bdukes:
Thanks @dimarobert, you're right, it works perfectly without the registrations.  I've pushed that change, @mitchelsellers.

@mitchelsellers do you think it's worth looking into using `CreateFactory` instead of `CreateInstance`?  We'd need to introduce a static `ConcurrentDictionary` or something like that to store the factory delegates.

mitchelsellers:
@bdukes I think we take this as it is now, as this gives us DI support while maintaining the current limitation.  I think a future change to use CreateFactory would be awesome, but it is a larger risk.

bdukes:
@dnnsoftware/approvers can I get a second approval?

valadas:
Any reason why this got closes @bdukes ?

bdukes:
@valadas sorry, I must have accidentally hit the _Close and comment_ button

ahawes-clarity:
Can you tell me how add the dependencies to a SchedulerClient? Do we just add them to the constructor?

bdukes:
Yes, if you've registered `IThingamajigger` with the `IServiceCollection` in your `IDnnStartup`, you can add that as a constructor argument to your `SchedulerClient` implementation. For example:

```
    public class AwsConfigurationTask : SchedulerClient
    {
        private readonly ISettingsService settingsService;

        private readonly IAwsService awsService;

        /// <summary>Initializes a new instance of the <see cref="AwsConfigurationTask"/> class.</summary>
        /// <param name="item">The schedule history item.</param>
        /// <param name="settingsService">The settings service.</param>
        /// <param name="awsService">The AWS service.</param>
        public AwsConfigurationTask(ScheduleHistoryItem item, ISettingsService settingsService, IAwsService awsService)
        {
            this.ScheduleHistoryItem = item;
            this.settingsService = settingsService;
            this.awsService = awsService;

            this.ProcessErrored += (client, exception) => Exceptions.LogException(exception);
        }

        public override void DoWork()
        {
            try
            {
                this.Progressing();
                
                this.ScheduleHistoryItem.AddLogNote("<ol>");
                // TODO: implement scheduler logic
                this.ScheduleHistoryItem.AddLogNote("</ol>");

                this.ScheduleHistoryItem.Succeeded = true;
                this.ScheduleHistoryItem.AddLogNote("<p><strong>AWS configuration completed for all sites</strong></p>");
            }
            catch (Exception exc)
            {
                this.ScheduleHistoryItem.Succeeded = false;
                this.ScheduleHistoryItem.AddLogNote("</ol>");
                this.ScheduleHistoryItem.AddLogNote("<p><strong>Unexpected error during AWS configuration</strong></p>");
                this.Errored(ref exc);
            }
        }
    }
```

PR #3697, opened 2020-04-22
Host users should bypass the whitelist

donker:
This came out of testing. Our recent work on Whitelists contained a flaw where hosts were still subject to their own whitelist which runs counter to the philosophy of the framework.


donker:
Agreed with Mitch. Retracting.

PR #3698, opened 2020-04-22
Provide better feedback in export

donker:
When trying to export a module, we write a file to disk. If this fails the error the user sees is non-descript. This PR allows the error message to bubble up to the end user.

### DIFF
--- a/DNN Platform/Website/admin/Modules/Export.ascx.cs
+++ b/DNN Platform/Website/admin/Modules/Export.ascx.cs
@@ -82,8 +82,8 @@ namespace DotNetNuke.Modules.Admin.Modules
                     {
                         var objObject = Reflection.CreateObject(Module.DesktopModule.BusinessControllerClass, Module.DesktopModule.BusinessControllerClass);
 
-						//Double-check
-						if (objObject is IPortable)
+                        //Double-check
+                        if (objObject is IPortable)
                         {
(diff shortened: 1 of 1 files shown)


DNN Platform/Website/admin/Modules/Export.ascx.cs
```diff
@@ -134,9 +134,16 @@ private string ExportModule(int moduleID, string fileName, IFolderInfo folder)
                             strMessage = Localization.GetString("ExportNotSupported", LocalResourceFile);
                         }
                     }
-                    catch
+                    catch (Exception ex)
```
mitchelsellers, line 137: Ideally we should capture a more specific error

valadas: Would you catch specifics and then the generic exception? Because the previous behaviour just catched anything and showed a generic error message.

donker: > 
> 
> Ideally we should capture a more specific error

You'd like me to rewrite this using catch (FooException ex) for each exception type instead of the if-then?

bdukes: FWIW, I'm happy with this as-is

mitchelsellers: Ideally I'd like to see an exception filter or otherwise used, but I'll let it go

PR #3719, opened 2020-05-01
Bump version to 9.6.0

donker:
Bring version up to date


SolutionInfo.cs
```diff
@@ -18,6 +18,6 @@
 [assembly: AssemblyProduct("https://dnncommunity.org")]
 [assembly: AssemblyCopyright("DNN Platform is copyright 2002-2019 by .NET Foundation. All Rights Reserved.")]
 [assembly: AssemblyTrademark("DNN")]
-[assembly: AssemblyVersion("9.4.4")]
-[assembly: AssemblyFileVersion("9.4.4.5")]
-[assembly: AssemblyInformationalVersion("9.4.4.5 Custom build")]
+[assembly: AssemblyVersion("9.6.0")]
+[assembly: AssemblyFileVersion("9.6.0.5")]
+[assembly: AssemblyInformationalVersion("9.6.0-rc.3+Branch.release-9.6.0.Sha.600e5a8e59d14c432645384f6cdeb6c97fadff36")]
```
bdukes, line 23: I don't think it makes sense for these to point to a specific build, how about this?
```suggestion
[assembly: AssemblyFileVersion("9.6.0.0")]
[assembly: AssemblyInformationalVersion("9.6.0.0 Custom build")]

```

Dnn.AdminExperience/ClientSide/Vocabularies.Web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "taxonomy",
-  "version": "9.4.4",
+  "version": "9.6.0-rc.3",
```
bdukes, line 3: These should be `9.6.0`, without `rc.3`, right?

SolutionInfo.cs
donker: Yeah, this was done using the GitVersion we use. I can also edit this by hand. Hopefully for the last time.

donker:
OK, I think it's good to go now

bdukes:
@valadas @donker we want to merge this back into develop now, right?

donker:
> 
> 
> @valadas @donker we want to merge this back into develop now, right?

Yes. But I am not confident enough I'll do that without breaking something ...

valadas:
Hmm, I was not involved in this one here, usually we only only want to merge stuff into master from a release branch. I don't think it would be a huge issue to merge this into develop, but in a few days when we do merge the release/9.6.1 into both master and develop, we will get 9.6.1 on both. My only concert is the issue template, I worked some automation (more coming up when I get some time) and we will need to try to avoid manual edits in the version list section. I am relying on the present of a line with "alpha" and a line with "latest supported release" to automate on 3 events:

1. Add the rc as soon as a release branch is created
2. Remove the rc and update the latest as soon as we publish a non-rc version
3. Maybe or maybe not automate for the future/xx branches (this changes less often)

valadas:
Oh, nevermind my comments here, I am late on notifications and this was already done...

PR #3723, opened 2020-05-03
Fix React build process (WIP)

donker:
Webpack now integrated with lerna workspaces


donker:
Sorry. This doesn't look like it's going anywhere. It won't build properly with this react project in our main monorepo approach. We'll have to take a different route.

kurtwilbies:
Why is this request closed?

PR #3733, opened 2020-05-05
Update bug report issue template from 9.6.0 release

bdukes:
Updated list of versions in bug report template


donker:
Errr ... wazzup with all the images?

bdukes:
Not sure why the action decided to run on everything, but looks like a good optimization regardless…

PR #3736, opened 2020-05-05
Fixed an issue with search results page rendering

### LINKED ISSUE #3712: Search Results distorted
## Description of bug
Search Results module look distorted

## Steps to reproduce
1. Install DNN 9.6.0 
2. enter a search term and submit
3. select "show more results"

## Current behavior
Explain the current behavior.

## Expected behavior
should look like before.

## Screenshots
image

## Error information
none

## Additional context
Might be caused by switch to JQuery 3.5.0

## Affected version

* [ ] 10.0.0 alpha build
* [ ]  9.5.1 alpha build
* [X]  9.6.0 latest supported release
* [ ]  9.5.0 

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [X] Microsoft Edge Chromium

valadas:
Closes #3712

Due to jQuery 3.5.0 not supporting shortcuts to closing html tags in templates:
ie: `<div class="something" />` is no longer supported and needs to be rewritten as `<div class="something></div>`

This PR fixes those instances used in the jQuery code for the search results.

### DIFF
--- a/DNN Platform/Website/Resources/Shared/scripts/dnn.searchBox.js
+++ b/DNN Platform/Website/Resources/Shared/scripts/dnn.searchBox.js
@@ -38,9 +38,9 @@
             var markup = '<div class="dnnSearchBox">';
             var advancedEnabled = this.options.showAdvanced && this.options.advancedId && $('#' + this.options.advancedId).length;
             if (advancedEnabled) {
-                markup += '<span class="dnnSearchBox_advanced_query" /><a class="dnnSearchBoxClearAdvanced"></a>';
+                markup += '<span class="dnnSearchBox_advanced_query"></span><a class="dnnSearchBoxClearAdvanced"></a>';
             }
-            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search" />' +
+            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search"></input>' +
                             '<a class="dnnSearchBoxClearText"></a>';
 
             markup += '<a class="dnnSearchButton"></a>';
--- a/DNN Platform/Website/Resources/Shared/scripts/dnn.searchBox.js
+++ b/DNN Platform/Website/Resources/Shared/scripts/dnn.searchBox.js
@@ -40,7 +40,7 @@
             if (advancedEnabled) {
                 markup += '<span class="dnnSearchBox_advanced_query"></span><a class="dnnSearchBoxClearAdvanced"></a>';
             }
-            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search"></input>' +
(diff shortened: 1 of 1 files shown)


DNN Platform/Website/Resources/Shared/scripts/dnn.searchBox.js
```diff
@@ -38,9 +38,9 @@
             var markup = '<div class="dnnSearchBox">';
             var advancedEnabled = this.options.showAdvanced && this.options.advancedId && $('#' + this.options.advancedId).length;
             if (advancedEnabled) {
-                markup += '<span class="dnnSearchBox_advanced_query" /><a class="dnnSearchBoxClearAdvanced"></a>';
+                markup += '<span class="dnnSearchBox_advanced_query"></span><a class="dnnSearchBoxClearAdvanced"></a>';
             }
-            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search" />' +
+            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search"></input>' +
```
bdukes, line 43: This change needs to be made, even though `<input />` is self-closing?

valadas: Yes, that is the main breaking change with jQuery 3.5.0, templates no longer support self closing tags

valadas: Described in https://jquery.com/upgrade-guide/3.5/

valadas: Well actually it's not that it is no longer supported, it's just that the behaviour has changed and the recommendation is to not use self closing tags for elements that may have content in them. in this instance, I guess just the span could solve it, but well, it also does not hurt for the input...

bdukes: In this case I'd prefer for `input` to remain self-closing, since it can't have children.  `<input></input>` isn't really valid.
```suggestion
            markup += '<input id="' + this.options.id + '_input" type="text" autocomplete="off" aria-label="Search" />' +
```

valadas: I agree, done, well at the same time as you actually :)

bdukes:
I saw https://github.com/marketplace/gitarborist which lists among its features:
>Merge PR once all checks pass (no more waiting around for CI after approval)
🤩

valadas:
>Merge PR once all checks pass (no more waiting around for CI after approval)

Well, I have mixed opinions, some PRs need to build and be downloaded and tested due to massive changes we cannot just review in the browser...

bdukes:
I think it would still require approval, but once approved we wouldn't need to wait for the build (as we are doing now).

PR #3756, opened 2020-05-12
Fix to settings saving in MVC modules

### LINKED ISSUE #3692: MVC style Settings control doesn't work
## Description of bug
Using a settings razor file and settingscontroller as per the MVC style of doing things does not persist settings to the database. The cause being that DNN insists of storing cached settings at the end of the update process, reverting previous changes.

## Steps to reproduce
Settings class in module:

``` cs
    public class ModuleSettings
    {
        [ModuleSetting]
        public string Foo { get; set; } = "Bar";

        public static ModuleSettings GetSettings(ModuleInfo module)
        {
            var repo = new ModuleSettingsRepository();
            return repo.GetSettings(module);
        }

        public void SaveSettings(ModuleInfo module)
        {
            var repo = new ModuleSettingsRepository();
            repo.SaveSettings(module, this);
        }
    }
    public class ModuleSettingsRepository : SettingsRepository<ModuleSettings>
    {
    }
```

Settings logic in SettingsController:

``` cs
        [HttpPost]
        [ValidateInput(false)]
        [DotNetNuke.Web.Mvc.Framework.ActionFilters.ValidateAntiForgeryToken]
        public ActionResult Settings(ModuleSettings settings)
        {
            settings.SaveSettings(ModuleContext.Configuration);
            return RedirectToDefaultRoute();
        }
```

## Current behavior
The above code runs and changes the values in the ModuleSettings database. But when the update logic completes those changes are reverted again.

## Error information
In the class MvcSettingsControl we see the following code:

``` cs
        public void UpdateSettings()
        {
            ExecuteModule();

            ModuleController.Instance.UpdateModule(ModuleContext.Configuration);
        }
```

The first line calls the update code in the settings controller of my module. The second line reverts this as the ModuleContext.Configuration still has the old ModuleSettings and writes them to the database.

The line where this happens is here: https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Entities/Modules/ModuleController.cs#L1966

Examining the code in the ModuleSettings control what strikes me is that there we already pass through the UpdateModule logic here:

https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Website/admin/Modules/Modulesettings.ascx.cs#L648

So there is no need to pass through it again in the MvcSettingsControl which kicks off a couple of lines below in UpdateSettings.

My proposal is to remove the extra UpdateModule in MvcSettingsControl.UpdateSettings.

## Affected version
* [x ]  9.4.4

donker:
Fixes #3692

### DIFF
--- a/DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
+++ b/DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
@@ -68,10 +68,12 @@ namespace DotNetNuke.Entities.Modules.Settings
                     if (attribute is ModuleSettingAttribute)
                     {
                         _moduleController.UpdateModuleSetting(moduleContext.ModuleID, mapping.FullParameterName, settingValueAsString);
+                        moduleContext.ModuleSettings[mapping.FullParameterName] = settingValueAsString; // temporary fix for issue 3692
                     }
                     else if (attribute is TabModuleSettingAttribute)
                     {
                         _moduleController.UpdateTabModuleSetting(moduleContext.TabModuleID, mapping.FullParameterName, settingValueAsString);
+                        moduleContext.TabModuleSettings[mapping.FullParameterName] = settingValueAsString; // temporary fix for issue 3692
                     }
                     else if (attribute is PortalSettingAttribute)
                     {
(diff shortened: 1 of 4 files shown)


mitchelsellers:
@donker it looks like this broke a whole bunch of tests, can you review?

bdukes:
Looks like `ModuleSettings` needs to be mocked out for the tests now

donker:
You're right, Brian. It fails because there are no ModuleSettings on the mocked ModuleInfo. And since you can't just create them empty there is no easy fix. Or is there?

bdukes:
I've pushed a fix

PR #3777, opened 2020-05-22
Improves deprecation of search indexes

### LINKED ISSUE #3776: IndexingProvider inheritance issue for Dnn10
## Some explanation
The IndexingProvider base class (abstract class) is built so 3rd parties can build their own indexing logic for module content.

This base class has a single **abstract** method `GetSearchIndexItems` that is marked as deprecated in Dnn7.1 and marked for removal in Dnn10 (used for the ISearchable interface). It is documented to use IndexSearchDocuments instead which is a **virtual** method.

Abstract methods **must** be implemented when the class is inherited, but virtual methods do not since they have a base implementation.

## Description of issue
3rd party developers are currently forced to implement a deprecated method to be able to build any custom indexer until it is removed. This is a 0-day breaking change, developers cannot just have a warning about the new implementation to use and be ready for then we remove the deprecated one.

## Additional context
I am not sure if there is a better solution and I am looking for opinions. I believe we should create a new base class (maybe call it IndexingProviderBase) and deprecate the old base class. On that new class we would make IndexSearchDocuments the one required method to override. Any better ideas?

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.06.01 release candidate
* [x] 09.06.00 latest supported release

valadas:
Closes #3776

Please read the issue during review. This is what comes to mind, but if there are better ideas, I am open to them :)

Basically my idea is to deprecate the whole base class, created a new copy of it which has the recommended method enforced and the deprecated method optional. That way 3rd parties will be able to change class now and be ready BEFORE the required method is removed.

### DIFF
--- a/DNN Platform/Library/DotNetNuke.Library.csproj
+++ b/DNN Platform/Library/DotNetNuke.Library.csproj
@@ -815,6 +815,7 @@
     <Compile Include="Services\Search\Entities\SortFields.cs" />
     <Compile Include="Services\Search\Entities\SearchStopWords.cs" />
     <Compile Include="Services\Search\Entities\SynonymsGroup.cs" />
+    <Compile Include="Services\Search\IndexingProviderBase.cs" />
     <Compile Include="Services\Search\Internals\Constants.cs" />
     <Compile Include="Services\Search\Internals\ISearchQueryStringParser.cs" />
     <Compile Include="Services\Search\Internals\LuceneResults.cs" />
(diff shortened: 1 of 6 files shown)


mitchelsellers:
It looks like a few public method signatures are changing, have we evaluated possibility of breaking change on this now?

valadas:
Which ones? What I did is a copy of the class, the old class is untouched. So, developers will get a warning that the old class is deprecated and they need to reference the new one. On the new one, the method that was mandatory is now optional (but still deprecated) and the one that people should implement (the new search as per Dnn 7.something when the ISearchable was replaced by ModuleSearchBase) is not mandatory. So, I do not see the potential breaking change unless I missed something?

mitchelsellers:
@valadas My bad, I missed the fact that there are no edits to the existing.  This looks great!

PR #3794, opened 2020-05-28
Unify CLS Compliance and Com Visibility

donker:
The CLSCompliance and ComVisible attributes are all over the place in the project. Because I was experiencing build issues locally, I decided to clean this up. Removing it from the individual AssemblyInfo files and putting it in the global SolutionInfo.cs.

### DIFF
--- a/SolutionInfo.cs
+++ b/SolutionInfo.cs
@@ -4,7 +4,9 @@
 // 
 #region Usings
 
+using System;
 using System.Reflection;
+using System.Runtime.InteropServices;
 
 #endregion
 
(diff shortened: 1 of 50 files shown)


SolutionInfo.cs
```diff
@@ -14,6 +16,8 @@
 
 // Review the values of the assembly attributes
 
+[assembly: CLSCompliant(false)]
```
bdukes, line 19: Looking at the removed attributes, it seems about 50/50 whether `CLSCompliant` was `true` or `false`.  What's the rationale behind setting everything to `false`?

mitchelsellers: Setting those that are existing as true to false could have some unexpected consequences for those building extensions with languages other than C#.  (https://docs.microsoft.com/en-us/dotnet/api/system.clscompliantattribute?view=netframework-4.7.2) Not that it is a huge audience, but I'm not sure of the long-term impacts here.

@donker Were there a number of errors due to CLS issues?

donker: I can't remember precisely any more. But if you guys want to leave this as is then let's skip this PR. Maybe it was on my end. My only goal was to unify what we do so we have less surprises. But if it makes sense that some dlls do and others don't then it should stay as it was.

donker:
Retracting due to lack of necessity/activity

PR #3795, opened 2020-05-28
Move Prompt to core libraries

donker:
This PR copies the Prompt classes and interfaces that are now scattered in the AE Library and Extensions projects to the DNN Core libraries. The goals are:

- to make the ability to create console commands more readily available to developers
- to clean up the code and make the coding experience for developers better and more intuitive

The new and the old system will run side by side until we deprecate the old system.


EPTamminga:
@mitchelsellers Am I correct if I say that the license text should also be incorporated in the official 5 Language packs?

sleupold:
@EPTamminga 
it should be contained in the manifest

mitchelsellers:
@EPTamminga For language packs @sleupold is correct, the manifest is the best place

EPTamminga:
@sleupold @mitchelsellers 
Of course, the license info is already in the manifest, but not (yet) with the text as listed here.

donker:
> 
> 
> The file headers for all files are incorrect, can we get them adjusted to follow the proper .NET Foundation standard from here: https://github.com/dotnet-foundation/foundation/blob/master/guidance/copyright.md
> 
> Which would be
> 
> // Licensed to the .NET Foundation under one or more agreements.
> // The .NET Foundation licenses this file to you under the MIT license.
> // See the LICENSE file in the project root for more information
> 
> Without additional comment lines

I just copied what we have in the other files. Doesn't it make more sense to do this once for all files?

mitchelsellers:
@donker I would rather NOT add more stuff that is wrong

sleupold:
there is no license file in DNN root folder, even not in /licences folder.
There is a license.txt in documentation folder, starting:
DotNetNuke - http://www.dnnsoftware.com
Copyright (c) 2002-2018
by DNN Corporation

sleupold:
PS: there is, of course, a license file in the GitHub root folder, but it is not included in release packages.

PR #3799, opened 2020-05-29
Makes the resource manager build as the other projects

valadas:
This PR makes the build consume our common components and build as the other modules build.

It is a work in progress, we still have browser errors that we need to look into using a debug build and we probably need to adjust many dependencies for a working build...


valadas:
@donker I just pushed some changes that resolve the css issue with the upload button. We still have another issue to resolve though, the dropzone is either not rendering or not working... Would you like that we merge this and make that fix as a separate PR since this PR goal was to make it build as part of the rest of the build process and I think this is now done ?

PR #3801, opened 2020-05-29
Copyright header

bdukes:
Based on @mitchelsellers' [comment on #3795](https://github.com/dnnsoftware/Dnn.Platform/pull/3795#pullrequestreview-420105436), I've updated all of the C# file headers.

This PR has a number of helper commits that I used, and then reverted.  This should be squashed when merging the PR.  However, I left them in place in case anyone thought they were useful and deserved their own PR.

In 810d73a and 096f19c, I added the [StyleCop.Analyzers](https://www.nuget.org/packages/StyleCop.Analyzers/) NuGet package to apply the header throughout the solution.  I know I've heard at least some folks discuss wanting to add StyleCop (@valadas? @mitchelsellers?), so we could add another PR which cherry-picks those changes (and once we have fewer in-progress PRs, we could apply some other style auto-fixes).

In 826bb68 and e78c244 I upgraded some dependencies that were causing an error when trying to add or update any other NuGet packages.  I didn't comb through these commits to reduce unnecessary changes, so they shouldn't be taken as-is, but it highlights a separate issue.


mitchelsellers:
@bdukes Addition of StyleCop would be AWESOME!  I am going to be setting up Sonar Cloud soon so StyleCop will help with scoring there

Issue #3599, opened 2020-02-27
Localization.GetString autoreplacing http to https


vmasanas:
## Description of bug
When site is running with https (settings SSLEnabled or SSLEnforce enabled) the method GetString will replace any http to https on content received from resource files.

I understand this should only be happening (if ever) on urls that are local to the portal. 
But a global replace on any resource is just wrong.

Since this has been added to the system a number of links we had on some resouces are failing because of this replacement.

Offending line is here:
https://github.com/dnnsoftware/Dnn.Platform/blob/7c790d82a6af27b28b1e2d65ba8737b2a450c257/DNN%20Platform/Library/Services/Localization/LocalizationProvider.cs#L83

## Steps to reproduce
1. Configure a site to run with https
2. Set either or both settings SSLEnabled or SSLEnforced to true
3. On any resx file add a link to any domain for example http://www.domain.com
4. Use the value of the resource in any place

## Current behavior
The text you'll get has been changed to https

## Expected behavior
The value that's on the resx file is used and not touched in any way

bdukes:
For reference, this change was made in 898cd7766b98f85abd20b3f91b25ded9a9a76604, historical discussion is in [DNN-10289](https://dnntracker.atlassian.net/browse/DNN-10289) and [DNN-8020](https://dnntracker.atlassian.net/browse/DNN-8020).

This fix is Not Good™ and now we're stuck needing to figure out a way to fix the fix without breaking folks relying on the fix.  Maybe instead of `resourceValue.Replace(@"http:", @"https:")` we could do `resourceValue.Replace(@"http://[Portal:URL]", @"https://[Portal:URL]")` to limit the scope a bit more?  Or even something like `resourceValue.Replace(@"http://[Portal:URL]", @"[Portal:FullURL]")`.

Any ideas @mitchelsellers @donker @sleupold?

mitchelsellers:
Wow @bdukes you are 100% this is beyond a Not Good™ fix, this is a true disaster.

Before I get into the specifics, this process also explains an interesting performance item that I noticed with regards to an excessive number of strings being generated on sites, I now know the root cause.  Given that .Replace() is used on EVERY single resource item, and strings are immutable in C# which means that we get 2 strings in memory for every resource loaded EVERY time we use HTTPS.  So lots of memory thrashing that in 90% of the requests isn't needed.

However, I'll let that go for a minute and focus on the task at hand.  A few thoughts.

[...]

PR #3808, opened 2020-06-03
Fix random replacement of http with https in resources (issue #3599)

donker:
An earlier "fix" for urls in emails added a blanket replace on each and every resource being retrieved. This is not just incorrect (as it also affects resources that should not be touched), it is also a performance hit.

This PR rolls back the fix and adds a "scheme" token to the portal settings. The [portal:scheme] token will resolve to either "https" or "http" depending on SSLEnabled for the portal. It is not used in any email in the framework currently, but if someone needs this in a template, it can be used.

Note that using the scheme token looks at the entire portal, so won't be a solution for those with mixed http/https situations. The [portal:fullurl] uses the current page's SSL setting, but obviously depends on context and so was never meant for emails.

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -699,15 +699,19 @@ namespace DotNetNuke.Entities.Portals
             var isPublic = true;
             switch (lowerPropertyName)
             {
+                case "scheme":
+                    propertyNotFound = false;
+                    result = SSLEnabled ? "https" : "http";
+                    break;
                 case "url":
                     propertyNotFound = false;
                     result = PropertyAccess.FormatString(PortalAlias.HTTPAlias, format);
                     break;
-                case "fullurl": //return portal alias with protocol
+                case "fullurl": //return portal alias with protocol - note this depends on HttpContext
                     propertyNotFound = false;
                     result = PropertyAccess.FormatString(Globals.AddHTTP(PortalAlias.HTTPAlias), format);
                     break;
-                case "passwordreminderurl": //if regsiter page defined in portal settings, then get that page url, otherwise return home page.
+                case "passwordreminderurl": //if regsiter page defined in portal settings, then get that page url, otherwise return home page. - note this depends on HttpContext
                     propertyNotFound = false;
                     var reminderUrl = Globals.AddHTTP(PortalAlias.HTTPAlias);
                     if (RegisterTabId > Null.NullInteger)
--- a/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
+++ b/DNN Platform/Library/Services/Localization/LocalizationProvider.cs
@@ -80,18 +80,7 @@ namespace DotNetNuke.Services.Localization
                 Logger.WarnFormat("Missing localization key. key:{0} resFileRoot:{1} threadCulture:{2} userlan:{3}", key, resourceFileRoot, Thread.CurrentThread.CurrentUICulture, language);
             }
 
-            return string.IsNullOrEmpty(resourceValue) ? string.Empty : RemoveHttpUrlsIfSiteisSSLEnabled(portalSettings, resourceValue);
-        }
-
-        private string RemoveHttpUrlsIfSiteisSSLEnabled(PortalSettings portalSettings, string resourceValue)
-        {
-
-            if (portalSettings != null && (portalSettings.SSLEnabled || portalSettings.SSLEnforced))
-            {
-                resourceValue = resourceValue.Replace(@"http:", @"https:");
-            }
-
-            return resourceValue;
+            return string.IsNullOrEmpty(resourceValue) ? string.Empty : resourceValue;
         }
 
         /// <summary>


sleupold:
once this PR has been included, I'll have a look to update texts of the resource files.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -699,15 +699,19 @@ public string GetProperty(string propertyName, string format, CultureInfo format
...
+                case "scheme":
+                    propertyNotFound = false;
+                    result = SSLEnabled ? "https" : "http";
+                    break;
                 case "url":
                     propertyNotFound = false;
                     result = PropertyAccess.FormatString(PortalAlias.HTTPAlias, format);
                     break;
-                case "fullurl": //return portal alias with protocol
+                case "fullurl": //return portal alias with protocol - note this depends on HttpContext
```
bdukes, line 710: Does `PortalAlias.HTTPAlias` depend on `HttpContext`?  I think if you use the `PortalSettings` constructor that takes a `PortalAliasInfo`, it shouldn't depend on context.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -699,15 +699,19 @@ public string GetProperty(string propertyName, string format, CultureInfo format
             var isPublic = true;
             switch (lowerPropertyName)
             {
+                case "scheme":
+                    propertyNotFound = false;
+                    result = SSLEnabled ? "https" : "http";
```
bdukes, line 704: The check in the code we're replacing is `portalSettings.SSLEnabled || portalSettings.SSLEnforced`.  I don't know that it hurts anything for it to remain also checking `SSLEnforced` (probably better to err on the side of more HTTPS rather than less).

bdukes:
So, there's a small chance for this to be a breaking change.  If someone has a modified version of an email template which uses the old `http://[Portal:URL]/default.aspx?ctl=PasswordReset&resetToken=[Membership:PasswordResetToken]` URL, it will no longer get auto-upgraded to `https`.  @donker @mitchelsellers do you think it's worth either doing a find/replace in the email code, or running an update of localized resources during upgrade (i.e. find/replace `http://[Portal:URL]` with `[Portal:FULLURL]`)?

DNN Platform/Library/Entities/Portals/PortalSettings.cs
sleupold: it should check for portalSettings.SSLEnabled && portalSettings.SSLEnforced - just SSLEnabled will deliver secured pages using http, if not called using https (e.g. if moved to a different location without https being bound to the IIS website.

sleupold:
> So, there's a small chance for this to be a breaking change. If someone has a modified version of an email template which uses the old `http://[Portal:URL]/default.aspx?ctl=PasswordReset&resetToken=[Membership:PasswordResetToken]` URL, it will no longer get auto-upgraded to `https`. @donker @mitchelsellers do you think it's worth either doing a find/replace in the email code, or running an update of localized resources during upgrade (i.e. find/replace `http://[Portal:URL]` with `[Portal:FULLURL]`)?

Yes, this is a breaking change, but should not affect too many sites. However, we should definetely mention it in release notes

DNN Platform/Library/Entities/Portals/PortalSettings.cs
donker: @bdukes : If you follow the AddHttp method, it uses HttpContext. I spent some time tracing this code and thought it was a good idea to include this in the comments as it means it is not safe for background tasks.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
donker: My thought was indeed to err on the side of https. Rapidly more and more technologies are refusing to work with plain http (e.g. mobile apps). So if a site has so much as allowed SSL, I suggest the default return of "scheme" should be https. Rather than waiting to see if it is enforced.

donker:
> 
> 
> So, there's a small chance for this to be a breaking change. If someone has a modified version of an email template which uses the old `http://[Portal:URL]/default.aspx?ctl=PasswordReset&resetToken=[Membership:PasswordResetToken]` URL, it will no longer get auto-upgraded to `https`. @donker @mitchelsellers do you think it's worth either doing a find/replace in the email code, or running an update of localized resources during upgrade (i.e. find/replace `http://[Portal:URL]` with `[Portal:FULLURL]`)?

Absolutely. Which is why it goes in 9.7. Like Sebastian says, chances are slim and frankly, it is IMHO worse that those who included external fixed links using http find that the http was changed. So I am favouring those that highlighted that latter issue.

bdukes:
Regarding the breaking change, the `CoreMessagingScheduler`, at least, does its own replace HTTP with HTTPS, so that should still continue to work:
https://github.com/dnnsoftware/Dnn.Platform/blob/91223d2b4307ea98a1c4e5b8a472e996c60a1c76/DNN%20Platform/Library/Services/Social/Messaging/Scheduler/CoreMessagingScheduler.cs#L572-L581

DNN Platform/Library/Entities/Portals/PortalSettings.cs
bdukes: Right, so `SSLEnabled || SSLEnforced` could be `true` more often than just `SSLEnabled` (though I'm not sure if there's any valid scenario where `SSLEnforced` is `true` while `SSLEnabled` is `false`)

bdukes:
But it doesn't look like `CoreMessagingScheduler` uses the standard token replacement, it has custom tokens (e.g. `[SITEURL]` instead of `[Portal:URL]`).  The scenario where we would expect breakage is the call to [`Mail.SendMail(newUser, MessageType.UserRegistrationPublic, PortalSettings)` in `UserModuleBase`](https://github.com/dnnsoftware/Dnn.Platform/blob/91223d2b4307ea98a1c4e5b8a472e996c60a1c76/DNN%20Platform/Library/Entities/Modules/UserModuleBase.cs#L449).  @mitchelsellers @valadas are you okay with this small breaking change in 9.7?  It's not a big deal to me.

`Globals.AddHTTP` does reference `HttpContent.Current`, but it checks for `null` first, so it should be safe to use from a scheduled task.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
bdukes: `Globals.AddHTTP` does reference `HttpContent.Current`, but it checks for `null` first, so it should be safe to use from a scheduled task.

mitchelsellers:
@bdukes Yes, I think we are fine with this one, it errs on the side of security essentially.  But I'm flagging as potentially breaking so we are sure to grab it in the release notes

DNN Platform/Library/Entities/Portals/PortalSettings.cs
mitchelsellers: Looking at this, and the option values, it seems that you could have SSL Enforced = true, but SSL Enabled = false.  For this, I think we have to rely only on the `SSLEnabled` flag though, it may be a bit aggressive, but for current standards should be acceptable

PR #3802, opened 2020-05-29
StyleCop

bdukes:
Add StyleCop.Analyzers NuGet package to all projects, reference a shared `stylecop.config` file in projects

PR #3810, opened 2020-06-05
StyleCop configuration

bdukes:
First follow-up to #3802 

I've added the default configuration inside `stylecop.json`, I'm going to propose some changes from the defaults as PR review comments, and we can try to come to a consensus.  There is also the option of [using a `ruleset` file to turn some rules off altogether](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#sharing-configuration-among-solutions).

I've also added an `.editorconfig` file, which is a mixture of the [`.editorconfig` file from StyleCop](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/.editorconfig) and the [`.editorconfig` files from `dotnet-format`](https://github.com/dotnet/format/blob/master/.editorconfig), with StyleCop "winning" where they conflict.  Some of the notable differences:
* Always use `this` 
* No prefix for static and instance fields (the `dotnet-format` style uses `s_` for static fields and `_` for fields)
* Prefer `var` only where type is obvious (the `dotnet-format` style prefers `var` everywhere)
* Require braces (the `dotnet-format` style prefers but does not enforce braces)

### DIFF
--- a/stylecop.json
+++ b/stylecop.json
@@ -16,7 +16,7 @@
       ],
       "systemUsingDirectivesFirst": true,
       "usingDirectivesPlacement": "insideNamespace",
-      "blankLinesBetweenUsingGroups": "allow"
+      "blankLinesBetweenUsingGroups": "require"
     },
     "namingRules": {
       "allowCommonHungarianPrefixes": true,
--- a/stylecop.json
+++ b/stylecop.json
@@ -29,7 +29,7 @@
       "topLevelTypes": ["class"]
     },
     "layoutRules": {
-      "newlineAtEndOfFile": "allow",
+      "newlineAtEndOfFile": "require",
       "allowConsecutiveUsings": true
     },
     "documentationRules": {
--- a/stylecop.json
+++ b/stylecop.json
@@ -1,10 +1,49 @@
 {
-    "$schema": "https://raw.githubusercontent.com/DotNetAnalyzers/StyleCopAnalyzers/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json",
-    "settings": {
-      "documentationRules": {
-        "companyName": ".NET Foundation",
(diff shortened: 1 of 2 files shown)


stylecop.json
```diff
@@ -1,10 +1,49 @@
...
+      "allowedHungarianPrefixes": [],
+      "allowedNamespaceComponents": [],
+      "includeInferredTupleElementNames": false,
+      "tupleElementNameCasing": "PascalCase"
+    },
+    "maintainabilityRules": {
+      "topLevelTypes": ["class"]
+    },
+    "layoutRules": {
+      "newlineAtEndOfFile": "allow",
```
bdukes, line 32:
```suggestion
      "newlineAtEndOfFile": "require",
```

stylecop.json
```diff
@@ -1,10 +1,49 @@
...
+      "elementOrder": [
+        "kind",
+        "accessibility",
+        "constant",
+        "static",
+        "readonly"
+      ],
+      "systemUsingDirectivesFirst": true,
+      "usingDirectivesPlacement": "insideNamespace",
+      "blankLinesBetweenUsingGroups": "allow"
```
bdukes, line 19:
```suggestion
      "blankLinesBetweenUsingGroups": "require"
```

stylecop.json
david-poindexter: Why should this be required?  Just for consistency's sake?

stylecop.json
david-poindexter: I agree with this one - it will certainly help with readability.

stylecop.json
bdukes: Yes.  Also if it's required then we can just auto-fix it and be done.

david-poindexter: Sounds like a good plan - thanks @bdukes - I'll go ahead and commit these suggestions.

PR #3811, opened 2020-06-05
StyleCop fixes

bdukes:
This is a follow-up to #3810.  I plan to add some more commits with auto-fixes of various StyleCop rules, so this is just a start.  That said, we do want to commit these changes quickly, so they don't get stale. 

Unfortunately, I've not been able to find an automated way to apply these fixes, so each change takes a fair amount of time.  The steps to apply this one change include the following:

1. Open `DNN_Platform.sln` in Visual Studio 2019
1. Find somewhere violating this rule (via the Error List if you want to wait for code analysis to run on the whole solution)
1. Select the Light Bulb icon and then choose to address the issue in the solution
1. Wait
1. Approve the preview of the change
1. Wait
1. Commit all of the changes files

Someone focused on this task could probably knock out a few thousand warnings in an hour.  I do suggest doing one commit per change, in case there's a consensus that one of the changes is one we don't want (some of them may be breaking changes, e.g. capitalization of non-private members)


bdukes:
I also started with one of the more contentious issues, to see if anyone has strong enough opinions to adjust.  This is a rule which differs between the StyleCop and the dotnet-format defaults, where the .NET guidelines suggest prefixing private fields with an underscore, and to prefer not using `this`, and StyleCop uses `this` everywhere and allows no prefixes.

valadas:
I am a fan of using this and no underscores because the underscores are used also as a discard object in both c# and typescript and I find it confusing for local variables... So I guess I prefer the stylecop way...

david-poindexter:
> I am a fan of using this and no underscores because the underscores are used also as a discard object in both c# and typescript and I find it confusing for local variables... So I guess I prefer the stylecop way...

Me too.

PR #3812, opened 2020-06-05
Use newer logic to send Host SMTP test emails

### LINKED ISSUE #3813: Testing SMTP uses old logic
## Description of bug
https://github.com/dnnsoftware/Dnn.Platform/pull/3237
## Steps to reproduce
1. send test emails with different configurations of portal/host
2. if host email is not set even when using portal smtp settings the test will fail

## Current behavior
The smtp test is using host settings to and from email addresses only for testing emails
Setting another domain that is not allowed to send emails as host email will potentially break portal smtp settings.

## Expected behavior
Portal settings for to and from addresses send emails from site security administrator account or host account.

## Error information
Blank Host Email in Host SMTP Settings will cause an error currently while trying to test smtp using portal settings.  Logged in as host during this test.  I set different email addresses

## Additional context

 I set different email addresses in a variety of ways and configurations to see which emails are being used and when and it reminded me of previous fix I had done.  I may have only tested while logged in as an administrator  and not as host so this is why it still acts this way while logged in as host.  It checks if host or not so things should work as intended similar to when you are logged in as admin (still need to test that theory of admin or host uses two different test SMTP methods/files as the names of the files may insist on) with the PR fix submitted. (still need to test that theory of admin or host uses two different test SMTP methods/files as the names of the files may insist on) 

This could be when logged in as host, when you are an admin maybe it worked.  I think the same logic should work in this location as well to set things right.  I know it worked before I will test as an administrator rather than host see if it makes a difference.

## Affected version

* [ ] 10.00.00 alpha build (in development)
* [ x ] 09.06.01 latest supported release

## Affected browser

* [ x ] Chrome
* [  x ] Firefox
* [  x] Safari
* [ x ] Internet Explorer 11
* [x ] Microsoft Edge (Classic)
* [ x] Microsoft Edge Chromium

thabaum:
Fixes #3813 
For Reference: logic was added here but only administrators got the fix... so this would be in addition to the following PR https://github.com/dnnsoftware/Dnn.Platform/pull/3237
## Summary

In the above referenced PR the logic was created to send emails to the user using host or portal while logged in as administrator.

A second controller is used for host accounts which also needed the same logic to allow testing different portals SMTP settings while logged in as a host similar to how you can while logged in as an administrator currently.  

The same logic needs added to the host users SMTP test method that is being used for administrator SMTP tests as it is still utilizing the previous logic for the Send From and To variables that get set.

This PR puts the same administrator SMTP test logic in the host method.

Additional Context:
I will work on making sure documentation supports this over the next week if this is accepted.

I thought this worked before but again maybe it was me testing logged in as an administrator, logged in as host currently only tests Host SMTP Settings even when Site SMTP Server Settings  is set to Portal.

One issue that still remains is error handling a blank Host Email in Host SMTP Settings...

Thanks for the support hope this helps!

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpHostController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpHostController.cs
@@ -126,8 +126,9 @@ namespace Dnn.PersonaBar.Servers.Services
         {
             try
             {
-                var mailFrom = Host.HostEmail;
-                var mailTo = request.SmtpServerMode == "h" ? Host.HostEmail : PortalSettings.UserInfo.Email;
+                var smtpHostMode = request.SmtpServerMode == "h";
+                var mailFrom = smtpHostMode ? Host.HostEmail : PortalSettings.Email;
+                var mailTo = UserInfo.Email;
 
                 var errMessage = Mail.SendMail(mailFrom,
                     mailTo,


thabaum:
I wondered the same thing... figured that could be a topic down the road if you wanted to keep a note on it.

PR #3824, opened 2020-06-08
Update copyright to satisfy .net foundation requirements

donker:
The old copyright text at the head of every .cs file was outdated. This PR updates it to the new requirements as per the .NET Foundation guidelines.


bdukes:
Most of these changes were resolved in #3801, so this commit ends up being full of mostly whitespace changes.  There are a few files that were missed in that PR, so we can take those, but I'm not sure if there's benefit to the rest of the changes.

donker:
OK, I'll retract

PR #3829, opened 2020-06-09
Create 09.06.02.config

Tychodewaard:
Add webp support


DNN Platform/Website/Install/Config/09.07.00.config
```diff
@@ -0,0 +1,7 @@
+<configuration>
+  <nodes configfile="Web.config">
+    <node path="/configuration/system.web/staticContent/ action="update" key="name" collision="overwrite">
```
bdukes, line 3:
```suggestion
    <node path="/configuration/system.web/staticContent/ action="update" key="fileExtension" collision="ignore">
```

bdukes:
Yeah, if we're doing a 9.6.2 release, let's include this.

valadas:
Ok, so let's decide on the next meeting and we can adjust the filename accordingly...

DNN Platform/Website/Install/Config/09.06.02.config
```diff
@@ -0,0 +1,7 @@
+<configuration>
+  <nodes configfile="Web.config">
+    <node path="/configuration/system.web/staticContent/ action="update" key="fileExtension" collision="ignore">
```
valadas, line 3:
```suggestion
    <node path="/configuration/system.web/staticContent/" action="update" key="fileExtension" collision="ignore">
```

There was a missing double-quote that made install/upgrades fail with invalid xml. I will accept my own change here and let it build and try again...

PR #3834, opened 2020-06-11
Change upgrade URL to extension manifest URL instead of the upgrade service redirect

### LINKED ISSUE #3823: Change upgrade URL to extension manifest URL instead of the upgrade service redirect
## Description of bug
Currently the "Upgrade?" URL in Settings > Extensions > Installed Extensions points to `https://dnnplatform.io/redirect.aspx?core=______&version=______&type=______&name=______` and that is an invalid URL.

## Steps to reproduce
1. Go to Settings > Extensions > Installed Extensions
2. View list of extensions
3. Hover over "Upgrade?" URL
4. See invalid URL

## Current behavior
Invalid URL.

## Expected behavior
This should point to the URL for the extensions as indicated in the extension's manifest.

## Screenshots
n/a

## Error information
n/a

## Additional context
n/a

## Affected version
* [x] 10.00.00 alpha build (in development)

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge Chromium

david-poindexter:
## Summary
Resolves #3823 

Several problems are being addressed regarding the `Upgrade?` column on the Installed Extensions tab.  

Prior to this PR, a `##.##.##` image is generated by the Update Service based on the existence of any reported instance of an extension version greater than the one currently installed.  That image would be used to create an image link to a redirect URL in the Update Service that is currently not functional.  Therefore, the user experience is not only misleading but also broken.

The system is also flawed in the sense that there are often valid use cases for there being a version of an extension that has been reported to the Update Service that is invalid or under development (i.e., v10 extensions for DNN).  Therefore, the version image created by the Update Service results in a false positive (e.g. "99.99.99" or "10.00.00").  

Therefore, the PR still checks to see that a potential upgrade is available, but instead of rendering a version image, the UI simply renders a "Check" text link that now points to the URL contained in the extension manifest instead of the **upgradeURL** (the broken redirect URL referenced earlier).  

With these changes, we could now update all core extension URLs to point to the appropriate URLs.  Also, we should notify extension vendors of this change so they can leverage this new tie-in in a different way if desired.

See below for a visual of the UI changes.

Screenshot 2020-06-10 20 23 00


mitchelsellers:
This change renders the update service and the associated community metrics useless.  As such, I do not approve this change

david-poindexter:
@mitchelsellers okay, it appears we had a gross misunderstanding among the leadership team then.  Let's discuss when you return from vacation.

david-poindexter:
Thanks for the quick conversation @mitchelsellers - I'll push a commit shortly to add the image back in, but I'll leave the URL as is.  Enjoy your vacation!

david-poindexter:
Okay, changes pushed and the version image is now back in.

Screenshot 2020-06-10 21 17 42

PR #3852, opened 2020-06-23
Fix casing issues in 9.6.2 data provider script

donker:
An error crept in with PR #3797 with casing of objectQualifier and databaseOwner


valadas:
@dnnsoftware/approvers Can we get a second approval and merge please....

PR #3859, opened 2020-06-25
Adjust webconfig on new installs in light of PR #3829

donker:
PR #3829 added support for webp images. This PR ensures this is not just the case for upgraded sites but also new sites.


valadas:
Tried #3861 on both upgrades and clean installs and the config applies, so this PR is not needed. Please review #3861 and we should be good to go.

PR #3882, opened 2020-06-30
SA12xx fixes - rearrange members

bdukes:
Also adjusted some whitespace and moved a couple of classes into
separate files


valadas:
Did'nt we discuss holding on these global auto-fixes after we finish with 9.6.2... We may have huge conflicts to resolve when we merge back release/9.6.2 into develop...

valadas:
Ok, not too bad, only one file conflict, see https://github.com/dnnsoftware/Dnn.Platform/pull/3888

I will try to resolve the conflict later today/tomorrow...

PR #3894, opened 2020-07-05
Modify dnn:Logo skin object to handle interactive SVG files

### LINKED ISSUE #3893: Modify Logo skin control to handle .svg files
## Description of problem
Logo control doesn't handle .svg files

## Description of solution
Modify the Logo skin control to handle a .svg as a logo.

mathisjay:
Resolves #3893

### DIFF
--- a/DNN Platform/Website/admin/Skins/Logo.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Logo.ascx.cs
@@ -67,6 +67,7 @@ namespace DotNetNuke.UI.Skins.Controls
 
                             HtmlGenericControl obj = new HtmlGenericControl("object");
                             obj.Attributes.Add("class", this.CssClass);
+                            obj.Attributes.Add("type", "image/svg+xml");
 
                             if (!string.IsNullOrEmpty(imageUrl))
                             {
--- a/DNN Platform/Website/admin/Skins/Logo.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Logo.ascx.cs
@@ -4,6 +4,7 @@
 namespace DotNetNuke.UI.Skins.Controls
 {
     using System;
+    using System.Web.UI.HtmlControls;
     using System.Web.UI.WebControls;
 
     using DotNetNuke.Abstractions;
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3894) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3894) before we can accept your contribution.<br/><br/>:x: mathisjay [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3894)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=3894) it.</sub>

valadas:
I was thinking we should actually read the file contents and render the real svg code. CSS support is different when used in an image, an object and a pure inline svg. Also we should probably read the file contents and cache it. And I also vote for dnn10 on this to not break existing themes.

valadas:
And if we go that route, I would like to also implement it for page icons in DDR menu at the same time so people don't have to rewrite their themes twice :)

mathisjay:
@david-poindexter SVGs were selectable pre 9.6.2 in the Logo control on within Site Settings, so I'm not sure how they would have set it, unless in the DB directly. perhaps?  After 9.6.2, it is now possible to select an SVG logo and it will display as defined by the SVG, but you can't manipulate it with CSS because it is using an  tag to load it.

@valadas  I thought about reading the contents of the SVG file and just parsing out the SVG code and pasting that but I I thought this approach was simpler.  I based my solution on this SO question: https://stackoverflow.com/questions/4476526/do-i-use-img-object-or-embed-for-svg-files

The main goal here was to support the 'currentColor' CSS variable so that the SVG logo can be colored to whatever the fore-color of the header is.  This is primarily seen when building a theme with a transparent fixed header over an image/video and the header then becomes opaque when scrolling down.

Something like this:
https://yevgenysim.github.io/incline/index.html

valadas:
Just for proper documentation on the actual issue here. I just had a chat with @mathisjay on Slack. As suspected css support is different with a real inline svg and an svg that is loaded in an img tag or an object tag. For instance a couple of popular css properties like fill and stroke can be overridden only when the svg is inline, these are important because many popular svg librarires do not define currentColor for the fill or stroke and for proper support of every possible svg, inline would really be ideal.

Also, some svg files have headers and we would probably need to parse the svg to only grab the svg node and cache that node as a string then just inject the string in a literal control...

I am closing this PR and copying this comment into the issue. I am willing to work on this but if we do it, it should be in 10.0.0 exactly so we can properly document the change as a possible breaking change for themes.

PR #3897, opened 2020-07-06
Tabs that are not Indexed shouldn't be in the SiteMap

### LINKED ISSUE #3896: Non Indexed Tabs should not be in the SiteMap
Ideally - a sitemap tells a crawler what pages to scan and index.  Therefore, any tab that is not to be indexed should NOT be included in the sitemap output.

The issue is then compounded if the Priority is kept at 0.5.  Because the sitemap states to crawl this page and and it means something, by being set as to not index - it's also telling the search engine to ignore it.  Google is negatively affecting rankings due to this issue

ajwaka:
Fixes #3896 

## Summary
Non-indexed tabs should not be included in the sitemap. This effectively tells the crawler two different statements.  

The only change needed is to validate that the tab is supposed to be index prior to exporting it to the sitemap.

### DIFF
--- a/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
+++ b/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
@@ -65,7 +65,7 @@ namespace DotNetNuke.Services.Sitemap
                         (Null.IsNull(tab.StartDate) || tab.StartDate < DateTime.Now) &&
                         (Null.IsNull(tab.EndDate) || tab.EndDate > DateTime.Now) && this.IsTabPublic(tab.TabPermissions))
                     {
-                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished)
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3897) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3897) before we can accept your contribution.<br/><br/>:x: ajwaka [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3897)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=3897) it.</sub>

PR #3898, opened 2020-07-06
Modify Logo skin object to handle SVGs

### LINKED ISSUE #3893: Modify Logo skin control to handle .svg files

mathisjay:
Resolves #3893


DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -22,16 +24,20 @@ namespace DotNetNuke.UI.Skins.Controls
...
         {
             this._navigationManager = Globals.DependencyProvider.GetRequiredService<INavigationManager>();
+            this._cacheKey = string.Format(DataCache.PortalCacheKey, this.PortalSettings.PortalId, this.PortalSettings.CultureCode) + "LogoSvg";
         }
 
         public string BorderWidth { get; set; }
 
         public string CssClass { get; set; }
 
+        public Nullable<bool> isSVG { get; set; }
```
valadas, line 39: I was thinking we should provide a property called injectSVG, if null or false we use the old behaviour, if true, we check the file extension and decide privatelly if it is an svg or not.

This would make it a non breaking change we can merge sooner and the way it is now, the theme developer would need to support only svg or not svg logos... which is not much reusable as it stands.

mitchelsellers: Just a note here that public properties should be Title Cased so InjectSvg or IsSvg (https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

My only concern with this property is that setting "IsSVG" in the skin object doesn't force the UI user to select an SVG image, so you could get unexpected results.

I think that InjectSvg as a property that then if an SVG image is provided it is injected would be the "least breaking" way to do this.

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -58,6 +64,38 @@ protected override void OnLoad(EventArgs e)
                         {
                             this.imgLogo.ImageUrl = imageUrl;
                             logoVisible = true;
+                        }
+
+                        if (this.isSVG.GetValueOrDefault() == true)
```
mitchelsellers, line 69: This entire process is going to need defensive coding around it due to the fact that the provided image, might not actually be an SVG image so you will get errors and other stuff as part of this.

Additionally I do not believe that fileInfo.PhysicalPath is actually safe for use with third-party folder providers.  (But need to check)

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: I renamed the property to InjectSVG, but it does exactly what @valadas says.

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: I added a check to make sure the file at least has an 'svg' extension.  If the person uploading a non-SVG as a .svg file, well... we can't do everything for them...  They could just as easily select a logo file that has a .jpg extension that is not actually a .jpg.

mathisjay: I don't know how the third party folder stuff works, so I'll need some help there.

valadas:
For the sanitization part I believe a simple try catch, if there is not svg node found we just log the info and don't show any logo.

As for third party, I think using the File manager class is the way to go? @mitchelsellers

mitchelsellers:
@mathisjay @valadas Yes, I think the FileManager class is the way we will have to go for the download of the file for parsing as PhysicalPath is not always going to be there and not always locally.

I'm going to throw a few more notes in on a review form a performance perspective.  

Lastly, in this situation if the provided SVG file is not valid, an empty string is going to be rendered, is this proper?  I think we need to give something better for a fallback?

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -58,6 +64,40 @@ protected override void OnLoad(EventArgs e)
                         {
                             this.imgLogo.ImageUrl = imageUrl;
                             logoVisible = true;
+                        }
+
+                        if (this.InjectSVG.GetValueOrDefault() == true)
+                        {
+                            logoVisible = false; // hide the <img>, we are going to inject an <svg> tag instead.
```
mitchelsellers, line 71: This isn't necessary as the logo is already not visible per the declaration above

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -58,6 +64,40 @@ protected override void OnLoad(EventArgs e)
                         {
                             this.imgLogo.ImageUrl = imageUrl;
                             logoVisible = true;
+                        }
+
+                        if (this.InjectSVG.GetValueOrDefault() == true)
+                        {
+                            logoVisible = false; // hide the <img>, we are going to inject an <svg> tag instead.
+
+                            hypLogo.CssClass = this.CssClass; // put the custom CSS class on the hyperlink instead
```
mitchelsellers, line 73: This provides an inconsistency for the designer by assigning the CSS class either to the image itself or the hyperlink if SVG.  Is this a desired/acceptable behavior change  @david-poindexter I'd love your $0.02 on this change

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -58,6 +64,40 @@ protected override void OnLoad(EventArgs e)
...
                             logoVisible = true;
+                        }
+
+                        if (this.InjectSVG.GetValueOrDefault() == true)
+                        {
+                            logoVisible = false; // hide the <img>, we are going to inject an <svg> tag instead.
+
+                            hypLogo.CssClass = this.CssClass; // put the custom CSS class on the hyperlink instead
+
+                            string svg = string.Empty;
```
mitchelsellers, line 75: This declaration isn't needed, please remove this if we can to keep string instantiation reduced

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -58,6 +64,40 @@ protected override void OnLoad(EventArgs e)
...
+                                        CachingProvider.Instance().Insert(this._cacheKey, svg);
+                                    }
+                                }
+                            }
+                            else
+                            {
+                                svg = cacheSvg;
+                            }
+
+                            Literal litSvg = new Literal();
```
mitchelsellers, line 97: We need some sort of fallback at this point if the file couldn't be obtained.

mathisjay:
I'll look into the FileNamager class, not used that before.
If the .svg is not valid, nothing will be injected into the hyperlink.  I guess we could inject the Site Name just so it is not empty. but we aren't doing anything special if they select an invalid .jpg for the logo, it will just show a broken image.  I'd say, just keep it simple...

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: actually, it will be set to visible in the logic block above.  I was trying to not change the existing change as much as possible.

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: This was on purpose.  It thought it desirable to include the CssClass somewhere and there was nowhere else to put it unless we want to create another html tag.  

Also, putting it on the <a> tag is at least consistent with how BootStrap expects the markup to look.
  <a class="navbar-brand" href="#">Navbar</a>

Since this is new functionality I think we have greenfield to do what we want here...

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: Is there a better way to do this? 

I am setting the value if the string inside an if block and referencing the value later from outside of the if block.

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mathisjay: What would you suggest here?  Inject the Portal Name as string literal?

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mitchelsellers: In all reality that means that we might want to change this process to be even further up the chain as there is no need to do all of that extra work on every load if we are doing it the SVG way

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mitchelsellers: Just use the cacheSvg value from below, no need for two strings

DNN Platform/Website/admin/Skins/Logo.ascx.cs
mitchelsellers: I'm not sure on this, but there needs to be something somewhere I think

mitchelsellers: Also, this still needs a try/catch around it

mathisjay:
Made the requested changes.  I just rewrote the whole process since the previous code was only expecting an image and not efficient when allowing other types.

The previous code was already using the FileManager class to load the Logo file, so I think we are good there.   Please verify.

mathisjay:
I made a few more optimizations.

I tested the current code with the following use cases;
1) No Logo file is specific -> empty hyperlink (current behavior)
2) Image type (jpg, png, gif) specified, but not found -> broken image shown (current behaviour) - this could only happen if the file were deleted after being selected
4) SVG file selected as logo but theme skin object DOES NOT request InjectSVG -> Displays SVG using an  tag (current behaviour)
5) SVG file selected and theme skin object DOES request InjectSVG -> Display SVG using an SVG tag and uses cache for SVG code
6) Skin object requests InjectSVG but Logo file is not an SVG file type -> hyperlink text shows the PortalName instead.
7) Skin object request InjctSVG and the file type is .svg but the file contains no <svg> tag -> hyperlink text shows the PortalName instead.

I believe this maintains all the current functionality and only enables the new behaviour if the skin object requests InjectSVG.

valadas:
@mathisjay as discussed earlier, i was working on this PR probably at the same time as you and just read your last comment after pushing to your branch... Let me know if you like my changes or if I messed up yours...

@mitchelsellers would you re-review, this should:
- Not change the existing behaviours (needs the InjectSvg opt-in)
- Will do some validation for an existing svg xml node
- Will add the recommended ADA compliance information.

I also fixed the Stylecop warnings on this file.

mathisjay:
Thanks @valadas 
I added a new opt-in property LinkCssClass that will put the Css class on the hyperlink instead of the `` or `<svg>`

DNN Platform/Website/admin/Skins/Logo.ascx.cs
```diff
@@ -3,82 +3,182 @@
...
                     var fileInfo = this.GetLogoFileInfo();
                     if (fileInfo != null)
-                    {
-                        string imageUrl = FileManager.Instance.GetUrl(fileInfo);
-                        if (!string.IsNullOrEmpty(imageUrl))
-                        {
-                            this.imgLogo.ImageUrl = imageUrl;
-                            logoVisible = true;
+                    {
+                        if (this.InjectSVG.GetValueOrDefault() && fileInfo.Extension == "svg")
```
valadas, line 74: I will adjust this here to do a case insensitive comparison on the file extension

mathisjay: good idea

valadas:
Ok, what do you think now @mitchelsellers  ?

bdukes:
I was having a hard time reviewing the changes because of how much of the original logic was adjusted or moved around.  I've re-created what y'all did, but arranged it a little differently in my fork 8428629000d254ffcb79fc8da1d3ad8bec872264.

Let me know if y'all see any issues in that.  When taken altogether, I think it results in a diff that's a lot cleaner.

The only issue I saw in my testing was that since we're caching SVG document, changing the `CssClass` property doesn't have an immediate effect until you do something to change the cache.  Is that just a trade-off we need to make for performance (@mitchelsellers)?

bdukes:
I've pushed a squashed version to ab2a618f4031d06fa5f02a05294887f1117c5f71, which should be easier to review.

bdukes:
Just wanted to make a reference in the PR to @mathisjay's comment on the commit  https://github.com/dnnsoftware/Dnn.Platform/commit/ab2a618f4031d06fa5f02a05294887f1117c5f71#commitcomment-40515132 /cc @valadas

valadas:
It is a bit hard to follow up on commits in a fork for a PR :)

But yeah, IMO the only issue with the changes on your fork is that if the svg has no classes to begin with but we give the skinObject a CssClass, then it would include a space with your changes as in " myclass" (hence why I put it in a list, added the class and then joined them). I don't believe this List is a huge performance hit, this would just be called on first load or cache clear it would not get hammered much...

bdukes:
@valadas thanks for the feedback, I've adjusted how we're generated the value of the `class` attribute: 
[`svgXmlNode.SetAttributeValue("class", string.IsNullOrEmpty(classes) ? this.CssClass : $"{classes} {this.CssClass}");`](https://github.com/dnnsoftware/Dnn.Platform/commit/4e90e245c1fc7cbaf4bdadf86d43caba59c4dd39#diff-ada7ef2c329065edf56c58dd914afcf3R149)

valadas:
> @valadas thanks for the feedback, I've adjusted how we're generated the value of the `class` attribute:
> [`svgXmlNode.SetAttributeValue("class", string.IsNullOrEmpty(classes) ? this.CssClass : $"{classes} {this.CssClass}");`](https://github.com/dnnsoftware/Dnn.Platform/commit/4e90e245c1fc7cbaf4bdadf86d43caba59c4dd39#diff-ada7ef2c329065edf56c58dd914afcf3R149)

Oh, yeah, love it.

bdukes:
@valadas @mitchelsellers can we get another :+1: and get this merged?

bdukes:
@valadas we need you to clear your "request changes" review before we can merge

valadas:
Sorry guys, done and done

PR #3902, opened 2020-07-08
SiteMapProvider - Fix to exclude non-indexed sites to avoid SEO ranking penalty

### LINKED ISSUE #3901: Submission 3897 Didn't properly Fix SiteMap issue
## Description of bug
The core sitemap provider is adding Tabs to the  SiteMap that are to not be indexed.  By not being allowed index - it is also telling the SearchEngine to ignore the file, while the SiteMap is stating it's a file to index.  Two different messages.

Submission for issue #3896 was incorrect as the tab's Indexed property is referring to a ContentItem, not the Tab's settings.

## Steps to reproduce
1. TabSettings - un check "Allow Index"

## Current behavior
Still adds the tab with the Priority to the SiteMap.  

## Expected behavior
Tab should be excluded from the SiteMap since it is added as a NOFOLLOW

* [ ] 10.00.00 alpha build
* [x ] 09.06.02 release candidate
* [ x] 09.06.01 latest supported release

ajwaka:
fixes #3901

## Summary
Previous submission wasn't correct.  It seems the tab's "Indexed" property is referring to a ContentItem and not the tab's setting.  This submission is referencing the proper TabSetting "AllowIndex"

### DIFF
--- a/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
+++ b/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
@@ -65,7 +65,7 @@ namespace DotNetNuke.Services.Sitemap
                         (Null.IsNull(tab.StartDate) || tab.StartDate < DateTime.Now) &&
                         (Null.IsNull(tab.EndDate) || tab.EndDate > DateTime.Now) && this.IsTabPublic(tab.TabPermissions))
                     {
-                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished)
+                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished && tab.Indexed)
                         {
                             try
                             {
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
```diff
@@ -65,7 +65,7 @@ public override List<SitemapUrl> GetUrls(int portalId, PortalSettings ps, string
                         (Null.IsNull(tab.StartDate) || tab.StartDate < DateTime.Now) &&
                         (Null.IsNull(tab.EndDate) || tab.EndDate > DateTime.Now) && this.IsTabPublic(tab.TabPermissions))
                     {
-                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished)
+                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished && Convert.ToBoolean(tab.TabSettings["AllowIndex"]))
```
bdukes, line 68: It looks like the behavior in other places is that `null` indicates `AllowIndex` is `true`, while using `Convert.ToBoolean` will convert it to `false`.  Can you copy the parsing logic from [`TabIndexer`](https://github.com/dnnsoftware/Dnn.Platform/blob/cc61549005ab5ea1c66fa91670931642a62893e4/DNN%20Platform/Library/Services/Search/TabIndexer.cs#L52-L55) (or, even better, introduce a computed property on `TabInfo` which encapsulates this logic from both places).

Also, you'll need to rebase this on the latest from `develop`, which includes your previous PR.  This PR will need to undo that other PR.  

Thanks!

bdukes:
Superseded by #3903

PR #3904, opened 2020-07-11
Fix MVC NuGet package and improve devsite build script

donker:
This PR fixes the MVC Nuget package definition as Microsoft.AspNet.Mvc v 5.1.3 depends on Microsoft.AspNet.WebPages and Microsoft.AspNet.Razor 3.1.2 or higher.

It also makes a small improvement to the devsite build scripts so we can test build without refreshing the dev site in case we want to test that locally.

### DIFF
--- a/Build/Cake/devsite.cake
+++ b/Build/Cake/devsite.cake
@@ -1,13 +1,19 @@
 // Tasks to help you create and maintain a local DNN development site.
 // Note these tasks depend on the correct settings in your settings file.
 
-Task("ResetDevSite")
+Task("BuildToTempFolder")
     .IsDependentOn("SetVersion")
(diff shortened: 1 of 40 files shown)


bdukes:
Version 3.1.1 is referenced all over the solution.  Can you update the other projects to 3.1.2, also?

bdukes:
I've adjusted this pull request to update all NuGet references for Microsoft.AspNet.* from 3.1.1 to 3.1.2 (thus moving the project into a "supported" configuration).

I've also removed all of the `allowedVersions` attributes from the `packages.config` files which were locking the projects from upgrading NuGet package versions using the tools inside Visual Studio.

PR #3919, opened 2020-07-17
Enable Scheduling - Add TZ to page StartDate and EndDate

### LINKED ISSUE #3918: Page schedule Start and End dates change after saving
## Description of bug
In the Pages panel, if the "Enable Scheduling" switch is on, a start and/or end dates are set, the page is saved and then the page is refreshed, the start/end dates will show different than the originally specified. Dates will be off depending on the user TZ. In my case they look 3 hours later because I'm on UTC-03.
If I set the correct date and save again, then it looks correct afterwards. It's the first time (when dates are still null) that fails.

## Steps to reproduce
1. On DNN 9.6.2 open the Pages panel
2. Pick any page and turn the "Enable Scheduling" switch on
3. Specify a Start Date (at midnight, to make it easy to verify)
4. Save
5. Hit F5 to refresh the page and wait for the Pages panel to open
6. Select the same page and observe the Start Date

## Current behavior
Unless you are at UTC timezone, the Start Date will be off by X number of hours where X is your TZ offset.

## Expected behavior
Start Date should not change after saving.

## Screenshots
image

## Error information
N/A

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.06.02 release candidate
* [X] 09.06.01 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

daguiler:
Fixes #3918

## Summary
This is both a back-end issue and a front-end issue.

Initially, when start/end datetime is null, and the user picks a datetime in the date picker, the front-end sends the datetimes correctly as UTC, and the back-end API saves the datetimes correctly in the database as UTC.

Later, during browser refresh, the `GetPageDetails` API is called, which fails to specify the `DateTimeKind` as UTC when filling the DTO from the database, and therefore the datetimes are serialized back as local datetimes, which produces the offset in the front-end.

At this point, if the user changes the datetime back to the original value, this time the date picker returns a `String` instead of a javascript `Date` object, and the TZ info is lost there, and the API then receives a local datetime. This sort of fixes the problem thereafter, but out of pure luck I would say.

This pull request basically adds the lost TZ both in the back-end (by specifying the `DateTimeKind`) and the front-end (using Date objects with the view model instead of strings).

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/services/pageService.js
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/services/pageService.js
@@ -118,7 +118,10 @@ const PageService = function () {
     const toFrontEndPage = function (pageResult) {
         return {
             ...pageResult.page,
-            schedulingEnabled: pageResult.page.startDate || pageResult.page.endDate,
+            schedulingEnabled: pageResult.page.startDate !== null || pageResult.page.endDate !== null,
+            // the API returns strings, but we need dates
+            startDate: pageResult.page.startDate === null ? null : new Date(pageResult.page.startDate),
+            endDate: pageResult.page.endDate === null ? null : new Date(pageResult.page.endDate),
             validationCode: pageResult.ValidationCode,
         };
     };
--- a/DNN Platform/Library/Common/Utilities/Null.cs
+++ b/DNN Platform/Library/Common/Utilities/Null.cs
@@ -215,6 +215,13 @@ namespace DotNetNuke.Common.Utilities
             return objValue != DBNull.Value ? Convert.ToDateTime(objValue) : NullDate;
         }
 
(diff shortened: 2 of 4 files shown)


Dnn.AdminExperience/ClientSide/Pages.Web/src/services/pageService.js
```diff
@@ -118,7 +118,10 @@ const PageService = function () {
     const toFrontEndPage = function (pageResult) {
         return {
             ...pageResult.page,
-            schedulingEnabled: pageResult.page.startDate || pageResult.page.endDate,
+            schedulingEnabled: pageResult.page.startDate !== null || pageResult.page.endDate !== null,
```
daguiler, line 121: This change was not strictly necessary, but React was complaining that `schedulingEnabled` was being assigned a string instead of a boolean value.

PR #3920, opened 2020-07-17
Dependency injection ordering

### LINKED ISSUE #3335: [Dependency Injection] Reliably override default instances through DI
## Description of problem
The IAppStartup pattern gives 3rd party vendors a means to register types and instances with the DI framework. This poses the following question:

#### What happens when more than one instance is registered under the same contract (interface)?

1. Is the new registration overriding any existing instance under same contract?
1.1 If so, how to reliably override a platform instance without running the risk of a race condition? The race condition exists because the winning instance is the one registered the last, but no registration order or priority is defined or enforced by DNN.

2. Is the new registration appended to other existing instances under the same contract?
2.1 If so, which instance is then returned at runtime? Is it a LIFO scheme? If it is, then in practical terms it's the same as overriding.

## Description of solution
Maybe the solution is to ensure all DNN built-in types are registered first.

Other alternatives include using named instances and then retrieving instances by name, but this is discouraged by most DI frameworks and I'm not sure if this would actually solve the problem.
Maybe a priority scheme where built-in DNN instances are registered with a certain priority and then 3rd party instances can choose a higher priority.

## Additional context
A case for this is overriding the Persona Bar container. This was explained here: https://github.com/dnnsoftware/Dnn.Platform/issues/3328

bdukes:
This updates the main DI code to run the startup code in a stable order, with DotNetNuke.* and DNN.* assemblies first.

Fixes #3335

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
+++ b/DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
@@ -131,7 +131,7 @@ namespace DotNetNuke.Web.Common.Internal
 
             var dependencyProvider = new LazyServiceProvider();
             Globals.DependencyProvider = dependencyProvider;
-            dependencyProvider.SetProvider(Startup.BuildServiceProvider());
+            dependencyProvider.SetProvider(DependencyInjectionInitialize.BuildServiceProvider());
             ServiceRequestScopeModule.SetServiceProvider(Globals.DependencyProvider);
 
             ComponentFactory.Container = new SimpleContainer();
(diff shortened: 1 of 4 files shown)


bdukes:
Thanks for the review @mitchelsellers, it ended up not working 100% (my `OrderBy` at the end negated the `OrderBy` for the types).  I also rearranged it a little bit.  I moved the `AddWebApi` call into its own `Startup` class that implements `IDnnStartup`, and have the rest of the DI initialization in its own class.

valadas:
Love this and the other one for the propmt command, now the question is, is we get a 9.6.3 out, should we hold this to 9.7.0 ? @bdukes @mitchelsellers  ?

mitchelsellers:
@valadas As much as this is an "enhancement" it could also be flagged as a bit of a bug, so it should go in 9.6.3 in my opinion

valadas:
Closing this since it is included in #3932

PR #3925, opened 2020-07-23
Adding a page URL of type Active (200) without specifying a site adds uneditable 301 URLs

### LINKED ISSUE #3924: Adding a page URL of type Active (200) without specifying a site adds uneditable 301 URLs
## Description of bug
In SEO options of a page, selecting a site alias is not enforced (in clustered systems) so that it ends up with bad records when saving with empty Site Alias.

## Steps to reproduce
- Webfarm consisting of 2 servers.
- Site alias added for each server on the webfarm.
- Secondary site of type directory.
- Log in
- Head to Content > Pages > [Select a page that has a parent page] > Advanced > SEO
- Attempt to add a URL without specifying a site alias.

## Current behavior
Multiple URLs are added with a path different than the one specified.

## Expected behavior
The URL is not added because no site alias is specified.

## Screenshots
Reproduction video: https://drive.google.com/drive/folders/1xvwctzL1dTAQwSWO6kuOo1QAlrfkUEZp?usp=sharing

## Affected version

* [x] 10.00.00 alpha build
* [ ] 09.06.02 release candidate
* [ ] 09.06.01 latest supported release

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #3924

## Summary
Null check is added into the Save button to disable it when Site Alias field is empty.

Demo: https://drive.google.com/file/d/12DpeK8Yf9AZJWjlKupUUC0T8KvyhtugY/view?usp=sharing

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/EditUrl.jsx
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/EditUrl.jsx
@@ -140,7 +140,10 @@ class EditUrl extends Component {
                         <Button type="secondary" onClick={onCancel} disabled={saving}>
                             {Localization.get("Cancel")}
                         </Button>
-                        <Button type="primary" onClick={onSave} disabled={!this.state.hasChanges || saving}>
+                        <Button
+                            type="primary"
+                            onClick={onSave}
+                            disabled={!this.state.hasChanges || url == null || url.siteAlias == null || url.siteAlias.Key == null || saving}>
                             {Localization.get("Save")}
                         </Button>
                     </div>


Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/EditUrl.jsx
```diff
@@ -140,7 +140,10 @@ class EditUrl extends Component {
                         <Button type="secondary" onClick={onCancel} disabled={saving}>
                             {Localization.get("Cancel")}
                         </Button>
-                        <Button type="primary" onClick={onSave} disabled={!this.state.hasChanges || saving}>
+                        <Button
+                            type="primary"
+                            onClick={onSave}
+                            disabled={!this.state.hasChanges || url == null || url.siteAlias == null || url.siteAlias.Key == null || saving}>
```
mitchelsellers, line 146: I don't believe that this check is proper, as it SHOULD be possible to set a redirect that is a 301 otherwise that applies to all site aliases.  This is what has support in the database within the TabUrls table as well as the URL Rewriting rules.

If anything this should be adjusted to validate based on type of redirect

berkarslan-xo: Hi @mitchelsellers , do you mean that Site Alias field is optional? When the user doesn't select a Site Alias and click on save button, it adds the URL for an Alias that's not selected at all (and it's not editable too). I don't think this is the desired behavior though, right? You can check this behavior in this [link](https://drive.google.com/drive/folders/1xvwctzL1dTAQwSWO6kuOo1QAlrfkUEZp) (between 0:35-0:52). I just made this change to ensure that a site alias is selected before saving the url.

@v-karbovnichy , could you also check this please?

mitchelsellers: @berkarslan-xo No, I don't believe an alias is required, it isn't at the database level and there are MANY reasons why you wouldn't want it in there for flexibility with redirects.

@david-poindexter or @valadas any thoughts?

bdukes: Agreed.  I think the bug is that the UI creates an alias-specific URL entry when you don't select an alias.  It should allow the alias to remain `null`.

sleupold:
By intent, the site alias is optional, i.e. selecting a page should not switch alias without reason, but use the current alias, the user is on, when selecting the page

Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/EditUrl.jsx
valadas: Yeah I agree this is optional

berkarslan-xo: Got it. I'm closing this PR then.

david-poindexter: Agreed for what it is worth.  ;-)

berkarslan-xo: Thanks everyone for the inputs btw.

PR #3933, opened 2020-07-24
Allow Dependency Injection in more places

bdukes:
## Summary
We can enable dependency injection in more places.  This is a work-in-progress PR which replaces `Activator.CreateInstance(type)` with `ActivatorUtilities.CreateInstance(serviceProvider, type)`.

The main complication with this is that we need to be able to dispose of services when the instance we've created is out of scope.  In some places there's a straightforward way to do this, and in some places there is not.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/MvcHostControl.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/MvcHostControl.cs
@@ -1,7 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
-
 namespace DotNetNuke.Web.Mvc
 {
     using System;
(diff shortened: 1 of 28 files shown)


bdukes:
@mitchelsellers This is the result of walking through the usages of `Activator.CreateInstance` in the framework and updating them to allow using services from the DI container.  Especially once we think about making sure to dispose of dependencies that have been injected, some of these become a little suspect (for some we're returning created instances without any mechanism to ensure they'd be disposed, in others we're caching them indefinitely).

If we move towards making more APIs that are available only through DI (e.g. `INavigationManager`), we'll have more and more scenarios where allowing DI usage will be important.  Any thoughts on the best way to enable that?  Do we need to adjust the APIs that are returning injected instances so that whatever we return is itself disposable?

bdukes:
I had no idea how many different components tried to create the business controller class.  This gets most of them, but there are TODOs left for:
* `DNN Platform/Library/Services/Search/Controllers/ModuleResultController.cs`
* `DNN Platform/Library/UI/Modules/Html5/Html5ModuleTokenReplace.cs`
* `DNN Platform/Modules/DDRMenu/Localisation/Generic.cs`
* `DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs`
* `Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs`

Most of these need to be adjusted to think about lifetime (i.e. to ensure that dependencies of the business controller class get disposed).

valadas:
Closing this PR in favor of multiple smaller ones as discussed in approvers meeting.

Issue #3934, opened 2020-07-27
Finish Dependency Injection Support for PortalController


ahoefling:
## Description of problem
Since support has been stabilized for Dependency Injection the goal is to add better support across the different static APIs. The `PortalController` current supports Dependency Injection via `IPortalController` however many APIs are missing because they are static APIs.

For example: If you need to decrypt portal settings you would use the following API
```c#
PortalController.GetEncryptedString(string settingName, int portalID, string passPhrase);
```

There is no instance level API method that supports this, which makes it difficult to use with Dependency Injection.

## Description of solution
Create a copy of every static API in the `PortalController` for the `IPortalController`. Then mark the existing static API obsolete while using the new instance version. 

If we were to use the example above with `GetEncryptedString`. Move the declaration to the IPortalController.
```c#
public interface IPortalController
{
    string GetEncryptedString(string settingName, int portalID, string passPhrase);    

    // omitted code
``` 

Then update the `PortalController` code to have both
```c#
public class PortalController
{
    [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.GetEncryptedString' instnace method. Scheduled removal in v11.0.0.")]
    public static string GetEncryptedString(string settingName, int portalID, string passPhrase) =>
        GetEncryptedString(settingName, portalID, passPhrase);

    string IPortalController.GetEncryptedString(string settingName, int portalID, string passPhrase)
    {
        // omitted code
    }
}
```

## Description of alternatives considered
N/A

## Screenshots
N/A

## Additional context
This change will effect the entire platform

## Affected browser
N/A

ahoefling:
This was never finished, I think it should be reopened and pinned. cc: @mitchelsellers

valadas:
Reopened and pinned.

PR #3935, opened 2020-07-27
Migrate Static APIs to from PortalController to IPortalController

ahoefling:
Closes: #3934

## Summary
Completes Dependency Injection support for the `IPortalController` by making the remaining static APIs available on the interface

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalController.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalController.cs
@@ -30,7 +30,6 @@ namespace DotNetNuke.Entities.Portals
     using DotNetNuke.Entities.Urls;
     using DotNetNuke.Entities.Users;
     using DotNetNuke.Entities.Users.Social;
-    using DotNetNuke.Framework;
     using DotNetNuke.Instrumentation;
     using DotNetNuke.Security.Membership;
     using DotNetNuke.Security.Permissions;
@@ -45,7 +44,7 @@ namespace DotNetNuke.Entities.Portals
     using DotNetNuke.Services.Search.Entities;
     using DotNetNuke.Web.Client;
     using ICSharpCode.SharpZipLib.Zip;
-
+    using Microsoft.Extensions.DependencyInjection;
     using FileInfo = DotNetNuke.Services.FileSystem.FileInfo;
     using IAbPortalSettings = DotNetNuke.Abstractions.Portals.IPortalSettings;
 
@@ -56,7 +55,7 @@ namespace DotNetNuke.Entities.Portals
     /// DotNetNuke supports the concept of virtualised sites in a single install. This means that multiple sites,
     /// each potentially with multiple unique URL's, can exist in one instance of DotNetNuke i.e. one set of files and one database.
     /// </remarks>
-    public partial class PortalController : ServiceLocator<IPortalController, PortalController>, IPortalController
+    public partial class PortalController : IPortalController
     {
         public const string HtmlText_TimeToAutoSave = "HtmlText_TimeToAutoSave";
         public const string HtmlText_AutoSaveEnabled = "HtmlText_AutoSaveEnabled";
@@ -64,6 +63,24 @@ namespace DotNetNuke.Entities.Portals
         protected const string HttpContextKeyPortalSettingsDictionary = "PortalSettingsDictionary{0}{1}";
 
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(PortalController));
+
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController' with Dependency Injection. Scheduled removal in v11.0.0.")]
+        protected static Func<IPortalController> Factory => () => Instance;
+
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController' with Dependency Injection. Scheduled removal in v11.0.0.")]
+        protected static Func<IPortalController> GetFactory() => Factory;
+
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController' with Dependency Injection. Scheduled removal in v11.0.0.")]
+        public static void SetTestableInstance(IPortalController instance) =>
+            throw new NotSupportedException("'PortalController.SetTestableInstance' is no longer supported, use Dependency Injection to resolve 'IPortalController'");
+
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController' with Dependency Injection. Scheduled removal in v11.0.0.")]
+        public static void ClearInstance() =>
+            throw new NotSupportedException("'PortalController.SetTestableInstance' is no longer supported, use Dependency Injection to resolve 'IPortalController'");
+
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController' with Dependency Injection. Scheduled removal in v11.0.0.")]
+        public static IPortalController Instance => Globals.DependencyProvider.GetService<IPortalController>();
+
 
         /// <summary>
         /// Adds the portal dictionary.
@@ -166,7 +183,7 @@ namespace DotNetNuke.Entities.Portals
             string message = string.Empty;
 
             // check if this is the last portal
-            var portals = Instance.GetPortals();
+            var portals = GetPortals();
             if (portals.Count > 1)
             {
                 if (portal != null)
@@ -319,7 +336,8 @@ namespace DotNetNuke.Entities.Portals
         }
 
         [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.GetPortalsByUser' instance method. Scheduled removal in v11.0.0.")]
-        public static ArrayList GetPortalsByUser(int userId) => Instance.GetPortalsByUser(userId);
+        public static ArrayList GetPortalsByUser(int userId) =>
+            Instance.GetPortalsByUser(userId);
 
         /// <inheritdoc />
         ArrayList IPortalController.GetPortalsByUser(int userId)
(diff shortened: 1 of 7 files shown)


DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -846,7 +1157,10 @@ public static void IncrementCrmVersion(int portalID)
             }
         }
 
-        public static void IncrementOverridingPortalsCrmVersion()
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.IncrementOverridingPortalsCrmVersion' instnace method. Scheduled removal in v11.0.0.")]
```
bdukes, line 1160:
```suggestion
        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.IncrementOverridingPortalsCrmVersion' instance method. Scheduled removal in v11.0.0.")]
```

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -835,7 +1143,10 @@ public static void UpdatePortalDefaultLanguage(int portalID, string CultureCode)
             DataProvider.Instance().EnsureLocalizationExists(portalID, CultureCode);
         }
 
-        public static void IncrementCrmVersion(int portalID)
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.IncrementCrmVersion' instnace method. Scheduled removal in v11.0.0.")]
```
bdukes, line 1146:
```suggestion
        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.IncrementCrmVersion' instance method. Scheduled removal in v11.0.0.")]
```

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -826,7 +1121,20 @@ public static string GetPortalDefaultLanguage(int portalID)
         /// <param name = "CultureCode"></param>
         /// <remarks>
         /// </remarks>
-        public static void UpdatePortalDefaultLanguage(int portalID, string CultureCode)
+        [Obsolete("Deprecated in DNN 9.7.0. Use the 'IPortalController.UpdatePortalDefaultLanguage' instnace method. Scheduled removal in v11.0.0.")]
+        public static void UpdatePortalDefaultLanguage(int portalID, string CultureCode) =>
+            UpdatePortalDefaultLanguage(portalID, CultureCode);
```
bdukes, line 1126: This is just a recursive loop, the static method will keep calling itself.  You'd need to create an `IPortalController` instance to delegate to, if that's your intention.

ahoefling: Thanks for catching this, a big mistake if we don't include the keyword `Instance`

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -747,7 +1024,18 @@ public static string CheckDesktopModulesInstalled(XPathNavigator nav)
...
+        public static string GetActivePortalLanguage(int portalID) => Instance.GetActivePortalLanguage(portalID);
+
+        /// <summary>
+        ///   function provides the language for portalinfo requests
+        ///   in case where language has not been installed yet, will return the core install default of en-us.
+        /// </summary>
+        /// <param name = "portalID"></param>
+        /// <returns></returns>
+        /// <remarks>
+        /// </remarks>
```
bdukes, line 1037: For the `IPortalController` implementations, let's just inherit docs from the interface:

```suggestion
        /// <inheritdoc />
```

bdukes:
Thanks @ahoefling!

DNN Platform/Library/Entities/Portals/PortalController.cs
ahoefling: This makes sense, do we have this written down somewhere in the contributing guide?

ahoefling: I'm going to mark this as resolved since the change has been implemented

valadas:
> Thanks! AFAIK the use of <inheritdoc /> is not discussed in the contributing docs

Yeah I don't remember seeing that or anything about public methods documentation at all in fact. But the stylecop config and editor config should just guide people here about it, the first suggesting in the flyout menu is "Inherit documentation", which does that.

valadas:
Hmm, tests are failing though but from a quick look, it looks like they are mostly in the Messaging aera. Gonna try to re-run the build to check if it is a fluke or what...

/azp run

valadas:
@ahoefling yeah, we have multiple tests I just took a quick look but did not pinpoint the issue. Do you have a clue and can resolve this quickly or do we hold a bit on this PR ?

ahoefling:
@valadas Thanks for pointing this out, I'll take a look at the failing unit tests

ahoefling:
The unit testing to this PR is much larger than I originally thought and it'll take some refactoring to get everything correct. I have pushed my latest changes and would like you all to check my direction as it is quite a shift from the original code. 

Looking through the history of `ServiceLocator<TContract, TImpl>` it appears it was built for the sole -purpose of unit testing the inner workings of objects such as `PortalController`, `MessageController`, etc. The old version of the `IPortalController` even has a XML comment that says it should only be used for testing and production code shouldn't depend on it. I removed that XML comment because the direction with dependency injection is to depend on contracts not implementations. With this change I don't think the `ServiceLocator` serves any value for the `PortalController` Implementation. But pulling out the base implementation will certainly cause problems. In my latest change I mimicked the `ServiceLocator` behavior in duplicate API names that exist in the `PortalController` and not a parent class. **This will break someone DNN site and isn't safe to merge.**. 

I'm curious if any of the reviewers have a recommendation on how we can properly deprecate the inherited parent class so the `PortalController` can safely remove it's dependency on the `ServiceLocator`.

### Potential Solutions
As I write this, I started to think about creating an independent implementation of `IPortalController` that is resolved when requested via the Dependency Injection Container. The current `PortalController` class will be re-written to encapsulate that dependency. This allows us to deprecate the entire class and solve the problem I highlighted above. 

I'll give this solution a shot later in the week or any discussions we come up with.

valadas:
It is a bit beyond my pay grade but your proposed solution sounds good to me... I kinda had a similar contradiction in the code for search, there was deprecated member in a base class and well you kinda had to use the class and forced to implement the deprecated members which would have caused a 0-day breaking whenever that member would be removed.

Not sure if it helps at all since the situation is different but here is how I solved that https://github.com/dnnsoftware/Dnn.Platform/pull/3777/files

bdukes:
Yeah, I think we should introduce a new `IPortalController` interface in `DotNetNuke.Abstractions`.

But your unit testing issue brings up a bigger mess we'll have to wade through, which is that when you're moving from service location to dependency injection, you create rippling changes by introducing new constructor parameters.  In order to resolve those constructor parameters, now everything that uses the changed dependency has to adjust how _they_  get created (either opting in to dependency injection or adjusting how they use service location).

I terms of non-breaking changes, I don't see how we can remove our dependency on the `ServiceLocator` base class.  That said, I need to dig into these unit test failures a little more to understand the issue we're running into.

ahoefling:
I like your idea of creating a new `IPortalController` in the `DotNetNuke.Abstractions` project this follows our guidelines of moving forward with dependency injection. Expanding on my "Potential Solution" above I think we can solve this problem without introducing a breaking change to the platform. The unit tests need quite a bit of work and it may need to be completely rewritten to properly execute tests.

Suppose we define the following interface in the `DotNetNuke.Abstractions` project
```c#
public interface IPortalController
{
    // omitted code
}
```

We currently have the following classes in the DotNetNuke.Library
```c#
public interface IPortalController
{
    // omitted code
}

public class PortalController : ServiceLocator<IPortalController, PortalController>, IPortalController
{
    // omitted code
}
```

We can create a new implementation of `DotNetNuke.Abstractions.IPortalController` called `PortalController2`
```c#
public class PortalController2 : DotNetNuke.Abstractions.IPortalController
{
    // omitted code
}
```

In the `Startup.cs` class for `DotNetNuke.Library` we will register the `DotNetNuke.Abstractions.IPortalController` side-by-side with the current `IPortalController`. 

```c#
{
    {
        // omtted code

        // Prevents any breaking API changes with the existing interface that will be deprecated

        // Registers the new interface
        services.AddTransient<DotNetNuke.Abstractions.IPortalController, PortalController2>();
}
```

Now that we have everything created and wired up we can make the following updates to the `PortalController` by injecting the `DotNetNuke.Abstractions.IPortalController`. This will leverage encapsulation to de-couple the existing API structure from the new API structure but still deprecate everything that needs to be deprecated.

```c#
[Obsolete("Deprecated in DNN 9.7.0. Use the 'DotNetNuke.Abstractions.IPortalController' interface resolved via Dependency Injection. Scheduled removal in v11.0.0.")]
public class PortalController : ServiceLocator<IPortalController, PortalController>, IPortalController
{
    private readonly DotNetNuke.Abstractions.IPortalController _portalController;

    // To ensure there are no breaking changes across the platform or unknown extensions
    // we will add a default constructor that will use the Dependency Injection container
    // to resolve the `DotNetNuke.Abstractions.IPortalController`
    public PortalController()
        : this(Globals.DependencyProvider.GetService<DotNetNuke.Abstractions.IPortalController>())
    {
    }
  
    public PortalController(DotNetNuke.Abstractions.IPortalController portalController)
    {
        _portalController = portalController;
    }

    [Obsolete("Deprecated in DNN 9.7.0. Use the 'DotNetNuke.Abstractions.IPortalController.GetPortalSetting' instance method. Scheduled removal in v11.0.0.")]
    public static string GetPortalSetting(string settingName, int portalID, string defaultValue) =>
        _portalController.GetPortalSetting(settingName, portalID, defaultValue);

    // omitted code
}
```

ahoefling:
@bdukes @valadas 
The `PortalController` is connected to just about everything when you start resolving all the dependencies. I am still working on this change and things are coming along as far as code organization. When I update this pull request we will need a very thorough review.

ahoefling:
I am still thinking about this and have a bunch of code for this. It may be best to close this PR and open smaller ones. Please take a look at how I migrated the `IHostController` in #3990 as that is the direction I think we should take for migrating APIs. My biggest concern for IPortalController is I think we made a mistake in the initial attempt. I am thinking of changing namespaces so all entity controllers exist in `DotNetNuke.Abstractions.Entities` following their exact same naming convention from `DotNetNuke.Library`.

Either way, to properly abstract the `PortalController` is a massive undertaking

bdukes:
Sounds good to me, thanks @ahoefling!

PR #3936, opened 2020-07-27
Resolves a logging issue

### LINKED ISSUE #3906: Clean Install of DNN v9.06.02 throwing a lot of errors from PersonaBar MenuController
**Note:** this was already confirmed by @valadas (Daniel Valadas) who ask me to post this issue/details.

## Description of bug and Steps to reproduce
After a new/clean install of DNN v9.06.02 and logged in as a super user, go immediately to the Logs (Settings, Servers, Server Settings Tab, then the LOGS sub-tab). Expand (cool new feature, eh?) today's 2020.07.NN.log.resources file. Skip the CKEditor issues, that will need a separate bug report.

## Current behavior
It appears that ever time DNN needs to iterate through the PersonaBar's Manage and Settings "pages" it throws 21 pairs menu related errors. 12 failed calls to AdminMenuController and 9 fails for HostMenuController all with Value cannot be null in parameter name Type. 

## Expected behavior
DNN should not be throwing errors like this "out of the box" and rapidly growing the size of the *.logs.resources files in /Portals/_default/Logs/.

## Error information
```
2020-07-12 18:15:02.242-05:00 [DS2035][D:11][T:24][ERROR] DotNetNuke.Framework.Reflection - Dnn.PersonaBar.Sites.MenuControllers.AdminMenuController, Dnn.PersonaBar.Extensions
System.TypeLoadException: Could not load type 'Dnn.PersonaBar.Sites.MenuControllers.AdminMenuController' from assembly 'Dnn.PersonaBar.Extensions'.
   at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, Boolean reflectionOnly, StackCrawlMarkHandle stackMark, IntPtr pPrivHostBinder, Boolean loadTypeFromPartialName, ObjectHandleOnStack type)
   at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, Boolean reflectionOnly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean loadTypeFromPartialName)
   at System.Type.GetType(String typeName, Boolean throwOnError, Boolean ignoreCase)
   at System.Web.Compilation.BuildManager.GetType(String typeName, Boolean throwOnError, Boolean ignoreCase)
   at DotNetNuke.Framework.Reflection.CreateType(String TypeName, String CacheKey, Boolean UseCache, Boolean IgnoreErrors)
2020-07-12 18:15:02.273-05:00 [DS2035][D:11][T:24][ERROR] Dnn.PersonaBar.Library.Controllers.PersonaBarController - System.ArgumentNullException: Value cannot be null.
Parameter name: type
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at Dnn.PersonaBar.Library.Controllers.PersonaBarController.GetMenuItemController(MenuItem menuItem)
```
 
## Additional context
I have a DNN install that started life as 9.04.04, was upgraded recently to 9.06.01, then upgraded to 9.06.02 and the errors (above) are not being thrown.

## Affected version
[ x ] 09.06.02 final 
 
## Affected browser - not browser related in any way

valadas:
Closes #3906

Apparently some files where deleted as part of #3843 which caused the extra logging.

This PR brings those files back and also fixes stylecop warnings. Some class names had to be changed both for clarity and to respect the stylecop rule that the file name should match the class name. Because of that change, the dnn manifest was also changed so it all works with the new names.


bdukes:
Looks good to me.  Thanks for tracking this down @valadas!

Was any functionality broken by these going missing?  Do we even need these classes?  I feel like it's a warning to look closer if we can lose a handful of classes and the only symptom is stuff filling the logs…

valadas:
@bdukes Well, each of these controllers control a menu item, from the looks of it it controls the visibility per role and on some it injects settings that I guess trickle down to the frontend initialization code. Those controllers are defined in the manifest and the logging issue was caused by them being defined but missing and reflection was choking trying to find them. It looks like in the case of not being defined in the manifest, it defaults to show to hosts only, so I guess for those who don't do anything else, it would be safe to remove, I decided to keep them because this is a bit cryptic to many people (including me) and I though it safer to leave it set as it was to prevent a possible change elsewhere that would change that default behaviour and expose menu items to users that should not see them...

bdukes:
Yeah, in terms of this PR as a bug fix, just reverting this change is definitely the right approach.

I'm just wondering out loud if :one: this is not cryptic for anyone, and :two: if it's worth a closer look to see if this makes sense or not.

PR #3939, opened 2020-07-28
Address StyleCop warnings for TabInfo and TabIndexer

bdukes:


bdukes:
I did the best I could on the documentation, but some of those properties I don't use and don't know the ins and outs of them.  Maybe @donker or @sleupold could provide some more details on the language GUIDs, perhaps.

PR #3948, opened 2020-07-31
Fixed an issue where it was impossible to select a page for non default sites in SEO

### LINKED ISSUE #3947: "Test a Page URL" doesn't work for non-default portal
## Description of bug

## Steps to reproduce
1. Create a new site
2. Access the new site
3. Access the "Test a Page URL" feature through "Persona Bar >> Setting >> S.E.O >> Test URL
4. Try to "Test a Page URL" by trying to choose a page to test

## Current behavior
Unable to select any pages available for the non-default portal causing this feature to not work at all for non-default portal.

## Expected behavior
Able to select any pages available to test its URL.

## Video steps
https://drive.google.com/file/d/1XruTq5clwrB1lb2h8MSRtPW3AsL5iaqq/view?usp=sharing

## Additional context
Issue already fixed - creating PR now

## Affected version

* [X] 10.00.00 alpha build
* [X] 09.06.02 release candidate
* [X] 09.06.01 latest supported release

## Affected browser

* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

sergeydryomin:
Fixes #3947

## Summary
Fixed PagePicker for non default sites to not deny selection in case of call from Seo.Web

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Dnn.React.Common/src/PagePicker/index.jsx
+++ b/Dnn.AdminExperience/ClientSide/Dnn.React.Common/src/PagePicker/index.jsx
@@ -278,7 +278,7 @@ class PagePicker extends Component {
 
     getSelected(page, isCurrentOrDescendant) {
         let className = "page-value";
-        if ((page.Selectable || this.props.isSeoPage) && !isCurrentOrDescendant)      
+        if ((page.Selectable || this.props.AllowSelectUnselectablePages) && !isCurrentOrDescendant)      
             className += (page.CheckedState !== 1 ? " selected" : "");
         else
             className += " non-selectable";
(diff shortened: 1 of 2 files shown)


bdukes:
I would prefer a more general approach, like naming the prop `AllowSelectUnselectablePages` instead of `isSeo`.  But, in general, I don't understand why those pages aren't selectable in the first place.  Is it really valid that the page selector is full of unselectable pages, or is there something wrong with the data?  Does the call to the server need some different context?

mitchelsellers:
@donker will be reviewing this one

donker:
This is unfortunately the wrong solution to the problem. The problem hides here:

https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/ClientSide/Seo.Web/src/components/testUrl/index.jsx#L89

The Test Url tab in the PB module has a hard coded reference to admin role for portal 0. This is off course completely wrong and should be repaired.

PR #3952, opened 2020-08-03
Reset invalidation link on email change

mikebigun:
## Summary

All details of this change are shared via private email.


mikebigun:
@zyhfish, I agree that setting is important but it doesn't fully protect the system. The user could change his email due to many reasons, including hacker attacks.. Email is a way we communicate, so after it is changed we must assure user the sensitive information we sent to his previous email address is not valid anymore.

PR #3963, opened 2020-08-06
Optimize WildcardToRegex for file search

david-poindexter:
## Summary
This PR optimizes `WildcardToRegex()` to remove unnecessary start/end wildcard characters per comments on #3962.  Thanks to @bdukes for the RegEx tip!

Also, since `Replace(oldValue, newValue)` just returns the current instance unchanged if the `oldValue` is not found, I thought we could further optimize by removing the check for `*` or `?` in the search term.


david-poindexter:
> LGTM - and my bad on merging ;)

Ah, no worries at all - more little green dots!  😆 👍

PR #3969, opened 2020-08-08
Implement Email Provider

### LINKED ISSUE #3967: Implement Mail Provider
## Description of problem
Currently, DNN does not implement a provider model for sending out mail. Sending email via SMTP is a slow process and does not allow API integrations with other services such as SendGrid or Amazon SES. 

## Description of solution
Implement a provider model for Mail so it can be easily switched out to a different provider to meet custom business needs. The existing code is refactored into CoreMailProvider extending the MailProvider.

meetmandeep:
## Summary

Closes #3967


DNN Platform/Library/Services/Mail/CoreMailProvider.cs
```diff
@@ -0,0 +1,216 @@
...
+            // translate semi-colon delimiters to commas as ASP.NET 2.0 does not support semi-colons
+            if (!string.IsNullOrEmpty(mailInfo.To))
+            {
+                mailInfo.To = mailInfo.To.Replace(";", ",");
+            }
+
+            if (!string.IsNullOrEmpty(mailInfo.CC))
+            {
+                mailInfo.To = mailInfo.To.Replace(";", ",");
+            }
```
valadas, line 52: should this not be mailInfo.CC instead of mailInfo.To ?

DNN Platform/Library/Services/Mail/CoreMailProvider.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            if (!string.IsNullOrEmpty(mailInfo.BCC))
+            {
+                mailInfo.To = mailInfo.To.Replace(";", ",");
+            }
```
valadas, line 57: Same here but for BCC

DNN Platform/Library/Services/Mail/MailInfo.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        public MailPriority Priority { get; set; }
+
+        /// <summary>
+        /// Gets or sets Body Encoding.
+        /// </summary>
+        public Encoding BodyEncoding { get; set; }
+
+        /// <summary>
+        /// Gets or sets Body Encoding.
```
bdukes, line 68:
```suggestion
        /// Gets or sets Body Format.
```

DNN Platform/Library/Services/Mail/SMTPInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+using System.Text;
+
+namespace DotNetNuke.Services.Mail
+{
+    public class SMTPInfo
```
bdukes, line 8: I think we should use PascalCase for the class name here.

```suggestion
    public class SmtpInfo
```

bdukes:
Yes, see 06264b671d3458a32a3620784736e6c47c89ded4 for the config needed to add a new provider on upgrades.

DNN Platform/Library/Services/Mail/CoreMailProvider.cs
meetmandeep: Good catch. Copy/Paste nightmare I'll resolve this in next commit.

DNN Platform/Library/Services/Mail/SMTPInfo.cs
meetmandeep: Resolved.

DNN Platform/Website/Install/Config/09.08.00.config
```diff
@@ -0,0 +1,15 @@
...
+    <node path="/configuration/configSections/sectionGroup[@name='dotnetnuke']" action="update" key="name" collision="ignore">
+      <section name="mail" requirePermission="false" type="DotNetNuke.Framework.Providers.ProviderConfigurationHandler, DotNetNuke" />
+    </node>
+    <node path="/configuration/dotnetnuke" action="update" targetpath="/configuration/dotnetnuke/mail" collision="ignore">
+      <tokens defaultProvider="CoreMailProvider">
+        <providers>
+          <clear />
+          <add name="CoreMailProvider" type="DotNetNuke.Services.Mail.CoreMailProvider, DotNetNuke" />
+        </providers>
+      </tokens>
```
valadas, line 12:
```suggestion
      <mail defaultProvider="CoreMailProvider">
        <providers>
          <clear />
          <add name="CoreMailProvider" type="DotNetNuke.Services.Mail.CoreMailProvider, DotNetNuke" />
        </providers>
      </mail>
```

sleupold:
@mitchelsellers wouldn't a provider just need a Settings UI and store the values in HostSettings table (using proper prefix, e.g.)?

valadas:
Yeah, that would be nice. now given that the email settings show up in the persona bar, if we do it like in the html editor provider than it would mean loading up that settings UI in webforms, which I think would be a step backwards.

I am thinking of 2 possible solutions:
1. We do load it up in a modal somehow (iframe or such) and we support the control just like any other module to support webforms, mvc or html5 in that modal.
2. We make the html5 module pattern work outside of a Dnn module (not sure how).

mitchelsellers:
I’m adding this to the discussion for the approvers meeting on Tuesday. 

For consistency we already have multiple providers that load a webforms view into the PersonaBar. Not saying that’s right, but it would be consistent with other documented processes. But we need to review as a whole

bdukes:
FWIW, my vote is to accept this PR for 9.8.0, with or without a UI.  If someone wants to create a SendGrid provider, it can come with its own UI or document how the SMTP settings map to SendGrid settings.  Having a customizable in-place UI is nice to have, but not required, IMO.

meetmandeep:
public abstract string SendMail(MailInfo mailInfo, SmtpInfo smtpInfo = null);

The signature on the SendMail basically requires just the MailInfo object and leaves the implementation to the method. We always add a generic in-place UI later but IMO it's not needed at this time. If anything, it restricts functionality. At best, it would allow a "Settings Config" at portal level. Right now the developer can route Emails based on priority via different Servers or Services.

If I were to create a MailProvider with load balancing option then the generic in-place UI is not sufficient. 

I agree with @sleupold that developer should handle their own UI and store in HostSettings or elsewhere as needed.

meetmandeep:
Please DO NOT merge this PR yet; we may want to create a generic Mapper for MailAttachment rather than using the System.Net.Mail class. I was mistaken that MailKit supports System.Net.Mail.Attachment. 

I'm thinking a generic MailAttachment object with following overloads:
MailAttachment(string Filename)
MailAttachment(string Filename, Byte[] Attachment)
MailAttachment(string Filename, Byte[] Attachment, string ContentType ) //Mime Type

Thoughts?

mitchelsellers:
I'm in 100% agreement that the developer should create their own UI, and not necessarily opposing this solution as-is for now, but it is an incomplete solution per how we handle other providers where additional settings and configuration are needed.  

Authentication Providers expose a UI for the addition/management of all settings.
HTML Providers do the same thing

Not all email is sent with SMTP settings, in fact, some of the best ways to manage email aren't.  A vendor can easily do something with this that is outside of everything else.  But then we have confusion from a user perspective when they see "SMTP Settings" but they don't do anything or have any impact in their environment.

Again - not saying no, just saying that in addition to this we have some additional items to think about.

meetmandeep:
> But then we have confusion from a user perspective when they see "SMTP Settings" but they don't do anything or have any impact in their environment.

I agree. Perhaps we can remove SMTP Settings when CoreMailProvider is not in use until we find a solution to in-place UI that is not webforms?

mitchelsellers:
Regarding the object for Attachments, we have done similar in our .NET Core model that we use, so I think those could be good choices.

And that COULD be a good idea for the provider.  But I'm thinking if we don't provide a UI option since the new provider COULD use it, that we almost just want to show a message "You are using a provider other than the default, the settings below may not apply."

valadas:
Well, a solution like for the Authentication providers would not be bad for now:

image

Maybe we could move the smtp settings in there too. Now this opens another discussion. The "core" has a fallback from site settings to host settings, I guess the providers would also need something like that...

mitchelsellers:
Discussing UI elements today in the approvers meeting, possibly adding a "Select Provider" option on the SMTP settings areas and a "SMTP Settings Required" or similar flag to the provider to show/hide SMTP configuration.

@donker is going to help look into this.

meetmandeep:
Changes from @donker were merged in this PR. 

@bdukes @valadas You may accept and merge this PR. The MailAttachment wrapper code is now part of the MailKit PR that will submit next week.

mitchelsellers (approved):
Changes were implemented as requested

bdukes:
Ready to merge/rebase once build is done

PR #3972, opened 2020-08-11
handled empty string comparison to null case for non-existent display…

### LINKED ISSUE #3971: User registration fails when Displayname field is not present in the custom register form
## Description of bug
User registration fails when Displayname field is not present in the custom register form

## Steps to reproduce
Precondition: Evoq Engage

1. Login as host.
2. Goto Settings> Security Page > Member Accounts > Registration Fields.
3. Set User Registration = Public
4. Set Registration Form Type = "Custom"
5. Add Registration Fields
- username
- Email
- password
- Passwordconfirm
- FirstName
- LastName
6. Click on Save.
7. Logout as host
8. Goto 'Register' Page
9. Enter all items in the fields
10. Click on Submit/Register

## Current behavior
Getting error "The Display Name is invalid."

## Expected behavior
No errors occur.
User is registered and logged in.

## Screenshots
image

## Error information

## Additional context

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.06.02 release candidate

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge Chromium

tauqeer-haider:
… name field in register form

Fixes #3971 

## Summary

Custom register form configured without Display Name field on it. Non-existent display name field has null value and after cleaning the input of display name, the cleaned value became empty string. A subsequent comparison between null and empty string marks this Display Name as invalid input.

### DIFF
--- a/DNN Platform/Website/DesktopModules/Admin/Security/Register.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/Admin/Security/Register.ascx.cs
@@ -587,7 +587,9 @@ namespace DotNetNuke.Modules.Admin.Users
                 if (string.IsNullOrEmpty(this.PortalSettings.Registration.DisplayNameFormat))
                 {
                     var cleanDisplayName = PortalSecurity.Instance.InputFilter(this.User.DisplayName, filterFlags);
-                    if (!cleanDisplayName.Equals(this.User.DisplayName))
+                    var participantsAreEmptyOrNull = string.IsNullOrWhiteSpace(cleanDisplayName)
+                        && string.IsNullOrWhiteSpace(this.User.DisplayName);
+                    if (!participantsAreEmptyOrNull && !cleanDisplayName.Equals(this.User.DisplayName))
                     {
                         this.CreateStatus = UserCreateStatus.InvalidDisplayName;
                     }


bdukes:
This scenario would only happen when there is a configured Display Name Format, but the result was empty, meaning that the fields specified in the Display Name Format weren't provided (or the `DisplayName` field wasn't properly set)

tauqeer-haider:
> This scenario would only happen when there is a configured Display Name Format, but the result was empty, meaning that the fields specified in the Display Name Format weren't provided (or the `DisplayName` field wasn't properly set)

Failing scenario: The "Display Name Format" is not set and DisplayName field is not included in custom form (i.e., is null).

bdukes:
Ah, sorry, I did read that conditional backwards.  The scenario is when Display Name Format is _not_ configured, and it's not set for the user (i.e. not included on the form).

My understanding is that the current code is correct, it is a requirement for a user to have a display name.

tauqeer-haider:
> I do not believe that we want to allow users to be created without a display name of any fashion. If anything, when using a custom registration form we might want to ensure that we have a better message to users that shows them that it is an invalid configuration.
> 
> DisplayName is important and used in multiple other locations, allowing a display name to be null/empty-string would cause many downstream issues.

The display name is worked out from FirstName, LastName combination or from email in later part of the code so it will not be null or empty as shown here in the code, the "else" part is for Public Form Type: 
image

mitchelsellers:
Sort of, but additionally the change that you made has some other implications from a security perspective it appears. Feel free to contact me directly if desired.  I'll chat with @bdukes today about this as well

mitchelsellers:
After review today, we are closing this issue, the more proper solution will be to improve the messaging for a site that is configured without a DisplayName field displayed and lacking a Display Name format.

PR #3976, opened 2020-08-12
Fix for issue Newly created sub-folder not shown if the parent folder…

### LINKED ISSUE #3975: Newly created sub-folder not shown if the parent folder name starts with "0"
## Description of bug
Newly created sub-folder not shown if the parent folder name starts with "0"

## Steps to reproduce
1. Login as host user
2. Navigate to CONTENT > Assets
3. Click on "Global Assets"
4. Click on "Add Folder" and provide name "01-test" and click on "Save"
5. Click on "01-test"
6. Click on "Add Folder" and provide name "subfolder" and click on "Save"

## Current behavior
"subfolder" is not displayed under parent folder "01-test"

## Expected behavior
"subfolder" should be displayed under parent folder "01-test"

## Screenshots

## Error information
Not Applicable

## Additional context
Not Applicable

## Affected version
* [x] 10.00.00 alpha build
* [x] 09.07.00 release candidate
* [x] 09.06.02 latest supported release

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge Chromium

kmuralidaran:
… name starts with "0"

Fixes #3975 

## Summary
Trimming the prefix "0" from folder name creates a folder with different name compared to the user provided name and the subfolder is also not created in the user provided folder name. 

Removing the code which trims the prefix "0" from folder name addresses this issue

### DIFF
--- a/DNN Platform/Library/Common/Utilities/PathUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/PathUtils.cs
@@ -222,7 +222,7 @@ namespace DotNetNuke.Common.Utilities
         }
 
         /// <summary>
-        /// Strips the original path by removing starting 0 or 0\\.
+        /// Strips the original path by removing starting 0\\.
         /// </summary>
         /// <param name="originalPath">The original path.</param>
         /// <returns>The stripped path.</returns>
@@ -235,7 +235,7 @@ namespace DotNetNuke.Common.Utilities
                 return FolderPathRx.Replace(originalPath, string.Empty);
             }
 
-            return originalPath.StartsWith("0") ? originalPath.Substring(1) : originalPath;
+            return originalPath;
         }
 
         internal string GetUserFolderPathElementInternal(int userId, UserFolderElement mode)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3976) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3976) before we can accept your contribution.<br/><br/>:x: kmuralidaran [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=3976)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=3976) it.</sub>

mitchelsellers:
I second this concern

mitchelsellers:
This item needs more review before merging, to validate usage.  It appears to be only called by the file manager, but possibly could have a check done at a later time to validate that it isn't an invalid path.

BUT, the true history of WHY this was done predates GitHub so we need to further determine why.

Also - Is this used in the new file manager?  if not, lets NOT change this.

donker:
The code checks if the path contains a backslash. If it does then it checks whether it starts with "0\\" (note TWO backslahses). In which case it gets cut off. Finally it checks if it starts with 0 in which case it gets cut off. It is puzzling why this is here (maybe @cnurse knows). Unfortunately the naming of the mehod or the documentation provide no clues. Very frustrating. The issue is valid though, so my vote goes to doing one round of research and then if it isn't found, we mark it as deprecated and clean up this code.

Going through various installations I cannot find a single record that has this pattern. Maybe others can do the same. SELECT * FROM dbo.Folders WHERE FolderPath LIKE '0%'. Then check those records. It must give us a hint. Maybe a specific folder provider uses this for something.

sleupold:
I can't remember ever having a DNN website (myself or any client), where a folder started with "0".
The only idea would be the default folder for the first website, which always starts with "0\" inside the website directory, but this should not be included in folder path.

mitchelsellers:
Funny enough...DNN will create folders that start with a Zero. 

Look under the /Users/ folder for an example. So it seems that code bypasses this

sleupold:
@mitchelsellers yes, but not directly in the portal folder, i.e. the path starts with "users/0%"

daguiler:
Hi guys.
I wonder what's so special about `\0`.
Yes, it's the null char, but there are dozens of other special characters that might produce similar issues, like `\r`, `\n`, `\a`, `\b`, `\f`, `\t`, etc., etc. Yet, we can create folders starting with these letters normally.

bdukes:
Has anyone looked into whether the new asset/resource manager uses this code?

valadas:
> Hi guys.
> I wonder what's so special about `\0`.
> Yes, it's the null char, but there are dozens of other special characters that might produce similar issues, like `\r`, `\n`, `\a`, `\b`, `\f`, `\t`, etc., etc. Yet, we can create folders starting with these letters normally.

True, true... Yeah, I must admit I am a bit clueless about the purpose here...

valadas:
>Has anyone looked into whether the new asset/resource manager uses this code?

It does not directly  but it calls AssetManager.Instance.CreateFolder which in turn calls this method. So I would say yeah.

kmuralidaran:
@valadas Do we have any use case to validate the side effects of this fix?

valadas:
@kmuralidaran  Well, I am really not sure, my guess was that it was to prevent \0 which is a null escape character, but maybe it's not since other replacement characters would also be a problem here, so really not sure what is special with 0 and why this was put in place...

bdukes (approved):
I support this change.  We allow folders that start with zero via other methods, and we have no specific scenario that is affected by this change.

mitchelsellers (approved):
I'm with Brian on this, lets approve!

mitchelsellers:
@valadas You good with this?

valadas:
> @valadas You good with this?

Yeah, I mean it looks it this must be a leftover from god knows what... I am approving.

valadas (approved):
Looks good to me

PR #3984, opened 2020-08-14
Support C# Async Constructs During Page Rendering

bdukes:
## Summary
Thanks @mitchelsellers for tracking down the answer to our `async` issues.

This turns on [`Async`](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ydy4x04a(v=vs.100)?redirectedfrom=MSDN#attributes) for `Default.aspx`.

Previously, when we attempted this in #2089, we ran into an issue with the culture not being preserved throughout the entire page lifecycle.  So, for example, the page may have been loaded in a non-English context, but the URLs were generated for en-US anyway.

The solution is to set the thread culture via an `IHttpModule` implementation, rather than via the page's `OnInit` method.  This allows the culture to persist throughout the page request.

### DIFF
--- a/DNN Platform/HttpModules/DotNetNuke.HttpModules.csproj
+++ b/DNN Platform/HttpModules/DotNetNuke.HttpModules.csproj
@@ -72,6 +72,7 @@
       <Link>SolutionInfo.cs</Link>
     </Compile>
     <Compile Include="DependencyInjection\ServiceRequestScopeModule.cs" />
+    <Compile Include="Localization\LocalizationModule.cs" />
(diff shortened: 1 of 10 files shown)


david-poindexter:
This is fantastic!

mitchelsellers:
This is awesome. I’ll review over the weekend

valadas:
Whohoo

valadas (approved):
Looks good to me.

david-poindexter (approved):

Issue #3985, opened 2020-08-15
Add Dependency Injection Support for ClientResourceManager


ahoefling:
## Description of problem
There is no interface for `ClientResourceManager` and it does not currently support Dependency Injection

## Description of solution
Extract interface and register with Dependency Injection container. 

## Description of alternatives considered
N/A

## Screenshots
N/A

## Additional context
This extension point will allow developers to add significant performance improvements to DNN. After modifying my ClientResourceManager with async scripts and adjusting orders I had DNN loading on average 100ms-300ms faster on each request. This performance change is a breaking change to DNN which is why I want to extract an interface instead. If we open up this extension point I can implement my performance changes as a module that can be used in current releases of DNN.

A proper implementation of this change will completely de-couple DNN from the 3rd party Client Resource Management tool used in DNN. It'll give us an opportunity to investigate new ways to handle client scripts that will work in a potential .NET Core solution of DNN.

## Affected browser
N/A

I will be submitting a pull request for this, I have implemented this in a fork already and it is working

david-poindexter:
@ahoefling this honestly sounds amazing!  I am looking forward to seeing what this leads to.

ahoefling:
I created a table on the initial comment of this issue to document the required features needed before work on this can start.

ahoefling:
Only 2 more items to convert to Dependency Injection and then we can submit a PR for this. Making progress!

ahoefling:
@bdukes @mitchelsellers The last bit of required work is for the `PortalController` which is a very large chunk of work. Any idea on how we can break this into more manageable chunks for submit as PRs? I am thinking evaluate all the properties that are used in the `PortalController` and it's dependencies and extract those to interfaces first. That alone will be touching many files.

Curious if you have any ideas of the path of least resistance here

mitchelsellers:
I think that makes sense

bdukes:
Does it make sense to look at what's used by the `ClientResourceManager`, and see if those make sense to be extracted into their own smaller interface?  We don't want to have 100 portal-related interfaces, but it could be nice and clean to have five or ten…

ahoefling:
When I was working on the `ClientResourceManager` I had everything ready to go, until I found this code that uses reflection to deal with a circular dependency.
[...]

PR #3988, opened 2020-08-16
Created IApplicationInfo and IApplicationStatusInfo for Dependency Injection

ahoefling:
Related to #3985 

## Summary
**This does not close the linked work item, but is required before I can submit a pull request.**

While working on changes to the Client Resource Manager I noticed circular dependency between `DotNetNuke.Libary` and `DotNetNuke.Web.Client`. The `ClientDependencySetting` utilizes reflection to workaround the circular dependency. Dynamically loading types like this is considered an anti-pattern, especially when we can use depenency injection to extract an interface.

This Pull Request extracts 2 interfaces from the Globals.cs
- `IApplicationInfo`
- `IApplicationStatusInfo`

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Common/LazyServiceProvider.cs
+++ b/DNN Platform/DotNetNuke.Web/Common/LazyServiceProvider.cs
@@ -5,28 +5,30 @@
 namespace DotNetNuke.Web.Common
 {
     using System;
-    using System.Collections.Generic;
-    using System.Linq;
-    using System.Text;
-    using System.Threading.Tasks;
+    using System.ComponentModel;
 
-    public class LazyServiceProvider : IServiceProvider
+    public class LazyServiceProvider : IServiceProvider, INotifyPropertyChanged
     {
-        private IServiceProvider _serviceProvider;
+        private IServiceProvider serviceProvider;
 
+        /// <inheritdoc/>
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        /// <inheritdoc/>
         public object GetService(Type serviceType)
         {
-            if (this._serviceProvider is null)
+            if (this.serviceProvider is null)
             {
                 throw new Exception("Cannot resolve services until the service provider is built.");
             }
 
-            return this._serviceProvider.GetService(serviceType);
+            return this.serviceProvider.GetService(serviceType);
         }
 
         internal void SetProvider(IServiceProvider serviceProvider)
         {
-            this._serviceProvider = serviceProvider;
+            this.serviceProvider = serviceProvider;
+            this.PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(serviceProvider)));
         }
     }
 }
--- a/DNN Platform/Library/Application/DotNetNukeContext.cs
+++ b/DNN Platform/Library/Application/DotNetNukeContext.cs
@@ -34,6 +34,11 @@ namespace DotNetNuke.Application
             : this(Globals.DependencyProvider.GetRequiredService<IApplicationInfo>())
         { }
 
+        [Obsolete("Deprecated in DotNetNuke 9.7.1. This constructor has been replaced by the overload taking an IApplicationInfo, which should be resolved via Dependency Injection. Scheduled removal in v11.0.0.")]
+    protected DotNetNukeContext(Application application)
+        : this((IApplicationInfo)application)
+    { }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="DotNetNukeContext" /> class using the provided application as base.
         /// </summary>
--- a/DNN Platform/Library/Application/DotNetNukeContext.cs
+++ b/DNN Platform/Library/Application/DotNetNukeContext.cs
@@ -24,6 +24,17 @@ namespace DotNetNuke.Application
         private readonly IApplicationInfo applicationInfo;
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="DotNetNukeContext"/> class.
+        /// </summary>
+        /// <remarks>
+        /// Initialize using the public constructor for Dependency Injection, this method will be removed.
+        /// </remarks>
+        [Obsolete("Deprecated in DotNetNuke 9.7.1. This constructor has been replaced by parameterized public constructor which is designed to be used with Dependency Injection. Resolve the new interface 'DotNetNuke.Abstractions.IDnnContext' instead. Scheduled removal in v11.0.0.")]
+        protected DotNetNukeContext()
+            : this(Globals.DependencyProvider.GetRequiredService<IApplicationInfo>())
+        { }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="DotNetNukeContext" /> class using the provided application as base.
         /// </summary>
         /// <param name="applicationInfo">The application.</param>
--- a/DNN Platform/Library/Application/DotNetNukeContext.cs
+++ b/DNN Platform/Library/Application/DotNetNukeContext.cs
@@ -61,7 +61,7 @@ namespace DotNetNuke.Application
         /// <summary>
         /// Gets get the application.
         /// </summary>
-        public Application Application { get => this.Application is Application app ? app : new Application(); }
+        public Application Application { get => this.applicationInfo is Application app ? app : new Application(); }
 
         /// <inheritdoc />
         IApplicationInfo IDnnContext.Application { get => this.applicationInfo; }
--- a/DNN Platform/Library/Application/DotNetNukeContext.cs
+++ b/DNN Platform/Library/Application/DotNetNukeContext.cs
@@ -1,41 +1,40 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
+
 namespace DotNetNuke.Application
 {
+    using System;
     using System.Collections.Generic;
 
+    using DotNetNuke.Abstractions.Application;
     using DotNetNuke.Collections.Internal;
+    using DotNetNuke.Common;
     using DotNetNuke.UI.Containers.EventListeners;
     using DotNetNuke.UI.Skins.EventListeners;
+
+    using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>
     /// Defines the context for the environment of the DotNetNuke application.
     /// </summary>
-    public class DotNetNukeContext
+    public class DotNetNukeContext : IDnnContext
     {
-        private static DotNetNukeContext _current;
-        private readonly Application _application;
-        private readonly IList<ContainerEventListener> _containerEventListeners;
-        private readonly IList<SkinEventListener> _skinEventListeners;
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DotNetNukeContext" /> class.
-        /// </summary>
-        protected DotNetNukeContext()
-            : this(new Application())
-        {
-        }
+        private static IDnnContext current;
+        private readonly IApplicationInfo applicationInfo;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DotNetNukeContext" /> class using the provided application as base.
         /// </summary>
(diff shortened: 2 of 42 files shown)


DNN Platform/Library/Application/AssemblyStatusAttribute.cs
```diff
@@ -26,27 +26,27 @@ public enum ReleaseMode
         /// <summary>
         /// Not asssigned
         /// </summary>
-        None,
+        None = 0,
```
ahoefling, line 29: I Added explicit numbers here to ensure this enum interops correctly with the new `DotNetNuke.Abstractions` enum. This will prevent any accidental breaking changes if the enums are ever re-ordered or changed. When we are converting the enums it is a simple cast as the int value matches, there is no need for a complex conversion method.

DNN Platform/Library/Application/AssemblyStatusAttribute.cs
```diff
@@ -26,27 +26,27 @@ public enum ReleaseMode
...
 
         /// <summary>
         /// Alpha release
         /// </summary>
-        Alpha,
+        Alpha = 1,
```
ahoefling, line 34: See above comment - https://github.com/dnnsoftware/Dnn.Platform/pull/3988/files#r471134926

DNN Platform/Library/Application/DotNetNukeContext.cs
```diff
@@ -45,30 +44,27 @@ public static DotNetNukeContext Current
         {
             get
             {
-                if (_current == null)
+                if (current == null)
                 {
-                    _current = new DotNetNukeContext();
+                    current = Globals.DependencyProvider.GetRequiredService<IDnnContext>();
```
ahoefling, line 49: Should we deprecate this function and ask consumers to use Dependency Injection? I decided not to deprecate this function because I didn't add the remaining members to the new `IDnnContext`. I think once we add the remaining members we can deprecate this method and recommend Dependency Injection.

If reviewers are in agreement I can create a work item

DNN Platform/Library/Application/DotNetNukeContext.cs
```diff
@@ -45,30 +44,27 @@ public static DotNetNukeContext Current
...
         /// Gets get the application.
         /// </summary>
-        public Application Application
-        {
-            get
-            {
-                return this._application;
-            }
-        }
+        public Application Application { get => this.applicationInfo is Application app ? app : new Application(); }
```
ahoefling, line 64: If the class is instantiated without Dependency Injection for some reason such as unit tests, I updated a default `Application` instantiation. Maybe we should update this to the `DependencyProvider`

DNN Platform/Library/Common/Globals.cs
```diff
@@ -316,28 +316,28 @@ public enum UpgradeStatus
             /// <summary>
             /// The application need update to a higher version.
             /// </summary>
-            Upgrade,
+            Upgrade = 0,
```
ahoefling, line 319: See above comment about enums - https://github.com/dnnsoftware/Dnn.Platform/pull/3988/files#r471134926

DNN Platform/Tests/DotNetNuke.Tests.Core/Controllers/Search/SearchControllerTests.cs
```diff
@@ -139,6 +141,10 @@ public void SetUp()
             ComponentFactory.Container = new SimpleContainer();
             MockComponentProvider.ResetContainer();
 
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddTransient<IApplicationStatusInfo>(container => new DotNetNuke.Application.ApplicationStatusInfo(Mock.Of<IApplicationInfo>()));
```
ahoefling, line 145: Many of the unit tests depend on File Operations that are generated from the `MapApplicationPath` function. It was difficult to mock this out and it was easier to use an actual implementation of `IApplicationStatusInfo` instead of a mock.

DNN Platform/Library/Common/Globals.cs
```diff
@@ -4061,5 +3830,15 @@ private static bool FilenameMatchesExtensions(string filename, string strExtensi
...
         }
+
+        /// <summary>
+        /// Resolves dependencies after the <see cref="Globals.DependencyProvider"/>
+        /// has been initialized. This should only be called once to resolve all
+        /// dependencies used by the <see cref="Globals"/> object.
+        /// </summary>
+        private static void OnDependencyProviderChanged(object sender, PropertyChangedEventArgs eventArguments)
+        {
+            applicationStatusInfo = DependencyProvider.GetRequiredService<IApplicationStatusInfo>();
```
ahoefling, line 3841: Creating a method to register dependencies once the provider has been configured is needed since the `DependencyProvider` uses our custom `LazyServiceProvider`. **It is critical that the `Global.cs` resolves only ONE instance of a dependency, as it can generate too many sql connections that causes DNN to stop functioning**. 

This finding has been documented - #3989

DNN Platform/DotNetNuke.Web/Common/LazyServiceProvider.cs
```diff
@@ -5,28 +5,30 @@
...
 
-            return this._serviceProvider.GetService(serviceType);
+            return this.serviceProvider.GetService(serviceType);
         }
 
         internal void SetProvider(IServiceProvider serviceProvider)
         {
-            this._serviceProvider = serviceProvider;
+            this.serviceProvider = serviceProvider;
+            this.PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(serviceProvider)));
```
ahoefling, line 31: I added `INotifyPropertyChanged` here to allow the `Globals.cs` to listen for an event when the Service Provider is ready for us. Since the `Globals.cs` is wrapping several Dependency Injection calls as we deprecate methods it is important that it can resolve dependencies once they are available. This is critical when using `Globals.cs` at startup.

I decided on `INotifyPropertyChanged` as it is a .NET Standard type and is used in the desktop/mobile space a lot for notifying objects when something has changed. If there is a better option, I am more than happy to change this

ahoefling:
This is ready for review

I added single line comments around sections of the code that needed extra explanation of why I made some changes. All unit tests are passing. I downloaded the build artifacts and validated that it does install DNN correctly for a new installation.

ahoefling:
I'm not sure why it shows the build is still in progress. I just checked the build server and the build for this PR is showed success
- https://dev.azure.com/dotnet/DNN/_build/results?buildId=37348&view=results

DNN Platform/Library/Application/DotNetNukeContext.cs
```diff
@@ -1,41 +1,40 @@
...
     {
-        private static DotNetNukeContext _current;
-        private readonly Application _application;
-        private readonly IList<ContainerEventListener> _containerEventListeners;
-        private readonly IList<SkinEventListener> _skinEventListeners;
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DotNetNukeContext" /> class.
-        /// </summary>
-        protected DotNetNukeContext()
```
bdukes, line 25: @mitchelsellers do you have any concern about a breaking change from removing this `protected` constructor?

ahoefling: @bdukes is this a breaking change? the class uses a `public` access modifier, by removing the `protected` constructor it opens the API structure up slightly. I had to remove the `protected` constructor so the Dependency Injection container could properly resolve `DotNetNukeContext`. If we want to keep this `protected` I could create a new child class with a public constructor and then register `IDnnContext` to the child class.

bdukes: The scenario that would break is when someone is inheriting from `DotNetNukeContext` and calling `base()` in its constructor.  It shouldn't be breaking to leave these two `protected` constructors and add the new `public` constructor, though.

ahoefling: I just tried adding 2 default constructors
```
public DotNetNukeContext() {}
protected DotNetNukeContext() {}
```

This has a compilation error as a class can only have 1 parameterless default constructor regardless of access level. I don't think this is a breaking change because child classes will automatically call the default constructor regardless if it is public or protected. If the constructor was private, then you start getting into different behaviors.

bdukes: I'm not seeing a default constructor on the new version of `DotNetNukeContext`, just `public DotNetNukeContext(IApplicationInfo applicationInfo)`, am I missing it?

ahoefling: 🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️

Glad we talked through this, my mistake. You are 100% correct, there is no long a default constructor and that is **wrong**. We need to add a public default constructor to prevent a breaking change.

DNN Platform/Library/Common/Globals.cs
```diff
@@ -369,13 +364,8 @@ public static string ApplicationPath
...
         /// The application map path.
         /// </value>
-        public static string ApplicationMapPath
-        {
-            get
-            {
-                return _applicationMapPath ?? (_applicationMapPath = GetCurrentDomainDirectory());
-            }
-        }
+        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.IApplicationStatusInfo instead.")]
```
ahoefling, line 367: @bdukes @mitchelsellers 

I'm thinking of changing the verbiage of this to give better instructions. Do you have any thoughts on my suggested change?
```suggestion
        [Obsolete("Deprecated in 9.7.1. Using Dependency Injections resolve 'DotNetNuke.Abstractions.IApplicationStatusInfo' instead. Scheduled for removal in v11.0.0.")]
```

If this format is good, I'll update all the deprecation notices in this PR

DNN Platform/Library/Application/DotNetNukeContext.cs
ahoefling: I have added the protected constructor back in, but have deprecated it. Marking this conversation as resolved.

DNN Platform/Library/Application/DotNetNukeContext.cs
```diff
@@ -1,105 +1,100 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DotNetNukeContext"/> class.
+        /// </summary>
+        /// <remarks>
+        /// Initialize using the public constructor for Dependency Injection, this method will be removed.
+        /// </remarks>
+        [Obsolete("Deprecated in DotNetNuke 9.7.1. This constructor has been replaced by parameterized public constructor which is designed to be used with Dependency Injection. Resolve the new interface 'DotNetNuke.Abstractions.IDnnContext' instead. Scheduled removal in v11.0.0.")]
+        protected DotNetNukeContext()
+            : this(Globals.DependencyProvider.GetRequiredService<IApplicationInfo>())
+        { }
```
bdukes, line 35:
```suggestion
        { }

        [Obsolete("Deprecated in DotNetNuke 9.7.1. This constructor has been replaced by the overload taking an IApplicationInfo, which should be resolved via Dependency Injection. Scheduled removal in v11.0.0.")]
    protected DotNetNukeContext(Application application)
        : this((IApplicationInfo)application)
    { }
```

ahoefling: This makes sense to me, thanks for the help on the verbiage!

DNN Platform/Library/Common/Globals.cs
ahoefling: This has been fixed in the latest code push, resolving conversation

ahoefling:
@bdukes I believe I have addressed all of your concerns in the latest update. Let me know if there is anything else I need to do to get this PR ready to merge

mitchelsellers:
@bdukes We good for this to be merged for 9.7.1?  I don't think it will be a big/breaking issue.

bdukes:
Hold on to your butts GIF

valadas:
Amazing work!

PR #3990, opened 2020-08-17
Created IHostSettingsService for Dependency Injection

ahoefling:
Related to #3985

## Summary
**This does not close the linked work item, but is required before I can submit a pull request.**

Added an `IHostController` an abstraction of the `HostController` that will be used with Dependency Injection. The DotNetNuke.Library code for this is a small sample of what I have been thinking would be a good approach for migrating controllers and entities from DotNetNuke.Library to DotNetNuke.Abstractions.

### DIFF
--- a/DNN Platform/DotNetNuke.Abstractions/Entities/Controllers/IHostController.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Entities/Controllers/IHostController.cs
@@ -1,0 +1,165 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Abstractions.Entities.Controllers
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// HostController provides business layer of host settings.
+    /// </summary>
+    /// <example>
+    /// 
+    /// <code lang="C#">
+    /// public class MySampleClass
+    /// {
+    ///     IHostController controller;
+    ///     public MySampleClass(IHostController controller)
+    ///     {
+    ///         this.controller = controller;
+    ///     }
+    ///
+    ///     public bool CheckUpgrade { get => this.controller.GetBoolean("CheckUpgrade", true);
+    /// }
+    /// </code>
+    /// </example>
+    public interface IHostController
+    {
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        bool GetBoolean(string key);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="defaultValue">this value will be return if setting's value is empty.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        bool GetBoolean(string key, bool defaultValue);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        double GetDouble(string key);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="defaultValue">this value will be return if setting's value is empty.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        double GetDouble(string key, double defaultValue);
+
+        /// <summary>
+        /// takes in a text value, decrypts it with a FIPS compliant algorithm and returns the value.
+        /// </summary>
+        /// <param name="key">the host setting to read.</param>
+        /// <param name="passPhrase">the pass phrase used for encryption/decryption.</param>
+        /// <returns>The setting value as a <see cref="string"/>.</returns>
+        string GetEncryptedString(string key, string passPhrase);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        int GetInteger(string key);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="defaultValue">this value will be return if setting's value is empty.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        int GetInteger(string key, int defaultValue);
+
+        /// <summary>
+        /// Gets all host settings.
+        /// </summary>
+        /// <returns>host setting.</returns>
+        Dictionary<string, IConfigurationSetting> GetSettings();
+
+        /// <summary>
+        /// Gets all host settings as dictionary.
+        /// </summary>
+        /// <returns>host setting's value.</returns>
+        Dictionary<string, string> GetSettingsDictionary();
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        string GetString(string key);
+
+        /// <summary>
+        /// Gets the setting value by the specific key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="defaultValue">this value will be return if setting's value is empty.</param>
+        /// <returns>host setting's value.</returns>
+        /// <exception cref="System.ArgumentException">key is empty.</exception>
+        string GetString(string key, string defaultValue);
+
+        /// <summary>
+        /// Increments the Client Resource Manager (CRM) version to bust local cache.
+        /// </summary>
+        /// <param name="includeOverridingPortals">If true also forces a CRM version increment on portals that have non-default settings for CRM.</param>
+        void IncrementCrmVersion(bool includeOverridingPortals);
+
+        /// <summary>
+        /// Updates the specified config.
+        /// </summary>
+        /// <param name="config">The config.</param>
+        void Update(IConfigurationSetting config);
+
+        /// <summary>
+        /// Updates the specified config.
+        /// </summary>
+        /// <param name="config">The config.</param>
+        /// <param name="clearCache">if set to <c>true</c> will clear cache after updating the setting.</param>
+        void Update(IConfigurationSetting config, bool clearCache);
+
+        /// <summary>
+        /// Updates the specified settings.
+        /// </summary>
+        /// <param name="settings">The settings.</param>
+        void Update(Dictionary<string, string> settings);
+
+        /// <summary>
+        /// Updates the setting for a specified key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="value">The value.</param>
+        void Update(string key, string value);
+
+        /// <summary>
+        /// Updates the specified key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="value">The value.</param>
+        /// <param name="clearCache">if set to <c>true</c> will clear cache after update settings.</param>
+        void Update(string key, string value, bool clearCache);
+
+        /// <summary>
+        /// Takes in a <see cref="string"/> value, encrypts it with a FIPS compliant algorithm and stores it.
+        /// </summary>
+        /// <param name="key">host settings key.</param>
+        /// <param name="value">host settings value.</param>
+        /// <param name="passPhrase">pass phrase to allow encryption/decryption.</param>
+        void UpdateEncryptedString(string key, string value, string passPhrase);
+    }
+}
--- a/DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
@@ -20,7 +20,7 @@ namespace DotNetNuke.Abstractions.Entities
         string Key { get; set; }
 
         /// <summary>
-        /// gets or sets configuraiton value.
+        /// gets or sets the configuraiton value.
         /// </summary>
         string Value { get; set; }
     }
--- a/DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
@@ -1,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Abstractions.Entities
+{
+    /// <summary>
+    /// The configuration setting.
+    /// </summary>
+    public interface IConfigurationSetting
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether if the configuration setting is secure.
+        /// </summary>
+        bool IsSecure { get; set; }
+
+        /// <summary>
+        /// Gets or sets the configuration key.
+        /// </summary>
+        string Key { get; set; }
+
+        /// <summary>
+        /// gets or sets configuraiton value.
+        /// </summary>
+        string Value { get; set; }
+    }
+}
--- a/DNN Platform/Library/Entities/Controllers/HostController.cs
+++ b/DNN Platform/Library/Entities/Controllers/HostController.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
+
 namespace DotNetNuke.Entities.Controllers
 {
     using System;
@@ -8,7 +9,8 @@ namespace DotNetNuke.Entities.Controllers
     using System.Data;
     using System.Globalization;
     using System.Linq;
-    using DotNetNuke.Abstractions.Entities;
+
+    using DotNetNuke.Abstractions;
     using DotNetNuke.Common;
     using DotNetNuke.Common.Utilities;
     using DotNetNuke.Data;
@@ -19,10 +21,8 @@ namespace DotNetNuke.Entities.Controllers
     using DotNetNuke.Services.Log.EventLog;
     using DotNetNuke.Web.Client;
 
-    using INewHostController = DotNetNuke.Abstractions.Entities.Controllers.IHostController;
-
     /// <inheritdoc/>
-    public partial class HostController : INewHostController
+    public partial class HostController : IHostSettingsService
     {
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(HostController));
 
@@ -115,7 +115,7 @@ namespace DotNetNuke.Entities.Controllers
         }
 
         /// <inheritdoc/>
-        Dictionary<string, IConfigurationSetting> INewHostController.GetSettings()
+        Dictionary<string, IConfigurationSetting> IHostSettingsService.GetSettings()
         {
             return CBO.GetCachedObject<Dictionary<string, IConfigurationSetting>>(
                                             new CacheItemArgs(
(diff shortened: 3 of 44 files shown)


ahoefling:
I am working on the unit tests for this one, but wanted to submit the PR so the business rules could get reviewed. There are a bunch of mocking issues with unit tests.

We should wait until #3988 is merged prior to merging this. I would like to rebase on that to ensure we have a good merge in the unit test files.

DNN Platform/Library/Entities/Controllers/HostController.cs
```diff
@@ -20,49 +19,27 @@ namespace DotNetNuke.Entities.Controllers
...
-    /// {
-    ///     get
-    ///     {
-    ///         return HostController.Instance.GetBoolean("CheckUpgrade", true);
-    ///     }
-    /// }
-    /// </code>
-    /// </example>
-    public class HostController : ComponentBase<IHostController, HostController>, IHostController
+    using INewHostController = DotNetNuke.Abstractions.Entities.Controllers.IHostController;
```
mitchelsellers, line 22: Any particular reason for this, rather than a traditional using?

DNN Platform/Library/Entities/Controllers/HostController.cs
```diff
@@ -159,36 +114,25 @@ public int GetInteger(string key, int defaultValue)
             return retValue;
         }
 
-        /// <summary>
-        /// Gets all host settings.
-        /// </summary>
-        /// <returns>host setting.</returns>
-        public Dictionary<string, ConfigurationSetting> GetSettings()
+        /// <inheritdoc/>
+        Dictionary<string, IConfigurationSetting> INewHostController.GetSettings()
```
mitchelsellers, line 118: This should be public

DNN Platform/Library/Obsolete/HostController.cs
```diff
@@ -0,0 +1,60 @@
...
+    // None of the APIs are deprecated, but the IHostController
+    // is deprecated and moved to the abstractions project. When
+    // it is time to remove APIs we should remove the parent
+    // classes listed here
+    /// <inheritdoc />
+    public partial class HostController : ComponentBase<IHostController, HostController>, IHostController
+    {
+        private static IHostController currentInstance;
+
+        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Controllers.IHostController instead.")]
```
mitchelsellers, line 27: It would be great if we could give more of of the "how" to fix so that people could know more than just what interface to look at.

DNN Platform/Library/Startup.cs
```diff
@@ -2,26 +2,31 @@
...
-            services.AddSingleton<ReflectedModuleControlFactory>();
-
-            services.AddTransient(x => PortalController.Instance);
-            services.AddTransient<INavigationManager, NavigationManager>();
-        }
-    }
-}
+    using Microsoft.Extensions.DependencyInjection;
+
+    using INewHostController = DotNetNuke.Abstractions.Entities.Controllers.IHostController;
```
mitchelsellers, line 17: I understand WHY this is done, but this smells really funny

DNN Platform/Library/Entities/Controllers/HostController.cs
ahoefling: ## Short Answer
In the deprecated partial class you will need to use the new abstractions interface. But now we have 2 interfaces with the exact same name. We can either alias the using statement as I did here, or use long fully qualified names.

## Long Answer
When I looked through the original history of the controller interfaces I discovered that they were originally built as a way to add more unit tests to DNN not ever designed for production use. The comment isn't shared across all controller interfaces. In addition to we need to extract an interface from our controller and move it over to the abstractions project so it is truly ready for dependency injection, as well as any future vNext of DNN.
https://github.com/dnnsoftware/Dnn.Platform/blob/b098a16d467ae3890ed13d654be092e4b5c0bad3/DNN%20Platform/Library/Entities/Portals/IPortalController.cs#L14-L18

My technique to organizing the code was to break it into a few sections
- Create interface in abstractions project
- Create partial class of controller implementation in Obsolete folder
- Update original class to partial class
- Remove deprecated inheritance to obsolete class (ServiceLocator and old interface)
- Add new interface to the original implementation
- **In the deprecated partial class you will need to use the new abstractions interface. But now we have 2 interfaces with the exact same name. We can either alias the using statement as I did here, or use long fully qualified names.**
  - All deprecated methods should use the new implementation and there shouldn't be 2 copies of the same implementation

DNN Platform/Library/Entities/Controllers/HostController.cs
ahoefling: I extracted the new `IConfigurationSetting` interface into the abstractions project. It is not possible to have 2 identical methods in a class with different return types. The compiler reports errors as it doesn't know how to handle it. Explicit interface implementations like this is a way to solve it. Considering the new interface will only be used for Dependency Injection I believe this change is safe. 

Once we move the deprecated method `Dictionary<string, ConfigurationSetting> GetSettings()` we can remove the explicit declaration and make this public

DNN Platform/Library/Startup.cs
ahoefling: See comment above - https://github.com/dnnsoftware/Dnn.Platform/pull/3990#discussion_r471475756. Let's have the discussion in my alias statement there to prevent thread splitting

DNN Platform/Library/Obsolete/HostController.cs
ahoefling: I agree, but I am not sure what to put down. The Dependency Injection story is slightly different depending on what part of the platform you are working in. Should it be as simple as this?

```suggestion
        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Controllers.IHostController via Dependency Injection instead.")]
```

ahoefling:
Thanks for the review @mitchelsellers . I would like us to hold off on merging this until we complete #3988 as the unit testing work done in that PR will help this one. When you get a chance could you review that one, I am interested in your thoughts on the new interfaces I added.

DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        bool IsSecure { get; set; }
+
+        /// <summary>
+        /// Gets or sets the configuration key.
+        /// </summary>
+        string Key { get; set; }
+
+        /// <summary>
+        /// gets or sets configuraiton value.
```
bdukes, line 23:
```suggestion
        /// Gets or sets the configuration setting value.
```

DNN Platform/Library/Obsolete/HostController.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <inheritdoc/>
+        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Controllers.IHostController instead.")]
+        public void Update(ConfigurationSetting config) =>
+            (this as INewHostController).Update(config);
+
+        /// <inheritdoc/>
+        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Controllers.IHostController instead.")]
+        public void Update(ConfigurationSetting config, bool clearCache) =>
+            (this as INewHostController).Update(config, clearCache);
```
bdukes, line 58: I would prefer that this use a regular cast instead of `as`, so that if anything goes wrong, it'll throw an invalid cast exception instead of a null reference exception.

ahoefling: Something like this?

```suggestion
            ((INewHostController)this).Update(config, clearCache);
```

bdukes: yup!

ahoefling:
Now that #3988 is merged, this PR needs to be rebased on `develop`. There is a lot of unit test work from that PR that this PR will need to update to follow the pattern. I can now dedicate my time to working towards the code review feedback

bdukes:
@ahoefling I've updated your branch with the rebase.  Hope it helps!

DNN Platform/Library/Obsolete/HostController.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <inheritdoc />
+    public partial class HostController : ComponentBase<IHostController, HostController>, IHostController
+    {
+        [Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Controllers.IHostController instead.")]
+        public static new IHostController Instance
+        {
+            get
+            {
+                var newHostController = Globals.DependencyProvider.GetRequiredService<INewHostController>();
```
ahoefling, line 31: I updated this from the initial PR submission to remove the static instance. This created a poorly implemented singleton that prevented the Dependency Injection container from operating correctly. We should offload calls like this to use the container as it will manage instance state for us automatically. If we want the `IHostController` to be a singleton for example, let's leave that responsibility to the container.

ahoefling:
@bdukes thanks for the assist, managing the rebase gave me a great head start on fixing all the broken unit tests. I have updated the PR with the requested changes and fixed all the unit tests. It was quite satisfying updating unit test mocks to use the new `IHostController` vs the old way of using `HostController.RegisterInstance`.


## Changes
[x] Fixed all failing unit tests, updated mocks
[x] Updated `HostController.Instance` to offload state management to the Dependency Injection Container
[x] Updated modified unit tests to following coding standards and guideline recommendations from stylecop

## What's Next
I have checked all the builds locally and all the unit tests are passing. I have not ran the generated installer build yet, but plan to do it on tomorrow (8/20/2020) when I get a chance. If someone wants to grab the install file and report back, that would save me some time and be a big help!

DNN Platform/Library/Startup.cs
```diff
@@ -26,6 +30,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<IDnnContext, DotNetNukeContext>();
 
             services.AddTransient(x => PortalController.Instance);
+            services.AddTransient<INewHostController, HostController>();
```
ahoefling, line 33: This should be registered as a scoped service and not Transient. Scoped services are re-used for the life of the request, where transient is a new instance every time requested.

ahoefling: This has been resolved in the latest code push and it now uses `Scoped`

ahoefling:
I just tested the installer and it appears to work without issue

DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        bool IsSecure { get; set; }
+
+        /// <summary>
+        /// Gets or sets the configuration key.
+        /// </summary>
+        string Key { get; set; }
+
+        /// <summary>
+        /// gets or sets the configuraiton value.
```
bdukes, line 23:
```suggestion
        /// Gets or sets the configuration value.
```

DNN Platform/DotNetNuke.Abstractions/Entities/Controllers/IHostController.cs
```diff
@@ -0,0 +1,165 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Abstractions.Entities.Controllers
```
bdukes, line 5: What if this interface was `DotNetNuke.Abstractions.IHostSettingsController` or `DotNetNuke.Abstractions.Settings.IHostSettingsController`?  Using `Entities.Controllers` doesn't mean anything to me, I'd like to simplify if we can (and set ourselves up with good patterns we want to continue).  Also, `IHostController` doesn't (to me) communicate the scope of the interface, I think `IHostSettingsController` clearly captures the responsibilities in this interface.

bdukes: And, as an added bonus, if we use `IHostSettingsController` as the interface name, we don't need to introduce the `using` alias between the old and new interface.

DNN Platform/Library/Obsolete/HostController.cs
bdukes: I might say _resolve_ instead of _use_, but I don't think we can add more details than _via Dependency Injection_, unless we add a link (to dnndocs?)

DNN Platform/Tests/DotNetNuke.Tests.Content/DotNetNuke.Tests.Content.csproj
```diff
@@ -108,6 +108,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\DotNetNuke.Abstractions\DotNetNuke.Abstractions.csproj">
       <Project>{6928A9B1-F88A-4581-A132-D3EB38669BB0}</Project>
+      <Project>{6928a9b1-f88a-4581-a132-d3eb38669bb0}</Project>
```
bdukes, line 111: Looks like a bad merge
```suggestion
```

DNN Platform/DotNetNuke.Abstractions/Entities/Controllers/IHostController.cs
ahoefling: ❤❤❤

This makes sense to me and it will allow us to remove all the aliases.

ahoefling: I was thinking about this a little bit more. As we migrate and shift interfaces to the abstractions project we should use the opportunity to standardize with naming conventions that are used throughout the rest of the .NET Ecosystem. 

Let's update the `IHostController` -> `IHostSettingsService`

DNN Platform/Library/Entities/Controllers/HostController.cs
ahoefling: @bdukes and I chatted about this and decided it would be best to name the new interface `IHostSettingsService`. I have updated the PR and there are no longer any using aliases. I am going to mark this conversation as resolved.

https://github.com/dnnsoftware/Dnn.Platform/pull/3990#discussion_r474019228

DNN Platform/Library/Startup.cs
ahoefling: We have renamed the interface to `IHostSettingsService` and there is no longer any need for the using alias. 

https://github.com/dnnsoftware/Dnn.Platform/pull/3990#discussion_r474019228

DNN Platform/Library/Obsolete/HostController.cs
ahoefling: This has been fixed in the latest code push

DNN Platform/DotNetNuke.Abstractions/Entities/Controllers/IHostController.cs
ahoefling: Added in latest code push, resolving conversation

DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
ahoefling: Duplicate, there is another item for this further down. I'm going to resolve this conversation

https://github.com/dnnsoftware/Dnn.Platform/pull/3990#discussion_r474004736

DNN Platform/Library/Obsolete/HostController.cs
ahoefling: The verbiage has been updated for the deprecation messages to

```
[Obsolete("Deprecated in 9.7.1. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.IHostSettingsService instead.")]
```

DNN Platform/DotNetNuke.Abstractions/Entities/IConfigurationSetting.cs
ahoefling: Updated in latest code push

ahoefling:
Since we are having conversations about migrating items such as `HostController` -> `IHostSettingService` to follow modern .NET conventions and best practices. Should we talk about the strategy for migrating Entity Models. In this PR there is the `ConfigurationSetting`, which maps to the `HostSettings` table. If I were building a greenfield application with that structure I would be creating an Entity class that maps 1:1 to that table. I would consider creating a business layer model or Data Transfer Object to build a bridge between the business layer and the database. If we want DNN to migrate towards an ultimate solution of Entity Framework Core, it would be useful to document what is appropriate. 

Perhaps something like this
```c#
public class HostSettingEntity
{
    public int Id { get; set; }
    public string Key { get; set; }
    public string Value { get; set; }
    public bool IsSecured { get; set; }
}
```

```c#
public class HostSetting
{
    public string Key { get; set; }
    public string Value { get; set; }
}
```

In this example I created a `HostSettingEntity` which maps 1:1 to the columns in the database table. We don't need to suffix it with `Entity`, I did that to illustrate the differences between entity and business model. When the IHostSettingsService queries the data it would read the data from the `IHostSettingEntity` and then it would transfer the data to `HostSettings` which would be returned. In this scenario the `IHostSettingsService` abstracts some of the rules such as encryption.

bdukes:
So, would you recommend moving `DotNetNuke.Abstractions.Users.IUserInfo` to `DotNetNuke.Abstractions.Models.IUserInfo`?  I think I would lean more towards keeping it with the rest of the `Users` namespace.  @mitchelsellers thoughts?  Or am I misunderstanding your idea, @ahoefling?  Thanks!

mitchelsellers:
I would tend to agree with @bdukes on this one, I would really like to try and keep like elements together, so I think that putting any sort of models into a separate namespace would be confusing, and additionally would clutter that namespace greatly in the future.

ahoefling:
@bdukes @mitchelsellers 
This makes sense, I made the recommendation as a convention I have seen in many MVC and MVVM apps in the .NET ecosystem. A downfall of a models folder is when you get 20+ objects in that namespace. As you pointed out the grouping makes it easier to find things.

I'm thinking of moving `IConfigurationSetting` from 
```
DotNetNuke.Abstractions.IConfigurationSetting
```
to
```
DotNetNuke.Abstractions.Application.IConfigurationSetting
```

I think this would follow our guidelines as it is a model used at the application level. Our current pattern appears to be keeping service related interfaces at the root level and models in the folders. Do you all have any thoughts on this?

mitchelsellers:
I really like the use of "Application" to hold those larger level items, honestly, in some ways the IHostController might be good as well to go there.

Might we get a quick call setup @ahoefling and @bdukes might be faster to just chat through this a bit?

bdukes:
I have a little time this morning (but @ahoefling just messaged me that he was switching projects, so he may not have time).

I agree that `IHostSettingsService` makes sense to go into the `Application` namespace.

My question with `IConfigurationSetting` is whether it's really going to only apply to host settings, or if we'll reuse it for other types of settings.  If it's only for host settings, let's call it `DotNetNuke.Abstractions.Application.IHostSetting`, otherwise `DotNetNuke.Abstrations.Settings.IConfigurationSetting` (or without `.Settings` in the namespace if we can't come up with more settings-specific interfaces we'll want).

ahoefling:
I have no problem moving 
```
DotNetNuke.Abstractions.IHostSettingService
``` 
to 
```
DotNetNuke.Abstractions.Application.IHostSettingService
``` 
I originally moved it to the root namespace as a place to put all the services. Considering the service is focused on "Application" level business rules it does make sense to put it back. (I think it was originally in there when I first submitted the PR and I moved it out to try it this way).

I don't have time for a call today unfortunately, but I can continue the dialogue on github

ahoefling:
@bdukes 
I really like the settings namespace
```
DotNetNuke.Abstrations.Settings.IConfigurationSetting
```

I would argue that it should keep the prefix of `Setting` because it would be confused with the `IConfiguration` object which reads the AppSettings from the appsettings.json or web.config

mitchelsellers:
I agree with the above and like the .Settings namespace as well

ahoefling:
Perfect! I'll make the following changes to the PR later today

Move: `IHostSettingService` -> into the `DotNetNuke.Abstractions.Application` namespace
Move: `IConfigurationSetting` -> into the `DotNetNuke.Abstractions.Settings` namespace

I'll do an audit of our changes and make sure I have no other questions. Once I push the update I'll request a re-review from both of you (@bdukes and @mitchelsellers). This should give us an opportunity to make sure there are no confusions or last minute questions on the change

ahoefling:
- Updated the IHostSettingsService and IConfigurationSetting with the new namespace recommendation
- Rebased branch on develop branch
- Requested re-review from @mitchelsellers and @bdukes

valadas:
Is this ready to merge @mitchelsellers @bdukes  ? And if so, in 9.7.1 or we hold it for 9.8.0 ?

mitchelsellers:
@bdukes @ahoefling thoughts? 

I think it is ok, as it isn't breaking

valadas:
@mitchelsellers sounds, good, merging...

ahoefling:
I think we have good test coverage on this and it doesn't introduce breaking changes. 🎉

Issue #3997, opened 2020-08-19
Add CRLF to .editorconfig to Prevent Mixed Line Endings


ahoefling:
## Description of bug
When contributing to DNN if you open up most files your IDE such as Visual Studio will display a message about mixed line endings. We should standardize the line endings in the editor config. 

This was discussed with @bdukes in #3995

ahoefling:
Closing this issue as the PR has been merged

PR #3998, opened 2020-08-19
Standardize on CR/LF line endings

bdukes:
Closes #3995


ahoefling:
This also closes #3997

donker (approved):

PR #4007, opened 2020-08-25
Fixes services framework usage in permission grid for the file manager

valadas:
@donker has been working integrating the persmissions grid into the new resource manager and had some issues with the persona bar specific services implementation integration.

This PR brings a copy of that implementation that should work inside or outside the persona bar.

I have tested that it does read the permissions properly and user search is also working. However there is still an issue saving those settings. But this PR brings us closer to have folder permissions working.

Thanks @david-poindexter  and @mitchelsellers  for co-coding on this one too.


valadas:
@donker the main change here from your fork is the addition of dnnSf.js in the app/globals folder and the changes in babel config, eslint config and package.json

donker (approved):

PR #4011, opened 2020-08-26
Fixes for issue "Searching users based on joined date is not working"

### LINKED ISSUE #4012: Searching users based on joined date is not working
## Description of bug
Searching users based on joined date is not working

## Steps to reproduce
1. Hover on 'MANAGE'
2. Click on 'Users'
3. Fill joining date in the "Search Users" box
4. Click on "Search" icon

## Current behavior
"No users found" displayed in search result

## Expected behavior
Users with the matching joining date should be listed

## Screenshots

**Before Search**
image

**After Search**
image

## Error information

## Additional context

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.07.00 release candidate
* [x] 09.06.02 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge Chromium

kmuralidaran:
Fixes #4012

## Summary

1. Updated Stored Procedure Personabar_GetUsersBySearchTerm to include CreatedOnDate(Joining date) values in  search criteria
2. Skipped modifying the search text when value is in date format

**After fix**

image


sleupold (changes requested):
the comparison does not use current UI locale and will fail for all date formats except US

sleupold (commented):
Besides, wouldn't it be more helpful to support "< 08/15/2020" and ">08/15/2020"

sleupold (changes requested):
SQL Code needs to go into a new file 09.07.01.SQLDataProvider, otherwise it will not be applied on upgrades.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/Helpers/SearchTextFilter.cs
```diff
@@ -16,6 +17,12 @@ public static string CleanWildcards(string searchText)
             {
                 return null;
             }
+
+            if (DateTime.TryParse(searchText.TrimEnd('*'), out var _))
```
mitchelsellers, line 21: If this is something to be supported, this should NOT be done in this method. as it is not cleaning a wildcard and the intention would not be clear

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/UsersController.cs
```diff
@@ -521,7 +521,9 @@ private bool CanUpdateUsername(UserInfo user)
 
             var parsedSearchText = string.IsNullOrEmpty(usersContract.SearchText) ? "" : SearchTextFilter.CleanWildcards(usersContract.SearchText.Trim());
 
-            usersContract.SearchText = string.Format("{0}*", parsedSearchText);
+            usersContract.SearchText = DateTime.TryParse(parsedSearchText, out var _)
```
mitchelsellers, line 524: This is now doing additional attempted date parsing and object creation, If we support this we should avoid multiple DateTime.TryParse calls over the same content

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/SqlDataProvider/03.00.00.SqlDataProvider
```diff
@@ -106,7 +106,13 @@ BEGIN
 						WHERE (LastName    Like @searchTerm)
 						  AND LastName  Is Not Null AND LastName  != N'' 
 						  AND (IsSuperUser = 1)
-						  AND (@isDeleted  Is Null OR IsDeleted = @isDeleted)) S 
+						  AND (@isDeleted  Is Null OR IsDeleted = @isDeleted)
+					   UNION SELECT UserID, Username, DisplayName, Email, CreatedOnDate, LastModifiedOnDate, IsDeleted, IsSuperUser
```
mitchelsellers, line 110: This needs to be discussed for international date & large DB performance

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/Helpers/SearchTextFilter.cs
david-poindexter: Agreed @mitchelsellers

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/UsersController.cs
david-poindexter: Agreed @mitchelsellers

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/SqlDataProvider/03.00.00.SqlDataProvider
david-poindexter: Also, if this is to be done, this should not be done here.  It should be done in a new `SqlDataProvider` file so it will fire for upgrades.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/UsersController.cs
sleupold: the main issue IMO: it is not supporting localized date formats

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/SqlDataProvider/03.00.00.SqlDataProvider
sleupold: @david-poindexter obviously, as stated before

sleupold:
> I don't really see any harm in supporting date-based search in addition to the existing capabilities, as long as we make sure to take into consideration other locale date formats. I could see this type of search having some value. I doubt there would be any serious performance considerations by this, but it may require some decent testing to make sure.

IMO the most common use case regarding dates is using "before.. " or "after" (and sometimes "from .. to ..", but not "at ...)

valadas:
Yeah, agree with all that, just to add my 2 cents, what about a date sort instead ?

sleupold:
@valadas, 
I agree, dates are best candidates for sorting, not as search filter.

kmuralidaran:
@valadas I will be creating new PR since this PR is dirty.

1. Requirement I got is to add provision to search based on joining date since it is displayed in the table and the existing functionality of search gives the impression that it will filter the rows based on the matching values displayed in the table.
2. Maybe we can add a separate feature to sort users based on the fields displayed in the table

kmuralidaran:
https://github.com/dnnsoftware/Dnn.Platform/pull/4017 created. 
@sleupold @david-poindexter @mitchelsellers @valadas Thanks for your suggestions. Could you please review the new PR #4017 and provide your inputs

PR #4016, opened 2020-08-26
Improves GitVersion speed

valadas:
Sets the maximum GitVersion lookup date to january 1st 2020 thus speeding up GitVersion calculations a lot. This means GitVersion will only look at commits from this year instead of well all the project history to find tags and read commit messages, etc.

### DIFF
--- a/gitversion.yml
+++ b/gitversion.yml
@@ -3,7 +3,7 @@ commit-date-format: 'yyyyMMdd'
 assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{CommitsSinceVersionSource}'
 mode: ContinuousDeployment
 ignore:
-  commits-before: 2019-01-01T00:00:00
+  commits-before: 2020-01-01T00:00:00
 branches:
   future:
     regex: future?[/-]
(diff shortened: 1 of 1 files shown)


gitversion.yml
```diff
@@ -2,8 +2,9 @@ next-version: 9.6.0
 commit-date-format: 'yyyyMMdd'
 assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{CommitsSinceVersionSource}'
 mode: ContinuousDeployment
+ignore:
+  commits-before: 2019-01-01T00:00:00
```
mitchelsellers, line 6: Should this be 2020-01-01?

valadas: Hmm, yeah, this guys did a mistake, let me run a try and change that date...

david-poindexter, line 6:
```suggestion
  commits-before: 2020-01-01T00:00:00
```

valadas:
Ok, I played a bit with different values and it looks like there is some unavoidable overhead no matter what value, but will it still improves the very very long times though. What's odd is running just `gitversion` is almost instant but in the build script (Cake) it never really get's under a minute... When I get more time maybe I can take a look at upgrading either cake or GitVersion

PR #4021, opened 2020-08-30
Created IPortalAliasService for Dependency Injection

ahoefling:
related to #3985

## Summary
- Adds a new interface `IPortalAliasService` which contains all the public APIs on the `PortalAliasController`. 
- Marked all static APIs on the `PortalAliasController` as deprecated
- Updated `IPortalInfo`
- Updated `IPortalAliasInfo`
  - Removed some APIs that were accidently added to the `IPortalAliasInfo`, see my review comments as it is technically a breaking change.
- Added new `PortalAliasCollection`

## Before Merging
| Task  | Status |
|-------|--------|
| Validate PR build | ✅ Completed |
| Unit Tests Passing | ✅ Completed |


DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+
+        /// <summary>
+        /// Gets or sets the skin.
+        /// </summary>
         string Skin { get; set; }
 
-        void Fill(IDataReader dr);
-        XmlSchema GetSchema();
-        void ReadXml(XmlReader reader);
-        void WriteXml(XmlWriter writer);
```
ahoefling, line 25: These APIs should never have been added to the `IPortalAliasInfo` in the first place. I removed them as they are contained within other .NET System level interfaces. This is technically a breaking change and I am not sure how we best manage this.

@bdukes @mitchelsellers any suggestions on this?

ahoefling:
Updated the PR to fix the failing unit tests to produce build artifacts. I tested the install build and portal aliases appear to update correctly.  Looking forward to our code review discussion 👍

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
valadas: Not saying it's not a breaking change but it was added in 9.7.0 and I doubt many 3rd parties call this... IMHO, I vote for removing it quick rather that keeping it in there deprecated for 2 full versions.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+        /// </summary>
         int PortalAliasID { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Portal ID.
+        /// </summary>
         int PortalID { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether if the portal is the primary portal.
```
bdukes, line 30:
```suggestion
        /// Gets or sets a value indicating whether this alias is the primary alias for the site/portal.
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+        /// </summary>
+        bool IsPrimary { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the poral is a redirect portal.
```
bdukes, line 35:
```suggestion
        /// Gets or sets a value indicating whether the alias is a redirect alias.
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+        /// </summary>
         bool Redirect { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Browser Type.
+        /// </summary>
+        BrowserTypes BrowserType { get; set; }
```
bdukes, line 42: This member was commented out before, what was the reason to add it back?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
         string Skin { get; set; }
 
-        void Fill(IDataReader dr);
-        XmlSchema GetSchema();
-        void ReadXml(XmlReader reader);
-        void WriteXml(XmlWriter writer);
+        /// <summary>
+        /// Gets or sets the key ID.
+        /// </summary>
+        int KeyID { get; set; }
```
bdukes, line 57: This member doesn't seem to be necessary for this interface.  Am I missing something?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+        /// </summary>
+        BrowserTypes BrowserType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the culture code.
+        /// </summary>
+        string CultureCode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the skin.
```
bdukes, line 50:
```suggestion
        /// Gets or sets the skin/theme for this alias, if there is one.
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,59 @@
...
+        /// </summary>
         bool Redirect { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Browser Type.
+        /// </summary>
+        BrowserTypes BrowserType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the culture code.
```
bdukes, line 45:
```suggestion
        /// Gets or sets the culture code for this alias, if there is one.
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <param name="tabId">The tab ID.</param>
+        /// <param name="portalAlias">The portal alias.</param>
+        /// <returns>Portal alias.</returns>
+        string GetPortalAliasByTab(int tabId, string portalAlias);
+
+        /// <summary>
+        /// Validates the alias.
+        /// </summary>
+        /// <param name="portalAlias">The portal alias.</param>
+        /// <param name="ischild">if set to <c>true</c> [ischild].</param>
```
bdukes, line 34:
```suggestion
        /// <param name="ischild">if set to <c>true</c>, validate the alias as a child alias (i.e. as a subdirectory of another alias).</param>
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// <value>E-mail of the portal.</value>
+        string Email { get; set; }
+
+        /// <summary>
+        /// Gets or sets date at which the portal expires.
+        /// </summary>
+        /// <value>Date of expiration of the portal.</value>
+        DateTime ExpiryDate { get; set; }
+
+
```
bdukes, line 122:
```suggestion
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// <summary>
+        /// Gets or sets gUID of the portal info object.
+        /// </summary>
+        /// <value>Portal info Object GUID.</value>
+        Guid GUID { get; set; }
+
+
```
bdukes, line 129:
```suggestion
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,317 @@
...
+        string KeyWords { get; set; }
+
+        /// <summary>
+        /// Gets or sets tabId with the login control, page to login.
+        /// </summary>
+        /// <value>TabId of the Login page.</value>
+        int LoginTabId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the portal has a logo (bitmap) associated with the portal. Teh admin can set the logo in the portal settings.
```
bdukes, line 161:
```suggestion
        /// Gets or sets the logo associated with the portal. The admin can set the logo in the portal settings.
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// Gets or sets the actual name of the Registerd Users group of the portal.
+        /// This name is retrieved from the RoleController object.
+        /// </summary>
+        /// <value>The name of the Registerd Users group.</value>
+        string RegisteredRoleName { get; set; }
+
+        /// <summary>
+        /// Gets or sets and sets the Key ID.
+        /// </summary>
+        int KeyID { get; set; }
```
bdukes, line 315: I don't think we want `KeyID` here either

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// </summary>
+        /// <param name="portalAliasId">The portal alias ID.</param>
+        /// <returns>Portal alias info.</returns>
+        IPortalAliasInfo GetPortalAliasByPortalAliasID(int portalAliasId);
+
+        /// <summary>
+        /// Get all portal aliases.
+        /// </summary>
+        /// <returns>A collection of portal aliases</returns>
+        PortalAliasCollection GetPortalAliases();
```
bdukes, line 77: If we're creating something new, do we want to use a custom collection class?  Or would it be simpler to just use `IDictionary<string, IPortalAliasInfo>`?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
bdukes: Ah, was it just commented out b/c the enum type wasn't available to this project?

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
```diff
@@ -1,68 +1,76 @@
...
-    }
-}
+    [Serializable]
+    [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.PortalAliasCollection instead.")]
+    public class PortalAliasCollection : DictionaryBase
+    {
+        /// <summary>
+        /// Gets a value indicating whether gets a value indicating if the collection contains keys that are not null.
+        /// </summary>
+        [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.PortalAliasCollection instead.")]
```
bdukes, line 17: I don't think we need to add `Obsolete` on each member if we've added it to the class

DNN Platform/Library/Entities/Portals/PortalAliasController.cs
```diff
@@ -1,390 +1,357 @@
...
+                default: // portal tab
+                    retValue = GetPortalAliasByPortal(intPortalId, portalAlias);
+                    break;
+            }
+
+            return retValue;
+        }
+
+        /// <inheritdoc />
+        bool IPortalAliasService.ValidateAlias(string portalAlias, bool ischild)
```
bdukes, line 112: While we're creating something new, can we fix the casing of this parameter?
```suggestion
        bool IPortalAliasService.ValidateAlias(string portalAlias, bool isChild)
```

DNN Platform/Library/Obsolete/PortalAliasController.cs
```diff
@@ -3,90 +3,197 @@
...
-        {
-            Instance.UpdatePortalAlias(portalAlias);
-        }
-    }
-}
+    using Microsoft.Extensions.DependencyInjection;
+
+    public partial class PortalAliasController : ServiceLocator<IPortalAliasController, PortalAliasController>, IPortalAliasController
+    {
+        [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.IPortalAliasSettingsService instead.")]
```
bdukes, line 26:
```suggestion
        [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.IPortalAliasService via dependency injection instead.")]

```

bdukes:
Thanks!

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
ahoefling: That is correct, the enum type was not in the abstraction project and it would have created a circular dependency. Since our pattern is creating a carbon-copy of the enums and ensuring each item maps to an int it is now safe to bring it over and handle the casts.

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
ahoefling: Do the compiler and intellisense warnings show up for every member if the class is deprecated?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
ahoefling: I agree with you on this, adding our own collection objects is something that I don't see as necessary anymore. I brought it over to keep consistency. I am going to take a 2nd look at the implementation and see if it is truly necessary. I'll report my findings here

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
mitchelsellers: I believe that we have to do it on the member for proper build warnings,

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
mitchelsellers: I would second the desire to rid of the custom collection in favor of a standard collection

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
mitchelsellers: I think that this will be an acceptable break due to the limited time, as users shouldn't be using it so that is the only time it would be truly breaking

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
mitchelsellers: Also, creating something new, would it be good to have a "GetDefaultPortalAlias" or similar with Portal & Culture?  To fix what people end up needing to do so often

DNN Platform/DotNetNuke.Abstractions/Portals/PortalAliasCollection.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    using System;
+    using System.Collections;
+
+    /// <summary>
+    /// The <see cref="IPortalAliasInfo"/> Collection which provides additional
+    /// helper functions for managing collections of <see cref="IPortalAliasInfo"/>.
+    /// </summary>
+    [Serializable]
+    public class PortalAliasCollection : DictionaryBase
```
mitchelsellers, line 15: looking at this implementation, at first glance I don't see any reason to have our own custom collection type

ahoefling: @bdukes and I were discussing this here - https://github.com/dnnsoftware/Dnn.Platform/pull/4021#discussion_r480122983

It sounds like the 3 of us agree that this implementation isn't necessary. I still plan to double check to make sure that we aren't overlooking anything.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
bdukes: Does this look right @mitchelsellers?
```c#
/// <summary>Gets the primary alias for the site.</summary>
/// <param name="portalId">The portal ID.</param>
/// <exception cref="ArgumentException">No portal has the given <paramref name="portalId"/>.</exception>
/// <returns>An <see cref="IPortalAliasInfo" /> instance.</returns>
/// <remarks>If no alias is marked as primary, returns the first neutral culture alias.</remarks>
IPortalAliasInfo GetPrimaryPortalAlias(int portalId);

/// <summary>Gets the primary alias for the site.</summary>
/// <param name="portalId">The portal ID.</param>
/// <param name="cultureCode">The culture code.</param>
/// <exception cref="ArgumentException">No portal has the given <paramref name="portalId"/>, or no alias matches the <paramref name="cultureCode"/>.</exception>
/// <returns>An <see cref="IPortalAliasInfo" /> instance.</returns>
/// <remarks>If no alias is marked as primary, returns the first alias matching the culture.</remarks>
IPortalAliasInfo GetPrimaryPortalAlias(int portalId, string cultureCode);
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
bdukes: `KeyID` is a property for the [`IHydratable`](https://github.com/dnnsoftware/Dnn.Platform/blob/cc61549005ab5ea1c66fa91670931642a62893e4/DNN%20Platform/Library/Entities/Modules/IHydratable.cs#L10) interface.

ahoefling: I agree, I don't think it should be included either. I'll remove it in the next update

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
ahoefling: I am marking this as resolved as we are deleting the items from the `IHydratable` interface in the abstractionsp roject

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
ahoefling: I have removed this in the latest update

ahoefling:
Thank you everyone for the feedback and assistance on this PR. I have pushed my changes based on the feedback everyone has left. Right now I am waiting for a final decision on the last remaining open discussion https://github.com/dnnsoftware/Dnn.Platform/pull/4021#discussion_r480128541.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets number of portal pages allowed in the portal. 0 means not limited.
+        /// </summary>
+        /// <value>Number of portal pages allowed.</value>
+        int PageQuota { get; set; }
+
+        /// <summary>
+        /// Gets or sets name of the Payment processor that is used for portal payments, e.g. PayPal.
+        /// </summary>
+        /// <value>Name of the portal payment processor.</value>
+        string PaymentProcessor { get; set; }
```
bdukes, line 174: @mitchelsellers is this feature deprecated?

mitchelsellers: Oh yeah...I don't think we even have an interface for it

ahoefling: Should I remove anything around Payments from `IPortalInfo`?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets tabId at which admin tasks start.
+        /// </summary>
+        /// <value>TabID of admin tasks.</value>
+        int AdminTabId { get; set; }
+
+        /// <summary>
+        /// Gets or sets image (bitmap) file that is used as background for the portal.
+        /// </summary>
+        /// <value>Name of the file that is used as background.</value>
+        string BackgroundFile { get; set; }
```
bdukes, line 74: @mitchelsellers this should be deprecated, right?  Which, I guess, means removed from the interface and (optionally) deprecated on `PortalInfo` itself

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: I think I would vote yes.  @bdukes

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// <summary>
+        /// Gets or sets current host version.
+        /// </summary>
+        string CrmVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets setting for the type of banner advertising in the portal.
+        /// </summary>
+        /// <value>Type of banner advertising.</value>
+        int BannerAdvertising { get; set; }
```
bdukes, line 85: @mitchelsellers this one doesn't appear to be used, remove/deprecate?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// <summary>
+        /// Gets or sets the Culture Code of the portal.
+        /// </summary>
+        string CultureCode { get; set; }
+
+        /// <summary>
+        /// Gets or sets curreny format that is used in the portal.
+        /// </summary>
+        /// <value>Currency of the portal.</value>
+        string Currency { get; set; }
```
bdukes, line 96: @mitchelsellers remove/deprecate?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets tabdId of the Home page.
+        /// </summary>
+        /// <value>TabId of the Home page.</value>
+        int HomeTabId { get; set; }
+
+        /// <summary>
+        /// Gets or sets amount of currency that is used as a hosting fee of the portal.
+        /// </summary>
+        /// <value>Currency amount hosting fee.</value>
+        float HostFee { get; set; }
```
bdukes, line 138: @mitchelsellers remove/deprecate?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
bdukes: Yeah, at least remove from this interface, perhaps deprecate on `PortalInfo` itself…

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets name of the portal. Can be set at creation time, Admin can change the name in the portal settings.
+        /// </summary>
+        /// <value>Name of the portal.</value>
+        string PortalName { get; set; }
+
+        /// <summary>
+        /// Gets or sets password to use in the payment processor.
+        /// </summary>
+        /// <value>Payment Processor password.</value>
+        string ProcessorPassword { get; set; }
```
bdukes, line 198: @mitchelsellers remove/deprecate?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        /// <summary>
+        /// Gets or sets payment Processor userId.
+        /// </summary>
+        /// <value>
+        string ProcessorUserId { get; set; }
```
bdukes, line 204: @mitchelsellers remove/deprecate?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets tabid of the User profile page.
+        /// </summary>
+        /// <value>TabdId of the User profile page.</value>
+        int UserTabId { get; set; }
+
+        /// <summary>
+        /// Gets or sets actual number of actual users for this portal.
+        /// </summary>
+        /// <value>Number of users for the portal.</value>
+        int Users { get; set; }
```
bdukes, line 282: @mitchelsellers this seems like an opportunity to not include this property on the interface, in an effort to avoid requiring an `IPortalInfo` implementation to take a dependency on `IUserController` (or `IUserService` or whatever)

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// This name is retrieved from the RoleController object.
+        /// </summary>
+        /// <value>The name of the Administrators group.</value>
+        string AdministratorRoleName { get; set; }
+
+        /// <summary>
+        /// Gets or sets actual number of pages of the portal.
+        /// </summary>
+        /// <value>Number of pages of the portal.</value>
+        int Pages { get; set; }
```
bdukes, line 301: @mitchelsellers as above, this seems like an opportunity to not include this property on the interface, in an effort to avoid requiring an `IPortalInfo` implementation to take a dependency on `ITabController` (or `ITabService` or whatever)

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// Gets or sets actual number of actual users for this portal.
+        /// </summary>
+        /// <value>Number of users for the portal.</value>
+        int Users { get; set; }
+
+        /// <summary>
+        /// Gets or sets dNN Version # of the portal installation.
+        /// </summary>
+        /// <value>Version # of the portal installation.</value>
+        string Version { get; set; }
```
bdukes, line 288: @mitchelsellers is this useful/meaningful?  I vote to remove

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <param name="portalId">The portal ID.</param>
+        /// <returns>Portal Alias Info.</returns>
+        IPortalAliasInfo GetPortalAlias(string alias, int portalId);
+
+        /// <summary>
+        /// Gets the portal alias by portal alias ID.
+        /// </summary>
+        /// <param name="portalAliasId">The portal alias ID.</param>
+        /// <returns>Portal alias info.</returns>
+        IPortalAliasInfo GetPortalAliasByPortalAliasID(int portalAliasId);
```
bdukes, line 71:
```suggestion
        IPortalAliasInfo GetPortalAliasByPortalAliasId(int portalAliasId);
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <param name="portalId">The portal ID.</param>
+        /// <returns>An IEnumerable of portal aliases.</returns>
+        IEnumerable<IPortalAliasInfo> GetPortalAliasesByPortalId(int portalId);
+
+        /// <summary>
+        /// Gets the portal by portal alias ID.
+        /// </summary>
+        /// <param name="PortalAliasId">The portal alias id.</param>
+        /// <returns>Portal info.</returns>
+        IPortalInfo GetPortalByPortalAliasID(int PortalAliasId);
```
bdukes, line 91:
```suggestion
        IPortalInfo GetPortalByPortalAliasId(int PortalAliasId);
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
ahoefling: Just took a look and that would be removing the following properties from the `IPortalInfo`
- `PaymentProcessor`
- `ProcessorPassword`
- `ProcessorUserId`

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,54 @@
...
+        /// Gets or sets the http alias.
+        /// </summary>
         string HTTPAlias { get; set; }
-        bool IsPrimary { get; set; }
-        int KeyID { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Portal Alias ID.
+        /// </summary>
         int PortalAliasID { get; set; }
```
bdukes, line 22: @mitchelsellers what do you think about enforcing a consistent naming convention for `Id` over `ID` in these new interfaces?

```suggestion
        int PortalAliasId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,54 @@
...
+
+        /// <summary>
+        /// Gets or sets the Portal ID.
+        /// </summary>
         int PortalID { get; set; }
```
bdukes, line 27:
```suggestion
        int PortalId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
bdukes: I guess this has already been released, so this would end up being a breaking change 😞

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
bdukes: @mitchelsellers but at least going forward, we can standardize on `Id` over `ID`, right?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
bdukes: Also `HostFee` and `Currency`

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
ahoefling: If we can't make an exception, I can strategically deprecate all the `ID` suffixes for `Id` and configure everything to flow through the `Id` value. Kind of a pain to implement and maintain, but we should do our best to follow our deprecation rules 😟

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
ahoefling: Right now the running list is:
- `PaymentProcessor`
- `ProcessorPassword`
- `ProcessorUserId`
- `HostFee`
- `Currency`

Are there any other fields that we are missing?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Yup!  Adding this to the other commentary for final inclusion

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Agreed!

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Totally agree, there are other API's for this

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Agreed

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Agreed

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
mitchelsellers: Final list - 

*  BackgroundFile 
*  BannerAdvertising 
*  PaymentProcessor
*  ProcessorPassword
*  ProcessorUserId
*  HostFee
*  Currency
*  Users
*  Version
*  Pages

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
mitchelsellers: @ahoefling That would be great, but otherwise I think we make note for 10.0.0

david-poindexter:
@mitchelsellers I am in agreement here.

david-poindexter:
> **Nullable Values**
> Not to possibly throw a truly crazy breaking change in here, but I think it should be discussed. DNN Platform, up to this point has been a huge fan of magic numbers. Things such as `Null.NullInteger` and `Null.NullDate` rather than supporting actually nullables. SHOULD we get away from this practice with these new APIs.

My vote would be yes, to get away from this poor practice.

mitchelsellers:
@ahoefling Let me know if you want to chat about the Nullables.....hate to throw crazy your way, but if we can make this "right" we can avoid breaking changes

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
sleupold: I would keep "Currency", because this might be used by numerous 3rd party extensions

sleupold:
AFAIR, DNN used its dedicated null values, as .Net 1 didn't include nullable types.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
bdukes: :+1: for adding `PortalAliasId` and deprecating `PortalAliasID`

bdukes:
👍🏻 for using nullable values (though it may be tricky to track down which fields are or are not nullable)

ahoefling:
Thanks everyone for the comments and concerns on this PR. There is a lot to unpack with all the additional comments

## Naming Convention

### Id vs ID for properties.
@mitchelsellers 
>I think for these new API's we should, as @bdukes recommended adopt a naming convention of Id rather than ID to match the current standards.

@bdukes 
> what do you think about enforcing a consistent naming convention for Id over ID in these new interfaces?

It is considered best practices to use `Id` instead of `ID` which is a more strict CamalCasing. 

✅ I agree

### Strict CamalCasing
@mitchelsellers 
>In addition to this, the HTTPAlias item then becomes a question, should this be HttpAlias? I believe yes.

As just mentioned it is considered a best practice.

✅ I agree

## Removing Unused Payment Properties
@mitchelsellers @bdukes and I have been discussing the removal of several payment gateway fields that are no longer used. We shouldn't bring these over to the new interface. @mitchelsellers put together this final list which will be removed from the new `IPortalInfo` interface. 

- BackgroundFile
- BannerAdvertising
- PaymentProcessor
- ProcessorPassword
- ProcessorUserId
- HostFee
- Currency
- Users
- Version
- Pages

✅ I agree

## Nullable Fields
@mitchelsellers 
> DNN Platform, up to this point has been a huge fan of magic numbers. Things such as Null.NullInteger and Null.NullDate rather than supporting actually nullables. SHOULD we get away from this practice with these new APIs.

The `Null.NullInteger` is an anti-pattern in how it leverages in magic numbers, this ultimately should be removed from DNN. I do not think this is the time and place to do it. We should create a separate work item to put on the backlog that captures this. As @sleupold and @bdukes it will be difficult to get this right. Especially ensuring that the database is updated correctly.

❌ I disagree, this should be created as a new work item. It goes **way** out of scope of this. (Still a great discussion to have)

## API Deprecation
@mitchelsellers 
> Functionality Removed prior to 9.7.x, API will be removed during the 10.00.00 release

I have no problem with removing deprecated fields at the 10.0 release instead of the 11.0 release. It actually makes things much easier for us to innovate and move the platform forward. In my opinion the 2 major release deprecation notice is kind of excessive, many OSS projects deprecate and remove APIs at their next major version release.

❓ I am not sure what you would like me to do with this. Should I be updating all active deprecation notices that are < 9.7.x (which is everything at this moment since this PR is targeting 9.7.2) to list they will be removed at v10.0?.
```
[Obsolete("Deprecated in 9.7.2. Scheduled for removal in v10.0.0, use IDictionary<string, IPortalAliasInfo> instead")]
public class PortalAliasCollection
```

## Next Steps
I have 1 major question about API deprecation to be clarified. Other than that, anything that I marked with a ✅ I will be implementing in the PR. Anything I marked with a ❌ I believe is out of scope.

mitchelsellers:
Regarding deprecating methods. Let’s keep the scope here simple and just flag the fields removed as Obsolete for removal later. 

I agree on all others. I would love to discuss the Nullable stuff at a near future time

bdukes:
Let's just remove those properties from `IPortalInfo`, we can talk about deprecating them on `PortalInfo` separately.

Agreed to handle nullable types separately.

ahoefling:
I have updated the PR with the requested changes that we have been discussing. Since there are a lot of conversations going on in separate threads I am asking everyone that has reviewed part of this PR to re-review the entire PR. If you feel like something you commented on got resolved and shouldn't be, please just start a new review and we can discuss it.

GitHub wouldn't let me request a re-review of everyone but I would love to get feedback and code reviews from everyone on this thread.
- @mitchelsellers 
- @bdukes 
- @valadas 
- @david-poindexter 
- @sleupold

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalInfo.cs
ahoefling: I have removed `Currency` as requested in our discussion, but I am leaving this thread open so there could be more discussion. I think marking `Currency` as deprecated will give developers enough time to migrate their code to use `PortalSettings`.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasService.cs
ahoefling: @bdukes @mitchelsellers is there any actionable item here that I need to take on the PR?

sleupold: shouldn't the method be named GetPortalAliasByPortalAliasId instead of GetPortalAliasByPortalAliasID?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
```diff
@@ -1,27 +1,54 @@
...
-        string HTTPAlias { get; set; }
+        /// <summary>
+        /// Gets or sets the http alias.
+        /// </summary>
+        string HttpAlias { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Portal Alias ID.
+        /// </summary>
+        int PortalAliasId { get; set; }
```
bdukes, line 22: @mitchelsellers this rename from `PortalAliasID` to `PortalAliasId` it technically breaking.  Is this one we can let slide b/c it's new enough, or do we need to have a deprecated `PortalAliasID` property on the interface also?

DNN Platform/Library/Entities/Portals/PortalAliasInfo.cs
```diff
@@ -10,58 +10,102 @@ namespace DotNetNuke.Entities.Portals
...
+
+        /// <inheritdoc />
+        public string HttpAlias { get; set; }
+
+        [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.IPortalAliasInfo.HttpAlias instead.")]
+        public string HTTPAlias
+        {
+            get => this.HttpAlias;
+            set => this.HttpAlias = value;
+        }
```
bdukes, line 48: :+1:

bdukes (approved):
Looks great @ahoefling!

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalAliasInfo.cs
mitchelsellers: I think we just leave it out

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
```diff
@@ -1,68 +1,76 @@
...
-        /// Adds an entry to the collection.
-        /// </summary>
-        public void Add(string key, PortalAliasInfo value)
-        {
-            this.Dictionary.Add(key, value);
-        }
-    }
-}
+    [Serializable]
+    [Obsolete("Deprecated in 9.7.2. Scheduled for removal in v11.0.0, use DotNetNuke.Abstractions.Portals.PortalAliasCollection instead.")]
```
mitchelsellers, line 11: Since we are not using the custom collection going forward, this Obsolete notice is not quite right as it just shouldn't be used going forward.

mitchelsellers (changes requested):
Just one thing, I noted a single instance, but there are a few Obsolete comments that need updating, otherwise looks good.

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
ahoefling: This is a great catch! I will go through all the obsolete notices that we added and fix them.

ahoefling:
When we are extracting interfaces should we be allowing gets/sets on all of these properties? Are we missing properties that should be readonly?

sleupold:
@ahoefling 
for calculated properties, of course.
for all other properties, which shouldn't be modified this would mean, you'd include these properties in the constructor?

bdukes:
I could see the primary key being read-only, but if it's settable on the underlying object, I'm not inclined to worry too much about it.

sleupold:
DNN uses a number of surrogate keys, like PortalSettings,PortalSettingsID, where all fields of the primary key (PortalId, CultureCode, SettingName) need to be preserved and should be readonly, to keep integrity.

DNN Platform/Library/Entities/Portals/PortalAliasCollection.cs
ahoefling: I have reviewed all the deprecation notices in this PR and updated them to match changes that we have iterated through in the PR. I am going to resolve this conversation, open a new one if there are others that need updates.

ahoefling:
I have made the requested changes by @mitchelsellers and went through all of the deprecation notices to certify they are correct to the best of my knowledge. Also, I have rebased this branch on the latest changes to `develop` and believe it is ready to be merged. 

If there is any other outstanding work that needs to be done, let me know and I'm happy to make the changes 👍

mitchelsellers (approved):

bdukes:
:tada: Thanks @ahoefling!

@mitchelsellers any concerns with merging this for 9.7.2?  It's ready to merge as far as I'm concerned.

ahoefling:
If we push this to 9.8.0 we will need to update all the deprecation notices, not a big deal but don't want to lose sight of it.

mitchelsellers:
No concerns here merging

Timo-Breumelhof:
> In addition to this, the `HTTPAlias` item then becomes a question, should this be `HttpAlias`? I believe yes.

Well, I just noticed that any VB code that uses HTTPAlias now crashes and there's no real way around this.

bdukes:
@Timo-Breumelhof what version of DNN are you compiling against?

Timo-Breumelhof:
@bdukes well, I once decided it would be better to not compile so one could include the SKO inside a skin package. So it's like many Skin Objects just an ascx and an ascx.vb file. This will change when I finally convert this to c# :-)

PR #4041, opened 2020-09-03
Fixed an issue with user management in portal groups

valadas:
- Fixed an issue with user management in portal groups
- Fixed many of the build warnings, there are a couple left about IPortalSettings vs PortalSettings which I did not fix because I think the correct solution would involve modifying other components so they can take IPortalSettings instead of PortalSettings and well, lack of time on that...

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -735,7 +735,7 @@ namespace DotNetNuke.Entities.Users
         }
 
         /// <summary>
-        /// Gets a user by it's vanity url.
+        /// Gets a user by its vanity url.
         /// </summary>
         /// <param name="portalId">The site (portal) id of the user.</param>
         /// <param name="vanityUrl">The vanity url to search.</param>
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -1211,22 +1211,31 @@ namespace DotNetNuke.Entities.Users
             int portalId = user.PortalID;
             user.PortalID = GetEffectivePortalId(portalId);
 
-            // Remove the User
             var retValue = MembershipProvider.Instance().RemoveUser(user);
 
             if (retValue)
             {
-                // Obtain PortalSettings from Current Context
                 var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
 
-                // Log event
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -649,35 +647,34 @@ public static string GetPassword(ref UserInfo user, string passwordAnswer)
             return user.Membership.Password;
         }
 
+        /// <summary>
+        /// Gets the unauthorized users for a portal.
+        /// </summary>
+        /// <param name="portalId">The site (portal) id on which to search for Unauthorized users.</param>
+        /// <param name="includeDeleted">A value indiciting whether to include deleted users.</param>
```
bdukes, line 654:
```suggestion
        /// <param name="includeDeleted">A value indicating whether to include deleted users.</param>
```

DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -716,70 +713,68 @@ public static UserInfo GetUserById(int portalId, int userId)
...
-        /// <returns>The User as a UserInfo object.</returns>
-        /// -----------------------------------------------------------------------------
+        /// <returns>The User as a <see cref="UserInfo"/> object.</returns>
         public static UserInfo GetUserByName(int portalId, string username)
         {
             return GetCachedUser(portalId, username);
         }
 
+        /// <summary>
+        /// Gets a user by it's vanity url.
```
bdukes, line 738:
```suggestion
        /// Gets a user by its vanity url.
```

DNN Platform/Library/Entities/Users/UserController.cs
```diff
@@ -716,70 +713,68 @@ public static UserInfo GetUserById(int portalId, int userId)
...
+        /// <param name="vanityUrl">The vanity url to search.</param>
+        /// <returns>A user as a <see cref="UserInfo"/> object.</returns>
         public static UserInfo GetUserByVanityUrl(int portalId, string vanityUrl)
         {
             portalId = GetEffectivePortalId(portalId);
             return MembershipProvider.Instance().GetUserByVanityUrl(portalId, vanityUrl);
         }
 
+        /// <summary>
+        /// Gets a single user by it's password reset token.
```
bdukes, line 750:
```suggestion
        /// Gets a single user by its password reset token.
```

bdukes (approved):
Thanks!

david-poindexter (approved):
Thanks @valadas!

valadas:
Thanks for the typo fixes 👍

mitchelsellers (approved):

PR #4048, opened 2020-09-03
Fixed bug with hierarchical vocabularies

### LINKED ISSUE #4047: Bug when adding terms to a hierarchical vocabulary
## Description of bug
Adding a term to a hierarchical vocabulary can lead to a broken hierarchy.

## Steps to reproduce
1. Open `SETTINGS` > `Vocabularies`
2. Click on `Create New Vocabulary`
3. Set `Name` to "Hierarchical" and `Type` to "Hierarchy" and click `Create`
4. Expand the newly created vocabulary and create the following hierarchy of terms:
```
Term 1
 + Term 11
 + Term 12
    + Term 121
```
5. Create a new `Term 111` below `Term 11`
6. Expand the hierarchy

## Current behavior
`Term 111` is missing. Refreshing the page makes it appear though.

## Expected behavior
Inserted term should be displayed below `Term 11`.

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.07.01 release candidate
* [X] 09.07.00 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

daguiler:
Fixes #4047

## Summary
Several problems in this `findInChildren` function.
1) The `forEach` loop cannot be broken, so the `return` statement at line 19 was not actually exiting the function.
2) The initial value of `parentTerm` was an empty object, so if the item was not found, an empty object was being returned by the function. However, empty objects evaluate to `true`, not to `false`.
3) Since this is a recursive function, and because of 2), line 21 was deciding to continue instead of deciding to stop. This was causing the `parentTerm` variable to get incorrectly overwritten in following cycles.

I replaced this buggy function with a more clear implementation that immediately breaks the loop if the item is found, or returns `null` otherwise.

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Vocabularies.Web/src/components/VocabularyList/RightPane/index.jsx
+++ b/Dnn.AdminExperience/ClientSide/Vocabularies.Web/src/components/VocabularyList/RightPane/index.jsx
@@ -10,20 +10,15 @@ import { GridCell, SvgIcons } from "@dnnsoftware/dnn-react-common";
 
 function findInChildren(list, parentTermId) {
     if (!list) {
-        return;
+        return null;
     }
-    let parentTerm = {};
-    list.forEach((_term) => {
-        if (_term.TermId === parentTermId) {
-            parentTerm = _term;
-            return parentTerm;
-        } else {
-            if (findInChildren(_term.ChildTerms, parentTermId)) {
-                parentTerm = findInChildren(_term.ChildTerms, parentTermId);
-                return parentTerm;
-            }
-        }
-    });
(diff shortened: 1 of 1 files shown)


bdukes:
Your description states that your new function "immediately breaks the loop if the item is found", but I don't see an immediate return.  Other than adding a return, this looks good to me. Thanks!

daguiler:
Hi @bdukes, the break is in the `for` condition.

bdukes (approved):

valadas (approved):
Looks good to me

PR #4053, opened 2020-09-04
Use request scope in PortalModuleBase

### LINKED ISSUE #4046: Use request scope in PortalModuleBase
## Description of problem
In the class `PortalModuleBase` it is possible to resolve services through the property `DependencyProvider`. 

In PR https://github.com/dnnsoftware/Dnn.Platform/pull/3930 the implementation changed which causes every control to have its own scope:

https://github.com/dnnsoftware/Dnn.Platform/blob/8bb34af67f456699cfba62a150f77b1055edec64/DNN%20Platform/Library/Entities/Modules/PortalModuleBase.cs#L46

Which is an improvement of the previous implementation (using the global scope) however it comes with a drawback:

Let's say I've registered a scoped service that uses a database connection and stores cache in a field.

- When I resolve the service in control A, the instance is created and the data is cached.
- When I resolve the same service in control B, it's creating a new instance because it's a different scope. The service retrieves the same data again.

## Description of solution
In https://github.com/dnnsoftware/Dnn.Platform/pull/3520 @dimarobert created `HttpContext.Current.GetScope()`. I would suggest to use the same scope in `PortalModuleBase` and if the HTTP Context doesn't exists, then it would create a new scope.

## Additional context
In ASP.Net Core, scoped services are created per request (see https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection#request-services). So I assumed this was also the case in DotNetNuke.

## Workaround
It's possible to use `Context.GetScope().ServiceProvider` instead of `DependencyProvider` to use the request scope.

GerardSmit:
Fixes #4046

## Summary
Currently every PortalModule create its own service scope. This PR changes this behaviour by using the request scope when possible.

If a PortalModule is created outside ASP.Net (or the HttpContext was not available) it'll still create a new scope.

I've placed the unit test in the same directory as `DnnDependencyResolverTests` since I was not sure where to place it. If this is incorrect, place a comment and I'll move it 😄.

### DIFF
--- a/DNN Platform/Library/Common/ServiceScopeContainer.cs
+++ b/DNN Platform/Library/Common/ServiceScopeContainer.cs
@@ -32,7 +32,7 @@ namespace DotNetNuke.Common
         /// </summary>
         public static ServiceScopeContainer GetRequestOrCreateScope()
         {
-            var requestScope = HttpContext.Current?.GetScope();
+            var requestScope = HttpContextSource.Current?.GetScope();
 
             if (requestScope != null)
             {
--- a/DNN Platform/Library/Common/ServiceScopeContainer.cs
+++ b/DNN Platform/Library/Common/ServiceScopeContainer.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Common
+{
+    using System;
+    using System.Web;
+    using DotNetNuke.Common.Extensions;
+    using Microsoft.Extensions.DependencyInjection;
+
+    internal struct ServiceScopeContainer : IDisposable
+    {
+        public ServiceScopeContainer(IServiceScope serviceScope, bool shouldDispose)
+        {
+            ServiceScope = serviceScope;
+            ShouldDispose = shouldDispose;
+        }
+
+        /// <summary>
+        ///     Service scope that the container holds.
+        /// </summary>
+        public IServiceScope ServiceScope { get; }
+
+        /// <summary>
+        ///     True if the <see cref="ServiceScope"/> should be disposed.
+        /// </summary>
+        public bool ShouldDispose { get; }
+
+        /// <summary>
+        ///     Get the service scope from <see cref="HttpContext"/> or create a new scope from <see cref="Globals.DependencyProvider"/> when the scope doesn't exists.
+        /// </summary>
+        public static ServiceScopeContainer GetRequestOrCreateScope()
+        {
+            var requestScope = HttpContext.Current?.GetScope();
+
+            if (requestScope != null)
+            {
+                return new ServiceScopeContainer(requestScope, false);
+            }
+
+            return new ServiceScopeContainer(Globals.DependencyProvider.CreateScope(), true);
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            if (ShouldDispose)
+            {
+                ServiceScope.Dispose();
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


DNN Platform/Library/Common/ServiceScopeContainer.cs
```diff
@@ -0,0 +1,54 @@
...
+        ///     True if the <see cref="ServiceScope"/> should be disposed.
+        /// </summary>
+        public bool ShouldDispose { get; }
+
+        /// <summary>
+        ///     Get the service scope from <see cref="HttpContext"/> or create a new scope from <see cref="Globals.DependencyProvider"/> when the scope doesn't exists.
+        /// </summary>
+        public static ServiceScopeContainer GetRequestOrCreateScope()
+        {
+            var requestScope = HttpContext.Current?.GetScope();
```
bdukes, line 35: Can you switch this to use [`HttpContextSource`](https://dnndocs.com/api/DotNetNuke.Common.HttpContextSource.html#DotNetNuke_Common_HttpContextSource_Current) instead of directly accessing `HttpContext`?  It'll also make your test simpler.  Thanks!
```suggestion
            var requestScope = HttpContextSource.Current?.GetScope();
```

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/Internals/ServiceContainerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        [TestFixtureTearDown]
+        public void FixtureTearDown()
+        {
+            Globals.DependencyProvider = null;
+        }
+
+        [Test]
+        public void CreateScopeWhenRequestDoesNotExists()
+        {
+            Assert.Null(HttpContext.Current?.GetScope());
```
bdukes, line 40:
```suggestion
            Assert.Null(HttpContextSource.Current?.GetScope());
```

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/Internals/ServiceContainerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+                HttpContext.Current.SetScope(scope);
+
+                Assert.NotNull(HttpContext.Current?.GetScope());
+
+                var container = ServiceScopeContainer.GetRequestOrCreateScope();
+
+                Assert.False(container.ShouldDispose);
+                Assert.AreEqual(scope, container.ServiceScope);
+            }
```
bdukes, line 64:
```suggestion
            var scope = Globals.DependencyProvider.CreateScope();

            HttpContextHelper.RegisterMockHttpContext();
            HttpContextSource.Current.SetScope(scope);

            Assert.NotNull(HttpContextSource.Current?.GetScope());

            var container = ServiceScopeContainer.GetRequestOrCreateScope();

            Assert.False(container.ShouldDispose);
            Assert.AreEqual(scope, container.ServiceScope);
```

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/Internals/ServiceContainerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Tests.Web.Api.Internals
+{
+    using System.Web;
+    using DotNetNuke.Abstractions;
+    using DotNetNuke.Abstractions.Application;
+    using DotNetNuke.Common;
+    using DotNetNuke.Common.Extensions;
+    using DotNetNuke.Tests.Instance.Utilities.HttpSimulator;
```
bdukes, line 12:
```suggestion
    using DotNetNuke.Tests.Utilities;
```

bdukes (approved):
This is great, thanks!

mitchelsellers (approved):

PR #4070, opened 2020-09-09
Issue 4038

### LINKED ISSUE #4038: Suppress Edit Bar
## Description of problem
It's easy to swap out the Control Panel by updating the Host Settings; however, there is no way to change the Edit Bar experience. 

Currently, the Edit Bar listens for SkinEvents via an HttpModule and automatically injects itself for each page load (even for anonymous users): https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/HttpModules/EditBarModule.cs#L62

The Edit Bar (UserControl) then removes itself it it's not need (not in Edit Mode): https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/Controllers/ContentEditorManager.cs#L120

This is a performance issue which should also be resolved. The Edit Bar does not get injected if the Application.SKU != "DNN"; however, that is a read-only property that cannot be changed so ultimately there's no way to prevent it from appearing on the page. 

## Description of solution
Proposed solution is to change the follow line: https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/HttpModules/EditBarModule.cs#L62

```
if (DotNetNukeContext.Current.Application.SKU != "DNN")
{
     return;
}
```

with: 

```
if (HostSettings["EnableEditBar"] != "TRUE") //Psuedo Code
{
     return;
}
```

## Additional context
We can also introduce a corresponding checkbox in Persona Bar with a note such as "Default Edit Bar is disabled and content editing is managed by a 3rd party component" 

Please suggest the best location in Persona Bar to add this setting. 

I'll be happy to contribute a PR for this issue. Let me know when we have consensus and I'll get started.

meetmandeep:
Closes #4038 

## To Disable Edit Bar
`insert into HostSettings values ('DisableEditBar', 'True',0,-1,GETDATE(),-1,GETDATE())`

Or (if setting already exists)

```
update HostSettings
set SettingValue = 'True'
where SettingName = 'DisableEditBar'
```

## To Enable Edit Bar
```
update HostSettings
set SettingValue = 'False'
where SettingName = 'DisableEditBar'
```

Or

delete from HostSettings where SettingName = 'DisableEditBar'

## Disabled View
If edit bar is disabled, the tooltip reads "Edit Bar is disabled." and help text reads "This DNN installation is running an alternative edit bar."

I couldn't find a way to swap the icon image with another way. Let me know if anyone has any pointers. 

image


mikesmeltzer:
Added some feedback to issue #4038.

david-poindexter (approved):
Looks okay to me - thanks @meetmandeep

Dnn.AdminExperience/Library/Dnn.PersonaBar.Library/Containers/PersonaBarContainer.cs
```diff
@@ -111,6 +111,7 @@ private static string GetBeaconUrl()
             settings.Add("debugMode", HttpContext.Current != null && HttpContext.Current.IsDebuggingEnabled);
             settings.Add("portalId", portalId);
             settings.Add("preferredTimeZone", preferredTimeZone);
+
```
bdukes, line 114: Let's remove this extra blank line
```suggestion
```

bdukes (approved):
LGTM, thanks!

mitchelsellers (approved):

PR #4071, opened 2020-09-09
Add IPortalSettingsManager for Abstractions Project

SkyeHoefling:
Related to #3985 

## Summary
Adds a new interface to Dependency Injection `DotNetNuke.Abstractions.Portals.IPortalSettingsManager` which is used for retrieving the current portal settings. The API structure in this PR has been simplified to use modern techniques in .NET such as generics and indexers. It still maintains the main functionality 
- Get
- Update
- Update Encrypted
- Delete

## Usage
This PR takes a few different interfaces and combines their implementations to work together. It will be consumed via dependency injection using the `IPortalSettingsManager`. 
```c#
var portalSettingsManager = Globals.DependencyProvider.GetRequiredService<IPortalSettingsManager>();
bool mySetting = portalSettingsManager
    .GetPortalSettings(portalId, cultureCode)
    .Get<bool>("My true/false portal setting");
```
The example above uses the service locator pattern, it is always recommended to use constructor injection where possible.

## Missing Items
Currently this PR is missing a few parts of the implementation. I submitted it early to get feedback prior on the general direction. Below is a list of missing features that I plan to implement after the initial review.

- `IConfigurationSetting` provides a Key-Value-Pair object which is used throughout DNN
- Get & Update dictionary. The old implementation allowed the user to get the entire dictionary or update the entire dictionary at once. Currently those APIs are not implemented

## Required to merge
This PR was submitted prior to working on Unit Tests or validating the build works. I am doing this to get early feedback on architectural decisions being made with the changes proposed here.

| Task | Status |
|------|--------|
| Unit Tests | ❌ Not Started |
| Test Installer | ❌ Not Started |


DNN Platform/Library/Entities/Portals/PortalSettingsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                cultureCode = PortalController.GetActivePortalLanguage(portalId);
+            }
+
+            var cachedSettings = this.GetCachedSettings(portalId, cultureCode);
+            var saveService = new PortalSaveSettingsService(portalId, cultureCode);
+            var deleteService = new PortalDeleteSettingsService(portalId, cultureCode);
+
+            return new PortalSettingsService(() => this.GetCachedSettings(portalId, cultureCode), saveService, deleteService);
+        }
```
ahoefling, line 38: This may be one of the focal points of this code review and I would like to highlight the change as well as my reasoning behind the change. 

I wanted to implement the `PortalSettingsService` as a very simple wrapper around an `IDictionary<string, string>`. Since DNN has static API calls everywhere I didn't want the `PortalSettingsService` to be tightly coupled to anything. In theory this implementation could live in a .NET Standard 2.0 project and doesn't need to exist in the `DotNetNuke.Library` project. To do this I needed to create an `ISaveSettingsService` and a `IDeleteSettingsService` The implementations of this interfaces are passed in at instantiation time. This allows the wrapper (`PortalSettingsService`) to easily invoke the functionality. 

The other item here is the callback function that retrieves the latest settings `GetCacheSettings(int, int)`. This is useful in the `PortalSettingsService` to reset the local `IDictionary<string, string>` whenever the cache is cleared. Due to the de-coupled structure we need a way for it to easily reload the data. Since the `PortalSettingsService` doesn't have direct knowledge how to construct the `IDictionary<string, string>` we are using the callback that stores the reference on how to load the data.

DNN Platform/DotNetNuke.Abstractions/Settings/ISaveSettingsService.cs
```diff
@@ -0,0 +1,45 @@
...
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Abstractions.Settings
+{
+    /// <summary>
+    /// The Save Settings service is used by implementations of
+    /// <see cref="ISettingsService"/>. It is not currently designed
+    /// to be injected via Dependency Injection. Please use the
+    /// <see cref="ISettingsService"/>.
+    /// </summary>
+    public interface ISaveSettingsService
```
bdukes, line 12: Should these interfaces be `internal` if they're not intended to be used via DI?

DNN Platform/DotNetNuke.Abstractions/Settings/ISettingsService.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// </summary>
+        /// <param name="clearCache">If true the cache will be cleared.</param>
+        void DeleteAll(bool clearCache);
+
+        /// <summary>
+        /// Adds or updates the setting for a specified key.
+        /// </summary>
+        /// <param name="key">The setting key to update.</param>
+        /// <param name="value">The value to set.</param>
+        void Update(string key, string value);
```
bdukes, line 102: Do we want generic Update APIs to match the generic Get APIs?

DNN Platform/Library/Entities/Portals/PortalDeleteSettingsService.cs
```diff
@@ -0,0 +1,79 @@
...
+                key + ((this.cultureCode == Null.NullString) ? string.Empty : " (" + this.cultureCode + ")"),
+                this.GetPortalSettings(),
+                this.userId,
+                EventLogController.EventLogType.PORTAL_SETTING_DELETED);
+            DataCache.ClearHostCache(clearCache);
+        }
+
+        /// <inheritdoc />
+        public void DeleteAll() =>
+            this.DeleteAll(false);
```
bdukes, line 54: I would expect `clearCache` to default to `true`

DNN Platform/Library/Entities/Portals/PortalSaveSettingsService.cs
```diff
@@ -0,0 +1,95 @@
...
+            Requires.NotNullOrEmpty("passPhrase", passPhrase);
+
+            var encryptedText = FIPSCompliant.EncryptAES(value, passPhrase, Host.Host.GUID);
+            this.Update(key, encryptedText, true, clearCache);
+        }
+
+        /// <inheritdoc />
+        public void UpdateEncrypted(string key, string value, bool clearCache)
+        {
+            this.Update(key, value, true, clearCache);
```
bdukes, line 58: Where does the encryption happen when calling this overload?

DNN Platform/Library/Entities/Portals/PortalSettingsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <inheritdoc />
+        public ISettingsService GetPortalSettings(int portalId, string cultureCode)
+        {
+            if (string.IsNullOrEmpty(cultureCode) && portalId > -1)
+            {
+                cultureCode = PortalController.GetActivePortalLanguage(portalId);
+            }
+
+            var cachedSettings = this.GetCachedSettings(portalId, cultureCode);
```
bdukes, line 33: This instance isn't getting used, right?
```suggestion
```

DNN Platform/Library/Entities/Portals/PortalSettingsService.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        /// <inheritdoc />
+        public void UpdateEncrypted(string key, string value, string passPhrase, bool clearCache)
+        {
+            this.SaveService.UpdateEncrypted(key, value, passPhrase, clearCache);
+
+            if (clearCache)
+            {
+                this.Settings = this.LoadSettingsCallback();
```
bdukes, line 185: Would it make more sense to set this to `null`, and then it'll only re-query if someone needs to read again?

DNN Platform/Library/Entities/Portals/PortalSettingsService.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+
+        /// <inheritdoc />
+        public void Update(string key, string value) => 
+            this.SaveService.Update(key, value, false);
```
bdukes, line 146: As above, I would expect `clearCache` to default to `true`

DNN Platform/Library/Entities/Portals/PortalSettingsService.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        /// <inheritdoc />
+        public T Get<T>(string key)
+            where T : IConvertible
+        {
+            try
+            {
+                // DNN uses special logic for boolean. It accepts Y and TRUE. We should make sure that is convertible
+                return (T)Convert.ChangeType(this[key], typeof(T));
```
bdukes, line 72: This is where I would like to see the logic from `SerializationController` reused (which will handle the special Boolean logic, for example).

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -414,21 +426,13 @@ public static string GetEncryptedString(string settingName, int portalID, string
...
-            catch (Exception exc)
-            {
-                Logger.Error(exc);
-            }
-
-            return retValue;
+            var portalSettingsManager = Globals.DependencyProvider.GetRequiredService<IPortalSettingsManager>();
+            return portalSettingsManager
+                .GetPortalSettings(portalID)
+                .Get<string>(settingName) ?? defaultValue;
```
bdukes, line 435: The previous logic is to return the `defaultValue` when the value is an empty string, is this an intentional change to only return `defaultValue` when the value is `null`?

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -2340,42 +2285,6 @@ private static void ParsePortalDesktopModules(XPathNavigator nav, int portalID)
             }
         }
 
-        private static void UpdatePortalSettingInternal(int portalID, string settingName, string settingValue, bool clearCache, string cultureCode, bool isSecure)
-        {
-            string currentSetting = GetPortalSetting(settingName, portalID, string.Empty, cultureCode);
-
-            if (currentSetting != settingValue)
```
bdukes, line 2347: Is this check for the new value being equal to the old value still in the new implementation?  I don't remember seeing something like that.

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -2340,42 +2285,6 @@ private static void ParsePortalDesktopModules(XPathNavigator nav, int portalID)
...
-                    var httpContext = HttpContext.Current;
-                    if (httpContext != null)
-                    {
-                        var cultureCodeForKey = GetActivePortalLanguageFromHttpContext(httpContext, portalID);
-                        var dictionaryKey = string.Format(HttpContextKeyPortalSettingsDictionary, portalID, cultureCodeForKey);
-                        httpContext.Items[dictionaryKey] = null;
-                    }
-                }
-
-                EventManager.Instance.OnPortalSettingUpdated(new PortalSettingUpdatedEventArgs
```
bdukes, line 2370: Is this event still being called?

bdukes (commented):
Thanks @ahoefling!

DNN Platform/DotNetNuke.Abstractions/Settings/ISaveSettingsService.cs
ahoefling: This is a really great question!

In my opinion these interfaces should remain public if we want to support making the `PortalSettingsService` extensible. If a developer wants to swap it out for their own implementation they should be able to use the following code.

```c#
public class CustomPortalSettingsManager : PortalSettingsManager { }
public class CustomPortalSaveSettingsService : PortalSaveSettingsService { }
public class CustomPortalDeleteSettingsService : PortalDeleteSettingsService { }
```
If the extension is just changing the save/delete rules it doesn't necessarily need to create another class for the `PortalSettingsService`.This approach gives us the highest degree of extensibility, but also makes it more complicated.

For many modules and implementations this will never be used. I see this being valuable for something like Evoq or custom installations that want to execute special logic during the save or delete sequence.

DNN Platform/DotNetNuke.Abstractions/Settings/ISettingsService.cs
ahoefling: Should the update API support full C# objects that get serialized into json or should it just support primitives such as
- string
- int
- double
- float
- long
- short
- bool

bdukes: [The `SerializationController` exposes an overload that takes a custom serializer](https://github.com/dnnsoftware/Dnn.Platform/blob/80f8a4dcb57ca63d2b4a7fedc01c7be000e500f9/DNN%20Platform/Library/Entities/Modules/Settings/SerializationController.cs#L20). otherwise it [checks for `DateTime` and `TimeSpan` and finally falls back to `Convert.ToString`](https://github.com/dnnsoftware/Dnn.Platform/blob/80f8a4dcb57ca63d2b4a7fedc01c7be000e500f9/DNN%20Platform/Library/Entities/Modules/Settings/SerializationController.cs#L131-L146).

ahoefling: WOW! That sounds really powerful, I will certainly swap out this implementation for that. Should we migrate that to the abstractions project as part of this PR. That sounds like something really useful to be available via Dependency Injection

bdukes: If you're up for it, that would make sense to me.  We can also do that as a separate PR if it makes sense to reduce the scope of this one.

ahoefling: I'll submit a new PR for the `SerializationController`

ahoefling: Summited PR #4087

ahoefling:
Since #4087 and #4159 have been merged this PR can be updated to reflect the latest and greatest code changes. Here is my list of things to do
- Update eventlogger to use `IEventLogger` via DI
- Update serialization to use DI
- Update PR based on review feedback

I am not 100% sure, but I think I am going to flatten the structure here as the PR is as written is kind of complicated and unnecessary

mitchelsellers:
Thanks for this contribution, given the time and other shifts these proposed changes have been noted, but a new pull request will be created if necessary to merge elements into the new 10.0 release plan.

PR #4076, opened 2020-09-10
Fixed an issue that prevents 9.7.0 to 9.7.1 upgrades

### LINKED ISSUE #4075: 9.7.0 upgrade script does not fire when going from before 9.7.0 to 9.7.1
## Description of bug
When upgrading from earlier than 9.7.0 to 9.7.1, the config merge script for 9.7.0 does not fire. This is probably caused by the lack of a 09.07.00.SqlDataProvider because that version did not have sql changes.

### If you are doing a clean 9.7.1 install
No workaround needed, the normal install works fine.

### If you are upgrading from 9.7.0 to 9.7.1
Nothing special is needed, this upgrade path will work.

### If you are upgrading from any version before 9.7.0
- Unzip the upgrade zip and add an empty text file named `09.07.00.SqlDataProvider` in the `Providers\DataProviders\SqlDataProvider` folder.
- Continue the upgrade as usual

### If you have already upgraded to 9.7.1 before encountering this issue
- Find the web.config file that is located in the root folder of the installation.
- Create a copy of this file as a backup just in case.
- Open the original web.config file that is in the root files of the installation in any text editor.
- In the `<sectionGroup name="dotnetnuke">` section add the following line:
```xml
<section name="tokens" requirePermission="false" type="DotNetNuke.Framework.Providers.ProviderConfigurationHandler, DotNetNuke" />
```
- In the `<dotnetnuke>` section, add the following lines :
```xml
<tokens defaultProvider="CoreTokenProvider">
  <providers>
    <clear />
    <add name="CoreTokenProvider" type="DotNetNuke.Services.Tokens.CoreTokenProvider, DotNetNuke" />
  </providers>
</tokens>
```

### Additional information
Token replace providers are a recent feature (Dnn 9.7.0) and is probably not yet used by many 3rd party modules. But if you do see other configured providers for tokens, you may want to consult with the 3rd party module vendor for instructions on how to best merge this config with theirs.

## Affected version

* [ ] 10.00.00 alpha build
* [x] 09.07.01 release candidate
* [ ] 09.07.00 latest supported release

valadas:
Closes #4075
- Adds missing SqlDataProvider files

### DIFF
--- a/DNN Platform/Website/DotNetNuke.Website.csproj
+++ b/DNN Platform/Website/DotNetNuke.Website.csproj
@@ -3343,9 +3343,10 @@
     <Content Include="Providers\DataProviders\SqlDataProvider\09.04.04.SqlDataProvider" />
     <None Include="Providers\DataProviders\SqlDataProvider\09.05.00.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.05.01.SqlDataProvider" />
-    <Content Include="Providers\DataProviders\SqlDataProvider\09.06.01.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.06.02.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.07.01.SqlDataProvider" />
+    <Content Include="Providers\DataProviders\SqlDataProvider\09.07.00.SqlDataProvider" />
+    <Content Include="Providers\DataProviders\SqlDataProvider\09.06.01.SqlDataProvider" />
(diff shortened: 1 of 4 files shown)


david-poindexter (changes requested):
@valadas not that it technically matters, but this should probably be re-ordered by version to increase readability and avoid confusion in the future.  Thoughts?

mitchelsellers (approved):
Approved. Don’t we need a 9.7.2 as well?

valadas:
Adjusted order and added 09.07.02 also for upcoming release @david-poindexter if you can approve please :)

david-poindexter (approved):
Awesome- thanks!

PR #4092, opened 2020-09-16
Fix deprecation message referencing wrong type

bdukes:
## Summary
The deprecation warnings on `IHostController` recommend using a new interface, but that interface's name changed during the course of the PR, so it's now referencing a type that doesn't exist.


david-poindexter (approved):
Looks good - thanks @bdukes 🎉

mitchelsellers (approved):

PR #4098, opened 2020-09-18
Dependency Injection in WebForms

### LINKED ISSUE #4097: Dependency Injection in WebForms
## Description of problem
With the release of .NET Framework 4.7.2, Dependency Injection was introduced for WebForms. Since DNN targets .NET Framework 4.7.2 we can use this feature.

## Description of solution
Register a service provider in `HttpRuntime.WebObjectActivator` to enable Dependency Injection

## Additional context
See https://devblogs.microsoft.com/dotnet/announcing-the-net-framework-4-7-2/#asp-net-dependency-injection-in-webforms for more information about Dependency Injection in .NET Framework 4.7.2.

_This issue was created because I'm making a PR for this enhancement._

GerardSmit:
Fixes #4097

## Summary
This PR implements Dependency Injection in WebForms. @ahoefling tried to implement this feature in https://github.com/dnnsoftware/Dnn.Platform/pull/2774#issuecomment-491825704 but ran in a lot of issues. Because this was only a comment and I haven't found the code maybe they can check if I'm doing the same thing.

~~The implementation is inspired by https://github.com/aspnet/AspNetWebFormsDependencyInjection/blob/master/src/UnityAdapter/ContainerServiceProvider.cs, which is a Unity implementation of this feature.~~
As of commit https://github.com/dnnsoftware/Dnn.Platform/pull/4098/commits/4fc3e0955fc00aaed251f613215c847f2ac3f0f3 we now use `Type.GetConstructors`. For more information, read the thread https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491260232.

From my test, DNN works as expected. I can log in, create pages etc. and the Dependency Injection is working as intended.

## Breaking change
This PR changes the property `HttpRuntime.WebObjectActivator`. It is possible that an extension implemented this feature as well and thus the DNN Service Providers gets overwritten.


ahoefling:
This is a great PR! Thank you @GerardSmit for contributing this as it has been a missing feature in the Dependency Injection support in DNN. When I first contributed the Dependency Injection PR #2774 I remember using this exact same technique with issues on WebForms controls. DNN's module system is heavily dependent on the WebForm controls which is why the site consistently crashed.  You can see the original comments between @bdukes 
 and I here - https://github.com/dnnsoftware/Dnn.Platform/pull/2774#issuecomment-491825704. Since the initial PR a lot has changed on how the container is created and it is not using scopes. This feature may work without issue.

I am excited to give this a try once the build finishes and see what happens

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <returns>New instance of the service type.</returns>
+        private static object DefaultCreateInstance(Type serviceType)
+        {
+            return Activator.CreateInstance(
+                serviceType,
+                BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.CreateInstance,
+                null,
+                null,
+                null);
+        }
```
ahoefling, line 73: This method will cause constant crashes in DNN since it is using the `Activator.CreateInstance`. In the scenarios where this would be called, the DI container does not have a registration for the dependency and it tries to resolve it. That dependency may be relying on Dependency Injection or not have a default constructor. At that point it will throw an exception when the `Activator.CreateInstance` is invoked.

I recommend updating this code to use `ActivatorUtilities.GetServiceOrCreateInstance` which will try to resolve it with the container first. If it does not exist on the container it will resolve all the parameters on the constructor with the container and then use the Activator to instantiate it.

Consider this example
```c#
public class MyClass
{
    public MyClass(IHostSettingsService settings)
    {
    }
}
```

if you try injecting the `MyClass` into a WebForms control without registering it, the code would fall through to your `DefaultCreateInstance` method. Even though the class isn't registered it is depending on the `IHostSettingsService` which is registered. The safest way to handle this is using the `ActivatorUtilities.GetServiceOrCreateInstance()`

DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
```diff
@@ -133,6 +133,7 @@ private void Application_Start(object sender, EventArgs eventArgs)
             Globals.DependencyProvider = dependencyProvider;
             dependencyProvider.SetProvider(DependencyInjectionInitialize.BuildServiceProvider());
             ServiceRequestScopeModule.SetServiceProvider(Globals.DependencyProvider);
+            HttpRuntime.WebObjectActivator = new RequestScopeServiceProvider();
```
ahoefling, line 136: I don't believe this code is correct, unless there is a special web forms technique I am missing. We already have an implementation of the `IServiceProvider` and that is the `LazyServiceProvider`. This code would effectively create 2 references to different `IServiceProvider` implementations. See comments on the `RequesScopeServiceProvider`

```suggestion
            HttpRuntime.WebObjectActivator = dependencyProvider;
```

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <inheritdoc />
+        public object GetService(Type serviceType)
+        {
+            if (this.typesCannotResolve.ContainsKey(serviceType))
+            {
+                return DefaultCreateInstance(serviceType);
+            }
+
+            var provider = HttpContextSource.Current?.GetScope()?.ServiceProvider;
```
ahoefling, line 28: Could you explain the purpose of this class as I don't understand why we need to have 2 separate implementations of the `IServiceProvider` in DNN.

This appears to be a wrapper around the `IServiceProvider` that is stored on the on the HttpContext. If you are looking for more details on how this is set, you can look at the `ServiceRequestScopeModule` which adds this onto the HttpContext. This then allows you to reference the `IServiceProvider` implementation here to resolve dependencies.

My questions is why can't we just use the `LazyServiceProvider` as suggested above?

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (provider == null)
+            {
+                return DefaultCreateInstance(serviceType);
+            }
+
+            var instance = provider.GetService(serviceType);
```
ahoefling, line 35: If we are going to use the `provider`  to resolve dependencies but add a bunch of checks in to create a default instance, we can just simplify this entire method to use the `ActivatorUtlities`

```c#
ActivatorUtilities.GetServiceOrCreateInstance()
```

ahoefling (changes requested):
I love that we are looking at supporting Dependency Injection for WebForms. I have some concerns about this change as it creates a wrapper around the existing `IServiceProvider` that is on the HttpContext. In the DNN core code we added the `LazyServiceProvider` which ultimately is stored into the `Globals.DependencyProvider`. This `IServiceProvider` is the implementation that is used throughout all of DNN. I think it makes the most sense to re-use that object instead of creating an additional wrapper.

In my review, I left comments and questions on particulars of the implementation. Clarification on these would be helpful and if there is a good reason for these decisions I am more than happy to change my review to an approve

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: The body of GetServiceOrCreateInstance is:

```csharp
provider.GetService(type) ?? CreateInstance(provider, type)
```

which is done on line 35 and line 44.

I can merge them together and use `ActivatorUtilities.GetServiceOrCreateInstance` (like the code below), but when the constructor throws a `InvalidOperationException` it'll failback `Activator.CreateInstance`. Is this desirable?

```csharp
try
{
    return ActivatorUtilities.GetServiceOrCreateInstance(provider, serviceType);
}
catch (InvalidOperationException)
{
    return CreateInstance(serviceType);
}
```

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: Every request a new scope is created by the `ServiceRequestScopeModule`. The new `RequestScopeServiceProvider` uses that scope and resolves the services requested by controls.

If we would use the instance of `LazyServiceProvider` here the services are being created in the global scope, not the request scope; causing them never to dispose. 

Additionally, whenever a service could not be found (for instance a HttpModule) it creates a new instance of that service. This is something that the `LazyServiceProvider` doesn't do, which causes ASP.Net to complain.

DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
GerardSmit: See the comment below. The variable `dependencyProvider` is in this case the global scope.

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: > That dependency may be relying on Dependency Injection or not have a default constructor. At that point it will throw an exception when the Activator.CreateInstance is invoked.

The problem is that not only services are being resolved here. Controls and HttpModules are also being resolved through this service provider. Some of them (for example `System.Web.UI.PageHandlerFactory` which I used in the Unit Test) are being resolved but don't have a public constructor (see https://referencesource.microsoft.com/#System.Web/UI/PageHandlerFactory.cs,31).

If we remove `Activator.CreateInstance` and only use `ActivatorUtilities.GetServiceOrCreateInstance` then these types cannot be resolved or created and will throw a `InvalidOperationException`. Which is invalid.

GerardSmit:
Thanks for the review @ahoefling. I've commented to them. If something isn't clear I'll try to use other words. English isn't my best language so bear with me 😄

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
ahoefling: This makes a lot of sense, thank you for clarifying this 👍

DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
ahoefling: This makes a lot a sense, thanks for explaining

ahoefling:
No problem and thanks for letting me know English isn't your first language. I'll do my best to use as much descriptive language as possible. So far your explanations have helped me understand the decisions made in your Pull Request. I will look at the remaining comments later tonight or this weekend

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
mitchelsellers: @GerardSmit I think that the try/catch method is a bit more intuitive to understand the flow

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
```diff
@@ -0,0 +1,75 @@
+﻿namespace DotNetNuke.Web.Common
```
mitchelsellers, line 1: Can we get the .NET FOundation Copyright notice here, you can grab it from other files

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+    using DotNetNuke.Common;
+    using DotNetNuke.Common.Extensions;
+    using Microsoft.Extensions.DependencyInjection;
+
+    /// <summary>
+    /// Service provider implementation for <see cref="HttpRuntime.WebObjectActivator"/>.
+    /// </summary>
+    internal class RequestScopeServiceProvider : IServiceProvider
+    {
+        private const int TypesCannotResolveCacheCap = 100000;
```
mitchelsellers, line 17: Is there a particular reason to limit this?  Just thinking if not necessary we could remove some operations/calcuation etc

mitchelsellers (changes requested):
Overall I think this is good, I would support the small noted change in flow.

Also, from a testing perspective, I would like to throw some simple load at this using a WebForms module on a page etc to help validate.

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: The only reason I can think of is whenever there is a service that is using a generic definition (e.g. `SomeControl<T>`). Every type (e.g. `SomeControl<int>` or `SomeControl<bool>`) will be added to this dictionary.

I've added this check because the Unity implementation (see https://github.com/aspnet/AspNetWebFormsDependencyInjection/blob/master/src/UnityAdapter/ContainerServiceProvider.cs#L71, created by a Microsoft developer as example) did this as well.

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
ahoefling: It is okay to use exception handling as a fallback mechanism, but it is an anti-pattern if this is handling standard control flow. 

Based on the other comment @GerardSmit made (https://github.com/dnnsoftware/Dnn.Platform/pull/4098/files#r491145227) it sounds like this exception will occur quite often when trying to resolve a `PageHandlerFactory`

ahoefling: I still don't like the solution of using the `Activator.CreateInstance` to instantiate **non-public** constructors, but it appears to be standard across the articles and github threads I have researched.

It almost appears as if there was an oversight in the implementation of the `WebObjectActivator`. Maybe we should think of it as more of a `WebObjectAdapter` instead of `Activator`. The recommendation I am reading is we must build a system that resolves the instance if it is on the container and if it isn't we will need to instantiate it using the `Activator.CreateInstance`. This is required due to some of the protected constructors that need to be resolved. I really do wish this was separated out into 2 objects, one to handle the `IServiceProvider` and 1 to handle the activation of these special objects with protected or internal constructors.

Here is the supporting documentation I found
- [Unity Sample for ASP.NET Web Forms](https://github.com/aspnet/AspNetWebFormsDependencyInjection/blob/c23cb5ba8b3ec4544008b498a1d0a6252c90ddfb/src/UnityAdapter/ContainerServiceProvider.cs#L102-L107)
- [StackOverflow: Wiring up simple injector in webforms in net 4.7.2](https://stackoverflow.com/questions/50632627/wiring-up-simple-injector-in-webforms-in-net-4-7-2) 
- [GitHub Thread Discussing WebForms Dependency Injection](https://github.com/Microsoft/dotnet/issues/727)

I really like the code I saw on Stack Overflow and don't think we should just copy-paste it in, but I think it will be useful as a point of reference.
```c#
class SimpleInjectorWebFormsServiceActivator : IServiceProvider
{
    private const BindingFlags flag =
        BindingFlags.Instance | BindingFlags.NonPublic |
        BindingFlags.Public | BindingFlags.CreateInstance;

    private readonly Container container;

    public SimpleInjectorWebFormsServiceActivator(Container container) =>
        this.container = container;

    public object GetService(Type serviceType) =>
        serviceType.GetConstructors().Length > 0
            ? this.container.GetInstance(serviceType)
            : Activator.CreateInstance(serviceType, flag, null, null, null);
}
```

This code is really elegant for several reasons
- It uses the container if there are public constructors
- If it can't be resolved by the container it uses the activator

I propose we update our `GetService` implementation to the following code
```c#

private const BindingFlags activatorFlags =
    BindingFlags.Instance | 
    BindingFlags.NonPublic |
    BindingFlags.Public | 
    BindingFlags.CreateInstance;

public object GetService(Type serviceType) =>
    serviceType.GetConstructors().Length > 0 ?
        ActivatorUtilities.GetServiceOrCreateInstance(this.container, serviceType) :
        Activator.CreateInstance(serviceType, activatorFlags, null, null, null);
```

If we aren't a fan of ternary statements we can update this to use an if statement, doesn't matter to me. 

Using the `GetConstructors()` API we can quickly determine if the `serviceType` has any public constructor ([GetConstructors() Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.type.getconstructors?view=netframework-4.7.2#System_Type_GetConstructors)). If the type has a public constructor using the `ActivatorUtilities` we can ensure there is no exception thrown if the object is not registered with the container. In the case that @GerardSmit pointed out with protected constructors we can use the ternary if logic to use the Activator to instantiate the object.

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
ahoefling: Thank you for giving me a concrete example, it really helped me in understanding the reasoning behind the `Activator.CreateInstance`. Let's finish the conversation in the other thread so we don't have diverging conversations. https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491128632

DNN Platform/DotNetNuke.Web/Common/DotNetNukeHttpApplication.cs
ahoefling: To clarify, no need to accept my suggested change. I think creating a special ServiceProvider as you originally recommended makes the most sense

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
mitchelsellers: This sounds like an even better compromise and should have better performance as no exceptions will be thrown

ahoefling:
We have a few diverging threads here, I put a lot of documentation in my research on https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491128632. 

I agree with the overall approach @GerardSmit is taking here. I have a few critiques that I think will make the code easier for us to maintain.

@GerardSmit if my recommendations don't make sense, let me know and I can post another suggestion in the code review

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
ahoefling: This appears to be used as a safe-guard in case the `RequestScopeServiceProvider` has already used the `Activator`. It tries to auto resolve the instance and not use the container. This is useful in the current implementation because it prevent performance degradation which is caused by the exception handling control flow.

The CacheCap is just a magic number to prevent the container from using up way too much memory. If we try to resolve that many objects that aren't on the container we are doing something wrong. The application shouldn't necessarily crash at that point, but it will go through the entire `GetService()` code on every request even if the object isn't in the container.

I don't believe the cache dictionary is needed, see my comment's above https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491128632

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: Thanks for the research. I've done some testing with the suggestion you made.

I've created a benchmark to test if `GetConstructors()` (aka reflection) is faster than try/catch and by how much.

While making this benchmark I've ran into a issue that it couldn't find a constructor for the interface `ITestService`, which is correct because it's a interface. Because of this, it goes through `Activator.CreateInstance` and it throws an exception:

```
System.MissingMethodException: 'Constructor on type 'ActivatorBenchmark.ActivatorBench+ITestService' not found.'
```

By doing `type.IsInterface || type.GetConstructors().Length > 0` we make sure that interfaces always go through `ActivatorUtilities`.

We cannot resolve enums and structs through the service provider due a bug. This bug is fixed in Microsoft.Extensions.DependencyInjection 5.0-rc2, before this version you'll get a AccessViolationException (see https://github.com/dotnet/runtime/issues/42037). This is why I specifically check if the type is an interface.

Back to benchmarking, the benchmark can be found at: https://gist.github.com/GerardSmit/fee505a7e7cbb25008f7b7d72b591fa8

In the test I've used Microsoft.Extensions.DependencyInjection 2.1.1 (through the dependency of package DotNetNuke.Core). My machine information:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.508 (2004/?/20H1)
Intel Core i7-6700K CPU 4.00GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4220.0), X64 RyuJIT
```

The results for my machine are as following:

```
|            Method |               Type |        Mean |     Error |    StdDev | Ratio |
|------------------ |------------------- |------------:|----------:|----------:|------:|
|          TryCatch |       ITestService |    22.10 ns |  0.158 ns |  0.132 ns |  1.00 |
|                   |                    |             |           |           |       |
| TryCatchWithCache |       ITestService |    37.53 ns |  0.190 ns |  0.169 ns |  1.00 |
|                   |                    |             |           |           |       |
|   ConstructorTest |       ITestService |    26.76 ns |  0.362 ns |  0.321 ns |  1.00 |
|                   |                    |             |           |           |       |
|          TryCatch | PageHandlerFactory | 8,932.25 ns | 26.429 ns | 22.070 ns |  1.00 |
|                   |                    |             |           |           |       |
| TryCatchWithCache | PageHandlerFactory |   435.58 ns |  5.026 ns |  4.702 ns |  1.00 |
|                   |                    |             |           |           |       |
|   ConstructorTest | PageHandlerFactory |   472.72 ns |  1.266 ns |  1.057 ns |  1.00 |
```

1. Try-catch with interface types is the fastest. The constructor test is only a little bit slower.
2. Try-catch is way slower when no cache is being used because it's throwing constantly an exception. With cache it's a little bit faster than the constructor test.

Because try-catch is only a little bit faster (a couple of nanoseconds), I would say we go for the constructor test because it's easier to maintain.

We have to check if all the types can be resolved correctly and we don't get `System.MissingMethodException`-exceptions or `System.InvalidOperationException`-exceptions whenever a service couldn't be resolved.

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: I've done some research with performance in https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491345847. Here we can see that using `Type.GetConstructors()` is faster than using a cache dictionary, so we can remove the dictionary (and the limit).

GerardSmit:
@ahoefling everything is crystal clear. I'll create a new commit where we use `Type.GetConstructors()` as it is faster than using try/catch (see https://github.com/dnnsoftware/Dnn.Platform/pull/4098#discussion_r491345847).

DNN Platform/DotNetNuke.Web/Common/RequestScopeServiceProvider.cs
GerardSmit: On a second thought, if we register a singleton or scoped service that's a class and doesn't have any constructors it'll always go through `Activator.CreateInstance`.

Example:

```csharp
public class TestService {}

{
	{
		services.AddSingleton<TestService>();
	}
}

public class TestModule : PortalModuleBase
{
    public TestModule(TestService service)
    {
        ConstructorService = service;
        DependencyService = DependencyProvider.GetRequiredService<TestService>();
    }

    public TestService ConstructorService { get; }

    public TestService DependencyService { get; }
}
```

In this case `ConstructorService` is created by `Activator` and `DependencyService` is created by the request scope. So we still have to try to get the service from the provider. If it's null then we can create the service from `Activator`.

I'll add this to the unit test.

GerardSmit:
I've changed the try/catch to `Type.GetConstructors`.

Noteworthy: when register a module with no constructor it'll always use `Activator.CreateInstance`. So with this change, the following won't work anymore:

```csharp
services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
    public string Property { get; set; }
}
```

Is this acceptable?

ahoefling:
I looked through all of your details on my phone this morning. Thank you! Amazing to include real numbers with the DotNetBenchmarking tool. I am going to take a closer look at this to provide real feedback but I think we are 90% there.

In short it is not acceptable for it to not resolve interfaces or the PortalModuleBase child classes. I'll make some recommendations on how to fix those

GerardSmit:
> In short it is not acceptable for it to not resolve interfaces or the PortalModuleBase child classes. I'll make some recommendations on how to fix those

Interfaces can be resolved since I've add a check if it's an interface ([`4fc3e09` (#4098)](https://github.com/dnnsoftware/Dnn.Platform/pull/4098/commits/4fc3e0955fc00aaed251f613215c847f2ac3f0f3#diff-8d6fc681ba9ace7b17068cee671c41d2R31)).

What I meant with the PortalModuleBase, is that it's not coming from the service provider if you have no public constructor. So registering it in the service collection will not do anything.

DNN Platform/DotNetNuke.Web/Common/WebFormsServiceProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+            BindingFlags.CreateInstance;
+
+        /// <inheritdoc />
+        public object GetService(Type serviceType)
+        {
+            var provider = HttpContextSource.Current?.GetScope()?.ServiceProvider;
+
+            return provider != null && (serviceType.IsInterface || serviceType.GetConstructors().Length > 0)
+                ? ActivatorUtilities.GetServiceOrCreateInstance(provider, serviceType)
+                : Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```
ahoefling, line 33: This is a great improvement and follows the recommendations I made. In https://github.com/dnnsoftware/Dnn.Platform/pull/4098#issuecomment-695230469 you (@GerardSmit) mentioned that this implementation doesn't work for classes that don't have a default constructor implementation. This demonstrates a major short-coming with this recommendation. Unfortunately, it is not an acceptable change because many classes registered with the container will not have a default constructor.

Let's evaluate a concrete example. Given a class called `MyService`
```c#
public class MyService
{
    // omitted implementation
}
```

If you try injecting this class into a WebForms module it will use the Activator and not the Dependency Injection container. This will become a problem if the `MyService` class was registered as a Singleton as it would generate a new instance every time and not re-use the instance as it is supposed to do if registered as a Singleton.

I think we can make some tweaks here to handle these different scenarios
```suggestion
            if (provider != null)
            {
                var service = provider.GetService(serviceType);
                if (service != null)
                {
                    return service;
                }
            }
            
            return Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```

My code suggestion here removes the `ActivatorUtilities` by design. I was recommending it as a nice to have but since WebForms is already heavily dependent on the use of the `Activator` for instantiating non-public constructors of public classes I don't think it will work for us. Here is the idea of my code suggestion
- First attempt to resolve the service and return it if it is not null
- Else use the activator, at this point the code is assuming that it must be activated by the `Activator`

**Concern:** Right now all of these implementations fallback to the Activator. This means dependencies can be resolved being registered with the Dependency Injection container. This scenario will be extremely buggy, and I don't think it is something we should support. 

@GerardSmit  could you give this change or similar a try and run it through the `DotNetBenchmarking` tool. I think this will be faster than using the checks for `IsInterface` and `GetConstructors` as it removes reflection API invocations from every `GetService` API invocation. This is very important to get right because DNN is all built on WebForms and anything we do here will effect **everything** in DNN except Web API controllers. That is a very significant risk/impact that we need to make sure we get right.

cc: @mitchelsellers @bdukes

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/Internals/WebFormsServiceProviderTests.cs
```diff
@@ -0,0 +1,145 @@
...
+    using DotNetNuke.Common;
+    using DotNetNuke.Common.Extensions;
+    using DotNetNuke.Entities.Modules;
+    using DotNetNuke.Tests.Utilities;
+    using DotNetNuke.Web.Common;
+    using Microsoft.Extensions.DependencyInjection;
+    using Moq;
+    using NUnit.Framework;
+
+    public class WebFormsServiceProviderTests
```
ahoefling, line 20: Could we add some performance tests in this `TestFixture` or create a new one. Maybe hook into the `DotNetBenchmarking` tool. I am very concerned about how this change affects the overall performance of DNN.

ahoefling (changes requested):
@GerardSmit thank you for updating the PR and taking our recommendations to simplify the `IServiceProvider` implementation. You provided great insight with performance metrics that are really helping in this PR.

I made some suggestions that remove all reflection code which should increase performance and resolve the issues you cited about classes with no constructor. When you update the PR, could you complete the following tasks?
- Generate benchmark results so we can analyze performance
- Run exact same benchmark on current DNN class instantiating for `PortalModuleBase` child classes

DNN is built on top of WebForms technology and by configuring the `HttpRuntime.WebObjectActivator` to use our `IServiceProvider` implementation we are introducing significant risk to the ecosystem. As far as I know this will instantiate all WebForms controls with the `IServiceProvider` implementation. Just about everything in DNN runs on a WebForms control except Web API controllers. This means if we do something wrong, it will affect MVC, SPA, and WebForms modules.

DNN Platform/DotNetNuke.Web/Common/WebFormsServiceProvider.cs
ahoefling: Forgot to include some additional research I did. 

I took a look at the `Activator.CreateInstance` implementation from `dotnet/runtime` project. I was worried the implementations fork depending and they all go through the same code path. This is useful for us, as it doesn't really matter how we invoke the API.

My first idea was to see if we can check the registered services on the container and then decide which code-path to go down. Once the `ServicePovider` is built there is no way to get back to the `ServiceCollection` and there is no additional extension method. It is impossible to retrieve an `IEnumerable` of all registered services. Instead I recommended attempting to resolve the requested `serviceType` and then performing additional rules if necessary.

mitchelsellers:
@GerardSmit I've been looking at this a bit more, and once we get things tweaked, I want to do a few basic "side-by-side" performance tests to see how this reacts.  I think everything is going to be fine, but @ahoefling is 100% correct on the potential impact.  (And we admittedly had a few "hiccups" with the MVC/WebAPI implementations of DI which is why we are being cautious.

My plan for this, would be to include as part of 9.8.0 just so you now, as there is a potential for some breaking changes, but I believe it to be minimal.  (Just wanting to set expectations.)

bdukes:
Thanks for everyone's hard work on this, reading through these comments is really encouraging, and this addition will be great to add to the platform!

Regarding resolving classes without public constructors, I want to try to add some clarity and make sure we're on the same page.  In C#, classes have a parameterless public constructor unless another constructor is declared.  So this example _does_ expose a public constructor:

```c#
services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
    public string Property { get; set; }
}
```

Having said that, I'm assuming that the container will not ever instantiate an instance of `MyModule` in this scenario, but will instantiate the generated class from the accompanying `ascx` file that _inherits_ from `MyModule`.  So, even if `MyModule` implemented a protected constructor, it should still work just fine.  That said, because we're not instantiating `MyModule`, I wouldn't expect the registration of `MyModule` in the container to have any effect.

bdukes:
Just to clarify, what scenarios does this PR enable (or, at least, what's the goal)?

- Add constructor injection for `IHttpModule` implementations
- Add constructor injection for `UserControl` implementations (e.g. `PortalModuleBase`)
- Add constructor injection for `Page` implementations (e.g. `PageBase`, `CDefault`)

```c#
public class SuperCoolHttpModule : IHttpModule
{
    public SuperCoolHttpModule(INavigationManager navigationManager)
    {
        this.navigationManager = navigationManager;
    }
}

public abstract class MySuperCoolModule : PortalModuleBase
{
    protected MySuperCoolModule(INavigationManager navigationManager)
    {
        this.navigationManager = navigationManager;
    }
}
```

Does that sound right?  Any additional features this will enable?

GerardSmit:
First of all, thanks all the comments. It's really motivational to work on this PR because of the amount of attention it gets!

However, this PR is getting a little bit out of hand. The comments on lines are handy but it's really annoying to see what the latest comment is and what the state is. So I want to sum up what's going on in this PR.

# The history of this PR
When this PR was created, I used a try-catch to get or create the type from the service provider:

```csharp
// WebFormsServiceProvider.cs

try
{
    return ActivatorUtilities.CreateInstance(provider, serviceType);
}
catch (InvalidOperationException)
{
    return DefaultCreateInstance(serviceType);
}
```

As suggested by @ahoefling I've replaced this with an if-statement to improve performance:

```csharp
// WebFormsServiceProvider.cs

return provider != null && (serviceType.IsInterface || serviceType.GetConstructors().Length > 0)
    ? ActivatorUtilities.GetServiceOrCreateInstance(provider, serviceType)
    : Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```

Because of this change it's not possible to register an HTTP module or control in the service provider **that doesn't have an constructor** because `serviceType.GetConstructors().Length > 0` would be false. So for example:

```csharp
// ExampleExtension.cs

services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
    public string Property { get; set; }
}
```

When we add a constructor then `serviceType.GetConstructors().Length > 0` will be true, causing the property `Property` to become `"Test"`:

```csharp
// ExampleExtension.cs

services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
	public MyModule() {}

    public string Property { get; set; }
}
```

# Current state of the PR
That brings us to the current state of the PR. We now have to decide what we want to do:

1. My idea was to add a check if the type is registered in the service collection:
   
   ```csharp
   // WebFormsServiceProvider.cs
   
   var service = provider?.GetService(serviceType);
   
   if (service != null)
   {
       return service;
   }
   
   return provider != null && (serviceType.IsInterface || serviceType.GetConstructors().Length > 0)
       ? ActivatorUtilities.GetServiceOrCreateInstance(provider, serviceType)
       : Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
   ```

   This will fix the current issue that classes without a constructor wouldn't get resolved.

   If we make a diagram of the flow:

   Diagram

   **Note:** WebForms will use `WebFormsServiceProvider` but when the services are being resolved or being created by `ActivatorUtilities`, then the request scope will be used. So it's not possible that scoped or singleton services are being created and not disposed.

2. @ahoefling suggested we remove `ActivatorUtilities`:

   > My code suggestion here removes the ActivatorUtilities by design. I was recommending it as a nice to have but since WebForms is already heavily dependent on the use of the Activator for instantiating non-public constructors of public classes I don't think it will work for us. Here is the idea of my code suggestion:

   ```csharp
   // WebFormsServiceProvider.cs
   
   if (provider != null)
   {
       var service = provider.GetService(serviceType);
       if (service != null)
       {
           return service;
       }
   }
   
   return Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
   ```
   
   This means we have to always register every control and HTTP module in the service collection, otherwise it'll will fall back to `Activator.CreateInstance` causing it not to resolve the parameters in the constructor and throw an exception.

   If we make a diagram of the flow:

   Diagram

ahoefling:
@GerardSmit WOW, it is not often I see this level of documentation in a PR and I can't tell you how much it is appreciated!

You are 100% correct and the flow-charts make this easy to understand, my latest recommendation requires registering the `PortalModuleBase` child classes with the Dependency Injection container. That was not my intent and the `ActivatorUtilities` would be more appropriate. Keep reading as I have some thoughts on how to close this gap that match the Dependency Injection implementation for MVC and Web API modules.

I gave the recommendation to remove all the reflection calls as your benchmarking data appeared to cause performance issues. While the reflection APIs are not super expensive they are still going to happen every single time we need to run this method, which can be multiple times on any given page. As a platform we need to optimize this as much as we can since so much relies on Web Forms controls.

When we updated the MVC and Web API implementations of Dependency Injection we originally used the `ActivatorUtilities` we decided it made more sense after the original implementation register every single controller with the Dependency Injection container. I wonder if a similar solution makes sense here. What if we add the following code to the startup routine for `DotNetNuke.Library`

**DISCLAIMER: This is only a suggestion to document my ideas, not necessarily the way we must go**

**Startup.cs**
```c#
{
    /// <inheritdoc />
    {
        // This is an extension method that registers all `Controls`
        // with the container so it can easily be resolved.
        services.AddWebForms();

        services.AddSingleton<WebFormsModuleControlFactory>();
        services.AddSingleton<Html5ModuleControlFactory>();

        services.AddScoped<IHostSettingsService, HostController>();
        services.AddScoped<INavigationManager, NavigationManager>();
        services.AddTransient<ISerializationManager, ISerializationManager>();

        services.AddScoped<IApplicationInfo, Application.Application>();
        services.AddScoped<IApplicationStatusInfo, ApplicationStatusInfo>();

        services.AddScoped<IPortalAliasService, PortalAliasController>();
    }
}
```

Then we would create a StartupExtensions class in an extensions namespace `DotNetNuke.Extensions`
```c#
internal static class StartupExtensions
{
    public static void AddWebForms(this IServiceCollection services)
    {
        // This will retrieve all objects in the AppDomain that inherit from Control
        // where Control is the System.Web.UI.Control which should be the parent 
        // class for every web forms control. 
        var webFormsControlTypes = AppDomain.CurrentDomain.GetAssemblies()
            .SelectMany(x => x.SafeGetTypes())
            .Where(x => typeof(Control).IsAssignableFrom(x) &&
                x.IsClass &&
                !x.IsAbstract);

        // Once we have all the controls we just try and register it with the container
        foreach (var controlType in webFormsControlTypes)
        {
            services.TryAddScoped(controlType);
        }
    }
}
```

The StartupExtension I documented here is very similar to how we implemented it for Web API Controllers. 
- https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web/Api/Extensions/StartupExtensions.cs

I think something like this will allow my proposed `WebFormsServiceProvider` to work without the `ActivatorUtilities` which would remove any reflection invocations happening from that provider. It is okay to use heavy reflection during application startup as it is a 1-time expense. The trade-offs here is doing more reflection at startup vs every `Control` instantiation. 

### Memory Leak Concerns
When we added this type of implementation for MVC controllers and Web API controllers the Dependency Injection container was not properly disposing of objects. After N number of requests depending on memory allocation of the server the DNN site would start throwing `OutOfMemoryExceptions`. We needed to update the Controller Factory that instantiated the different MVC and Web API controllers to use the service provider. I believe the `HttpRuntime.WebActivatorObject` should handle that, but I wanted to raise it as a risk that we overlooked in other parts of the platform.

GerardSmit:
@ahoefling  Thanks for the kind words and thanks for explaining it in detail what you meant 🙏.

It now it makes sense for me. I haven't thought of registering every HTTP Module/Control in the service collection and that sounds like a great idea to increase performance.

This opens one more question, do we also have to support Dependency Injection in modules that are being loaded from DesktopModules or App_Start?

For example, in DNN6 the login code-behind was in the file `AuthenticationServices\DNN\Login.ascx.cs` and from what I know is that each .cs-file gets loaded when the view gets loaded. When we register all the controls, these classes are not loaded and cannot be registered. So extensions that define their code-behind in the webroot will not support Dependency Injection.

bdukes:
To help clarify what we're looking at here, I added some logging to the `WebFormsServiceProvider` to see what it's being asked to resolve.  Here's the list:

1. `ClientDependency.Core.CompositeFiles.Providers.CompositeFileProcessingProvider` :warning:
1. `DotNetNuke.Web.Client.Providers.DnnCompositeFileProcessingProvider` :warning:
1. `DotNetNuke.Web.Client.Providers.DnnBodyProvider` :warning:
1. `DotNetNuke.Web.Client.Providers.DnnPageHeaderProvider` :warning:
1. `DotNetNuke.Web.Client.Providers.DnnFormBottomProvider` :warning:
1. `ClientDependency.Core.FileRegistration.Providers.PageHeaderProvider` :warning:
1. `ClientDependency.Core.FileRegistration.Providers.LazyLoadProvider` :warning:
1. `ClientDependency.Core.FileRegistration.Providers.LoaderControlProvider` :warning:
1. `System.Web.Util.RequestValidator` :red_square:
1. `System.Web.Caching.OutputCacheModule` :warning:
1. `System.Web.SessionState.SessionStateModule` :warning:
1. `System.Web.Security.WindowsAuthenticationModule` :warning:
1. `System.Web.Security.FormsAuthenticationModule` :warning:
1. `System.Web.Security.DefaultAuthenticationModule` :warning:
1. `System.Web.Security.RoleManagerModule` :warning:
1. `System.Web.Security.UrlAuthorizationModule` :warning:
1. `System.Web.Security.FileAuthorizationModule` :warning:
1. `System.Web.Security.AnonymousIdentificationModule` :warning:
1. `System.Web.Profile.ProfileModule` :warning:
1. `System.Web.UrlMappingsModule` :warning:
1. `System.Web.Routing.UrlRoutingModule` :warning:
1. `System.Web.Handlers.ScriptModule` :warning:
1. `DotNetNuke.HttpModules.RequestFilter.RequestFilterModule` :warning:
1. `DotNetNuke.HttpModules.UrlRewriteModule` :warning:
1. `DotNetNuke.HttpModules.MobileRedirectModule` :warning:
1. `DotNetNuke.HttpModules.Exceptions.ExceptionModule` :warning:
1. `DotNetNuke.HttpModules.Membership.MembershipModule` :warning:
1. `DotNetNuke.HttpModules.Personalization.PersonalizationModule` :warning:
1. `DotNetNuke.HttpModules.Analytics.AnalyticsModule` :warning:
1. `DotNetNuke.HttpModules.Services.ServicesModule` :warning:
1. `DotNetNuke.Web.Mvc.MvcHttpModule` :warning:
1. `ClientDependency.Core.Module.ClientDependencyModule` :warning:
1. `DotNetNuke.HttpModules.OutputCaching.OutputCacheModule` :warning:
1. `Dnn.EditBar.UI.HttpModules.EditBarModule` :warning:
1. `Dnn.PersonaBar.UI.HttpModules.PersonaBarModule` :warning:
1. `DotNetNuke.HttpModules.DependencyInjection.ServiceRequestScopeModule` :warning:
1. `System.Web.WebPages.WebPageHttpModule` :warning:
1. `System.Web.UI.PageHandlerFactory` :green_circle:
1. `System.Web.Compilation.PageBuildProvider` :green_circle:
1. `System.Web.Compilation.PageBuildProvider` :green_circle:
1. `System.Web.Compilation.PageBuildProvider` :green_circle:
1. `ASP.default_aspx` :green_circle:
1. `System.Web.UI.WebControls.Literal` :green_circle:
1. `System.Web.UI.WebControls.Literal` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlTitle` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlMeta` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `DotNetNuke.Common.Controls.Form` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.WebControls.Label` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlInputHidden` :green_circle:
1. `System.Web.UI.HtmlControls.HtmlInputHidden` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `System.Web.UI.WebControls.PlaceHolder` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.ClientResourceLoader` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `ASP.portals__default_skins_xcillion_home_ascx` :green_circle:
1. `ASP.admin_skins_meta_ascx` :green_circle:
1. `ASP.admin_skins_search_ascx` :green_circle:
1. `System.Web.UI.WebControls.RadioButton` :green_circle:
1. `System.Web.UI.WebControls.RadioButton` :green_circle:
1. `System.Web.UI.WebControls.TextBox` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `DotNetNuke.Web.UI.WebControls.DnnImage` :green_circle:
1. `System.Web.UI.WebControls.TextBox` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `ASP.admin_skins_language_ascx` :green_circle:
1. `System.Web.UI.WebControls.Literal` :green_circle:
1. `System.Web.UI.WebControls.DropDownList` :green_circle:
1. `System.Web.UI.WebControls.Repeater` :green_circle:
1. `System.Web.UI.WebControls.Literal` :green_circle:
1. `ASP.admin_skins_search_ascx` :green_circle:
1. `System.Web.UI.WebControls.RadioButton` :green_circle:
1. `System.Web.UI.WebControls.RadioButton` :green_circle:
1. `System.Web.UI.WebControls.TextBox` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `DotNetNuke.Web.UI.WebControls.DnnImage` :green_circle:
1. `System.Web.UI.WebControls.TextBox` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `ASP.admin_skins_login_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `ASP.admin_skins_user_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `ASP.admin_skins_logo_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `System.Web.UI.WebControls.Image` :green_circle:
1. `System.Web.UI.WebControls.Literal` :green_circle:
1. `ASP.desktopmodules_ddrmenu_menu_ascx` :green_circle:
1. `ASP.admin_skins_copyright_ascx` :green_circle:
1. `System.Web.UI.WebControls.Label` :green_circle:
1. `ASP.admin_skins_linktomobilesite_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `ASP.admin_skins_terms_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `ASP.admin_skins_privacy_ascx` :green_circle:
1. `System.Web.UI.WebControls.HyperLink` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnCssInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnCssInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnCssInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnCssInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnJsInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnJsInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnJsInclude` :green_circle:
1. `DotNetNuke.Web.Client.ClientResourceManagement.DnnJsInclude` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `ASP.portals__default_containers_xcillion_notitle_ascx` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.SourceFileBuildProvider` :green_circle:
1. `ASP.admin_menus_moduleactions_moduleactions_ascx` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `System.Web.UI.WebControls.Panel` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `ASP.desktopmodules_html_htmlmodule_ascx` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNLabelEdit` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBar` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `ASP.portals__default_containers_xcillion_notitle_ascx` :green_circle:
1. `ASP.admin_menus_moduleactions_moduleactions_ascx` :green_circle:
1. `System.Web.UI.WebControls.LinkButton` :green_circle:
1. `System.Web.UI.WebControls.Panel` :green_circle:
1. `ASP.desktopmodules_html_htmlmodule_ascx` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNLabelEdit` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBar` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `DotNetNuke.UI.WebControls.DNNToolBarButton` :green_circle:
1. `System.Web.Compilation.UserControlBuildProvider` :green_circle:
1. `ASP.desktopmodules_admin_dnn_personabar_usercontrols_personabarcontainer_ascx` :green_circle:
1. `System.Web.UI.WebControls.Panel` :green_circle:

A couple of takeaways:
* When rendering an ASCX control, the auto-generated class from the ASCX file is resolved, not the codebehind class.
  * Regarding the question of registering every control in the service collection, I think it's still an open question as to whether we can truly know all of those at start time.
* The item with 🟥, `System.Web.Util.RequestValidator`, is requested without a current `HttpContext` and so bypassed the service provider.
  * Do we need to support resolving this via the container, or do we only support resolving with an `HttpContext`?  (Otherwise, how would we manage lifetime?)
* The items with ⚠️ were requested with a current `HttpContext`, but the `GetScope` call returned `null`.  Do we need to move the scope creation here?

mitchelsellers:
@bdukes Can we chat on this in the next few days

bdukes:
Yup.  I think we just need to setup a couple of tests and try out our different methods (primarily whether we're adding control types into the container or not).

bdukes:
I've [pushed a branch to my repo](https://github.com/GerardSmit/Dnn.Platform/compare/feature/webforms-dependency-injection...bdukes:feature/webforms-dependency-injection) which adjust this PR to define the HTTP Context scope within the `WebFormsServiceProvider`.  It also updates a bunch of WebForms controls to use constructor injection to resolve `INavigationManager`.  I think this provides a good base to run some performance tests regarding whether we want to add `Control` instances into the container.

GerardSmit:
Sorry for the inactivity on my part. I had quite a lot on my plate the last weeks.

I've looked at @bdukes his fork and see that we now create a new scope when it's necessary. At first I got confused by `SetScope(ServiceProvider)` but after looking at the implementation I see the scope is created in there. Nice solution 👍

If you want I can make a performance test this weekend.

My suggestion is we make 3 benchmarks:

1. Default implementation of ASP.NET, which uses `Activator.CreateInstance` only: https://referencesource.microsoft.com/#System.Web/HttpRuntime.cs,3413.
2. Our new service provider (`WebFormsServiceProvider`).
3. Our new service provider (`WebFormsServiceProvider`) with all controls and modules pre-registered to the service provider and with fallback to `Activator.CreateInstance`.

These benchmarks will try to load all the types @bdukes mentioned in https://github.com/dnnsoftware/Dnn.Platform/pull/4098#issuecomment-696298061 with the exception of types in the `ASP.`-namespace, since these are generated by WebForms.

mitchelsellers:
For your first scenario can we test against the current implementation in DNN?  Not default ASP.NET?

GerardSmit:
By my knowledge DNN doesn't set `HttpRuntime.WebObjectActivator` so it falls back to `Activator.CreateInstance`: https://referencesource.microsoft.com/#System.Web/HttpRuntime.cs,3384 or am I missing something?

bdukes (approved):

GerardSmit:
I see this was approved but the following issue is still present in this PR:

If you add an module/control without an constructor to the dependency injection:

```csharp
// ExampleExtension.cs

services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
    public string Property { get; set; }
}
```

It'll not be resolved. In this case, the property is null when it gets resolved.
Whereas if you add a constructor:

```csharp
// ExampleExtension.cs

services.AddTransient(provider => new MyModule
{
    Property = "Test"
});

public class MyModule : PortalModuleBase
{
	public MyModule() {}

    public string Property { get; set; }
}
```

Then the module/control from the DI is being used. The whole story can be found here: https://github.com/dnnsoftware/Dnn.Platform/pull/4098#issuecomment-696210844

Should I apply my proposed fix where we first try to resolve the service, and if it doesn't exists that we create it from DI?

```cs
var service = provider?.GetService(serviceType);

if (service != null)
{
    return service;
}

return provider != null && (serviceType.IsInterface || serviceType.GetConstructors().Length > 0)
    ? ActivatorUtilities.GetServiceOrCreateInstance(provider, serviceType)
    : Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```

bdukes:
@GerardSmit thanks for the quick reply and helping us get back in context. Your proposal sounds good to me, my only question is if there's actually a scenario where `GetService(serviceType)` returns `null` and `serviceType` is an interface. For your workflow (in the post you linked to), wouldn't the code be something like this:

```
var service = provider?.GetService(serviceType);
if (service != null)
{
    return service;
}

if (provider != null)
{
	return ActivatorUtilities.CreateInstance(provider, serviceType);
}

return Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```

I'm sure I'm missing the nuances of our discussion before. In what scenario was `ActivatorUtilities.CreateInstance(provider, serviceType)` throwing `InvalidOperationException` in your initial version? Only when `provider` is `null`, or was there another reason? Is that where the `serviceType.GetConstructors().Length > 0` piece comes from?

GerardSmit:
> @GerardSmit thanks for the quick reply and helping us get back in context. Your proposal sounds good to me, my only question is if there's actually a scenario where `GetService(serviceType)` returns `null` and `serviceType` is an interface. 

I'm not quite sure if there is any scenario in ASP.NET where it would try to resolve an interface. In your test I don't see any interfaces; https://github.com/dnnsoftware/Dnn.Platform/pull/4098#issuecomment-696298061. We could remove it or let it stay, just to be sure 🤷

> I'm sure I'm missing the nuances of our discussion before. In what scenario was `ActivatorUtilities.CreateInstance(provider, serviceType)` throwing `InvalidOperationException` in your initial version? Only when `provider` is `null`, or was there another reason? Is that where the `serviceType.GetConstructors().Length > 0` piece comes from?

Because there are ASP.NET classes that are being resolved that doesn't have a public constructor (also shown in https://github.com/dnnsoftware/Dnn.Platform/pull/4098#issuecomment-696298061). In this case the `ActivatorUtilities.CreateInstance` will fail: https://github.com/dotnet/aspnetcore/blob/64be3b197fe3c65f5eb56119f1b12814910753a4/src/Shared/ActivatorUtilities/ActivatorUtilities.cs#L48-L52

bdukes:
My point with the interfaces is that we won't be able to construct an interface unless it's registered with the container. There's no valid fallback if we're given an interface.

My testing when a module doesn't have an explicit constructor is that `GetConstructors()` still returns an implicit constructor, so I don't think the scenario of a control that inherits from `PortalModuleBase` without an explicit constructor will cause a true issue.

This seems to me like it covers every scenario, then:

```
var service = provider?.GetService(serviceType);
if (service != null)
{
	return service;
}

if (provider != null && serviceType.GetConstructors().Length > 0)
{
	return ActivatorUtilities.CreateInstance(provider, serviceType);
}

return Activator.CreateInstance(serviceType, ActivatorFlags, null, null, null);
```

bdukes:
I've reviewed these changes and it all appears to be working for me. I was able to instantiate an `IHttpHandler` without a constructor, to take in a dependency via constructor injection in an `IHttpHandler`, and didn't notice any performance issues (site felt fast). @mitchelsellers are we good to include this? Having this merged in can simplify some of my other DI work.

valadas (approved):
This looks fine to me, I'll let @mitchelsellers chime in on it since it's kind of an important one, but if it was tested working, this is an awesome enhancement! And we have plenty of time to test and iron out minor issues later on...

mitchelsellers (approved):
Given this is tested, and the benefit, I approve.  Lets use/dogfood and list as a test item for 10.x

bdukes:
Thanks @GerardSmit, this contribution is a great improvement and you've been wonderful to work with!

Issue #3875, opened 2020-06-28
Replace momentjs with dayjs


david-poindexter:
## Description of problem
We are currently using Moment.js in various areas of DNN.  

## Description of solution
We should evaluate replacing with `dayjs` (which is much more lightweight and has a similar API).

The following are all the areas that need to be updated.  We'll use this as a checklist for PRs (one PR per area).

### Admin Experience
* [x] Dnn.EditBar.UI #5082 
* [x] Dnn.PersonaBar.UI #5107 
* [x] Dnn.React.Common #4466 
* [x] AdminLogs.Web #4464 
* [x] Pages.Web #4463 
* [x] Prompt.Web #4465
* [x] Roles.Web #4462
* [x] Servers.Web #4104
* [x] SiteImportExport.Web #4461
* [x] Sites.Web #4459 
* [x] Users.Web #4108

### Platform
* [x] DotNetNuke.Web #5083
* [x] Website #5083

## Description of alternatives considered
n/a

## Screenshots
n/a

## Additional context
We could go from approximately 18.2 KB to 2 KB.

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

valadas:
Yep, but... We need to test for calendars in other cultures, i know this is used some places and there has already been some reported issues about it. Other than that I totally vote for that.

bdukes:
https://momentjs.com/docs/#/-project-status/
>We now generally consider Moment to be a legacy project in maintenance mode. It is not _dead_, but it is indeed _done_.

david-poindexter:
@bdukes thanks for documenting that here.  I am going to go ahead and start seeing what I can do to replace it with dayjs.  I am hoping it will be a fairly smooth transition.  I have experienced this in several other projects so far and I am hoping this one to be the same.

bdukes:
@david-poindexter do you want this pinned?

david-poindexter:
That would be great- thanks @bdukes

david-poindexter:
[...]

PR #4104, opened 2020-09-20
Replace momentjs with dayjs for Servers.Web

david-poindexter:
## Summary
This is the first of several PRs for #3875.  This one replaces `momentjs` with `dayjs` for `Servers.Web` in Dnn.AdminExperience.


valadas (approved):
Looks good to me and saw it working fine

mitchelsellers (changes requested):
Can we update the license documentation to include this new third party to keep compliant.

valadas:
> Can we update the license documentation to include this new third party to keep compliant.

My suggestion is a bit beyond the scope of this PR but what do you guys think if we use https://classic.yarnpkg.com/en/docs/cli/licenses/ to generate all the mentions for the stuff we grab using npm. If we are in agreement, I can work on a PR that would do this on build... It would no cover stuff we use without npm but that would be a lot less maintenance for what does.

mitchelsellers:
Possibly the generate disclaimer option could work. But for the interim id like to get it added here the normal way.

david-poindexter:
@mitchelsellers yes, the license stuff is on my list.  This is the first of many PRs to switch from moment.js to dayjs.  Do you want the license mention to be added to this specific PR?  Also, I have seen license information in multiple places, so at this point I am not clear where it should be based on the usage context.

@valadas I really like your idea and it would really help to have this automated like this.  To me, the sooner we do that, the better off we'll be covered for attribution.

mitchelsellers:
Just needs to be in the master list of licenses for the project. 

No specific place for this PR

david-poindexter:
> Just needs to be in the master list of licenses for the project.

@mitchelsellers I just pushed a commit to put it where I think it needs to be.  Let me know if that is not the correct place.

mitchelsellers (approved):

PR #4096, opened 2020-09-18
Merge CKEditor Provider

bdukes:
This PR merges the `development` branch from https://github.com/dnn-connect/CKEditorProvider into DNN Platform (moving that source code to `DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE`).  This does _not_ yet do anything to include it in the main build process.

PR #4105, opened 2020-09-20
Adjusts build scripts to build CKEditor

valadas:
This is a follow up PR on #4096 which merged CKEditor into this repository. This PR adjusts the build to no longer fetch it externally and build it as the rest of the project.


mitchelsellers (approved):

bdukes (approved):
I pushed one commit to add StyleCop.Analyzers to the `packages.config` for the CKEditorProvider.  Once the build is done this can be merged 👍🏻

valadas:
@bdukes I thought we had a main packages.config and projects would just reference the common packages folder, am I wrong about that?

bdukes:
Yeah, each package has its own `packages.config` file (except for the couple of newer projects that use `<ProjectReference />` inside the `csproj` file).  There's a shared `packages` folder in the root (next to the `sln`), but each project needs its own list of NuGet packages (in addition to the references).

valadas:
Hmm, ok, thanks, I guess it makes sense if you only build one project. I thought for some reason there was a single packages.config to make it easier to bump dependency verisons...

PR #4109, opened 2020-09-20
Fixed an issue where the React common components would interfere with container css class

### LINKED ISSUE #4107: Dnn.React.Common bundle makes containers white
## Description of bug
Whenever one uses the Dnn.React.Common bundle outside of the Persona Bar, it makes the `container` class have a white background. Since `container` is such a generic and popular css class name, it should be at least a bit more scoped to prevent making any random container have a white background.

## Steps to reproduce
1. Install Dnn 9.7.1
2. Get the new Resource Manager module from any build on the feature/resource-manager branch
3. Put the module on a age which has the default theme (Bootstrap based)
4. Notice on the page that has the module, the top bar is now white instead of black/dark-grey

## Screenshots
image

## Additional context
I will submit a PR shortly to fix that.

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.07.02 release candidate
* [x] 09.07.01 latest supported release

valadas:
Closes #4107
This scopes the `container` styles inside the `area` class for the Scrollbars common component. It looks like this component is only used once in all our modules in the Pages module for the list of pages scroll area. I have done some testing and it looks all like before.

Before:
image

After:
image

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Dnn.React.Common/src/ScrollBar/styles.less
+++ b/Dnn.AdminExperience/ClientSide/Dnn.React.Common/src/ScrollBar/styles.less
@@ -1,18 +1,18 @@
-.container {
-  width: 100%;
-  height: 100%;
-  box-sizing: border-box;
-  position: relative;
-  background: white;
-
-  padding: 0px;
-}
 
 .area {
   margin: 0 auto;
   width: 175px;
   min-height: 109px;
   background: white;
+  .container {
+    width: 100%;
+    height: 100%;
+    box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


valadas:
Just adding a little more context for reviewers, the `.container` in question is in the `.area` div so this keeps working in the component while not affecting bootstrap (and other) themes. And that common component appears to be only used in the Pages module for the page list, I tested it there and no UI changes where noticed.

david-poindexter (approved):
Looks good - thanks @valadas 🎉🎉🎉

mitchelsellers (approved):

PR #4110, opened 2020-09-21
Work to improve developer experience for upcoming Telerik Removal

mitchelsellers:
#  Summary

Upon review of the existing code warnings to developers were missing regarding items that will be removed in 10.x.  I've added messaging to this effect that will show as build warnings.  Additionally, a Control skinning feature was adjusted to be able to exist beyond telerik support.

This is part of the larger "Telerik Removal" that is tracked on multiple issues.


DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/TermsList.cs
```diff
@@ -1,173 +1,173 @@
...
-        }
-
-        private void TreeViewNodeClick(object sender, RadTreeNodeEventArgs e)
-        {
-            // Raise the SelectedTermChanged Event
-            this.OnSelectedTermChanged(new TermsEventArgs(this.SelectedTerm));
-        }
-    }
-}
+    public class TermsList : WebControl
```
valadas, line 14: I think this one got missed, but other than that it all looks very good to me

valadas (changes requested):
It all looks good to me except I think TermsList was forgotten.
I know it does not rely on Telerik, but it is still in the deprecated assembly that we would completelly remove right ?

DNN Platform/DotNetNuke.Web.Deprecated/UI/WebControls/TermsList.cs
mitchelsellers: Just pushed that one

valadas (approved):

valadas:
Awesome, looks all good now...

bdukes (approved):

PR #4117, opened 2020-09-22
Removal of references to the Dashboard Project Type

### LINKED ISSUE #4114: Dashboard Functionality Removed - API's Not Updated
## Description of bug
When DNN Platform was migrated to using the Persona Bar, starting with version 9.0.0, the older Dashboard extension point was not brought forward.  

However, all of the dashboard API's, methods, and implementation remain part of the DNN API and could be used by developers with no real results 

## Steps to reproduce
Create a new Dashboard Extension and notice that it is not displayed anywhere.

## Current behavior
Developers can build a Dashboard Extension without any warnings, and the functionality will not work.

## Expected behavior
Developers should get a build warning and notice that the API's will be removed.

## Discussion Point
Traditionally, these API's would be something that should be flagged for removal in version 11.0.0 given the current version.  However, since this functionality is not working now, I would propose that we flag for removal in 10.0.0 to coincide with all other breaking changes already planned.  

## Affected version

* [x] 09.07.01 latest supported release

mitchelsellers:
Fixes #4114 

Upon further review, since version 8.0.0 it has been impossible to install, or uninstall Dashboard components.  However, there were lingering references.  Given this, I have opted in this PR to remove the references rather than simply flagging as there wasn't anything easily flaggable.  I can adjust if needed.

### DIFF
--- a/DNN Platform/Library/Services/Installer/Util.cs
+++ b/DNN Platform/Library/Services/Installer/Util.cs
@@ -616,12 +616,6 @@ namespace DotNetNuke.Services.Installer
         public static string CONFIG_Committed = GetLocalizedString("CONFIG_Committed");
         public static string CONFIG_RolledBack = GetLocalizedString("CONFIG_RolledBack");
         public static string CONFIG_Updated = GetLocalizedString("CONFIG_Updated");
-        public static string DASHBOARD_ReadSuccess = GetLocalizedString("DASHBOARD_ReadSuccess");
-        public static string DASHBOARD_SrcMissing = GetLocalizedString("DASHBOARD_SrcMissing");
-        public static string DASHBOARD_Registered = GetLocalizedString("DASHBOARD_Registered");
-        public static string DASHBOARD_KeyMissing = GetLocalizedString("DASHBOARD_KeyMissing");
-        public static string DASHBOARD_LocalResourcesMissing = GetLocalizedString("DASHBOARD_LocalResourcesMissing");
-        public static string DASHBOARD_UnRegistered = GetLocalizedString("DASHBOARD_UnRegistered");
         public static string DNN_Reading = GetLocalizedString("DNN_Reading");
         public static string DNN_ReadingComponent = GetLocalizedString("DNN_ReadingComponent");
         public static string DNN_ReadingPackage = GetLocalizedString("DNN_ReadingPackage");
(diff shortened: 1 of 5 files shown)


Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Extensions/Constants.cs
```diff
@@ -2,58 +2,57 @@
...
-    {
-        internal const string DefaultExtensionImage = "icon_extensions_32px.png";
-        internal const string DefaultLanguageImage = "icon_languagePack.gif";
-        internal const string DefaultAuthenicationImage = "icon_authentication.png";
-        internal const string DefaultContainerImage = "icon_container.gif";
-        internal const string DefaultSkinImage = "icon_skin.gif";
-        internal const string DefaultProviderImage = "icon_provider.gif";
-        internal const string DefaultLibraryImage = "icon_library.png";
-        internal const string DefaultWidgetImage = "icon_widget.png";
-        internal const string DefaultDashboardImage = "icon_dashboard.png";
```
bdukes, line 54: Do we want to delete this image also?  Or leave it since someone might be using it for something else?

DNN Platform/Library/Services/Installer/Util.cs
```diff
@@ -616,12 +616,6 @@ private static string ValidateNode(string propValue, bool isRequired, Logger log
         public static string CONFIG_Committed = GetLocalizedString("CONFIG_Committed");
         public static string CONFIG_RolledBack = GetLocalizedString("CONFIG_RolledBack");
         public static string CONFIG_Updated = GetLocalizedString("CONFIG_Updated");
-        public static string DASHBOARD_ReadSuccess = GetLocalizedString("DASHBOARD_ReadSuccess");
```
bdukes, line 619: Let's also remove the resources from the resx file: https://github.com/dnnsoftware/Dnn.Platform/blob/f5240f8148ad6bdae497fcba1964f49fd98cd756/DNN%20Platform/Website/App_GlobalResources/SharedResources.resx#L1407-L1424

bdukes: And this one: https://github.com/dnnsoftware/Dnn.Platform/blob/80e48fcde4694d70960253d401a8fcd6851608eb/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Extensions/App_LocalResources/Extensions.resx#L252-L254

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Extensions/Constants.cs
mitchelsellers: I think we have to leave  the image due to others possibly using it

DNN Platform/Library/Services/Installer/Util.cs
mitchelsellers: Just pushed a commit for this one

bdukes (approved):

donker (approved):

PR #4126, opened 2020-09-24
Purge Expired JWT tokens task

donker:
This PR introduces a scheduled task in DNN to remove expired JWT tokens using the existing (and until now unused) DeleteExpiredTokens SPROC.


mitchelsellers (approved):
Removed double approval

mitchelsellers (approved):
Nice to see that this was a simple solution.

For those with history, this job will fail all the time as it will not execute in the 30 second timeout that we have.  However, I'm not 100% convinced that requires adjustment to this PR.

david-poindexter (approved):

PR #4129, opened 2020-09-25
Address React invariant violation

### LINKED ISSUE #4128: Invariant Violation: Minified React error #152;
Found this when upgrading multiple DNN's to 9.7.2 but then found out that this is allready there at least since 9.4.4
We have not found any functional problems so probably nothing big but here it is.

If you open console and you have to preserve log to see it.
Then go to server in the personabar and click clear cache

There it is. 

VM37 export-bundle.js:90 Invariant Violation: Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=t for the full message or use the non-minified dev environment for full errors and additional helpful warnings. 
 at https://SOMEDOMAIN/DesktopModules/admin/Dnn.PersonaBar/scripts/exports/export-bundle.js?cdv=304:90:423
....

bdukes:
Fixes #4128

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/App.jsx
+++ b/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/App.jsx
@@ -33,7 +33,7 @@ class App extends Component {
         const buttonVisible = utils.isHostUser();
         if (this.props.reloadPage) {
             window.top.location.reload();
-            return;
+            return null;
         }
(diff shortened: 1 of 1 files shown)


bdukes:
Thanks @sboshuis!

mitchelsellers (approved):

david-poindexter (approved):

PR #4133, opened 2020-09-27
Swap CodeMirror in CssEditor for monaco-editor

### LINKED ISSUE #4132: Swap CodeMirror in CssEditor for monaco-editor
## Description of problem
The current `CssEditor` PersonaBar module uses is `CodeMirror`.  This editor is antiquated and lacking many useful features of a modern code editor.  It needs to be replaced.

## Description of solution
Replace `CodeMirror` in the `CssEditor` PersonaBar module with [monaco-editor](https://microsoft.github.io/monaco-editor).

n/a

## Screenshots
n/a

## Additional context
The CssEditor module in the PersonaBar is not a `*.Web` project, so for now `monaco-editor` will not be managed via `Yarn`.  This should go on the wishlist though - to convert the CssEditor project to a proper front end project with package management.

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

david-poindexter:
## Summary
Resolves #4132 

Thanks to @valadas for his great help on this PR - it was a lot of fun working on this one!

### Light Theme
Screenshot1

### Dark Theme
Screenshot2

### Find & Replace
Screenshot3

### Color Picker
Screenshot4

### Command Palette
Screenshot5


mitchelsellers (commented):
Two questions before I approve, this looks awesome by the way!

1. It looks like an XCOPY inclusion was done here, which most likely makes sense due to our project structure, but wanted to be sure.

2. If my previous comment is correct, should we strip language support for things that will not be possible, such as PHP, etc.?

david-poindexter:
1. As mentioned in #4132 this is a stepping stone, using the current pattern.  Since the `CssEditor` module is not a proper `*.Web` project with package management, the `monaco-editor` is placed in the shared components.

2. Currently, we are controlling the editor language support in the config for the implementation.

valadas:
In my opinion this (which is awesome by the way) is a first implementation but ideally we would rework this and others together to make it use it as a node_module like the rest of the projects. But this feels like a nice stepping stone. We can use this too in sql console and config editor. When we do move it to a node_module then webpack will just grab the languages that are consumed (css, xml, sql)... As saving does not allow actually saving a php file, I see no concern here.

valadas:
I will review this in more details by the way a bit later today...

mitchelsellers:
> As saving does not allow actually saving a php file, I see no concern here.

My only concern is files with PHP in them can show up in security scans, even if they do nothing.  We already have some issues like that with the CKEditor.

All good here for now.

mitchelsellers (approved):

valadas (approved):
Looks awesome to me!

bdukes (approved):

ahoefling:
👋 Hey everyone, I have quite a bit of experience using monaco in DNN. Last year I built my own IDE built into DNN as a proof of concept. Would love to take a look at this before merging. Hoping I can take a look today or tomorrow.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
```diff
@@ -97,26 +98,82 @@ define([
...
                 return true;
             });
 
             viewModel.portal.subscribe(portalChanged);
         };
 
+        var initCssEditor = function() {
+            var monacoEditorLoaderScript = document.createElement('script');
+            monacoEditorLoaderScript.type = 'text/javascript';
+            monacoEditorLoaderScript.src = '/Resources/Shared/components/MonacoEditor/loader.js';
```
ahoefling, line 124: This PR is maintaining the MonocoEditor minified source code under our shared components. This will make it difficult to maintain the version of the editor DNN is using. Did you try using any of the npm packages that exist for MonacoEditor?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
```diff
@@ -97,26 +98,82 @@ define([
...
+            document.body.appendChild(monacoEditorLoaderScript);
+
+            require.config({ paths: { 'vs': '/Resources/Shared/components/MonacoEditor' }});
```
ahoefling, line 127: Using requireJS is a supported code path for MonacoEditor and it is a great way to get started. The DNN PersonaBar core modules use ReactJS and the MonacoEditor has a ReactJS component that is easy to use in that context. If we are adding this control to a ReactJS page I would consider using that component.

I am not 100% sure how this particular module uses the ReactJS components that is the DNN persona bar modules. In the context of ReactJS I find it much easier to configure the editor and manage it than calling the JS apis. That is my preference and I defer to the maintainers opinions on this one.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
```diff
@@ -97,26 +98,82 @@ define([
...
+                wordWrapColumn: 80,
+                wordWrapMinified: true,
+                wrappingIndent: "indent",
+                lineNumbers: "on",
+                roundedSelection: false,
+                scrollBeyondLastLine: false,
+                readOnly: false,
+                theme: theme,
+                automaticLayout: true
+            });
```
ahoefling, line 174: See comment about about ReactJS component.

ahoefling (commented):
@david-poindexter this will be a great addition to DNN, adding the core VSCode editor the Monaco Editor will be amazing!

I have a few concerns which I highlighted in my comments and will summarize here
- The Monaco Editor is pulled into the DNN source code. We really should use a npm package reference as it will be easier to update in the future. VSCode iterates at an incredibly fast pace and if we need to update the version as this PR exists, it will be a time consuming process. **Unless there is a build script or some form of automation**
- This implementation uses the getting started guide's recommendations of RequireJS. Since the DNN Persona Bar core modules all use ReactJS I think it would make sense to use the Monaco Editor's ReactJS component. This will make it easy to drop onto the page and integrate in other core modules.

Is there plans to allow people to switch from dark to light theme or vice-versa? It may be as simple as placing a tooltip explaining how to use the command palette to adjust theme.

### Useful NPM Links
- [Monaco Editor](https://www.npmjs.com/package/monaco-editor)
- ReactJS Monaco Component - This is where things get tricky, there are several different versions out there that all have their pros/cons
  - [react-monaco-editor](https://www.npmjs.com/package/react-monaco-editor)
  - [@Monaco-Editor/React](https://www.npmjs.com/package/@monaco-editor/react) - not an official one even though it has `@monaco-editor` in name. A good implementation that hides a lot of the webpack config

I have used both of the react components that wrap the Monaco Editor in the past. If I recall correctly I landed on the `react-monaco-editor`. Even though it required more webpack configuration it gave me more control.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
bdukes: The CSS Editor extension for the Persona Bar does not use React and doesn't have a Webpack config at this point, so this approach was used as a first implementation.  The plan is to switch it to use a version from npm.

@david-poindexter @valadas @mitchelsellers Do we want to avoid adding these scripts to the Shared Resources folder if we're planning to remove them soon?  We don't want to give developers the impression that they're intended for broader use, right?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
bdukes: This Persona Bar extension (CSS Editor) does not use React.  I'm not sure if the other extensions which might use Monaco use React or not (SQL and Config Manager).

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
```diff
@@ -97,26 +98,82 @@ define([
...
+        }
+
+        var initMonacoEditor = function() {
+            var theme = "vs-light";
+            if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+                theme = "vs-dark";
+            }
+            window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
+                theme = e.matches ? "vs-dark" : "vs-light";
+            });
```
bdukes, line 159: @ahoefling regarding your question about switching themes, it looks like the current support is making use of [the OS dark/light theme preference](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
mitchelsellers: @bdukes Maybe for this version it would make sense to include under the /Dnn.CssEditor extension until we include this for usage in other locations and properly pull it in?  @david-poindexter thoughts?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
ahoefling: If it follows the Operating System theme that is probably good enough for now

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
ahoefling: Thanks for the clarification on the CSS Editor, I thought something was odd when I didn't see any React code. 

Is there an item already on the backlog for updating this to use an NPM package instead of adding the minified source code? Was there a barrier to getting it working?

ahoefling:
@bdukes shared some insight about the lack of ReactJS in the CSS Editor module. We can disregard all my comments on the ReactJS npm module from my review.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
david-poindexter: @ahoefling please refer to #4132 - this is a "stepping stone" solution.  This is obviously not the "right way" to do this, but the current CssEditor module in the PersonaBar is not a proper solution with package management.  Therefore, that is a task for a future PR.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
david-poindexter: @ahoefling this module currently is not using ReactJS.  It is plain HTML, JS, CSS.  Therefore, this is a stepping stone solution as stated in #4132.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
david-poindexter: See #4132.  This is a stepping stone solution.  Moving this to a proper module with package management will a future task.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
david-poindexter: Correct @bdukes - it detects the user's preference and themes monaco-editor accordingly.

david-poindexter:
> @bdukes shared some insight about the lack of ReactJS in the CSS Editor module. We can disregard all my comments on the ReactJS npm module from my review.

Yes, this will be a future task, and we will have the same issue with the ConfigConsole and SqlConsole modules.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.CssEditor/scripts/CssEditor.js
valadas: Furthermore, I would love that we don't make it React only :) But agree we should pull it as an npm package, which is a future step after this stepping stone.

PR #4121, opened 2020-09-22
Remove the Improvement Program

mitchelsellers:
This PR removes all aspects of the Improvement program from the DNN Platform including

* Removal of existing settings
* Removal of injection into Persona Bar & Control Bar
* Removal of all localization keys

@valadas and @david-poindexter Can you verify my work on the ClientSide stuff please.

PR #4141, opened 2020-09-29
Remove beacon setting on upgrade

bdukes:
This change was missed from #4121


valadas (approved):
Excellent, thank you @bdukes

mitchelsellers (approved):

PR #4142, opened 2020-09-29
Remove Telerik references

bdukes:
This PR removes remaining references to Telerik (primarily project references that aren't used).


bdukes:
I didn't find any significant references to Telerik in the HTML or HTML Editor Manager modules, which I had heard mentioned as one last usage.  If anyone knows something I missed, let's look at getting it out all the way, so we don't have any issues in 9.8.0 removing Telerik.

valadas (approved):
Awesome, looks good to me!

mitchelsellers (approved):

PR #4152, opened 2020-10-06
Implemented a new optional file manager called Resource Manager

valadas:
This new file manager allows us to no longer have any code that relies on Telerik libraries.
This does not replace the current file manager automatically but allows those who want to replace it. Doing so with this release means (providing no other 3rd party modules use Telerik) that the Telerik libraries and bindings can be completely removed.


mitchelsellers (approved):
I have reviewed this PR, and the final solution, on the feature branch which was just brought inline from the develop branch.

I've reviewed the contributions from a security perspective and will continue to review during the 9.7.0 RC.  Given the time spent, and individuals involved I am requesting an independent review from either @bdukes or @mtrutledge  to be the final merging/approving vote on this inclusion.  As such, I'm adding the on-hold label to ensure that we don't accidentally merge.

david-poindexter (approved):
O glorious day!  🎉 🎉 🎉

bdukes:
I pushed some small changes, doing a little testing still.

bdukes (approved):

bdukes:
@mitchelsellers are we good to remove On Hold?

mitchelsellers:
Yes we are!

PR #4153, opened 2020-10-06
Add host setting option and ability to use settings outside module context

donker:
This PR does two things:

1. It adds the (IMO missing) HostSettingsAttribute which persists the value to the HostSettings table
2. It allows you to construct the settings repository with a simple PortalId to retrieve a settings class with only portal/host settings

Because of #2 you can also create a settings class using -1 for portal ID which allows you to use only HostSettings.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/Settings/ISettingsRepository.cs
+++ b/DNN Platform/Library/Entities/Modules/Settings/ISettingsRepository.cs
@@ -7,8 +7,36 @@ namespace DotNetNuke.Entities.Modules.Settings
     public interface ISettingsRepository<T>
         where T : class
     {
+        /// <summary>
+        /// Retrieve module settings. This can optionally include TabModule settings, Portal settings and
+        /// host settings as well. Note the result is cached.
+        /// </summary>
+        /// <param name="moduleContext">Your module's context.</param>
+        /// <returns>Your settings class.</returns>
         T GetSettings(ModuleInfo moduleContext);
 
+        /// <summary>
+        /// Retrieve portal/host settings. This will ignore Module and TabModule settings.
+        /// Note the result is cached.
+        /// </summary>
+        /// <param name="portalId">The portal ID for which to retrieve the settings.</param>
+        /// <returns>Your settings class.</returns>
+        T GetSettings(int portalId);
+
+        /// <summary>
+        /// Save your module's settings. This can optionally include TabModule settings, Portal settings and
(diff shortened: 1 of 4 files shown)


bdukes (approved):

sleupold:
wouldn't it make sense at this state to introduce multilingual support for HostSettings as well?

DNN Platform/Library/Entities/Modules/Settings/ISettingsRepository.cs
```diff
@@ -7,8 +7,36 @@ namespace DotNetNuke.Entities.Modules.Settings
...
+        /// <returns>Your settings class.</returns>
+        T GetSettings(ModuleInfo moduleContext);
+
+        /// <summary>
+        /// Retrieve portal/host settings. This will ignore Module and TabModule settings.
+        /// Note the result is cached.
+        /// </summary>
+        /// <param name="portalId">The portal ID for which to retrieve the settings.</param>
+        /// <returns>Your settings class.</returns>
+        T GetSettings(int portalId);
```
mitchelsellers, line 24: Given the differences in behaviors here, I think it might be better for these two new methods to be added with different names, as we are blending/mixing Module and Portal settings, in a manner that is only clear if you dive into the comments of the particular method. Additionally, the differentiation between Portal & Host settings is hard to understand here.  If a behavior is host settings with a -1 value, we should document this.

Given we also have ModuleSettings, TabSettings, and TabModuleSettings "GetSettings" is just too ambiguous in my opinion.

mitchelsellers (changes requested):

DNN Platform/Library/Entities/Modules/Settings/ISettingsRepository.cs
bdukes: This is part of the settings repository, which retrieves a "settings object".  That object's properties will be marked with attributes to determine which settings they're retrieved from.  So, typically, most of the settings will be module settings and used within the context of a module, but potentially the same pattern can be used in instances where you know that there aren't module settings, or you only care about properties that you know are portal settings.  But `GetSettings` does retrieve all relevant settings with the information it's been given.

mitchelsellers: Ok, I'm fine, just seems confusing to mix portal/module/host here with the same API even with it being generic

mitchelsellers (approved):

PR #4155, opened 2020-10-06
Enable dependency injection for IServiceRouteMapper

bdukes:
This allows types that implement `IServiceRouteMapper` to utilize constructor injection from the DI container.

@mitchelsellers unfortunately, this is the only commit from #3933 that doesn't have any potential lifetime management issues.

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
@@ -1,7 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
-
 namespace DotNetNuke.Web.Api.Internal
 {
     using System;
@@ -22,6 +21,8 @@ namespace DotNetNuke.Web.Api.Internal
     using DotNetNuke.Web.Api.Auth;
     using DotNetNuke.Web.Api.Internal.Auth;
     using DotNetNuke.Web.ConfigSection;
+
+    using Microsoft.Extensions.DependencyInjection;
 
     public sealed class ServicesRoutingManager : IMapRoute
     {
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):
I’ll take what I can get

ahoefling:
Is this adding feature parity with ASP.NET either Core or Framework? I am not sure if I fully agree with adding this level of support as I've seen it abused as a startup event.

In the scenarios I have seen in the field, developers are abusing different parts of the platform to get access to startup events. If this is trying to solve that problem, I think we should add an interface specific to startup logic. 

@bdukes could you explain a little bit more about intended use cases?

bdukes:
`IServiceRouteMapper` is a DNN concept, so there's no feature parity to consider (unless I'm misunderstanding).  I think `IStartup` should start becoming the interface to use to run logic during startup.

I don't have any specific use cases in mind, just trying to replace `Activator.CreateInstance` calls with DI.

ahoefling:
@bdukes 
> IServiceRouteMapper is a DNN concept

🤦‍♂️It is really starting to show how little web work I do these days as I spend all my time on mobile and desktop 🤦‍♂️

You are absolutely correct 💯. What I was remembering was the static class that we typically used in an ASP.NET WebAPI project for configuring routes. It didn't have a specific interface or anything, it was just invoked from the startup routine. Here is an example of what the class would look like

```c#
public static class RouteConfig
{
    public static void RegisterRoutes(RouteCollection routes)
    {
    }
}
```

In DNN as you mentioned the `IServiceRouteMapper` is used instead.

I see why you are adding support for this, by adding Dependency Injection for the `IServiceRouteMapper` it allows the container to manage the lifecycle instead of using reflection. This should give us a little bit of a performance benefit. I'll leave some comments in the PR as I have a few more thoughts on this.

DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
```diff
@@ -258,12 +261,11 @@ private IEnumerable<IServiceRouteMapper> GetServiceRouteMappers()
                 IServiceRouteMapper routeMapper;
                 try
                 {
-                    routeMapper = Activator.CreateInstance(routeMapperType) as IServiceRouteMapper;
+                    routeMapper = ActivatorUtilities.CreateInstance(serviceProvider, routeMapperType) as IServiceRouteMapper;
```
ahoefling, line 264: I like adding the ActivatorUtlities here it enables us to use either reflection or the container. I was wondering if it made sense to register all implementations of `IServiceRouteMapper` as a transient service. That way we can just depend on the container and it would never use reflection here.

bdukes: I had been thinking about that (especially with some of the trickier scenarios I looked at in #3933).  In this case, we're only instantiating these instances one time on startup, so it doesn't seem to me like it'd be worth it.  But definitely a route we can look at for other extension points that are used multiple times.

ahoefling: That makes sense, the only scenario I can think of which would make sense would be route localization. If you want to inject a localization service to the `IServiceRouteMapper` which would handle generating localized routes.

The developer could still just register the mapper in their own implementation of `IDnnStartup`.

ahoefling (approved):
Looks good to me 🛳

donker (approved):

PR #4156, opened 2020-10-07
Migrate to MailKit from System.Net.Mail

### LINKED ISSUE #3768: TimeOut when testing host SMTP parameters on SSL
## Description of bug
Using SMTP connection on port 465 with SSL and testing the specified parameters always results to a Timeout if target host only supports implicit SSL.

## Steps to reproduce
Just try using the following parameters and test to see a timeout:
host: smtps.aruba.it:465
enable ssl: on
smtp username: set at any value, test will never get to the point of using it
smpt password: set at any value, test will never get to the point of using it
smtp authentication: basic

## Current behavior
Test hangs until a timeout occurs.

## Expected behavior
A proper connection and a successfull send of the test email.

## Error information
Test will add the following error messate to the EventLog:
AbsoluteURL:/API/PersonaBar/ServerSettingsSmtpHost/SendTestEmail
ExceptionGUID:c74b358f-b88b-49cc-87e1-2132543fb65a
AssemblyVersion:
PortalId:-1
UserId:-1
TabId:-1
RawUrl:
Referrer:
UserAgent:
ExceptionHash:YtkkCJ66Rwp9mcQRpSQZr7zQ+MI=
Message:The operation has timed out.
StackTrace:
   at System.Net.Mail.SmtpClient.Send(MailMessage message)
   at DotNetNuke.Services.Mail.Mail.SendMailInternal(MailMessage mailMessage, String subject, String body, MailPriority priority, MailFormat bodyFormat, Encoding bodyEncoding, IEnumerable`1 attachments, String smtpServer, String smtpAuthentication, String smtpUsername, String smtpPassword, Boolean smtpEnableSSL)
InnerMessage:
InnerStackTrace:
Source:System
FileName:
FileLineNumber:0
FileColumnNumber:0
Method:
Server Name: XXXXXX

## Additional context
The timeout error is caused by this library: System.Net.Mail. According to this post (http://www.systemnetmail.com/faq/5.3.aspx) this library only supports explicit SSL, that is it expects port 25 to be open so that it can send a start TLS on it and switch to port 465 then after. If port 25 is not open, the library times out.
This is not a bug of DNN in itself, but because it is using a System DLL that does not support implicit SSL (directly starting with encryption), it makes impossible to use it with providers like Aruba (in Italy one of the biggest providers of internet services and cloud solutions) that expects the use of implicit SMTP on their SMTP service.

## Affected version

* [ X] 10.00.00 alpha build
* [ X] 09.06.01 release candidate
* [ X] 09.06.00 latest supported release

### LINKED ISSUE #3974: Migrate to MailKit from System.Net.Mail
## Description of problem
See #3768 

## Description of solution
Migrate to MailKit in CoreMailProvider Implementation of MailProvider

meetmandeep:
Closes #3974 
Possibly Closes #3768 (**Need someone to test with Office 365**)
## Summary
Migrated to MailKit from System.Net.Mail

Uses the new MailAttachment wrapper class instead of relying on System.Net.Mail

## Deprecation & Backwards Compatible
Few methods have been marked as obsolte with scheduled removal in v11; these methods are using the System.Net.Mail Attachment. New methods have been introduced with MailAttachment parameter as alternatives. Backwards compatibility supported. 

## Missing Functionality
I couldn't find the equivalent functionality in MailKit for following: 

```
mailMessage.SubjectEncoding = mailInfo.BodyEncoding;
mailMessage.BodyEncoding = mailInfo.BodyEncoding;
```

My understanding is that MailKit automatically encodes both Subject/Body as needed. 

```
smtpClient.ServicePoint.MaxIdleTime = Host.SMTPMaxIdleTime;
smtpClient.ServicePoint.ConnectionLimit = Host.SMTPConnectionLimit;
```

I couldn't find any relevant properties in MailKit. We do have UI settings for these 2 items so need to make a decision on that also. 

## Testing
**We really need to test this extensively since sending mail is a critical component of the platform.** 

I've tested using [Mailtrap ](https://mailtrap.io/)using the following test code for both new and backwards compatible functionality. 

```
 try
                {
                    using (var sr = new StreamReader(this.Page.MapPath("Robots.txt")))
                    {
                        using (MemoryStream ms = new MemoryStream())
                        {
                            sr.BaseStream.CopyTo(ms);
                            sr.BaseStream.Seek(0, SeekOrigin.Begin);

                            Attachment attach1 = new Attachment(this.Page.MapPath("403-3.gif"));
                            Attachment attach2 = new Attachment(sr.BaseStream,"Robots.txt");

                            MailAttachment mattach1 = new MailAttachment(this.Page.MapPath("403-3.gif"));
                            MailAttachment mattach2 = new MailAttachment("Robots.txt", ms.ToArray());

                            List<Attachment> attachments = new List<Attachment>();
                            attachments.Add(attach1);
                            attachments.Add(attach2);

                            List<MailAttachment> mattachments = new List<MailAttachment>();
                            mattachments.Add(mattach1);
                            mattachments.Add(mattach2);

                            Mail.SendEmail("user1@test.com", "user2@test.com", "Test Email 1", "Test Body 1");
                            Mail.SendEmail("user1@test.com", "sender@test.com", user2@test.com", "Test Email 2", "Test Body 2");
                            Mail.SendEmail("user1@test.com", "sender@test.com", "user2@test.com", "Test Email 3", "Test Body 3", attachments);
                            Mail.SendEmail("user1@test.com", "sender@test.com", "user2@test.com", "Test Email 4", "Test Body 4", mattachments);
                        }
                    }
                }
                catch (Exception ex)
                {
                    string mes = ex.Message;
                }
```


DNN Platform/Library/Services/Mail/CoreMailProvider.cs
```diff
@@ -161,32 +161,21 @@ public override string SendMail(MailInfo mailInfo, SmtpInfo smtpInfo = null)
...
                                 {
-                                    smtpClient.UseDefaultCredentials = false;
-                                    smtpClient.Credentials = new NetworkCredential(smtpInfo.Username, smtpInfo.Password);
+                                    smtpClient.Authenticate(smtpInfo.Username, smtpInfo.Password);
                                 }
 
                                 break;
-                            case "2": // NTLM
-                                smtpClient.UseDefaultCredentials = true;
+                            case "2": // NTLM (Not Supported by MailKit)
```
mitchelsellers, line 168: This is a breaking change technically.  Are we ok with this?  I think so but calling attention to it.

DNN Platform/Library/Services/Mail/MailAttachment.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        }
+        private void MailAttachmentInternal(string Filename, Byte[] Content, string ContentType)
+        {
+            this.Filename = Filename;
+            this.Content = Content;
+            this.ContentType = ContentType;
+        }
+
+        private string GetContentType(string fileExtension)
```
mitchelsellers, line 39: Can we do this the .net way? https://docs.microsoft.com/en-us/dotnet/api/system.web.mimemapping.getmimemapping

mitchelsellers (changes requested):

ahoefling:
@meetmandeep thanks for contributing this PR 🎉 

I was wondering if we could update this PR to add Dependency Injection support for emails instead of just swapping out the implementation in the existing DNN Provider Framework. This will make it easier for developers to use their own implementation in the future (which is going to be pretty rare). I think it would be a big win for the platform to have this critical piece of the puzzle with Dependency Injection Support.

**This is only a recommendation as a non-maintainer, please only do this if you agree and the maintainers agree**

### What I'm thinking and only do this if you agree and the maintainers agree
Update the current EmailProvider to be an `IEmailService` that is registered with the Dependency Injection container. We can then define a contract of methods that we expect all email services to implement. All interfaces would be included in a namespace in the Abstractions project such as `DotNetNuke.Abstraction.Mail`.

Once this is done we would use your implementation as the registered implementation, but it allows a framework for developers to swap out for their own implementation as long as they implement the contract

mitchelsellers:
@ahoefling I'm all for DI support, but I think it would need to be done at a different area than this.

The provider model is our extension point for configuration allowing an implementation to be switched out.  (Authentication Providers, Permission Providers, etc.).  I would see DI usage here being for the consumers to get away from Mail.Mail static members, which would be a different effort, although a good one.

DNN Platform/Library/Services/Mail/CoreMailProvider.cs
meetmandeep: Yes. We'll just have to document it.

> You cannot use CredentialCache.DefaultNetworkCredentials with MailKit because when MailKit asks the credential to provide the username and password strings that it needs to send to the server, the DefaultNetworkCredentials tells MailKit that the username and password strings are empty and so MailKit tries to authenticate with empty strings.
> 
> System.Net.Mail seems to be able to use an internal API to get the real strings that MailKit does not have access to.

Here's the reference thread I found: https://stackoverflow.com/questions/38113437/is-it-possible-to-use-default-network-credentials-with-mailkit-and-exchange

mitchelsellers:
Upon further review of this PR, this introduces a breaking change that would be in violation of our release rules, by removing a supported authentication method for users.  (This requires notice and inclusion in a major release.)

The preferred approach would be to leave the existing Core provider as-is, and introduce MailKit as a new provider that could be selected by end-users.  This should be a small change in theory.

@meetmandeep feel free to reach out to me directly tomorrow on this.

valadas:
We need to rebase this to resolve the conflicts, @bdukes if you decide to handle this I would like to learn how to do it properly, ping me up.

bdukes:
@dnnsoftware/approvers this PR is ready for review now. Does anyone have an Office365 SMTP server to test with? There were notes about testing with a null and empty from name, as well as a comma-separated to email list, @meetmandeep did you test those scenarios, or are we still needing verification?  Thanks!

mitchelsellers (changes requested):
Can we quickly add the license information for these new packages to the /licenses folder to make sure we follow that pattern

bdukes:
I've added license files, let me know if you need any adjustments @mitchelsellers

mitchelsellers (approved):
This looks good, I'll try to test with an O365 account later today as well

valadas:
Can we add a merge config file so that we have both providers available in the web config and users can easily just change the defaultProvider value ?

```diff
<mail defaultProvider="CoreMailProvider">
      <providers>
        <clear />
        <add name="CoreMailProvider" type="DotNetNuke.Services.Mail.CoreMailProvider, DotNetNuke" hideCoreSettings="False" />
+        <add name="MailKitMailProvider" type="DotNetNuke.Services.Mail.MailKitMailProvider, DotNetNuke" hideCoreSettings="False" />
      </providers>
    </mail>
```

bdukes:
I've added the config change

mitchelsellers (approved):
Thanks @bdukes

valadas (commented):
Looks good to me

valadas (approved):

PR #4158, opened 2020-10-07
Resolves an infinite redirect loop issue

### LINKED ISSUE #4157: Website with multiple aliases create infinite redirect loop on mobile
On a website with multiple aliases, navigating to the site in Chrome simulating a mobile device, we get an infinite redirect loop.

Thanks to @patrickryan79 I was able to reproduce the issue with the following steps.

Here you go @valadas.

Steps to reproduce in a local environment.

1. Edit hosts file and add two domains pointing to 127.0.0.1 (e.g. mobiletest.com, sub.mobiletest.com)
2. Setup site files and configure in IIS, adding the two bindings for mobiletest.com, sub.mobiletest.com
2. Install DNN (I used 9.7.2) by loading mobiletest.com
3. Add sub.mobiletest.com as portal alias
4. Leave all other settings as default
5. Using Chrome, test the second alias
6. With developer tools, trigger mobile view and reload

Worth mentioning here that there is no SSL certificate, no redirects, etc.

_Originally posted by @patrickryan79 in https://github.com/dnnsoftware/Dnn.Platform/discussions/4137#discussioncomment-91870_

Versions affected:
[x] 9.7.2
Maybe others ? did not try.

valadas:
Closes #4157

An infinite redirect loop could occur even in situations where the portal aliasses where setup in canonical mode.

Relates to this discussion: https://github.com/dnnsoftware/Dnn.Platform/discussions/4137

### DIFF
--- a/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
@@ -254,7 +254,7 @@ namespace DotNetNuke.Entities.Urls
             {
                 var portalAliases = PortalAliasController.Instance.GetPortalAliasesByPortalId(result.PortalId).ToList();
 
-                if (queryStringCol != null && queryStringCol["forceAlias"] != "true")
+                if (queryStringCol != null && queryStringCol["forceAlias"] != null && queryStringCol["forceAlias"] != "true")
                 {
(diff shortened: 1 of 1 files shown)


david-poindexter (approved):
This looks good to me - thanks @valadas 🎉 🎉

mitchelsellers (approved):

Issue #3161, opened 2019-10-17
Add Support for Microsoft.Extensions.Logging


ahoefling:
## Description of Problem
Logging in DNN is using log4net and I believe we are using a DNN specific fork of it. **This works and I don't intend to change how this works at all for the .NET Framework version of DNN.** As we get closer to .NET Core migrations, we need to consider how to decouple the platform from any hard dependencies on .NET Framework to make the transition easier. 

DNN currently has an `ILog` interface that exists in the `DotNetNuke.Instrumentation` library. This is the interface that is used for logging throughout the platform and ecosystem.

```c#
public interface ILog
{
    bool IsDebugEnabled { get; }
    bool IsInfoEnabled { get; }
    bool IsTraceEnabled { get; }
    bool IsWarnEnabled { get; } 
    bool IsErrorEnabled { get; }
    bool IsFatalEnabled { get; }
        
    void Debug(object message);
    void Debug(object message, Exception exception);
    void DebugFormat(string format, params object[] args);
    void DebugFormat(IFormatProvider provider, string format, params object[] args);

    void Info(object message);
    void Info(object message, Exception exception);
    void InfoFormat(string format, params object[] args);
    void InfoFormat(IFormatProvider provider, string format, params object[] args);

    void Trace(object message);
    void Trace(object message, Exception exception);
    void TraceFormat(string format, params object[] args);
    void TraceFormat(IFormatProvider provider, string format, params object[] args);

    void Warn(object message);
    void Warn(object message, Exception exception);
    void WarnFormat(string format, params object[] args);
    void WarnFormat(IFormatProvider provider, string format, params object[] args);

    void Error(object message);
    void Error(object message, Exception exception);
    void ErrorFormat(string format, params object[] args);
    void ErrorFormat(IFormatProvider provider, string format, params object[] args);

    void Fatal(object message);
    void Fatal(object message, Exception exception);
    void FatalFormat(string format, params object[] args);
    void FatalFormat(IFormatProvider provider, string format, params object[] args);
}
```

This topic was discussed in some capacity in the following item:
* #2801 - The discuss was more focused on upgrading log4net

## Proposed Solution Option 1
[...]

Issue #2801, opened 2019-05-17
RFC: DNN & Log4Net


mitchelsellers:
Currently, the DNN Platform is using Log4Net as the backend logging provider, this provider is used to write the log files to the /Portals/_default/logs directory.  Users can customize the configuration of this using a configuration file.

## Description of Problem
There are a number of concerns with the current Log4Net implementation:

* It is a fork, and an older fork at that, of the Log4Net project and has not been maintained
* There is a bug in the file provider, whereby log files are not cleaned up automatically.  This resulted in various "fun" solutions outside of DNN to clean up the directories.

## Proposed Solution Option 1
Update the fork to be current with the Log4Net implementation as it is today.

I'm not a fan of this, as the management of forks is hard on the community and takes a lot of effort to manage.

## Proposed Solution Option 2
Replace the `DotNetNuke.Log4Net` with a NuGet reference to the actual Log4Net project.  It appears that the actual code included in this project is nothing more than a direct copy/paste, including namespaces, so the level of effort here is NIL and the benefit would be better management of assets.

This is a much better option, however, looking at .NET Core future the Log4Net solution doesn't fully adhere to the new .NET Core items such as configuration.

## Proposed Solution Option 3
Replace Log4Net with another logging solution, such as [SeriLog](https://serilog.net/).  This library is licensed under the same license as the Log4Net solution but has a large number of appenders that allow data to be shipped to other locations with minimal effort.  Including Email, File, Database, Azure Blob Storage, Application Insights, and more.

This solution also has the benefit of better supporting .NET Core standards for configuration loading etc.  

## Alternatives Researched
We can leave it "as is" but eventually have to fix this for the future.

mitchelsellers:
My personal recommendation would be to leverage option 3.  I don't believe that this would be a breaking change for users, as the usage of the Log4Net library is all internal to DNN.  But the added bonus features that we get with SeriLog would outweigh the differences.

We have been using this product for a number of years on our .NET Core projects and I will not create a new project without it.  The flexibility has been amazing.

liffeyd:
Usage of Log4Net is not only internal to DNN. I, and I am sure others, also use it in development.

valadas:
Yes, I did use log4net in some modules. If I remember correctly there is a dnn wrapper around that logging maybe we could update the wrapper to use the new solution? Or announce it in advance and document how to update if you where using it...

mitchelsellers:
[...]

PR #4159, opened 2020-10-07
Adds IEventLogger to Dependency Injection

ahoefling:
## Summary
Extracts a new `IEventLogger` interface from the `EventLogController` and registers it with the dependency injection container. This will make it easier for developers to log events with DNN anywhere in the platform.

Looking to get feedback before I spend too much time polishing this PR up, I need to look at the following items before merging
- Unit Tests
- XML docs

## Required to Merge
| Task | Status |
|------|--------|
| Unit Test |  ✔ Completed |
| Test Installer | ✔ Completed |
| Update XML Documentation | ✔ Completed |


DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace DotNetNuke.Abstractions.Logging
+{
+    using System;
+    using System.Collections;
+    using System.Xml;
+
+    /// <summary>
+    /// add xml docs.
+    /// </summary>
+    public interface ILogProperties : IList, ICollection, IEnumerable, ICloneable
```
bdukes, line 13: Could we implement `ICollection<ILogDetailInfo>`?

ahoefling: This makes sense, I didn't create that interface and it should be added

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace DotNetNuke.Abstractions.Logging
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// add xml docs.
+    /// </summary>
+    public interface ILogProperties : IList<ILogDetailInfo>, ICollection<ILogDetailInfo>, IEnumerable<ILogDetailInfo>, IList, ICollection, IEnumerable, ICloneable
```
ahoefling, line 13: @bdukes I updated this to explicitly inherit from `IList<ILogDetailInfo>, ICollection<ILogDetailInfo>, IEnumerable<<ILogDetailInfo>>`. This entire inheritance structure can be simplified but it may have benefit to be explicit

```c#
public interface ILogProperties : IList<ILogDetailInfo>, IList, IClonable
```

DNN Platform/Library/Services/Log/EventLog/LogProperties.cs
```diff
@@ -25,8 +29,54 @@ public string Summary
...
+            {
+                base.Remove(item);
+            }
+            catch
+            {
+                return false;
+            }
+
+            return true;
+        }
```
ahoefling, line 78: @bdukes do you have any thoughts on this implementation. The interface requires a bool return yet `ArrayList` is just void. I wrapped it to catch any exception and return false.

Open to ideas here.

bdukes: I think you'd probably need to do a contains check first.  Unless the `ArrayList` is going to throw if the item doesn't exist.

ahoefling: I just looked at the ArrayList implementation for .NET 5 which may differ slightly, but it is a useful reference. It doesn't appear to throw any exceptions with this method. If it can't find the item it just slightly continues.

https://github.com/dotnet/runtime/blob/de6f9114a54587ea1c3e13c1aba2dbaff3c39fc5/src/libraries/System.Private.CoreLib/src/System/Collections/ArrayList.cs#L536-L557
```c#
        public virtual void Remove(object? obj)
        {
            int index = IndexOf(obj);
            if (index >= 0)
                RemoveAt(index);
        }


        // Removes the element at the given index. The size of the list is
        // decreased by one.
        //
        public virtual void RemoveAt(int index)
        {
            if (index < 0 || index >= _size) throw new ArgumentOutOfRangeException(nameof(index), SR.ArgumentOutOfRange_Index);


            _size--;
            if (index < _size)
            {
                Array.Copy(_items, index + 1, _items, index, _size - index);
            }
            _items[_size] = null;
            _version++;
        }
```

## Next Steps
I agree with you, let's update this to check if the item exists and then delete it

DNN Platform/Library/Services/Log/EventLog/LogProperties.cs
```diff
@@ -25,8 +29,52 @@ public string Summary
...
+        {
+            var index = base.IndexOf(item);
+            if (index >= 0)
+            {
+                this.RemoveAt(index);
+                return true;
+            }
+
+            return false;
+        }
```
ahoefling, line 76: @bdukes I updated this code in the latest push to use `IndexOf` instead of `Contains` because it uses the native `Array.IndexOf` implementation which will be slightly faster. 

In this code snippet we will iterate through all of the items in the collection once O(n) vs the other approach we would iterate through everything twice O(2n). Granted the big-O simplifies down to O(n) for both scenarios, but this should be slightly faster.

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Gets or sets the log user id.
+        /// </summary>
+        int LogUserID { get; set; }
+
+        /// <summary>
+        /// Gets or sets the log event id.
+        /// </summary>
+        int LogEventID { get; set; }
```
bdukes, line 36: Did we decide with `IPortalAliasInfo` that we would standardize on `Id` over `ID` in these new interfaces?  Note that it _did_ create an issue using the class from VB.NET when both `PortalId` and `PortalID` were exposed, so the `Id` implementation will need to be an explicit interface implementation.

```suggestion
        string LogGuid { get; set; }

        /// <summary>
        /// Gets or sets the log file ID.
        /// </summary>
        string LogFileId { get; set; }

        /// <summary>
        /// Gets or sets the log type key.
        /// </summary>
        string LogTypeKey { get; set; }

        /// <summary>
        /// Gets or sets the log user id.
        /// </summary>
        int LogUserId { get; set; }

        /// <summary>
        /// Gets or sets the log event id.
        /// </summary>
        int LogEventId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Gets or sets the log username.
+        /// </summary>
+        string LogUserName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the log portal id.
+        /// </summary>
+        int LogPortalID { get; set; }
```
bdukes, line 46:
```suggestion
        int LogPortalId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Gets or sets the log server name.
+        /// </summary>
+        string LogServerName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the log config id.
+        /// </summary>
+        string LogConfigID { get; set; }
```
bdukes, line 81:
```suggestion
        string LogConfigId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
bdukes: I think I'd lean towards only requiring `ICollection<ILogDetailInfo>`, so we don't cause issues with other implementations that aren't based on `ArrayList`.  Is the `ICloneable` interface used, or just something that happens to be available on `ArrayList`?

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets the notification threshold time.
+        /// </summary>
+        NotificationThresholdTimeType NotificationThresholdTimeType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Id.
+        /// </summary>
+        string ID { get; set; }
```
bdukes, line 51:
```suggestion
        string Id { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets the log file name with path.
+        /// </summary>
+        string LogFileNameWithPath { get; set; }
+
+        /// <summary>
+        /// Gets or sets the log type portal Id.
+        /// </summary>
+        string LogTypePortalID { get; set; }
```
bdukes, line 71:
```suggestion
        string LogTypePortalId { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// Log type info.
+    /// </summary>
+    public interface ILogTypeInfo
+    {
+        /// <summary>
+        /// Gets or sets log type css class.
+        /// </summary>
+        string LogTypeCSSClass { get; set; }
```
bdukes, line 14:
```suggestion
        string LogTypeCssClass { get; set; }
```

DNN Platform/DotNetNuke.Abstractions/Logging/LoggingProviderReturnType.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Log info objects.
+        /// </summary>
+        LogInfoObjects = 0,
+
+        /// <summary>
+        /// XML.
+        /// </summary>
+        XML = 1,
```
bdukes, line 19:
```suggestion
        Xml = 1,
```

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
ahoefling: I forgot to update that, thanks for catching this. Can we create a stylecop setting that is specific to the Abstractions project to catch these in the future?

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
```diff
@@ -0,0 +1,196 @@
...
+{
+    using System.Collections;
+    using System.Collections.Generic;
+
+    using DotNetNuke.Abstractions.Portals;
+
+    /// <summary>
+    /// Event Logger.
+    /// </summary>
+    public interface IEventLogger
```
bdukes, line 14: My only other feedback is wondering whether `IEventLogger` should be primarily focused on `AddLog` and related methods, and the log config APIs could be put into a separate interface (later).

ahoefling: Thanks for all the feedback here, it is much appreciated. I am glad you commented on this as it gives me an opportunity to explain my thoughts.

## My Reasoning behind IEventLogger
I decided to make this PR a lift and shift from `IEventLogController` to `IEventLogger`. I think it makes sense to have 2 different types of logging interfaces: log4net file logging vs database event logging. The event logging appears to be designed around tracking business rule events that happen from different modules or core components. Where log4net is more for tracking exceptions and errors. Ultimately my goal would be to submit a PR that removes log4net and uses `Microsoft.Extensions.Logging` that would use the `ILogger<T>` that ships with that NuGet. I just don't think we are at the point where we can do that yet.

I have quite a bit of experience with event tracking as it is very popular in mobile and desktop apps. The problem I see most of the time is an `ILogger` API structure doesn't line up very well with something like the `IEventLogger` API structure. I think it makes sense to keep them separate. Once we get better support for logging in DNN with Dependency Injection I think it would make sense to add an AggregateLogger or extension methods so you can use general Logging and Event Logging from the same interface. I just don't think we are there, hence the lift and shift PR here.

## Future Plans
I am not a big fan of how hard it is to get up and running with the `EventLogController` and now the `IEventLogger`. There is a lot of guess and check as the database doesn't like null values. I would like to simplify this API structure if possible, but I think the goal of this PR would be to just do a lift and shift since it just needs some minor improvements.

For example I think it would make more sense to use `IDictionary<string, string>` for `LogProperties` instead of `ArrayList`. I'm sure this was first implemented at a time when generics weren't available so this was the best way to handle it at the time.

If we accept this lift and shift PR I plan to create a new backlog item where we can discuss changes to the API structure

ahoefling: To directly answer your question, I agree we should move non-logging methods to a separate interface. Any thoughts on naming?

Perhaps we call it the `IEventLogManager` or `IEventLogService`. The challenge with a separate interface will be too many interfaces with similar names as it could cause confusion.

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
ahoefling: `IClonable` is just something that appears on the `ArrayList` interface.

I can update my implementation to just use `ICollection<ILogDetailInfo>` and the existing non-generic interfaces

bdukes: I don't think we need to non-generic interfaces, either.  What benefit/need is there for those?

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
bdukes: How about `IEventLogConfigService`?

mitchelsellers: I just want to say a few things here....

1.  Totally support this, but I think we non-logging stuff to `IEventLogConfigManager` as I think that follows our new convention?  (Or would _Service be more appropriate.)
2. I'm 1000000% on board with `ILogger<T>` support!  As ultimately I want to be able to add SeriLog

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
mitchelsellers: I would support `ICollection<>` ArrayList support is from pre-generics era

DNN Platform/Library/Services/Log/EventLog/LogProperties.cs
```diff
@@ -2,17 +2,21 @@
...
-    using System.Xml;
+    using System.Xml;
 
+    using DotNetNuke.Abstractions.Logging;
     using DotNetNuke.Common.Utilities;
 
-    public class LogProperties : ArrayList
-    {
+    /// <inheritdoc />
+    public class LogProperties : ArrayList, ILogProperties
```
mitchelsellers, line 17: As we moving forward, do we really need to create our own collection property?  Should this possible be marked obsolete to get around it?  Not sure of upstream usage

mitchelsellers (changes requested):

DNN Platform/Library/Services/Log/EventLog/LogProperties.cs
bdukes: If it wasn't for that `Summary` property, we could easily replace this with `IDictionary<string, string>` as @ahoefling suggested in a different chain.

One thing we need to keep in mind as we look toward future improvements is that adding members to interfaces is a breaking change, so we can't just create `AddLog(ILogProperties)` now and then deprecate it and add `AddLog(IDictionary<string, string>)` without it being a breaking change.

ahoefling:
Thank you @bdukes and @mitchelsellers for the review on this PR. I am glad to see excitement around this PR.

@mitchelsellers 
> I'm 1000000% on board with ILogger<T> support! As ultimately I want to be able to add SeriLog

My intentions with DNN logging is to isolate the IEventLogger from a general ILogger. We have 2 major types of logging in DNN  and they serve very different purposes. The Event Logger is a database logger which is designed to log application events as the name suggests. Where the file logger uses log4net to log application errors and exceptions to a file. In modern applications it is best practice to use a file logger for that as database logging will affect database performance and indexes since log tables grow incredibly fast.

In my experience from the mobile and desktop world Event Logging is incredibly useful to track feature usage among other things. It isn't really a technique I've seen leveraged as much in the web world but is still applicable. As it stands right now I am not planning to implement the ILogger<T> for the IEventLogger as the API structures are quite different. 

When you log an event you typically need a couple items
- Event Name
- Properties - IDictionary<string, string>

When you use ILogger<T> you are using one of the methods below (this isn't a complete list)
- Info
- Trace
- Warning
- Debug
- Exception
- Critical

Each of these statements display the log differently in the file. 

## Combining the logging interfaces
The challenge with wrapping the IEventLogger into the ILogger<T> is the interfaces just don't line up. I think it would be more confusing on what the logger is actually doing. If the maintainers think this is the best direction then we can create an AggregateLogger which would use many ILogger<T> to log messages. For example it could invoke both a file log and an event log.

I would propose that we have an IEventLogger and an ILogger to handle the different logging scenarios. It may be useful to create an `IDnnLogger` that inherited from both to create a simple interface developers can use, but I am not convinced on that idea.

mitchelsellers:
@ahoefling 100% agreed here, two separate things!  I'm more excited about the second one than this one, but that's incredibly self serving!

bdukes:
I _think_ @mitchelsellers was just expressing excitement for [your statement](https://github.com/dnnsoftware/Dnn.Platform/pull/4159#discussion_r501148301)
>Ultimately my goal would be to submit a PR that removes log4net and uses Microsoft.Extensions.Logging that would use the ILogger<T> that ships with that NuGet.

I'm not expecting (and I don't think @mitchelsellers is either) that it would be part of the event logger.  We've discussed replacing Log4Net with Serilog in #2801, and your issue #3161 documents a path to deprecate `ILog` and migrate to `ILogger`.  It'll be a good day when we get there 😃

ahoefling:
Thanks @mitchelsellers and @bdukes for the clarification. I wanted to be thorough in my explanation as these are concepts that are often confused when to use them. They both serve very important purposes and want to make sure we are all on the same page.

**updates are on the way**

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
ahoefling: Fixed in latest code push, resolving thread

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
ahoefling: Fixed in latest code push, resolving thread

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
ahoefling: Fixed in latest code push, resolving thread

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeInfo.cs
ahoefling: Fixed in latest code push, resolving thread

DNN Platform/DotNetNuke.Abstractions/Logging/LoggingProviderReturnType.cs
ahoefling: Fixed in latest code push, resolving thread

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
ahoefling: This makes sense, I didn't create that interface and it should be added

DNN Platform/DotNetNuke.Abstractions/Logging/ILogInfo.cs
ahoefling: This makes sense, I didn't create that interface and it should be added

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
ahoefling: This makes sense, I didn't create that interface and it should be added

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
ahoefling: I have added an `IEventLogConfigService` in the latest code push. I am going to resolve this thread so we can focus the conversation on individual APIs in that interface

DNN Platform/Library/Services/Log/EventLog/LogProperties.cs
ahoefling: I have a lot of thoughts about this, I really don't like the `LogProperties` and understand it was built during a time before C# generics. In the mobile/desktop world we typically have an event logger interface with the following API

```c#
void TrackEvent(string name, IDictionary<string, string> properties);
```

I would argue that `LogProperties` should be updated to meet that need but I don't think it is appropriate for this PR. My goal for this PR was a lift and shift for Dependency Injection. Once we accept this PR I think it will make sense to create a new API specification that simplifies EventLogging. I think the current API creates a barrier to entry as it requires a lot of guess and check work to create custom events. When using the API it should be as simple as the code snippet in this comment.

Is there any actionable item I need to take on this thread?

ahoefling:
I have updated the PR with the changes suggested. If there are no other code changes we need to make can we start looking at the XML comments of the new interfaces in `DotNetNuke.Abstractions.Logging`. None of the objects I migrated to Abstractions had any API documentation and I was not very thorough. I would like us to try and get property explanations prior to merging this PR.

Thanks everyone for all the feedback so far!

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Adds the log.
+        /// </summary>
+        /// <param name="logType">The log type.</param>
+        void AddLogType(ILogTypeInfo logType);
+
+        /// <summary>
+        /// Clears the log.
+        /// </summary>
+        void ClearLog();
```
bdukes, line 37: Is this a config API?  I would think of this as a (dangerous) log-level config.  I'd be okay leaving this out altogether (some implementations might not have a way to clear the log).

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <summary>
+        /// Deletes the log.
+        /// </summary>
+        /// <param name="logInfo">the log info.</param>
+        void DeleteLog(ILogInfo logInfo);
```
bdukes, line 43: This one doesn't seem like it's config related either.  Does this fit better on `IEventLogger`?

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Get the logs.
+        /// </summary>
+        /// <param name="portalID">The portal id.</param>
+        /// <param name="logType">The log type.</param>
+        /// <param name="pageSize">The page size.</param>
+        /// <param name="pageIndex">The page index.</param>
+        /// <param name="totalRecords">The total records.</param>
+        /// <returns>The logs.</returns>
+        IEnumerable<ILogInfo> GetLogs(int portalID, string logType, int pageSize, int pageIndex, ref int totalRecords);
```
bdukes, line 66: I think this should be on `IEventLogger`.

Or a new interface, e.g. `IEventLogService`, could have a more complete CRUD interface, while `IEventLogger` is focused on the 99% use case of just logging stuff.  Not sure what's the right conceptual limit of interfaces.

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        /// <param name="log">The log to retrieve.</param>
+        /// <param name="returnType">The return type.</param>
+        /// <returns>The single log.</returns>
+        object GetSingleLog(ILogInfo log, LoggingProviderReturnType returnType);
+
+        /// <summary>
+        /// Purge the log buffer.
+        /// </summary>
+        void PurgeLogBuffer();
```
bdukes, line 98: I think this should be on `IEventLogger` or `IEventLogService`

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// Add setting log.
+        /// </summary>
+        /// <param name="logTypeKey">The log type key.</param>
+        /// <param name="idFieldName">The id field name.</param>
+        /// <param name="idValue">The id value.</param>
+        /// <param name="settingName">The setting name.</param>
+        /// <param name="settingValue">The setting value.</param>
+        /// <param name="userId">The user id.</param>
+        void AddSettingLog(EventLogType logTypeKey, string idFieldName, int idValue, string settingName, string settingValue, int userId);
+
```
bdukes, line 24: I know @mitchelsellers' already expressed his displeasure before, but this seems like a scenario that'd be perfect for an extension method.  A specific use case that's really just sugar over calling another overload.

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace DotNetNuke.Abstractions.Logging
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// add xml docs.
+    /// </summary>
+    public interface ILogProperties : ICollection<ILogDetailInfo>, IList, ICollection, IEnumerable
```
bdukes, line 13: `IEnumerable` is already part of `ICollection<T>`, so it's fine to include it or not, but it doesn't change anything.  `IList` seems too restrictive to me (on top of `ICollection` methods it also requires access by index).  And, if we're already implementing `ICollection<T>`, I'd prefer to avoid encouraging non-generic usages if we can.

```suggestion
    public interface ILogProperties : ICollection<ILogDetailInfo>
```

DNN Platform/Library/Services/Log/EventLog/LogInfo.cs
```diff
@@ -30,37 +32,64 @@ public LogInfo(string content)
...
-        }
-
-        public string LogGUID { get; set; }
-
-        public string LogFileID { get; set; }
-
+        }
+
+        /// <inheritdoc />
+        public string LogGuid { get; set; }
```
bdukes, line 38: These renamed properties need to use explicit interface implementation so that we're not exposing multiple members that only differ by case (for VB.NET support)
```suggestion
        string ILogInfo.LogGuid { get; set; }
```

DNN Platform/Library/Startup.cs
```diff
@@ -28,6 +30,8 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<ReflectedModuleControlFactory>();
             services.AddSingleton<IDnnContext, DotNetNukeContext>();
 
+            services.AddScoped<IEventLogger, EventLogController>();
```
bdukes, line 33:
```suggestion
            services.AddScoped<IEventLogger, EventLogController>();
            services.AddScoped<IEventLogConfigService, EventLogController>();
```

bdukes (approved):

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: In the purest sense of a logger, you should never be able to delete a log statement only add log statements. I think it may make sense to have a 3rd interface such as `IEventLogDeleteService`. This would allow us to have all the delete APIs and clear APIs

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: See comment below https://github.com/dnnsoftware/Dnn.Platform/pull/4159#discussion_r501757829

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: This might be the direction to go

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: The `IEventLogger` should be kept as clean as possible and only be used for adding log statements. I like the idea of an `IEventLogService`

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
ahoefling: This one kind of confused me and I wasn't sure where to put it. I'll take a closer look after I look at the rest of the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
ahoefling: This makes sense to me, I grabbed the non-generics as they come from `ArrayList`. I'll make the change

DNN Platform/Library/Services/Log/EventLog/LogInfo.cs
ahoefling: Is it still okay to deprecate the old ones?

DNN Platform/Library/Startup.cs
ahoefling: Thanks for catching the missed registration, this needs to be updated.

mitchelsellers:
I think moving "Clear" and "Delete" into a Service interface is a good compromise.  Those are needed items, but not every day etc.

ahoefling:
@bdukes thanks for the detailed review, I will go through and make the updates. Based on several discussion threads it appears that we will have a total of 3 different event logging interfaces
| Interface | Description |
|----------|---------------|
| `IEventLogger` | Used when adding new event log statements to the database. Does not contain any config or delete APIs |
| `IEventLogConfigService` | Used for creating new event log configuration types and managing those types |
| `IEventLogService` | Used for clearing logs, deleting logs and additional management APIs. This can inherit from `IEventLogger` |

With this API structure my question is can we use `IEventLogger` without the `IEventLogConfigService` is this possible?
- I think it is possible, but want to ask the question

mitchelsellers:
> With this API structure my question is can we use `IEventLogger` without the `IEventLogConfigService` is this possible?

Absolutely, general exceptions etc.

ahoefling:
Thanks for the quick response. I'll take this direction as documented in my table above. I'll update this PR later today with changes

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets the log file name with path.
+        /// </summary>
+        string LogFileNameWithPath { get; set; }
+
+        /// <summary>
+        /// Gets or sets the log type portal Id.
+        /// </summary>
+        string LogTypePortalId { get; set; }
```
ahoefling, line 71: @mitchelsellers @bdukes Do you know why this property is a `string`? Since the `PortalId` is type of `int` should we update this for type safety?

Otherwise code might look like this
```c#
ILogTypeConfigInfo configInfo = new LogTypeConfigInfo();
configInfo.LogTypePortalId = $"{PortalSettings.PortalID}";
```

ahoefling: To accomplish this we would need to add a new property `AllowOnAllPortals` something to indicate the wildcard.

I retract my statement and don't think we should make any changes. I'll leave this conversation open for a little bit in case anyone wants to comment on it

sleupold: Adding a second property could easily lead into inconsistency. I suggest using a dedicated value instead - either Integer.Null or -1 as usual.

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
ahoefling: I tried moving this to an extension method but it opens up a lot of circular dependency issues. We have moved interfaces to the Abstractions project and we can't really move the extension method to the abstraction project.

Should I add a new extension method in `DotNetNuke.Library`. If you have thoughts on this feel free to push right to the repository

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
ahoefling: Thanks @sleupold for the input but I think it is best to just leave this as is for now. I don't want to make SQL changes to the database and don't think we should be making any changes to stuff like this unless we are doing a major overhaul of the database

DNN Platform/Library/Startup.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/Library/Services/Log/EventLog/LogInfo.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/ILogProperties.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
ahoefling: This has been updated in the latest code push, thanks for the feedback

DNN Platform/DotNetNuke.Abstractions/Logging/ILogTypeConfigInfo.cs
sleupold: @ahoefling I agree, but IMO object properties should be mapped 1:1 to database columns (or rows, if stored serialized, lke settings), otherwise it becomes confusing or even inconsistent. 
If we don't modify the database, AllowOnAllPortals wuld need to be added as a calculated column based on PortalID.

ahoefling: @sleupold we are in agreement, technically the `string LogTypePortalId` is mapped 1:1 to the database column. It lacks any foreign key constraint and can't be indexed correctly.  This is a classic example of a SQL anti-pattern, but it has been in DNN for so long it will cause more harm today to fix it and make a breaking change.

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Delete the log type config info.
+        /// </summary>
+        /// <param name="logTypeConfig">The log type config.</param>
+        void DeleteLogTypeConfigInfo(ILogTypeConfigInfo logTypeConfig);
+
+        /// <summary>
+        /// Gets the log type config info.
+        /// </summary>
+        /// <returns>The logs.</returns>
+        ArrayList GetLogTypeConfigInfo();
```
ahoefling, line 50: I think it will make more sense to use an `IEnumerable<>` instead of `ArrayList`

```suggestion
        IEnumerable<ILogTypeConfigInfo> GetLogTypeConfigInfo();
```

bdukes: Definitely

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogConfigService.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Delete an existing <see cref="ILogTypeConfigInfo"/> from the database..
+        /// </summary>
+        /// <param name="logTypeConfig">The log type config.</param>
+        void DeleteLogTypeConfigInfo(ILogTypeConfigInfo logTypeConfig);
+
+        /// <summary>
+        /// Gets all the <see cref="ILogTypeConfigInfo"/> from the database.
+        /// </summary>
+        /// <returns>The logs.</returns>
+        IEnumerable<ILogTypeConfigInfo> GetLogTypeConfigInfo();
```
ahoefling, line 50: @bdukes @mitchelsellers 

I changed the API from `ArrayList` to `IEnumerable<>` as it made more sense to use generics in the new abstractions API. Do you have any problems with this change?

bdukes: 👍🏻

ahoefling:
After reviewing the changes here since this is a lift and shift I don't think we need to add any new unit tests. Just need to ensure the existing tests are still passing. If anyone disagrees, I am more than willing to add unit tests in particular areas

bdukes:
Works for me

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <returns>The logs.</returns>
+        IEnumerable<ILogInfo> GetLogs(int portalID, string logType, int pageSize, int pageIndex, ref int totalRecords);
+
+        /// <summary>
+        /// Get a single log.
+        /// </summary>
+        /// <param name="log">The log to retrieve.</param>
+        /// <param name="returnType">The return type.</param>
+        /// <returns>The single log.</returns>
+        object GetSingleLog(ILogInfo log, LoggingProviderReturnType returnType);
```
bdukes, line 42: Sorry I've missed this in previous reviews.  I know this effort is lift-and-shift, but I'm going to have a hard time letting this one pass…

I'd lean towards getting rid of the `LoggerProviderReturnType` altogether, and just always returning `ILogInfo`.  We can then add a `ToXmlNode` method to `ILogInfo`.

How does that feel to y'all, @ahoefling @mitchelsellers?

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
```diff
@@ -0,0 +1,49 @@
...
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Abstractions.Logging
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The event log service provides APIs for
+    /// managing existing Event Logs in the database.
+    /// </summary>
+    public interface IEventLogService
```
bdukes, line 12: @ahoefling you had talked about `IEventLogService` implementing `IEventLogger` also.  Did I misunderstand, or was that just missed?

DNN Platform/Library/Obsolete/LogTypeConfigInfo.cs
```diff
@@ -0,0 +1,26 @@
...
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Services.Log.EventLog
+{
+    using System;
+
+    using DotNetNuke.Abstractions.Logging;
+
+    public partial class LogTypeConfigInfo
+    {
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogTypeConfigInfo.Id' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 12:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogTypeConfigInfo.Id' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogTypeConfigInfo.cs
```diff
@@ -0,0 +1,26 @@
...
+        public string ID
+        {
+            get => ((ILogTypeConfigInfo)this).Id;
+            set => ((ILogTypeConfigInfo)this).Id = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogTypeConfigInfo.LogTypePortalID' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 19:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogTypeConfigInfo.LogTypePortalId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Services.Log.EventLog
+{
+    using System;
+
+    using DotNetNuke.Abstractions.Logging;
+
+    public partial class LogInfo
+    {
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogGuid' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 12:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogGuid' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        public string LogGUID
+        {
+            get => ((ILogInfo)this).LogGuid;
+            set => ((ILogInfo)this).LogGuid = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogFileId' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 19:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogFileId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        public string LogFileID
+        {
+            get => ((ILogInfo)this).LogFileId;
+            set => ((ILogInfo)this).LogFileId = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogUserId' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 26:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogUserId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        public int LogUserID
+        {
+            get => ((ILogInfo)this).LogUserId;
+            set => ((ILogInfo)this).LogUserId = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogEventId' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 33:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogEventId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        public int LogEventID
+        {
+            get => ((ILogInfo)this).LogEventId;
+            set => ((ILogInfo)this).LogEventId = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogPortalId' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 40:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogPortalId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        public int LogPortalID
+        {
+            get => ((ILogInfo)this).LogPortalId;
+            set => ((ILogInfo)this).LogPortalId = value;
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogInfo.LogConfigId' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 47:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogInfo.LogConfigId' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/LogTypeInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Services.Log.EventLog
+{
+    using System;
+
+    using DotNetNuke.Abstractions.Logging;
+
+    public partial class LogTypeInfo
+    {
+        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Services.Log.EventLog.LogTypeCSSClass.LogTypeCSSClass' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 12:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use 'DotNetNuke.Abstractions.Logging.ILogTypeInfo.LogTypeCssClass' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+            WEBSERVER_CREATED = 148,
+            WEBSERVER_UPDATED = 149,
+            WEBSERVER_DISABLED = 150,
+            WEBSERVER_ENABLED = 151,
+            WEBSERVER_PINGFAILED = 152,
+            FOLDER_MOVED = 153,
+        }
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
+
```
bdukes, line 181:
```suggestion
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
+        public void AddLog(object businessObject, IPortalSettings portalSettings, int userID, string userName, string logType) =>
+            this.EventLogger.AddLog(businessObject, portalSettings, userID, userName, logType);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
+        public void AddLog(LogInfo logInfo) =>
+            this.EventLogger.AddLog(logInfo);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 254:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void AddLogType(string configFile, string fallbackConfigFile) =>
+            this.EventLogConfigService.AddLogType(configFile, fallbackConfigFile);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 258:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void AddLogType(LogTypeInfo logType) =>
+            this.EventLogConfigService.AddLogType(logType);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 262:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void AddLogTypeConfigInfo(LogTypeConfigInfo logTypeConfig) =>
+            this.EventLogConfigService.AddLogTypeConfigInfo(logTypeConfig);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 266:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void ClearLog() =>
+            this.EventLogService.ClearLog();
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
+        public void DeleteLog(LogInfo logInfo) =>
+            this.EventLogService.DeleteLog(logInfo);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 274:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void DeleteLogType(LogTypeInfo logType) =>
+            this.EventLogConfigService.DeleteLogType(logType);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 278:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void DeleteLogTypeConfigInfo(LogTypeConfigInfo logTypeConfig) =>
+            this.EventLogConfigService.DeleteLogTypeConfigInfo(logTypeConfig);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 282:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public List<LogInfo> GetLogs(int portalID, string logType, int pageSize, int pageIndex, ref int totalRecords) =>
+            this.EventLogService.GetLogs(portalID, logType, pageSize, pageIndex, ref totalRecords).Cast<LogInfo>().ToList();
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 286:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public ArrayList GetLogTypeConfigInfo() =>
+            LogController.Instance.GetLogTypeConfigInfo();
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 290:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public LogTypeConfigInfo GetLogTypeConfigInfoByID(string id) =>
+            (LogTypeConfigInfo)this.EventLogConfigService.GetLogTypeConfigInfoByID(id);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 294:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public Dictionary<string, LogTypeInfo> GetLogTypeInfoDictionary() =>
+            this.EventLogConfigService
+                .GetLogTypeInfoDictionary()
+                .ToDictionary(key => key.Key, value => (LogTypeInfo)value.Value);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 300:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public object GetSingleLog(LogInfo log, LoggingProvider.ReturnType returnType) =>
+            this.EventLogService.GetSingleLog(log, (LoggingProviderReturnType)returnType);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 304:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void PurgeLogBuffer() =>
+            this.EventLogService.PurgeLogBuffer();
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 308:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

DNN Platform/Library/Obsolete/EventLogController.cs
```diff
@@ -0,0 +1,322 @@
...
+        public virtual void UpdateLogTypeConfigInfo(LogTypeConfigInfo logTypeConfig) =>
+            this.EventLogConfigService.UpdateLogTypeConfigInfo(logTypeConfig);
+
+        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogger' instead. Scheduled for removal in v11.0.0.")]
```
bdukes, line 312:
```suggestion
        [Obsolete("Deprecated in 9.8.0. Use Dependency Injection to resolve 'DotNetNuke.Abstractions.Logging.IEventLogConfigService' instead. Scheduled for removal in v11.0.0.")]
```

bdukes (approved):

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
ahoefling: I am always happy to remove stuff that doesn't make sense anymore. While working through these APIs this is one that caught my attention and I have all the questions on why we still need

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
ahoefling: That was missed and I think still think the `IEventLogService` should implement `IEventLogger`. This will simplify integrations to only have 1 dependency if the code needs to log and manage the log table

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
bdukes: The XML return type is in use by the Event Log Persona Bar component, when it's creating the attachment to email log events:
https://github.com/dnnsoftware/Dnn.Platform/blob/80b6b2c56c06a058a8995196f8fa8375a28c9508/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/AdminLogs/AdminLogsController.cs#L256-L270

The implementation always fills a `LogInfo` object, and then optionally converts it to an `XmlNode` by calling `LogInfo.Serialize()` and feeding it to `XmlDocument.LoadXml`.

Also, rather than taking a whole `ILogInfo`, we can really just take a `Guid` (or `string`, since that's how we represent the GUID on `ILogInfo`), since that's all that's used to retrieve the data.

ahoefling:
@bdukes I think I covered all your comments on the deprecation notices. Can you take a 2nd look to make sure nothing was missed

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
ahoefling: Thank you for the explanation, is there any action I can take on this right now or are we waiting for @mitchelsellers to respond?

bdukes: I would say you're free to tackle it if you have time now.

ahoefling: I just pushed changes to address this concern. Let's open up new threads to comment on the code

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogService.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="totalRecords">The total records.</param>
+        /// <returns>The logs.</returns>
+        IEnumerable<ILogInfo> GetLogs(int portalID, string logType, int pageSize, int pageIndex, ref int totalRecords);
+
+        /// <summary>
+        /// Retrieves a single event log via the Log Guid.
+        /// </summary>
+        /// <param name="logGuid">A string reprenstation of the log Guid.</param>
+        /// <returns>The <see cref="ILogInfo"/>.</returns>
+        ILogInfo GetLog(string logGuid);
```
ahoefling, line 42: @bdukes @mitchelsellers I add this new API following the standards that are already in place from the `DBLoggingProvider`. I am not sure of the history on this one but it appears that searching by a varchar guid representation vs the primary key of `LogEventId` will be slower. I originally wanted to add 2 methods
- GetLog(string logGuid)
- GetLog(int logEventId)

I noticed the underlying technique to read the event log is using a stored procedure. When I saw this, I decided adding the other method isn't really a good idea right now. 

I wanted to document this and share my findings

DNN Platform/DotNetNuke.Web/InternalServices/EventLogServiceController.cs
```diff
@@ -33,17 +33,16 @@ public HttpResponseMessage GetLogDetails(string guid)
 
             try
             {
-                var logInfo = new LogInfo { LogGUID = guid };
-                logInfo = EventLogController.Instance.GetSingleLog(logInfo, LoggingProvider.ReturnType.LogInfoObjects) as LogInfo;
-                if (logInfo == null)
+                var log = EventLogController.Instance.GetLog(guid);
+                if (log == null || !(log is LogInfo))
```
bdukes, line 37:
```suggestion
                if (!(EventLogController.Instance.GetLog(guid) is LogInfo log))
```

bdukes (approved):

mitchelsellers (approved):

DNN Platform/DotNetNuke.Web/InternalServices/EventLogServiceController.cs
ahoefling: 😕 I'm not sure if I agree with this recommendation
- Isn't this doing exactly the same thing?
- Is this a faster computation?

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
ahoefling: @bdukes @mitchelsellers is there any action to take on this one?

bdukes: I did move this to an extension method.  I created `EventLogExtensions` in `DotNetNuke.Services.Log.EventLog`.

DNN Platform/DotNetNuke.Web/InternalServices/EventLogServiceController.cs
bdukes: It is doing exactly the same thing (`is` pattern matching won't match `null` for a type pattern), and it is faster b/c it's storing the cast instead of doing the cast a second time.

ahoefling: That makes sense, I much rather prefer the 2 statements and remove the null check then. I find it easier to read if we declare the `var log` on a separate line from the if statement.
```suggestion
                var log = EventLogController.Instance.GetLog(guid);
                if (!(log is LogInfo))
```

Is this a good compromise?

bdukes:
```suggestion
                var log = EventLogController.Instance.GetLog(guid);
                if (!(log is LogInfo logInfo))
```

The main missing piece is adding a variable to store the `LogInfo` instance after casting.  This will let us remove the cast later on.

DNN Platform/DotNetNuke.Web/InternalServices/EventLogServiceController.cs
```diff
@@ -33,17 +33,16 @@ public HttpResponseMessage GetLogDetails(string guid)
...
                 {
                     return this.Request.CreateResponse(HttpStatusCode.BadRequest);
                 }
 
                 return this.Request.CreateResponse(HttpStatusCode.OK, new
                 {
                     Title = Localization.GetSafeJSString("CriticalError.Error", Localization.SharedResourceFile),
-                    Content = this.GetPropertiesText(logInfo),
+                    Content = this.GetPropertiesText((LogInfo)log),
```
bdukes, line 45:
```suggestion
                    Content = this.GetPropertiesText(logInfo),
```

bdukes (approved):

DNN Platform/DotNetNuke.Web/InternalServices/EventLogServiceController.cs
ahoefling: This makes more sense, sorry about the confusion. We are going to save 1 object allocation by doing it in 1 line.

I'll make the necessary changes as you originally recommended.

ahoefling:
I just ran through a new installation with the PR build and it appears to install without issue. The event logs are being added correctly and it is visible from the persona bar 'Admin Logs' page.

I think we are good to go with this 🚀

bdukes:
Thank for all the hard work on this one @ahoefling!

PR #4161, opened 2020-10-07
Adds Dependency Injection to WebAPI Action Filters using Property Injection

ahoefling:
## Summary
This adds property injection for WebAPI filters using the current Dependency Injection model. It mimics similar techniques that are common in the Unity Dependency Injection library by creating a new `BuildUp()` extension method to manage the Property Injection.

## Why
Suppose you have an `ActionFilter` that decorates one of your WebAPI methods and you need to leverage dependency injection to access a specific service. In ASP.NET WebAPI you can't use constructor injection due to how the filters are generated in the pipeline. It is considered a best practice to use Property Injection instead of the standard constructor injection.

This can be dangerous for us to adopt into the platform as I can see property injection and the `BuildUp()` method being abused. I would like to take a closer look at the `AttributeUsage` and see if we can tightly-couple the `[Dependency]` attribute just to `ActionFilter`. **I do not want and do not think we should support property injection in DNN, this is a discussion I have had with @mitchelsellers and I am not trying to bring it back up with this PR.** I am using property injection here as it appears to be required to get Dependency Injection support in ActionFilters

## Example
```c#
public class MyCustomFilter : ActionFilterAttribute
{
    [Dependency]
    public IMyCustomService MyCustomService { get; set; }

    public override void OnActionExecuted(HttpActionExecutedContext actionExecutedContext)
    {
        // When this method executes MyCustomService will be injected and ready for use
    }
}
```

## Steps to Merge
| Task | Status |
|------|-------|
| Unit Tests | ✔ Success |
| Test Installer | ✔ Success |
| XML Documentation | ✔ Success |

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
@@ -4,6 +4,7 @@
 
 namespace DotNetNuke.Web.Api.Internal
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Web.Http;
@@ -11,13 +12,20 @@ namespace DotNetNuke.Web.Api.Internal
     using System.Web.Http.Filters;
 
     using DotNetNuke.Common;
+    using DotNetNuke.DependencyInjection.Extensions;
 
     internal class DnnActionFilterProvider : IFilterProvider
     {
+        private IServiceProvider container;
+
+        public DnnActionFilterProvider(IServiceProvider container)
+        {
+            this.container = container;
+        }
+
         /// <inheritdoc/>
         public IEnumerable<FilterInfo> GetFilters(HttpConfiguration configuration, HttpActionDescriptor actionDescriptor)
         {
-            // Requires.NotNull("configuration", configuration);
             Requires.NotNull("actionDescriptor", actionDescriptor);
 
             var controllerFilters = actionDescriptor.ControllerDescriptor.GetFilters().Select(instance => new FilterInfo(instance, FilterScope.Controller));
@@ -30,6 +38,11 @@ namespace DotNetNuke.Web.Api.Internal
             if (!overrideFilterPresent)
             {
                 allFilters.Add(new FilterInfo(new RequireHostAttribute(), FilterScope.Action));
+            }
+
+            foreach (var filterInfo in allFilters)
+            {
+                this.container.BuildUp(filterInfo.Instance);
             }
 
             return allFilters;
--- a/DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
@@ -134,7 +134,7 @@ namespace DotNetNuke.Web.Api.Internal
                 GlobalConfiguration.Configuration.Services.Replace(typeof(ITraceWriter), new TraceWriter(IsTracingEnabled()));
 
                 // replace the default action filter provider with our own
-                GlobalConfiguration.Configuration.Services.Add(typeof(IFilterProvider), new DnnActionFilterProvider());
+                GlobalConfiguration.Configuration.Services.Add(typeof(IFilterProvider), Globals.DependencyProvider.GetRequiredService<IFilterProvider>());
                 var defaultprovider = GlobalConfiguration.Configuration.Services.GetFilterProviders().Where(x => x is ActionDescriptorFilterProvider);
                 GlobalConfiguration.Configuration.Services.Remove(typeof(IFilterProvider), defaultprovider);
 
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -32,10 +32,10 @@ namespace DotNetNuke.DependencyInjection.Extensions
             var properties = GetDependencyProperties(filter.GetType());
             foreach (var property in properties)
             {
-                var service = container.GetService(property.GetType());
+                var service = container.GetService(property.PropertyType);
                 if (service != null)
                 {
-                    property.SetValue(service, filter);
+                    property.SetValue(filter, service);
                 }
             }
 
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.DependencyInjection.Extensions
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+    using System.Web.Http.Filters;
+
+    using DotNetNuke.Common.Utilities;
+
+    /// <summary>
+    /// Adds property injection extension methods.
+    /// </summary>
+    public static class BuildUpExtensions
+    {
+        /// <summary>
+        /// Injects property dependency for public or proetected
+        /// properties that are decorated with <see cref="DependencyAttribute"/>.
+        /// </summary>
+        /// <param name="container">The service provider.</param>
+        /// <param name="filter">The <see cref="IFilter"/> to inject properties.</param>
+        public static void BuildUp(this IServiceProvider container, IFilter filter)
+        {
+            if (container == null || filter == null)
+            {
+                return;
+            }
+
+            var properties = GetDependencyProperties(filter.GetType());
+            foreach (var property in properties)
+            {
+                var service = container.GetService(property.GetType());
+                if (service != null)
+                {
+                    property.SetValue(service, filter);
+                }
+            }
+
+            IEnumerable<PropertyInfo> GetDependencyProperties(Type type)
+            {
+                return CBO.Instance.GetCachedObject<IEnumerable<PropertyInfo>>(
+                    new CacheItemArgs($"WebApiDependencyProperties_{type.FullName}"),
+                    (args) => type
+                        .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                        .Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
+                        .ToList(),
+                    false);
+            }
+        }
+    }
+}
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -10,12 +10,15 @@ namespace DotNetNuke.DependencyInjection.Extensions
     using System.Web.Http.Filters;
 
     using DotNetNuke.Common.Utilities;
+    using DotNetNuke.Instrumentation;
 
     /// <summary>
     /// Adds property injection extension methods.
     /// </summary>
     public static class BuildUpExtensions
     {
+        private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(BuildUpExtensions));
+
         /// <summary>
         /// Injects property dependency for public or proetected
         /// properties that are decorated with <see cref="DependencyAttribute"/>.
@@ -35,7 +38,14 @@ namespace DotNetNuke.DependencyInjection.Extensions
                 var service = container.GetService(property.PropertyType);
                 if (service != null)
                 {
-                    property.SetValue(filter, service);
+                    try
+                    {
+                        property.SetValue(filter, service);
+                    }
+                    catch (Exception exception)
+                    {
+                        Logger.Error(exception);
+                    }
                 }
             }
 
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -20,8 +20,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(BuildUpExtensions));
 
         /// <summary>
-        /// Injects property dependency for public or proetected
-        /// properties that are decorated with <see cref="DependencyAttribute"/>.
+        /// Injects property dependency for properties that are decorated with <see cref="DependencyAttribute"/>.
         /// </summary>
         /// <param name="container">The service provider.</param>
         /// <param name="filter">The <see cref="IFilter"/> to inject properties.</param>
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -45,7 +45,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
                     new CacheItemArgs($"WebApiDependencyProperties_{type.FullName}"),
                     (args) => type
                         .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
-                        .Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
+                        .Where(propertyInfo => propertyInfo.GetSetMethod(true) != null && propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
                         .ToList(),
                     false);
             }
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -56,7 +56,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
                         .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
                         .Where(propertyInfo => propertyInfo.GetSetMethod(true) != null && propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
                         .ToList(),
-                    false);
+                    saveInDictionary: false);
             }
         }
     }
--- a/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
+++ b/DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
@@ -15,7 +15,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
     /// <summary>
     /// Adds property injection extension methods.
     /// </summary>
-    public static class BuildUpExtensions
+    internal static class BuildUpExtensions
     {
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(BuildUpExtensions));
 
@@ -25,7 +25,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
         /// </summary>
         /// <param name="container">The service provider.</param>
         /// <param name="filter">The <see cref="IFilter"/> to inject properties.</param>
-        public static void BuildUp(this IServiceProvider container, IFilter filter)
+        internal static void BuildUp(this IServiceProvider container, IFilter filter)
         {
             if (container == null || filter == null)
             {
--- a/DNN Platform/DotNetNuke.Web/Startup.cs
+++ b/DNN Platform/DotNetNuke.Web/Startup.cs
@@ -3,7 +3,10 @@
 // See the LICENSE file in the project root for more information
 namespace DotNetNuke.Web
 {
+    using System.Web.Http.Filters;
+
     using DotNetNuke.DependencyInjection;
+    using DotNetNuke.Web.Api.Internal;
     using DotNetNuke.Web.Extensions;
 
     using Microsoft.Extensions.DependencyInjection;
@@ -14,6 +17,7 @@ namespace DotNetNuke.Web
         /// <inheritdoc />
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddTransient<IFilterProvider, DnnActionFilterProvider>();
             services.AddWebApi();
         }
     }
--- a/DNN Platform/Tests/DotNetNuke.Tests.Web/Api/DnnActionFilterProviderTests.cs
+++ b/DNN Platform/Tests/DotNetNuke.Tests.Web/Api/DnnActionFilterProviderTests.cs
@@ -4,12 +4,16 @@
 
 namespace DotNetNuke.Tests.Web.Api
 {
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
     using System.Collections.ObjectModel;
     using System.Linq;
+    using System.Reflection;
     using System.Web.Http;
     using System.Web.Http.Controllers;
     using System.Web.Http.Filters;
-
+    using DotNetNuke.Common.Utilities;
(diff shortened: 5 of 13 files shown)


DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace DotNetNuke.DependencyInjection.Extensions
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+
+    /// <summary>
+    /// Adds property injection extension methods.
+    /// </summary>
+    public static class BuildUpExtensions
```
mitchelsellers, line 13: One possible thought here, not that this is a good thing...but a thought.  Should this be moved as an internal to the implementation in the action filter area to prevent mis-use?

DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
```diff
@@ -3,17 +3,25 @@
...
+    {
+        private IServiceProvider container;
+        public DnnActionFilterProvider(IServiceProvider container)
+        {
+            this.container = container;
+        }
+
         public IEnumerable<FilterInfo> GetFilters(HttpConfiguration configuration, HttpActionDescriptor actionDescriptor)
         {
             // Requires.NotNull("configuration", configuration);
```
mitchelsellers, line 27: While we are in here can we remove the commented out code

mitchelsellers (changes requested):
Overall, I really like this idea as it helps with ActionFilters to use the newer API's and constructs.  But I am VERY concerned about attempted misuse.  However, by only implementing BuildUp in the code for ActionFilters we protect by proxy from misuse.

.NET Does not provide a method for us to limit attribute usage to a specific type however.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: This is something I was thinking, but it does not work without making a few other changes. The Web API code is contained in the `DotNetNuke.Web` project, which means we need to add `[InternalsVisibleTo]` assembly declaration in the Dependency Injection project. 

I think this would be an acceptable solution

DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
ahoefling: Yes we can, no problem

ahoefling:
@mitchelsellers Thanks for the review, do you have any opinions on performance implications of this PR?

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
mitchelsellers: Honestly, I believe that is a great solution for this, and I'm sure other things that we need to do, but only for internal benefit.  it would prevent us from just making everything public all the time

mitchelsellers:
> @mitchelsellers Thanks for the review, do you have any opinions on performance implications of this PR?

At this point in time, nothing concrete necessarily.  There would be some impact, but traditionally Action Filters will not have many members/properties so it shouldn't be too big of a deal.

I think I'd want to try and profile this one

ahoefling:
My biggest concern is action filters that run before or during a request. If I remember correctly from some enterprise projects I worked on years ago a blocking thread through cause performance issues with the `ApiController`. 

@mitchelsellers 
> I think I'd want to try and profile this one

I agree and I think we should test the following items prior to accepting this PR
- Performance Profile
- DotNetBenchmarking Results (if they are easy to obtain)
- Memory Leak Analysis - I don't want another DI memory leak to come up when we should be able to load test it prior to merging

Are we in agreement that we should move forward and turn this PR into a mergable PR?

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Injects property dependency for public or proetected
```
bdukes, line 16:
```suggestion
        /// Injects property dependency for public or protected
```

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static void BuildUp(this IServiceProvider container, object myObject)
+        {
+            if (container == null || myObject == null)
+            {
+                return;
+            }
+
+            var properties = myObject
+                .GetType()
+                .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
```
bdukes, line 30: It might be worth profiling whether caching would help here.  If we're doing this many times per request, looking up the properties to fill might become a hot path to avoid.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public static class BuildUpExtensions
+    {
+        /// <summary>
+        /// Injects property dependency for public or proetected
+        /// properties that are decorated with <see cref="DependencyAttribute"/>.
+        /// </summary>
+        /// <param name="container">The service provider.</param>
+        /// <param name="myObject">The object to inject properties.</param>
+        public static void BuildUp(this IServiceProvider container, object myObject)
```
bdukes, line 21: Do we want this method to take an `ActionFilter`, to better document the expected usage?  We'd need to be explicit to create an overload if there was another type we wanted to support.

DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
```diff
@@ -3,17 +3,25 @@
...
 
-    using DotNetNuke.Common;
+    using DotNetNuke.Common;
+    using DotNetNuke.DependencyInjection.Extensions;
 
     internal class DnnActionFilterProvider : IFilterProvider
-    {
+    {
+        private IServiceProvider container;
+        public DnnActionFilterProvider(IServiceProvider container)
```
bdukes, line 20: Is DI going to resolve the global container or a scoped container here?  Do we need to think about scoping?  I would hope it would automatically pull the scoped container, which should already be in use, but want to make sure we don't make the wrong assumption here.

bdukes (approved):

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: I think that is a great idea. As far as I understand all WebAPI `ActionFilter` objects implement the `IFilter` interface. It should be safe to type check on that.

I would like to extend this code for MVC filters as well let's focus on WebAPI for now

mitchelsellers: I really like the IFilter or similar limitation here.  Great idea @bdukes

DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
ahoefling: As far as I understand it will resolve the scoped provider. @mitchelsellers do you know more about this?

This is the technically correct way to access the `IServiceProvider` instead of using the `Globals.DependencyProvider`. I thought since we can register the `DnnActionFilterProvider` let's just resolve the `IServiceProvider` in the constructor

ahoefling: How do we verify this? Do we have an easy way where we can just attach a debugger and check the `IServiceProvider`?

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: Won't caching the objects cause side-effects if 2 different users make the same request? Wouldn't user 1 get the scoped service from user 2 or vice-versa?

bdukes: I'm thinking about caching the property lookup, not the object itself.

```c#
var properties = GetDependencyProperties(myObject.GetType());

private IEnumerable<PropertyInfo> GetDependencyProperties(Type type)
{
    return CBO.GetCachedObject<IEnumerable<PropertyInfo>(
        new CacheItemArgs($"DependencyProperties_{type.TypeFullName}"),
        (args) => 
            type
                .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
                .Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
                .ToList());
}
```

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: This needs to be moved into `DotNetNuke.Web` because I don't want `DotNetNuke.DependencyInjection` to take a dependency on `System.Web`. 

The new plan will be keeping the exact same namespace but move it into the `DotNetNuke.Web` project. This is a common technique I've seen in various DI projects across .NET

ahoefling: I'm going to resolve this conversation since I have moved the code. Please open a new one if you think there needs more discussion

DNN Platform/DotNetNuke.Web/Api/Internal/DnnActionFilterProvider.cs
ahoefling: Removed stale code in latest update

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: I have added CBO caching in the latest code push

ahoefling:
I made changes as requested in the initial review and fixed the failing unit tests. This should generate a build for us to take a closer look at and see how and if this all works. After investigating the unit tests I think we can add quite a few unit tests around this new code to ensure it is thoroughly tested. I would like to do that prior to merging it

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+                }
+            }
+
+            IEnumerable<PropertyInfo> GetDependencyProperties(Type type)
+            {
+                return CBO.Instance.GetCachedObject<IEnumerable<PropertyInfo>>(
+                    new CacheItemArgs($"WebApiDependencyProperties_{type.FullName}"),
+                    (args) => type
+                        .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                        ?.Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null),
```
bdukes, line 48: I'm removing the `?.` from the `Where` call b/c [`GetProperties`](https://docs.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=netframework-4.8#System_Type_GetProperties_System_Reflection_BindingFlags_) should never return `null`.

I'm also adding `ToList` b/c we want to cache the results of the query.  Otherwise we're just caching the query, and it'll go back to the data source each time it's called.

```suggestion
                        .Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
                        .ToList(),
```

bdukes (approved):
Looks great, thanks!

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
ahoefling: This makes sense, thank you for the explanation

DNN Platform/DotNetNuke.DependencyInjection/Extensions/BuildUpExtensions.cs
ahoefling: This has been updated in the latest code push to use `IFilter`. I'm going to resolve this conversation

bdukes:
@ahoefling is our current status that we're waiting on unit tests for this one?  Should I mark this as On Hold until those come in?

ahoefling:
Correct we are waiting on a few things for this PR
- Unit Tests
- Sample Module
- Test PR installer

I haven't had a chance to work on those yet. If we want to get this in for 9.8.0 I can find time tonight

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (container == null || filter == null)
+            {
+                return;
+            }
+
+            var properties = GetDependencyProperties(filter.GetType());
+            foreach (var property in properties)
+            {
+                var service = container.GetService(property.GetType());
```
ahoefling, line 35: This is incorrect, it will return the PropertyInfo as the type and not the correct property type. This needs to be updated to

```suggestion
                var service = container.GetService(property.PropertyType);
```

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                if (service != null)
+                {
+                    property.SetValue(service, filter);
```
ahoefling, line 38: The parameters on this are backwards

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+            IEnumerable<PropertyInfo> GetDependencyProperties(Type type)
+            {
+                return CBO.Instance.GetCachedObject<IEnumerable<PropertyInfo>>(
+                    new CacheItemArgs($"WebApiDependencyProperties_{type.FullName}"),
+                    (args) => type
+                        .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                        .Where(propertyInfo => propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
+                        .ToList(),
+                    false);
+            }
```
ahoefling, line 51: We should update this to check for properties that have an accessor (set method). If this is not true we will most likely see runtime exceptions.

I would expect the code snippet below to throw an exception during the buildup

```c#
public IEventLogger Logger { get; }
```

ahoefling: This has been fixed in the latest code push

ahoefling:
I have updated the accessor of the BuildUp API to internal and added unit tests.

If there is no other feedback on this, I think it is good to go

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Adds property injection extension methods.
+    /// </summary>
+    internal static class BuildUpExtensions
+    {
+        private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(BuildUpExtensions));
+
+        /// <summary>
+        /// Injects property dependency for public or proetected
+        /// properties that are decorated with <see cref="DependencyAttribute"/>.
```
bdukes, line 24:
```suggestion
        /// Injects property dependency for properties that are decorated with <see cref="DependencyAttribute"/>.
```

DNN Platform/DotNetNuke.Web/Startup.cs
```diff
@@ -14,6 +17,7 @@ public class Startup : IDnnStartup
         /// <inheritdoc />
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddTransient<IFilterProvider, DnnActionFilterProvider>();
```
bdukes, line 20: Does it make sense for this to be a singleton, since it's only manually accessed once and then stored in global configuration?

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            IEnumerable<PropertyInfo> GetDependencyProperties(Type type)
+            {
+                return CBO.Instance.GetCachedObject<IEnumerable<PropertyInfo>>(
+                    new CacheItemArgs($"WebApiDependencyProperties_{type.FullName}"),
+                    (args) => type
+                        .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                        .Where(propertyInfo => propertyInfo.GetSetMethod(true) != null && propertyInfo.GetCustomAttribute<DependencyAttribute>() != null)
+                        .ToList(),
+                    false);
```
bdukes, line 60: Every time I look at this I'm confused by this `false`
```suggestion
                    saveInDictionary: false);
```

DNN Platform/Tests/DotNetNuke.Tests.Web/DependencyInjection/BuildUpExtensionsTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            var mockCbo = new Mock<ICBO>();
+            mockCbo
+                .Setup(x => x.GetCachedObject<IEnumerable<PropertyInfo>>(It.IsAny<CacheItemArgs>(), It.IsAny<CacheItemExpiredCallback>(), It.IsAny<bool>()))
+                .Returns(properties);
+
+            CBO.SetTestableInstance(mockCbo.Object);
+        }
+
+
```
bdukes, line 154: We don't need to duplicate the logic from the production code in order to test with the cache, we just need to make sure the callback gets used.  I guess this can be moved into the `SetUp`, since it no longer depends on the type being retrieved.
```suggestion
            var mockCbo = new Mock<ICBO>();
            mockCbo
                .Setup(x => x.GetCachedObject<IEnumerable<PropertyInfo>>(It.IsAny<CacheItemArgs>(), It.IsAny<CacheItemExpiredCallback>(), It.IsAny<bool>()))
                .Returns<CacheItemArgs, CacheItemExpiredCallback, bool>((args, callback, _) => (IEnumerable<PropertyInfo>)callback(args));

            CBO.SetTestableInstance(mockCbo.Object);
        }

```

bdukes (approved):
A couple of little things, but overall approach looks good.  I believe this will be in 9.8.1 or 9.9.0, since 9.8.0 RC has already started, FYI.

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
ahoefling: makes sense to me, thanks!

DNN Platform/DotNetNuke.Web/Startup.cs
ahoefling: If it is only instantiated once and then stored in the global configuration I think it makes sense to keep it as a transient. Otherwise it would store that singleton instance in the Dependency Injection container. If you think the `IFilterProvider` will be needed outside the global configuration then I agree we should make it a singleton.

DNN Platform/DotNetNuke.Web/DependencyInjection/BuildUpExtensions.cs
ahoefling: this certainly makes the code more readable. Let's do it!

DNN Platform/Tests/DotNetNuke.Tests.Web/DependencyInjection/BuildUpExtensionsTests.cs
ahoefling: 🤦‍♂️ I knew there was a way to do this but couldn't figure it out. Thanks for the tips, I'll update the unit tests

ahoefling:
@bdukes 
> I believe this will be in 9.8.1 or 9.9.0, since 9.8.0 RC has already started

Thanks for the update on when this will be merged. If the maintainers are happy with how this PR is looking I am thinking of starting a 2nd PR for MVC filters.

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/DnnActionFilterProviderTests.cs
```diff
@@ -3,24 +3,45 @@
...
     public class DnnActionFilterProviderTests
-    {
+    {
+        [SetUp]
+        public void SetUp()
+        {
+            var mockCache = new Mock<ICBO>();
+            mockCache
+                .Setup(x => x.GetCachedObject<IEnumerable<PropertyInfo>>(It.IsAny<CacheItemArgs>(), It.IsAny<CacheItemExpiredCallback>(), It.IsAny<bool>()))
+                .Returns(new PropertyInfo[0]);
```
bdukes, line 31: This `SetUp` could probably also use a pass-through logic, rather than hardcoding the return value.  Here's a generic setup for `ICBO.GetCachedObject`
```suggestion
            var mockCache = new Mock<ICBO>();
            mockCache
                .Setup(x => x.GetCachedObject<It.IsAnyType>(It.IsAny<CacheItemArgs>(), It.IsAny<CacheItemExpiredCallback>(), It.IsAny<bool>()))
                .Returns(new InvocationFunc(invocation => ((CacheItemExpiredCallback)invocation.Arguments[1])((CacheItemArgs)invocation.Arguments[0])));
```

bdukes: This should probably be put in a test helper class, since it's useful in many places.

bdukes:
:+1: go ahead on the PR for MVC filters

mitchelsellers (approved):

DNN Platform/Tests/DotNetNuke.Tests.Web/DependencyInjection/BuildUpExtensionsTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            this.container.BuildUp(filter);
+            Assert.Throws<NullReferenceException>(() => filter.OnActionExecuted(null));
+
+            this.VerifyEventLoggerInvoked(Times.Never());
+        }
+
+        void VerifyEventLoggerInvoked() => this.VerifyEventLoggerInvoked(Times.Once());
+
+        void VerifyEventLoggerInvoked(Times times) =>
```
bdukes, line 139:
```suggestion
        private void VerifyEventLoggerInvoked() => this.VerifyEventLoggerInvoked(Times.Once());

        private void VerifyEventLoggerInvoked(Times times) =>
```

DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
```diff
@@ -20,7 +20,8 @@ namespace DotNetNuke.Web.Api.Internal
     using DotNetNuke.Services.Localization;
     using DotNetNuke.Web.Api.Auth;
     using DotNetNuke.Web.Api.Internal.Auth;
-    using DotNetNuke.Web.ConfigSection;
+    using DotNetNuke.Web.ConfigSection;
+    using Microsoft.Extensions.DependencyInjection;
 
     using Microsoft.Extensions.DependencyInjection;
```
valadas, line 26: Looks like there is a duplicate using here

PR #4171, opened 2020-10-11
Reset redirect url for redirecting to 'Redirect After Login' page

### LINKED ISSUE #3661: Redirect after login not working if Login page has a different name than login
## Description of bug
If a page to be redirected after login is set in Settings > Security tab Login settings but the login page name is called different from Login (i.e. Enter) the redirect page is ignored.

## Steps to reproduce
1. Create a login page not named Login (buy the way this is trapped by DNN and one cannot create a page called Login!) and insert the login module in it. Then set that page to be the login page of the portal.
2. Set a page to be redirected to after login
3. Then try to login 

## Current behavior
The user is redirected to the home page instead of the intended page (the redirect after login page)

## Expected behavior
The user should be directed to the redirect after login page

## Error information
N/A

## Additional context
The error is caused by the following function in this source file Dnn.Platform-9.5.0\DNN Platform\Website\DesktopModules\Admin\Authentication\Login.ascx.cs at line 256:

private bool IsRedirectingFromLoginUrl()
{
      return Request.UrlReferrer != null &&
                Request.UrlReferrer.LocalPath.ToLowerInvariant().EndsWith(LOGIN_PATH);
}
because LOGIN_PATH is declared as follows
private const string LOGIN_PATH = "/login";
If the page is called Enter the LocalPath will be something like "/en-Us/Enter" and of course will never end with "/Login" thus returning always false if the login page is not called Login.

## Affected version
* [x]  9.5.0 latest supported release

kmuralidaran:
Fixes #3661

## Summary
Reset redirectURL value so that it will be empty for populating redirectURL with 'Redirect After Login' value

### DIFF
--- a/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
@@ -80,9 +80,8 @@ namespace DotNetNuke.Modules.Admin.Authentication
             get
             {
                 var redirectURL = string.Empty;
-                var redirectAfterLoginUrl = Convert.ToInt32(GetSetting(this.PortalId, "Redirect_AfterLogin"));
-                var isValidRedirectAfterLoginUrl = this.NeedRedirectAfterLogin
-                    && redirectAfterLoginUrl != Null.NullInteger;
+
+                var setting = GetSetting(this.PortalId, "Redirect_AfterLogin");
 
                 // first we need to check if there is a returnurl
                 if (this.Request.QueryString["returnurl"] != null)
@@ -112,12 +111,6 @@ namespace DotNetNuke.Modules.Admin.Authentication
                     redirectURL = UrlUtils.ValidReturnUrl(redirectURL);
                 }
 
-                // Reset redirect url for redirecting to 'Redirect After Login' page
-                if (isValidRedirectAfterLoginUrl)
-                {
-                    redirectURL = string.Empty;
-                }
-
                 var alias = this.PortalAlias.HTTPAlias;
                 var comparison = StringComparison.InvariantCultureIgnoreCase;
 
@@ -127,9 +120,12 @@ namespace DotNetNuke.Modules.Admin.Authentication
 
                 if (string.IsNullOrEmpty(redirectURL) || isDefaultPage)
                 {
-                    if (isValidRedirectAfterLoginUrl && (isDefaultPage || this.IsRedirectingFromLoginUrl()))
+                    if (
+                        this.NeedRedirectAfterLogin
+                        && (isDefaultPage || this.IsRedirectingFromLoginUrl())
+                        && Convert.ToInt32(setting) != Null.NullInteger)
                     {
-                        redirectURL = this._navigationManager.NavigateURL(redirectAfterLoginUrl);
+                        redirectURL = this._navigationManager.NavigateURL(Convert.ToInt32(setting));
                     }
                     else
                     {
--- a/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
@@ -806,13 +806,13 @@ namespace DotNetNuke.Modules.Admin.Authentication
             }
 
             var tab = TabController.Instance.GetTab(this.PortalSettings.LoginTabId, this.PortalId);
-            return tab != null ? $"/{this._navigationManager.NavigateURL(tab.TabName)}" : LOGIN_PATH;
+            return tab != null ? new Uri(this._navigationManager.NavigateURL(tab.TabName)).LocalPath : LOGIN_PATH;
         }
 
         private bool IsRedirectingFromLoginUrl()
         {
             return this.Request.UrlReferrer != null &&
-                this.Request.UrlReferrer.LocalPath.EndsWith(GetLoginPath(), StringComparison.InvariantCultureIgnoreCase);
+                this.GetLoginPath().StartsWith(this.Request.UrlReferrer.LocalPath, StringComparison.InvariantCultureIgnoreCase);
         }
 
         private void AddLoginControlAttributes(AuthenticationLoginBase loginControl)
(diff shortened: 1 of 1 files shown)


DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
```diff
@@ -109,6 +110,12 @@ protected string RedirectURL
                     redirectURL = UrlUtils.ValidReturnUrl(redirectURL);
                 }
 
+                // Reset redirect url for redirecting to 'Redirect After Login' page
+                if (isValidRedirectAfterLoginUrl)
```
mitchelsellers, line 114: I do not believe that this is the proper solution, as with RedirectAfterLogin you don't want to lose the functionality of a ReturnUrl.

@david-poindexter or @hismightiness any thoughts?

mitchelsellers (changes requested):

DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
hismightiness: You're correct, Mitch.  Not retaining `ReturnUrl` would be a breaking change.  There are many workflows on other DNN sites that depend on it.

kmuralidaran: @mitchelsellers @hismightiness 
1. Thanks for you feedback. It will be helpful if you could please provide the scenarios where returnUrl should be retained when RedirectAfterLogin flag is set and we have valid redirectAfterLoginUrl value
2. I believe redirectAfterLoginUrl should be the top priority, Please correct me if I am wrong

mitchelsellers: Based on current functionality I would argue that it is the opposite of your priority.

If you have a return URL, the user is trying to login to a specific secured resource, they should be returned back there.

kmuralidaran: @mitchelsellers I feel that is the root cause of the issue https://github.com/dnnsoftware/Dnn.Platform/issues/3661. Either we should close the issue as "Function as Designed" or we should find alternate solution. 
@mitchelsellers @hismightiness Could you please validate https://github.com/dnnsoftware/Dnn.Platform/issues/3661?

valadas: I think the main cause is that LOGIN_PATH is hardcoded to /login where it should be read from whichever page is the login page in the settings. The suggested fix in the (I think 3rd) comment on the issue, looked appropriate to me I think.

Basically , getting the path from the settings instead of harcoded "login", then check if we come from there, do the redirectAfterLogin, if not respect the RetrunUrl. I believe that would fix the bug without discarding the returnUrl all together

kmuralidaran: IsRedirectingFromLoginUrl() call not hit in my case. https://github.com/dnnsoftware/Dnn.Platform/pull/4171/files#diff-09d8eba5524d6de9ef99d20e271369edL119 returns false

hismightiness: I like your suggestion, @valadas

kmuralidaran: @valadas @mitchelsellers @hismightiness Comments addressed and applied proposed solution from https://github.com/dnnsoftware/Dnn.Platform/issues/3661#issuecomment-608557613

DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
```diff
@@ -796,10 +798,21 @@ private static string ReplaceLanguage(string Url, string originalLanguage, strin
...
 
+        private string GetLoginPath()
+        {
+            if (this.PortalSettings.LoginTabId == Null.NullInteger)
+            {
+                return LOGIN_PATH;
+            }
+
+            var tab = TabController.Instance.GetTab(this.PortalSettings.LoginTabId, this.PortalId);
+            return tab != null ? $"/{tab.TabName}" : LOGIN_PATH;
```
bdukes, line 809: I don't think `$"/{tab.TabName}"` is going to work for all scenarios (e.g. what if the page name has a space in it).  The URL should be generated via `INavigationManager.NavigateURL` and then compare its `LocalPath` to `UrlReferrer.LocalPath`

kmuralidaran: LocalPath EndsWith comparison not matching

image

kmuralidaran: this.Request.UrlReferrer.LocalPath EndsWith comparison of GetLoginPath() fails whereas GetLoginPath() StartsWith comparison of this.Request.UrlReferrer.LocalPath succeeds. Updated the comparison and verified the fix

DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
```diff
@@ -796,10 +798,21 @@ private static string ReplaceLanguage(string Url, string originalLanguage, strin
...
 
+        private string GetLoginPath()
+        {
+            if (this.PortalSettings.LoginTabId == Null.NullInteger)
+            {
+                return LOGIN_PATH;
+            }
+
+            var tab = TabController.Instance.GetTab(this.PortalSettings.LoginTabId, this.PortalId);
+            return tab != null ? new Uri(this._navigationManager.NavigateURL(tab.TabName)).LocalPath : LOGIN_PATH;
```
bdukes, line 809: [`NavigateURL(string)`](https://dnndocs.com/api/DotNetNuke.Common.Globals.html#DotNetNuke_Common_Globals_NavigateURL_System_String_) takes a control key, not a tab name, which is why the generated URL is not correct.
```suggestion
            return tab != null ? new Uri(this._navigationManager.NavigateURL(this.PortalSettings.LoginTabId), UriKind.RelativeOrAbsolute).LocalPath : LOGIN_PATH;
```

kmuralidaran: Applied suggestion and reverted comparison

bdukes (approved):
Looks alright to me

valadas (approved):
Yep, that looks good to me

valadas:
@mitchelsellers can you re-review please

mitchelsellers (approved):

Issue #4167, opened 2020-10-10
Add Startup Events - IDnnStartupConfigure


ahoefling:
## Description of problem
`IDnnStartup` interface is designed to mimic the behaviors of the `Startup.cs` class from AspNetCore. It currently only provides APIs for registering services with the container. We should add a `Configure` method that runs during application startup. 

This will enable DNN extensions to access the DNN Application's startup code and have access to Dependency Injection all at the same time.

## Description of solution
Add new interface `IDnnStartupConfigure` that will define the following contract or similar
```c#
public class IDnnStartupConfigure
{
    void Configure(IApplicationBuilder app, IWebHostEnvironment env);
}
```

## Description of alternatives considered
Not applicable

## Screenshots
Not applicable

## Additional context
Not applicable

## Affected browser
Not applicable

bdukes:
What sorts of things do you expect a developer would be able to do with an `IApplicationBuilder` instance?

I like this idea as a way to run startup logic after the services container is built.  What are some scenarios you see this fulfilling?

ahoefling:
The idea here isn't so much what we can do today, but what we will be able to do tomorrow. Having the plumbing in place will go a very long way. For example Dependency Injection has been in the platform since 9.4 but is really starting to shine with all the enhancements we have made.

Some things I would like to be able to do with `IApplicationBuilder`
- Complete access to the DI container, this allows code to interact with any singletons. The reason why I am proposing this is so we can add an `AggregateEventLogger`. This would enable developers to implement their own `IEventLogger` implementation and when any API is invoked it will write to the DNN event log and wherever the other implementations go. I see an `AggregateLogger` in our future for general logging that we have been talking about
- Adding middleware support where applicable (this is a tricky one)
- This could allow a module or extension to swap out the Dependency Injection container
- Configure a DbContext for entity framework
- Add swashbuckle to some or all Web API routes

This list can go on, but these are some concepts that come to mind

This API structure is using the `Startup.cs` API from AspNetCore 2.1.1 (which makes what is used in DNN today). I think it makes the most sense to include the necessary abstration projects to try and implement this. It will get us 1-step closer to being ready for a .NET Core migration in the future. 

### Implementation Notes
I think a MVP implementation would give us the most basic functionality with these interfaces. Many of the properties on them would throw a `NotSupportedException`. The main goal here is to have a startup event that is triggered during the Application Start sequence after the container is fully built.

### Intent of Issue
[...]

PR #4174, opened 2020-10-12
Adds Application Startup Events with Dependency Injection

ahoefling:
Fixes: #4167 

## Summary
Adds new `IDnnStartupConfigure` interface to `DotNetNuke.Abstractions` project which allows developers to add startup code to their extension with Dependency Injection. 

In modern AspNetCore applications there is a `Startup.cs` file that has a `Configure` method. This method provides advanced startup initialization such as
- Registering middleware
- Configuring special loggers based on environments
- Configured connection strings for Entity Framework
- Setting up swashbuckle Web API docs
- Much More

This adds the most basic implementation via a new interface `IDnnStartupConfigure` which adds a carbon copy of the what is being used in AspNetCore 2.1

## Notable Changes
This PR adds several AspNetCore 2.1 assemblies which is the LTS build but has since been out of date. 

## Draft PR
This PR has been created as a draft as it demonstrates how we can accomplish this with AspNetCore assemblies. I want to share this with the maintainers so they can see how this would be implemented. 

### Alternative
If we decide this adds too much of a footprint in DNN, I recommend we update the interface to use custom dnn objects
```c#
public interface IDnnApplicationBuilder {}
public interface IDnnApplicationInfo {}
```

### DIFF
--- a/DNN Platform/DotNetNuke.Abstractions/Application/IApplicationInfo.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Application/IApplicationInfo.cs
@@ -5,10 +5,12 @@ namespace DotNetNuke.Abstractions.Application
 {
     using System;
 
+    using Microsoft.Extensions.Hosting;
+
     /// <summary>
     /// The Application class contains properties that describe the DotNetNuke Application.
     /// </summary>
-    public interface IApplicationInfo
+    public interface IApplicationInfo : IHostingEnvironment
     {
         /// <summary>
         /// Gets the company to which the DotNetNuke application is related.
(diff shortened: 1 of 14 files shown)


ahoefling:
The PR builds will fail until I add the following to the web.config. (existing nodes are omitted in the snippet below)
```xml
<system.web>
  <compilation debug="false" strict="false" targetFramework="4.7.2">
    <assemblies>
      <add assembly="netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" />
    </assemblies>
  </compilation>
</system.web>
```

ahoefling:
cc: @bdukes @mitchelsellers

mitchelsellers:
I believe the inclusion of .NET Core 2.1 assemblies will be a non-starter for us.  As we have no upgrade path to get to later versions.

ahoefling:
I'll close this PR and open a new one with my other technique which creates our own interfaces

Thanks for the quick response

Issue #4175, opened 2020-10-12
EventLogging During Application Startup Fails Silently


ahoefling:
## Description of bug
If you add event logging to application start prior to the `HttpContext.Current` being initialized the logs will silently fail

## Steps to reproduce
Add the following code to an application startup
```c#
var logger = Globals.DependencyProvider.GetRequiredService<IEventLogger>();
logger.AddLog("Title", "testing startup log", EventLogType.ADMIN_ALERT);
```

## Current behavior
The code will silently fail and not report any useful information to the log4net file logs (at least this was my experience)

## Expected behavior
This should properly log to the event logger even during application startup

## Screenshots
not applicable

## Error information
This bug is caused in https://github.com/dnnsoftware/Dnn.Platform/blob/0cc69d61f0966ee02d4d0a85a94a185ffd35c3d6/DNN%20Platform/Library/Services/Log/EventLog/LogController.cs#L47-L56

The root cause is because `HttpContext.Current` isn't currently accessible the website throws an exception. When this exception is thrown the log is never written. I believe this is in error since many event logs don't have a username. We should add specific exception handling to the username if block referenced here so the exception can be handled and the event log can be written correctly

## Additional context
Provide any additional context that may be helpful in understanding and/or resolving the bug.

## Affected version
* [x] 09.08.00 alpha build

## Affected browser
not applicable

ahoefling:
I'm planning to PR a fix for this

PR #4178, opened 2020-10-12
Fixes Event Log from Failing Silently during Application Startup

ahoefling:
Fixes: #4175

## Summary
Fixes a bug where the event log will silently fail if the username is not provided and the `HttpContext` is not set. This will happen if the event log is being used during the Application Startup which executes prior to the initialization of the `HttpContext`.

### DIFF
--- a/DNN Platform/Library/Services/Log/EventLog/LogController.cs
+++ b/DNN Platform/Library/Services/Log/EventLog/LogController.cs
@@ -13,6 +13,7 @@ namespace DotNetNuke.Services.Log.EventLog
     using System.Threading;
     using System.Web;
     using System.Xml;
+
     using DotNetNuke.Abstractions.Logging;
     using DotNetNuke.Common;
     using DotNetNuke.Common.Utilities;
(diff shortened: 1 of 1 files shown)


mitchelsellers (changes requested):
Do we have an ability to catch a more granular exception here?  Was it a null exception on a particular member?  (Just wanting to be sure we don't hide something unnecessarily)

ahoefling:
I took a 2nd look and the exception is thrown when access `HttpContext.Current.Request`. At this part in the life-cycle we can't do a null check, it will still throw an exception. I updated the code to now check for an `HttpException`.

To test this, if you add the following code to the last line of `DotNetNukeHttpApplication.Application_Start()`
```c#
Globals.DependencyProvider.GetRequiredService<IEventLogController>().AddLog("Test", "TEST", EventLogController.EventLogType.ADMIN_ALERT);
```

mitchelsellers (approved):
Thanks for the revision!

david-poindexter (approved):
Cool cool - thanks @ahoefling 🎉 🎉 🎉

Issue #3703, opened 2020-04-27
Removal of the WebSlice option


valadas:
## Description of problem
Well, you know IE 8 technology and such, I think we could clean that up and remove it.

## Description of solution
Kill it :)

## Description of alternatives considered
Live in 1999

## Additional context
Since we no longer run tests for the browsers (well the one) that had support for this, I think we should remove it because we are not going to really support that anyway.

We must not forget to remove the localization keys relating to this feature too.

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [x] Internet Explorer
* [ ] Edge

sleupold:
I agree, this was an IE only feature.

sleupold:
AFAIK, this is still an issue.

joshgk00:
I'm working on this issue. @sleupold  and @valadas can you point me to an example of how to include database changes? It looks like at the very least the vw_Modules would need to be updated to remove the WebSlice related columns, as well as the underlying table(s) the view is created from. As a new contributor, I'm not sure the best way to capture these changes.

david-poindexter:
@joshgk00 thanks so much for our interest in contributing!  Database changes are typically handled in an upgrade script file for DNN or the module within DNN.  I am on my phone at the moment and cannot easily point to examples, but I'll try to circle back a little later and provide some examples.

joshgk00:
@david-poindexter I see there's a Upgrade.SqlDataProvider. I take it I would add the changes to that script? Since it looks like this issue is not slated until 10.0 release, I assume when this gets merged it'll be part of the upgrade script for the release it's merged into?

david-poindexter:
Actually, just found.  @joshgk00 - you'll find the DNN scripts here for each version.  

https://github.com/dnnsoftware/Dnn.Platform/tree/develop/DNN%20Platform/Website/Providers/DataProviders/SqlDataProvider

For this one, it is marked for removal I believe in 10.00.00.  So you would need to add the necessary script to the appropriate file.  If it doesn't exist already, you would need to create it.

joshgk00:
@david-poindexter Thanks!

mitchelsellers:
@joshgk00 FOr this one, since the removal is in 10.0.0, it might be best to hold on this for a bit.  As we will be unable to merge necessarily for quite a while..  (Just making you aware)

joshgk00:
[...]

PR #4181, opened 2020-10-12
Marked all public WebSlice members as obsolete

mitchelsellers:
Part of #3703, marks items for removal in 10.x


mitchelsellers:
Made a mistake on branch, will resubmit

PR #4184, opened 2020-10-12
Resolved issue with Keyword and Description Defaults

### LINKED ISSUE #3541: Page description leaves to blank is showing Site Description
## Description of bug
Page description leaves to blank is showing Site Description
## Steps to reproduce
1. Open Persona Bar > Site Settings
2. Enter "My Website Description" in Site Description text box and save the settings
3. Open Persona Bar > Pages
4. Create a new Pages with Name: page1 and leave all other field to blank or default
5. Open Persona Bar > Pages
6. Click page1
7. Observe the page description

## Current behavior
Explain the current behavior.
page description is showing "My Website Description" which is the site description

## Expected behavior
Provide a clear and concise description of the expected behavior.
page description should show blank

## Screenshots

## Error information

## Additional context
It seems the code written so here - https://github.com/dnnsoftware/Dnn.Platform/blob/80e48fcde4694d70960253d401a8fcd6851608eb/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/Converters.cs#L104
However, it creating confusion to the user especially during site export / import because the page description changed after import to the next site (both site have different site description)

## Affected version

* [x] 10.0.0 alpha build
* [x] 9.5.0 alpha build

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

mitchelsellers:
Closes #3541 

The existing behavior when editing a tab would add the Portal level description and keyword fields.  If the user then clicked saved it would be saved to the page as content.

The defaulting of descriptions when rendered is correct, but this behavior resulted in improper values being set without the user knowing.

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/Converters.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/Converters.cs
@@ -79,9 +79,6 @@ namespace Dnn.PersonaBar.Pages.Components
             }
 
             var pageManagementController = PageManagementController.Instance;
-
-            var description = !string.IsNullOrEmpty(tab.Description) ? tab.Description : PortalSettings.Current.Description;
-            var keywords = !string.IsNullOrEmpty(tab.KeyWords) ? tab.KeyWords : PortalSettings.Current.KeyWords;
             var pageType = GetPageType(tab.Url);
 
             var file = GetFileRedirection(tab.Url);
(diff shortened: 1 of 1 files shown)


valadas (approved):
Awesome, thanks for spotting that one

david-poindexter (approved):
This goes against the originally intended behavior, but I think this approach is fine given the way `description` and `keywords` work today with search engines. They provide such little value anyway but duplicating it across all pages when page-level description and keywords are blank provides absolutely no value (actually worse).  Thanks for the PR @mitchelsellers 🎉

PR #4185, opened 2020-10-12
Anonymous users denied access to profile picture when configured as n…

tauqeer-haider:
…on-public through ImageHandler

## Summary

Anonymous users denied access to profile picture when configured as non-public through ImageHandler.


mitchelsellers (approved):

valadas (approved):
Looks good to me

PR #4194, opened 2020-10-15
Resolves multiple stylcop warnings in Library and HttpModules projects

valadas:
Resolves multiple stylcop warnings in Library and HttpModules projects


DNN Platform/Library/WebControls/WebControlBase.cs
```diff
@@ -1,95 +1,123 @@
...
+            get
+            {
+                return Globals.ResolveUrl("~/Resources/");
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the control is currently in the host menu.
+        /// </summary>
+        [Obsolete("There is no longer the concept of aa Host Page.  All pages are controlled by Permissions. Scheduled removal in v11.0.0.")]
```
bdukes, line 36: I'm not sure that it's true that there isn't a concept of a host page anymore.  It's still possible that have a page with a -1 portal ID, which would only be accessible to super users

DNN Platform/Library/WebControls/WebControlBase.cs
```diff
@@ -1,95 +1,123 @@
...
+        /// <summary>
+        /// Gets the portal settings for the portal where the WebControl is displayed.
+        /// </summary>
+        [Browsable(false)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        public PortalSettings PortalSettings
+        {
+            get
+            {
+                return (PortalSettings)PortalController.Instance.GetCurrentSettings();
```
bdukes, line 54: Rather than casting the interface, can we just mark this as obsolete, indicating they should call `PortalController.Instance.GetCurrentSettings()` directly?

bdukes (commented):
Thanks @valadas!

This build produced [37,352 warnings](https://dev.azure.com/dotnet/DNN/_build/results?buildId=40132&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=87231), compared with the previous build producing [37,507 warnings](https://dev.azure.com/dotnet/DNN/_build/results?buildId=40119&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=87619) (i.e. this removes 155 warnings).

DNN Platform/Library/WebControls/WebControlBase.cs
valadas: Makes sense, I removed the [Obsolete]

DNN Platform/Library/WebControls/WebControlBase.cs
valadas: Ok, I have deprecated PortalSettings in here, changed our usage of it in this file to TabController.CurrentPage but I left the cast in so we can have this file with 0 warnings, does that work out ?

bdukes: While the cast should practically work for the foreseeable future, it's creating a dependency that the interface usage is attempting to break (`IPortalSettings` doesn't _have to_ always be a `PortalSettings` instance).  I'd prefer to suppress the warning in this instance (in addition to deprecating).

```suggestion
#pragma warning disable 612, 618 // GetCurrentPortalSettings is obsolete
                return PortalController.Instance.GetCurrentPortalSettings();
#pragma warning restore 612, 618 
```

valadas: Works for me

bdukes (approved):

valadas:
conflicts resolved

mitchelsellers (approved):

PR #4195, opened 2020-10-15
Cleaned up UserProfile Class

valadas:
- Resolves many stylecop warnings in UserProfile
- Makes constants for retrieving profile properties obsolute and re-implements them as private constants
- Updated some upstream classes to use IPortalSettings instead of PortalSettings
- Implemented automated stylecop fixes on many other files in the same project

### DIFF
--- a/DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
+++ b/DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
@@ -58,6 +58,21 @@ namespace DotNetNuke.Entities.Users
         /// <param name="accessingUser">The accessing user.</param>
         /// <param name="targetUser">The target user.</param>
         /// <returns><c>true</c> if property accessible, otherwise <c>false</c>.</returns>
+        [Obsolete("Deprecated in 9.8, Use the overload that takes IPortalSettings instead. Scheduled removal in v10.0.0")]
+        public static bool CheckAccessLevel(PortalSettings portalSettings, ProfilePropertyDefinition property, UserInfo accessingUser, UserInfo targetUser)
+        {
+            var portalSettingsAsInterface = (IPortalSettings)portalSettings;
+            return CheckAccessLevel(portalSettingsAsInterface, property, accessingUser, targetUser);
+        }
+
+        /// <summary>
+        /// Checks whether profile property is accessible.
+        /// </summary>
+        /// <param name="portalSettings">The portal settings.</param>
+        /// <param name="property">The property.</param>
+        /// <param name="accessingUser">The accessing user.</param>
+        /// <param name="targetUser">The target user.</param>
+        /// <returns><c>true</c> if property accessible, otherwise <c>false</c>.</returns>
         public static bool CheckAccessLevel(IPortalSettings portalSettings, ProfilePropertyDefinition property, UserInfo accessingUser, UserInfo targetUser)
         {
             var isAdminUser = IsAdminUser(portalSettings, accessingUser, targetUser);
--- a/DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
+++ b/DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
@@ -1,17 +1,18 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
+
 // ReSharper disable CheckNamespace
 namespace DotNetNuke.Entities.Users
 
 // ReSharper restore CheckNamespace
 {
     using System;
-    using System.Collections.Generic;
     using System.Globalization;
     using System.Linq;
     using System.Web;
 
+    using DotNetNuke.Abstractions.Portals;
     using DotNetNuke.Common;
     using DotNetNuke.Common.Internal;
     using DotNetNuke.Common.Lists;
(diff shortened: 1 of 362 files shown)


DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
```diff
@@ -49,7 +58,7 @@ public CacheLevel Cacheability
         /// <param name="accessingUser">The accessing user.</param>
         /// <param name="targetUser">The target user.</param>
         /// <returns><c>true</c> if property accessible, otherwise <c>false</c>.</returns>
-        public static bool CheckAccessLevel(PortalSettings portalSettings, ProfilePropertyDefinition property, UserInfo accessingUser, UserInfo targetUser)
+        public static bool CheckAccessLevel(IPortalSettings portalSettings, ProfilePropertyDefinition property, UserInfo accessingUser, UserInfo targetUser)
```
bdukes, line 61: We need to keep and deprecate the `PortalSettings` overload

DNN Platform/Library/Services/Tokens/PropertyAccess/IPropertyAccess.cs
```diff
@@ -7,10 +7,32 @@ namespace DotNetNuke.Services.Tokens
...
         CacheLevel Cacheability { get; }
-
-        string GetProperty(string propertyName, string format, CultureInfo formatProvider, UserInfo accessingUser, Scope accessLevel, ref bool propertyNotFound);
+
+        /// <summary>
+        /// Gets a property.
+        /// </summary>
+        /// <param name="propertyName">The property name to get.</param>
+        /// <param name="format">The format string to format the property with.</param>
+        /// <param name="formatProvider">The <see cref="CultureInfo"/> format profider to use.</param>
```
bdukes, line 25:
```suggestion
        /// <param name="formatProvider">The format provider (e.g. <see cref="CultureInfo"/>) to use.</param>
```

DNN Platform/Library/Services/Tokens/PropertyAccess/IPropertyAccess.cs
```diff
@@ -7,10 +7,32 @@ namespace DotNetNuke.Services.Tokens
...
+        /// </summary>
         CacheLevel Cacheability { get; }
-
-        string GetProperty(string propertyName, string format, CultureInfo formatProvider, UserInfo accessingUser, Scope accessLevel, ref bool propertyNotFound);
+
+        /// <summary>
+        /// Gets a property.
+        /// </summary>
+        /// <param name="propertyName">The property name to get.</param>
+        /// <param name="format">The format string to format the property with.</param>
```
bdukes, line 24:
```suggestion
        /// <param name="format">The format string to format the property with, or <c>null</c> to use the default format.</param>
```

bdukes (commented):
Thanks!  We just need to add a deprecated `PortalSettings` overload for that `public` method, otherwise it's all good (with a couple of XML doc suggestions).

I've verified this produced [37,378 warnings](https://dev.azure.com/dotnet/DNN/_build/results?buildId=40133&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=87345) while the previous build produced [37,507 warnings](https://dev.azure.com/dotnet/DNN/_build/results?buildId=40119&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=a04569be-1b08-5c9a-7e65-a83076350a0c&l=87619).

DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
```diff
@@ -3,688 +3,710 @@
...
+    /// <summary>
+    /// The UserProfile class provides a Business Layer entity for the Users Profile.
+    /// </summary>
+    [Serializable]
+    public class UserProfile : IIndexable
+    {
+#pragma warning disable SA1310 // Field names should not contain underscore
+#pragma warning disable SA1600 // Elements should be documented
+        // Name properties
+        [Obsolete("Deprecated in 9.8. Scheduled removal in v10.0.0.")]
```
bdukes, line 32: What's the suggested replacement if you were using these fields previously?

bdukes: Maybe just say `Deprectated in v9.8.  Use the properties on this class instead. Scheduled removal in v10.0.0`?

valadas:
>I've verified this produced 37,378 warnings while the previous build produced 37,507 warnings (i.e. this removes 129 warnings).

Yeah, so we are almost done here 🤣

DNN Platform/Library/Entities/Users/Profile/ProfilePropertyAccess.cs
valadas: Makes total sense, implemented that suggestion.

DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
valadas: I have applied the comments suggestion, I do believe these should have never been made public in the first place. I would gladly be overruled, but I see no reason to expose those as we have much easier way to get to those properties.

bdukes (approved):
Looks great, thanks!

DNN Platform/Library/Entities/Users/Profile/UserProfile.cs
bdukes: Agreed, those shouldn't need to be exposed.

bdukes:
I did some analysis on the build log, and if we ignore the SA16xx documentation warnings and the CS0618 obsolete warnings, we're almost there with only a little over 10,000 warnings left.

I am this close GIF

mitchelsellers (approved):

PR #4203, opened 2020-10-16
Fixed all build warnings in Console module

valadas:
- Fixed all stylecop warnings in the Console module project
- Enabled TreatWarningsAsErrors for the project

### DIFF
--- a/DNN Platform/Admin Modules/Dnn.Modules.Console/Components/BusinessController.cs
+++ b/DNN Platform/Admin Modules/Dnn.Modules.Console/Components/BusinessController.cs
@@ -7,19 +7,13 @@ namespace Dnn.Modules.Console.Components
     using System;
 
     using DotNetNuke.Entities.Modules;
-    using DotNetNuke.Entities.Modules.Definitions;
-    using DotNetNuke.Services.Upgrade;
 
     /// <summary>
-    ///
+    /// Implements de module business controller interfaces.
     /// </summary>
     public class BusinessController : IUpgradeable
     {
-        /// <summary>
-        ///
-        /// </summary>
-        /// <param name="version"></param>
-        /// <returns></returns>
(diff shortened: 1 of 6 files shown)


mitchelsellers (approved):

DNN Platform/Admin Modules/Dnn.Modules.Console/Components/BusinessController.cs
```diff
@@ -2,41 +2,35 @@
...
-            catch (Exception)
-            {
-                return "Failed";
-            }
-        }
-    }
-}
+    /// <summary>
+    /// Implements de module business controller interfaces.
+    /// </summary>
```
bdukes, line 13:
```suggestion
    /// <summary>Implements the module's business controller interface(s).</summary>

```

DNN Platform/Admin Modules/Dnn.Modules.Console/ViewConsole.ascx.cs
```diff
@@ -1,539 +1,583 @@
...
-            {
-                if (this._consoleCtrl == null)
-                {
-                    this._consoleCtrl = new ConsoleController();
-                }
-
-                return this._consoleCtrl;
-            }
-        }
-
```
bdukes, line 72: Was this removed on purpose?

bdukes (approved):

DNN Platform/Admin Modules/Dnn.Modules.Console/ViewConsole.ascx.cs
valadas: Yes, nothing calls that and since it's not an extension point but a codebehind file, I thought it was ok to remove, but I can put it back if you think it was not a good idea...

PR #4206, opened 2020-10-16
Removed the unused json2.js from the codebase

### LINKED ISSUE #2725: Removal of json2.js
We have json2.js file here https://github.com/dnnsoftware/Dnn.Platform/blob/development/Website/Resources/Shared/scripts/json2.js but I do not see that we use it anywhere. This was used for very old browsers that did not support json serialization and deserialization (IE7 and bellow). 

Can this be safely removed or do you see any risk in removing that ?

joshgk00:
Fixes #2725 

## Summary
Removing the json2.js and associated license file from the codebase as it was only used for older browsers (IE7 and below). This file was primarily used in the 

## How to Test

1. Login either as a Super User or a regular user
2. Click on your username to view your profile
image

3. You'll see a box with a placeholder "Tell the world something" 
image

4. Type a message and click Share to post it to the Journal 
image

5. You should be able to post a message without any errors
image


valadas (approved):
This looks good to me, Anyone sees a risk doing this in a patch release or we hold into a minor release ? Do we consider the resources in the Resources/Shared forler as 3rd party extension points or not ?

bdukes:
Yeah, I think `Resources/Shared` implies it's intended to be reused.  We haven't officially supported old IE for many years, but I'd have a hard time removing the script altogether in a patch release.

valadas:
Ok, let's slate this for 9.9.0 then

mitchelsellers (approved):

valadas:
We discussed this removal in the approvers meeting and since we don't officially support the browsers this would affect, we agreed on getting this into a minor release.

bdukes:
Thanks @joshgk00!

PR #4210, opened 2020-10-18
Remove impractical rule

david-poindexter:
## Summary
@dnnsoftware/approvers I'm thinking this is a bit overkill, as I can't think of a reason why it would matter who actually clicks the merge button.  The important thing is that there are two approvals.  Do you all agree?


valadas (approved):
Yes I approve, very often we have 2 approvals and need to wait quite some time to keep working on next steps while waiting just for CI. And well the persons who can actually click that button are also approvers. So I think this makes sense.

mitchelsellers (approved):

PR #4216, opened 2020-10-18
Resolves 101 build warnings on CoreMessaging project

valadas:
- Resolves 101 warnings
- Prevents sending detailed errors to frontend
- Updates documentation and settings here and there to make it build and correct related documentation
- Enabled TreatBuildWarningsAsErrors on project

### DIFF
--- a/DNN Platform/Library/Services/Exceptions/Exceptions.cs
+++ b/DNN Platform/Library/Services/Exceptions/Exceptions.cs
@@ -34,7 +34,7 @@ namespace DotNetNuke.Services.Exceptions
     /// (Event Log) or only to SuperUsers (hosts) in the server logs.
     /// </para>
     /// <para>
-    /// Current we immplement a lot of custom exception to use in different levels:
+    /// Currently we implement a lot of custom exception to use in different levels:
     /// <list type="bullet">
     /// <item><see cref="ModuleLoadException"/></item>
     /// <item><see cref="ObjectHydrationException"/></item>
--- a/DNN Platform/Library/Services/Exceptions/Exceptions.cs
+++ b/DNN Platform/Library/Services/Exceptions/Exceptions.cs
@@ -27,15 +27,14 @@ namespace DotNetNuke.Services.Exceptions
     /// </summary>
     /// <remarks>
     /// <para>
-    /// For most developers, there is a hard problem need to face to is that our product will run on many and many servers with
-    /// much different environment, such as hardware, network, system version, framework version and so on, so there is many of reasons
-    /// will make our application throw lof of exceptions,even will stop our app to working. so when some error occured, we need a way
-    /// to find out the reason, we know we need to log all the exception, but the point is how to log useful information, you should log
-    /// the information what you need to location the code caught the error, but DONOT just log 'ERROR'. so we provide a full support of
-    /// exception log system. when error occured, we can found the detail information in event log and can locationt the error quickly.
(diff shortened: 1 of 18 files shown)


DNN Platform/Library/Services/Exceptions/Exceptions.cs
```diff
@@ -26,16 +26,15 @@ namespace DotNetNuke.Services.Exceptions
...
+    /// <para>
+    /// Logging is an import part of a good architecture. A good practice is to provide enough logging to accurately be able to identify which
+    /// code caused an exception. Some exceptions are expected and you can display information to the user to help him figure out what happened.
+    /// It is however bad practice to catch any and all exceptions (general exception) and display this to users. In such situations it is best
+    /// to provide the user with a general message and only log the details in a way that the details are either just visible to administrators
+    /// (Event Log) or only to SuperUsers (hosts) in the server logs.
     /// </para>
     /// <para>
-    /// Current we immplement lot of custom exception to use in different levels:
+    /// Current we immplement a lot of custom exception to use in different levels:
```
bdukes, line 37:
```suggestion
    /// Currently we implement a lot of custom exception to use in different levels:
```

DNN Platform/Modules/CoreMessaging/Services/MessagingServiceController.cs
```diff
@@ -1,440 +1,556 @@
...
+                var portalId = PortalController.GetEffectivePortalId(UserController.Instance.GetCurrentUserInfo().PortalID);
+
+                messageBoxView.TotalNewThreads = InternalMessagingController.Instance.CountUnreadMessages(this.UserInfo.UserID, portalId);
+                messageBoxView.TotalConversations = InternalMessagingController.Instance.CountConversations(this.UserInfo.UserID, portalId);
+
+                return this.Request.CreateResponse(HttpStatusCode.OK, messageBoxView);
+            }
+            catch (Exception ex)
+            {
+                var message = "An unexpected error occurred while trying to fetch the inbox, consult the server logs for more information.";
```
bdukes, line 60: Does this message display to end users?  It'll need to be loclizable, if so.

bdukes (approved):
LGTM, thanks!

DNN Platform/Modules/CoreMessaging/Services/MessagingServiceController.cs
valadas: Well IMO, it was not localized to begin with. My goal here was for security reasons to not display the whole details of a general exception which could potentially contain sensitive information. So going from an exception dump to a non-localized human-readable and safer string IMO is a step forward. It would obviously be nice to localize them but I think it may be a separate issue beyound the scope of reducing the build warnings.

valadas: Are we ok with this, or do we need to add localized exception messages everywhere ? If so I would prefer we make that a separate issue or TODOs for a second run, the goal of this PR (an the others) was really to reduce the build warnings, I just rewrote this because I do believe it is bad practice to send details of general exceptions to the frontend...

david-poindexter (approved):
Excellent - thanks @valadas 🎉

DNN Platform/Modules/CoreMessaging/Services/MessagingServiceController.cs
david-poindexter: @valadas we should probably create an issue to localize these exception messages, but let this PR pass for the intended purpose.

PR #4224, opened 2020-10-19
Resolves some build warnings in DDRMenu Project

valadas:
Resolves multiple build warnings on the DDRMenu Project

This is still a work in progress, some of it I want to undo, just submitting the draft to no forget continuing this a bit later this week...

### DIFF
--- a/DNN Platform/Library/UI/Containers/ActionBase.cs
+++ b/DNN Platform/Library/UI/Containers/ActionBase.cs
@@ -21,7 +21,20 @@ namespace DotNetNuke.UI.Containers
     /// </remarks>
     public abstract class ActionBase : UserControl, IActionControl
     {
-        private bool supportsIcons = true;
+        /// <summary>
+        /// Defines if the action supports icons.
+        /// </summary>
+        [Obsolete("Deprecates in v9.8.1, sheduled removal in v11.")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage(
+            "StyleCop.CSharp.NamingRules",
+            "SA1308:Variable names should not be prefixed",
+            Justification = "Keeping the name to prevent a breaking change, will be removed in v11.")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage(
+            "StyleCop.CSharp.MaintainabilityRules",
+            "SA1401:Fields should be private",
+            Justification = "In v11, we will make this private and rename.")]
+        protected bool m_supportsIcons = true;
+
         private ActionManager actionManager;
         private ModuleAction actionRoot;
 
@@ -46,7 +59,7 @@ namespace DotNetNuke.UI.Containers
         {
             get
             {
-                return this.supportsIcons;
+                return this.m_supportsIcons;
             }
         }
 
--- a/DNN Platform/Library/UI/Containers/ActionBase.cs
+++ b/DNN Platform/Library/UI/Containers/ActionBase.cs
@@ -13,11 +13,6 @@ namespace DotNetNuke.UI.Containers
     using DotNetNuke.UI.Modules;
     using DotNetNuke.UI.WebControls;
 
-    /// -----------------------------------------------------------------------------
-    /// Project  : DotNetNuke
-    /// Namespace: DotNetNuke.UI.Containers
-    /// Class    : ActionBase
-    /// -----------------------------------------------------------------------------
     /// <summary>
     /// ActionBase is an abstract base control for Action objects that inherit from UserControl.
     /// </summary>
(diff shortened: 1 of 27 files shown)


valadas:
Ok, this should be ready for review now

DNN Platform/Modules/DDRMenu/Common/DNNContext.cs
```diff
@@ -40,73 +42,122 @@ public DNNContext(Control hostControl)
...
+        {
+            if (this.isDisposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                // free managed resources.
+                this.page.Dispose();
```
bdukes, line 149: I'm not familiar with this `DNNContext` class, but I doubt that it's responsible for disposing of the passed control's `Page` reference.

DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -15,48 +15,68 @@ namespace DotNetNuke.Web.DDRMenu
...
+        /// </summary>
         public TemplateDefinition TemplateDef { get; set; }
 
+        /// <summary>
+        /// Gets the portal settings for the current portal.
+        /// </summary>
         internal PortalSettings HostPortalSettings
         {
-            get { return this.hostPortalSettings ?? (this.hostPortalSettings = PortalController.Instance.GetCurrentPortalSettings()); }
+            get { return (PortalSettings)(this.hostPortalSettings ?? (this.hostPortalSettings = PortalController.Instance.GetCurrentSettings())); }
```
bdukes, line 57: Rather than casting, I think this should be marked as `Obsolete`, return `PortalController.Instance.GetCurrentPortalSettings()`, and use a `#pragma` to avoid the warning about using an obsolete member.

DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -374,7 +410,7 @@ private void ApplyNodeManipulator()
             this.RootNode =
                 new MenuNode(
                     ((INodeManipulator)Activator.CreateInstance(BuildManager.GetType(this.menuSettings.NodeManipulator, true, true))).
-                        ManipulateNodes(this.RootNode.Children, this.HostPortalSettings));
+                        ManipulateNodes(this.RootNode.Children, (PortalSettings)this.HostPortalSettings));
```
bdukes, line 413: Isn't `HostPortalSettings` still returning `PortalSettings`?
```suggestion
                        ManipulateNodes(this.RootNode.Children, this.HostPortalSettings));
```

david-poindexter (approved):
This looks good to me - thank you @valadas - you are rocking it out sir! 🎉

DNN Platform/Modules/DDRMenu/Common/DNNContext.cs
valadas: Ok, it's been addressed in 6af2a81

DNN Platform/Modules/DDRMenu/MenuBase.cs
valadas: Does this make sense ? https://github.com/dnnsoftware/Dnn.Platform/pull/4224/commits/45da40c51e44f89fefa2c67929141f5861054139

DNN Platform/Modules/DDRMenu/MenuBase.cs
valadas: Addressed in https://github.com/dnnsoftware/Dnn.Platform/pull/4224/commits/45da40c51e44f89fefa2c67929141f5861054139

DNN Platform/Library/UI/Containers/ActionBase.cs
```diff
@@ -26,13 +21,16 @@ namespace DotNetNuke.UI.Containers
     /// </remarks>
     public abstract class ActionBase : UserControl, IActionControl
     {
-        protected bool m_supportsIcons = true;
-        private ActionManager _ActionManager;
-        private ModuleAction _ActionRoot;
+        private bool supportsIcons = true;
```
bdukes, line 24: There's no evidence that this is in use (via [a GitHub search](https://github.com/search?q=m_supportsIcons&type=Code)), but this change from `protected` to `private` is a breaking change.  I've never used `ActionBase`, so I'm not sure how likely it is that someone would make use of this field.

bdukes (approved):
:+1:

DNN Platform/Library/UI/Containers/ActionBase.cs
valadas: Well since it was protected and not used from anywhere else in this library, stylecop said it should be private. I can revert it if you think it may be a breaking change, but a third party would already not have access to a protected member right ?

valadas: Oh, wait, a derived class could access it, let me change that back then...

valadas: Ok, been addresses in https://github.com/dnnsoftware/Dnn.Platform/pull/4224/commits/e790b39f3b48f9acad819e373172f00639c1288f

bdukes (approved):

DNN Platform/Modules/DDRMenu/MenuBase.cs
bdukes: Yup.  I looked through this and didn't find anything to change (except for some typos 🙂)

PR #4225, opened 2020-10-19
Moved UserMode from PortalSettings to Personalization class

### LINKED ISSUE #4213: IPortalSettings lacking UserMode
While trying to resolve build warnings on DotNetNuke.Common.Globals, I found the method IsEditMode() had a warning about using the deprecated PortalController.Instance.GetCurrentPortalSettings and replaced it with .GetCurrentSettings which returns an IPortalSettings instead of a PortalSettings object.

IPortalSettings is lacking a UserMode property that PortalSettings had.

Should we :
🅰️ Add UserMode to the interface ?
🅱️ Deprecated UserMode in PortalSettings with some suggestion on how to do it the new way ?

Paging @bdukes @ahoefling

valadas:
Closes #4213

I am open to any feedback on this, I do believe this way would not be a breaking change...

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -105,10 +105,24 @@ namespace DotNetNuke.Entities.Portals
             ModuleEditor,
         }
 
+        /// <summary>
+        /// Enumerates the possible view modes of a page.
+        /// </summary>
         public enum Mode
         {
+            /// <summary>
+            /// The user is viewing the page in normal mode like a visitor.
+            /// </summary>
             View,
+
+            /// <summary>
+            /// The user is editing the page.
+            /// </summary>
             Edit,
+
+            /// <summary>
+            /// The user is viewing the page in layout mode.
+            /// </summary>
             Layout,
         }
 
@@ -201,6 +215,10 @@ namespace DotNetNuke.Entities.Portals
             }
         }
 
+        /// <summary>
+        /// Gets the mode the user is viewing the page in.
+        /// </summary>
+        [Obsolete("Deprecated in v9.8.1, use Personalization.GetUserMode() instead, Scheduled for removal in v10.")]
         public Mode UserMode
         {
             get
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -221,33 +221,7 @@ namespace DotNetNuke.Entities.Portals
         [Obsolete("Deprecated in v9.8.1, use Personalization.GetUserMode() instead, Scheduled for removal in v10.")]
         public Mode UserMode
         {
-            get
-            {
-                Mode mode;
-                if (HttpContext.Current != null && HttpContext.Current.Request.IsAuthenticated)
-                {
-                    mode = this.DefaultControlPanelMode;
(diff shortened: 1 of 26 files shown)


DNN Platform/Library/Services/Personalization/Personalization.cs
```diff
@@ -147,6 +152,38 @@ public static void SetSecureProfile(PersonalizationInfo personalization, string
...
+        }
+
+        /// <summary>
+        /// Gets the mode the user is viewing the page in.
+        /// </summary>
+        /// <returns><see cref="Mode"/>.</returns>
+        public static Mode GetUserMode()
+        {
+            Mode mode;
+            if (HttpContext.Current != null && HttpContext.Current.Request.IsAuthenticated)
```
bdukes, line 164:
```suggestion
            if (HttpContextSource.Current?.Request.IsAuthenticated == true)
```

DNN Platform/Library/Services/Personalization/Personalization.cs
```diff
@@ -147,6 +152,38 @@ public static void SetSecureProfile(PersonalizationInfo personalization, string
...
+            {
+                mode = PortalSettings.Current.DefaultControlPanelMode;
+                string setting = Convert.ToString(Personalization.GetProfile("Usability", "UserMode" + PortalController.Instance.GetCurrentSettings().PortalId));
+                switch (setting.ToUpper())
```
bdukes, line 168:
```suggestion
                switch (setting.ToUpperInvariant())
```

DNN Platform/Library/Services/Personalization/Personalization.cs
```diff
@@ -2,22 +2,27 @@
...
-{
+{
+    using System;
     using System.Web;
 
     using DotNetNuke.Common.Utilities;
     using DotNetNuke.Entities.Portals;
     using DotNetNuke.Entities.Users;
     using DotNetNuke.Security;
+    using static DotNetNuke.Entities.Portals.PortalSettings;
```
bdukes, line 13: Doesn't look like this is in use
```suggestion
    using static DotNetNuke.Entities.Portals.PortalSettings;
```

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -201,6 +215,10 @@ public UserInfo UserInfo
             }
         }
 
+        /// <summary>
+        /// Gets the mode the user is viewing the page in.
+        /// </summary>
+        [Obsolete("Deprecated in v9.8.1, use Personalization.GetUserMode() instead, Scheduled for removal in v10.")]
         public Mode UserMode
         {
             get
```
bdukes, line 224: Should the body of this property just return `Personalization.GetUserMode()`, so we're not duplicating the implementation?

bdukes (approved):
LGTM, thanks!

DNN Platform/Library/Services/Personalization/Personalization.cs
valadas: Yes it is used for `PortalSettings.Current.DefaultControlPanelMode`

DNN Platform/Library/Entities/Portals/PortalSettings.cs
valadas: Yep, forgot to do that, it is pushed in my latest commit. Thanks.

david-poindexter (approved):

Issue #4239, opened 2020-10-23
Add Page ID to the UI somewhere?


kieran23101:
## Description of problem
I have been creating allot of client websites where they want to edit it allot more but I always have to add some extra code somewhere to allow them to find the page id much easier so I thought that it might be handy to have it in the page details in page management?

Anyone else have any other way to do this?

Regards.

PR #4240, opened 2020-10-23
Adding Page ID to the Page Management

kieran23101:
This PR is in relation to #4239 

## Summary

I have made a simple addition to the Persona Bar so that you can see the Page ID without having to go the the SEO -> URL Test service to get it.

Unless there is somewhere else where people would see it fit I can re-do the change elsewhere however this is the most logical place I could think of...


bdukes (approved):
Thanks!  Can we get a screenshot of what this ends up looking like?

kieran23101:
image

here is what it looks like, it just added an extra section on the right on the page details

PR #4244, opened 2020-10-26
Deprecates IEventLogController in favor of DI

### LINKED ISSUE #4242: IEventLogController enforces usage of deprecated enum
Not sure if this should be an issue or a discussion, but since I guess it will involve some coding, let's start it up as an issue.

The IEventLogController has the EventLogType enum that is deprecated, but it also has non-deprecated methods that rely on it.

What should we do with this ? I know some work has been done for DI and logging. should we:

🅰️ Mark the whole interface deprecated and provide some short DI guidance ?
🅱️ Add more overloads so it supports the DotNetNuke.Abstractions.Logging.EventLogType one ?

valadas:
Resolves #4242

The PR:
- Deprecates the interface
- Deprecates the implementation
- The implementation beeing internal and the interface having a warning to not implement by 3rd parties, we can replace it's usage in other downstream classes without too much worry
- I did only one usage replacement in RelationshipController as a proof of concept and adjusted the related test on it.
- There are many many usages of this in Platform, so if thise PR is ok, we have a plan forward for the other usages, any feedback welcome.

### DIFF
--- a/DNN Platform/Library/Entities/Users/Social/RelationshipControllerImpl.cs
+++ b/DNN Platform/Library/Entities/Users/Social/RelationshipControllerImpl.cs
@@ -5,13 +5,14 @@ namespace DotNetNuke.Entities.Users.Social
 {
     using System.Collections.Generic;
     using System.Linq;
-
+    using DotNetNuke.Abstractions.Logging;
     using DotNetNuke.Common;
     using DotNetNuke.Common.Utilities;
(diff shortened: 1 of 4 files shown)


valadas:
@bdukes Does this make sense in the spirit of the issue comments ?

bdukes (approved):
:+1: looks good to me

mitchelsellers (approved):

PR #4245, opened 2020-10-26
Relaxed compile warnings on Tests projects

valadas:
Tests projects are not consumed by 3rd parties but where contrained to the same stylecop code style rules as the rest of the platform code (like documentation requirements and name styles and whitespace time. Since we are on a hunt to reduce our build warnings, this PR relaxes those rules on the Tests projects for now.

I would like to revisit this at a later date and actually correct some of those but for now it removes us 959 warnings quickly.


bdukes (approved):
👍🏻

mitchelsellers (approved):

PR #4246, opened 2020-10-27
Resolves multple build auto-fixable warnings in solution.

valadas:
This resolves many many many (6227) build warnings across the whole solution by implementing stylecop auto-fixes on some easy to fix warnings.

I do recommend reviewers to 'hide whitespace changes' while reviewing as this affects many many many files...

The gift that keeps on giving....


mitchelsellers (approved):

bdukes (approved):

valadas:
Ok, I rebased this and re-read the diff, didn't see anything wrong in it if we can get some more reviews and then a merge when built.

valadas:
Hmm, the build did not run though... Not sure why yet

PR #4250, opened 2020-10-27
Fix some integration tests

bdukes:
## Summary
We don't automatically run the integration tests, so we weren't notified that some of them are failing because they aren't setting up the services container.  This commit fixes that for a couple projects.


valadas (approved):
Thanks, looks good to me

david-poindexter (approved):
Looks good to me - thanks @bdukes 🎉

PR #4251, opened 2020-10-27
Ensure new Localization HTTP module is running soon enough

bdukes:
## Summary

This PR places the new Localization HTTP module right after URL Rewrite (which determines the culture from the URL), and then all modules after it can make use of that culture info.  The current behavior puts it at the end for upgrades, which potentially causes issues.

### DIFF
--- a/DNN Platform/Website/Install/Config/09.08.00.config
+++ b/DNN Platform/Website/Install/Config/09.08.00.config
@@ -1,6 +1,6 @@
 <configuration>
   <nodes configfile="web.config">
-    <node path="/configuration/system.webServer/modules" action="update" key="name" collision="overwrite">
+    <node path="/configuration/system.webServer/modules/add[@name='UrlRewrite']" action="insertafter">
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

valadas (approved):
Looks good to me, I'll do some localization tests on WebAPI as soon as possible too.

daguiler:
Hi @bdukes 
apparently this doesn't work because it creates a duplicate entry in the Web.config file.

Here's the error I got during install (installation cannot complete):

image

If this change will add the entry to the Web.config file, then maybe you need to remove it from the static files [here](https://github.com/dnnsoftware/Dnn.Platform/blob/b2db9a58bf2a47272542c63b9923eb75026cef9a/DNN%20Platform/Website/development.config#L87) and [here](https://github.com/dnnsoftware/Dnn.Platform/blob/b2db9a58bf2a47272542c63b9923eb75026cef9a/DNN%20Platform/Website/release.config#L87).

PR #4252, opened 2020-10-28
Merges latest 9.8.0 changes into develop branch

valadas:
Merges latest changes in 9.8.0 into the develop branch


valadas:
I'll redo this a different way because of the conflict...

PR #4258, opened 2020-10-28
Fix data reader not getting disposed

### LINKED ISSUE #3840: SqlDataProvider.IsConnectionValid leaks connections, exhausts pool
## Description of bug
Our production instance has been having issues with the NumberOfReclaimedConnections performance metric increasing. There appears to be a strong correlation between the NumberOfReclaimedConnections metric increasing and the number of times SqlDataProvider.IsConnectionValid is accessed.

This appears to be because a datareader created in https://github.com/dnnsoftware/Dnn.Platform/blob/future/10/DNN%20Platform/Library/Data/SqlDataProvider.cs#L64 is not being properly disposed.

## Steps to reproduce
1. Execute any code which directly or indirectly calls SqlDataProvider.IsConnectionValid

## Current behavior
The number of active connections in the connection pool increases with each access of IsConnectionValid, and the NumberOfReclaimedConnections metric increases once a high enough gen GC is triggered to finalize the undisposed datareaders.

## Expected behavior
The number of active connections in the connection pool remains constant and the NumberOfReclaimedConnections metric remains 0.

## Screenshots
image

image

## Affected version
DNN 9.5.0, but the behavior appears to be present in master as well as in 10-alpha source code.

* [x] 10.00.00 alpha build (in development)

bdukes:
Fixes #3840

### DIFF
--- a/DNN Platform/Library/Data/SqlDataProvider.cs
+++ b/DNN Platform/Library/Data/SqlDataProvider.cs
@@ -21,13 +21,7 @@ namespace DotNetNuke.Data
         private static DatabaseConnectionProvider _dbConnectionProvider = DatabaseConnectionProvider.Instance() ?? new SqlDatabaseConnectionProvider();
 
         /// <inheritdoc/>
-        public override bool IsConnectionValid
-        {
-            get
-            {
-                return CanConnect(this.ConnectionString, this.DatabaseOwner, this.ObjectQualifier);
-            }
-        }
+        public override bool IsConnectionValid => CanConnect();
 
         /// <inheritdoc/>
         public override Dictionary<string, string> Settings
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me, basically just to learn something from this, it's the change to remove static that makes it disposable and the rest is to support that change right?

bdukes:
No, the `static` change was just because the method was reimplementing an instance method for no good reason.  The fix is the switch from `ExecuteReader` (which returns an `IDataReader` that needs to be disposed) to `ExecuteNonQuery` which does handles disposing the connection/reader on its own.

valadas:
Ha thanks for the info, makes sense.

mitchelsellers (approved):

PR #4259, opened 2020-10-28
Remove deprecation for some DataProvider methods

### LINKED ISSUE #3830: Deprecation for ExecuteDataSet
On the current code for the DataProvider there is an ExecuteDataSet method with this deprecation comment:

[Obsolete("Deprecated in 7.0.0.  This method is unneccessary.  You can get a reader and convert it to a DataSet. Scheduled removal in v10.0.0.")]
public virtual DataSet ExecuteDataSet(string procedureName, params object[] commandParameters);

I think this is wrong and will break many module that might depend on it.

By defintion a DataSet can contain multiple tables so on the current call you will get all result sets from the SP into the dataset. See for example:

CREATE PROC spsample
as
select * from t1
select * from t2
...

So the change from ExecuteDataSet to GetReader is not correct, because you will end up loosing some of the data returned.

I don't see why this method should be deprecated when it has some valid use cases.

bdukes:
Fixes #3830

### DIFF
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -4123,7 +4123,6 @@ namespace DotNetNuke.Data
             this.ExecuteNonQuery("AuthCookies_DeleteOld", FixDate(utcExpiredBefore));
         }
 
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me

mitchelsellers (approved):

valadas (approved):
Looks good to me.

PR #4262, opened 2020-10-28
Fixed redirection after login

### LINKED ISSUE #4261: Redirect After Login does not work for Login.aspx
## Description of bug
Redirection after login works fine if current URL is `/login`, but it does not if current URL is `Login.aspx`.

## Steps to reproduce
1. Create a page named **Welcome**. Switch off `Display in Menu` and grant **VIEW** permission to **All Users**.
2. Open SETTINGS > Site Settings > Site Behavior > DEFAULT PAGES
3. Set `Redirect After Login` to the newly created **Welcome** page.
3. Click **Save**.
4. Log out and navigate to /Login.aspx. The login page should be displayed.
5. Log in as host.

## Current behavior
DNN does not redirect to **Welcome** page.

## Expected behavior
DNN should redirect to **Welcome** page.

## Screenshots
N/A

## Error information
N/A

## Additional context
N/A

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.08.00 release candidate
* [X] 09.07.02 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

daguiler:
Fixes #4261

## Summary
In the redirection after login logic, there is this `IsRedirectingFromLoginUrl()` method checking whether current URL ends with `/login` or not (hard coded), and DNN would redirect after login based on this condition.

Redirection after login should occur no matter which referrer URL, provided there is no `returnurl` parameter in the query string.

I executed 184 test cases to make sure this is now working as expected. These cases are the permutations of:

Portal: [ None | subsite ]
Login Page: [ None | My Login ]
Redirect After Login: [ None | Welcome ]

across following login URL formats:

/?ctl=login
/?ctl=login&returnurl=/Test-Page
/?ctl=login&returnurl=/Test-Page&popup=true
/login
/login?returnurl=/
/login?returnurl=/&popUp=true
/login?returnurl=/Test-Page
/login?returnurl=/Test-Page&popUp=true
/login.aspx
/login.aspx?returnurl=/
/login.aspx?returnurl=/&popUp=true
/login.aspx?returnurl=/Test-Page
/login.aspx?returnurl=/Test-Page&popUp=true
/my-login
/my-login?returnurl=/
/my-login?returnurl=/&popUp=true
/my-login?returnurl=/Test-Page
/my-login?returnurl=/Test-Page&popUp=true
/my-login.aspx
/my-login.aspx?returnurl=/
/my-login.aspx?returnurl=/&popUp=true
/my-login.aspx?returnurl=/Test-Page
/my-login.aspx?returnurl=/Test-Page&popUp=true

See attached spreadsheet for details.

### DIFF
--- a/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
+++ b/DNN Platform/Website/DesktopModules/Admin/Authentication/Login.ascx.cs
@@ -54,7 +54,6 @@ namespace DotNetNuke.Modules.Admin.Authentication
     /// </remarks>
     public partial class Login : UserModuleBase
     {
-        private const string LOGIN_PATH = "/login";
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(Login));
 
         private static readonly Regex UserLanguageRegex = new Regex(
@@ -80,8 +79,6 @@ namespace DotNetNuke.Modules.Admin.Authentication
             get
             {
                 var redirectURL = string.Empty;
-
-                var setting = GetSetting(this.PortalId, "Redirect_AfterLogin");
 
                 // first we need to check if there is a returnurl
                 if (this.Request.QueryString["returnurl"] != null)
@@ -120,12 +117,16 @@ namespace DotNetNuke.Modules.Admin.Authentication
 
                 if (string.IsNullOrEmpty(redirectURL) || isDefaultPage)
                 {
-                    if (
-                        this.NeedRedirectAfterLogin
-                        && (isDefaultPage || this.IsRedirectingFromLoginUrl())
-                        && Convert.ToInt32(setting) != Null.NullInteger)
+                    var redirectAfterLogin = Convert.ToString(GetSetting(this.PortalId, "Redirect_AfterLogin"));
+
+                    if (!int.TryParse(redirectAfterLogin, out var redirectAfterLoginTabId))
                     {
-                        redirectURL = this._navigationManager.NavigateURL(Convert.ToInt32(setting));
+                        redirectAfterLoginTabId = Null.NullInteger;
+                    }
+
+                    if (this.NeedRedirectAfterLogin && redirectAfterLoginTabId != Null.NullInteger)
+                    {
+                        redirectURL = this._navigationManager.NavigateURL(redirectAfterLoginTabId);
                     }
                     else
                     {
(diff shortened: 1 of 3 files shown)


daguiler:
[4262.xlsx](https://github.com/dnnsoftware/Dnn.Platform/files/5455125/4262.xlsx)

daguiler:
Here's a [link](https://docs.google.com/spreadsheets/d/1OgHg5n2r4CAw0hEq8VqP_-Qo7zpx2rL18RuOMEAlTZU/edit?usp=sharing) to the test cases spreadsheet.

daguiler:
And btw, I noticed that this line in the Web.config file is not used at all by DNN:

image

You can set anything you want as the "loginURL" and nothing will happen.

mitchelsellers:
@daguiler Can we add to this a change to remove that value from the web.config files then?

mitchelsellers (approved):
I think this is a great improvement overall, this simplifies the logic as well to use the ReturnUrl as the method to determine redirection.  

If possible, I'd like to remove that rogue Web.Config items though

daguiler:
Hi @mitchelsellers 
I added a commit to remove this entry (and a related entry, also unused).
I couldn't test the final installer because of [this issue](https://github.com/dnnsoftware/Dnn.Platform/pull/4251#issuecomment-718335337).
I'll try again tomorrow with a fresh brain.

daguiler:
Well, it wasn't that hard after all 😄
I just removed the `Localization` entry from the Web.config before installing and it worked.

So I was able to test all the basic login redirection scenarios, and all passed.
So I think this is good to go.

mitchelsellers (approved):
Thanks!  @bdukes should we clear this on upgrades too?

daguiler:
I can't fix the upgrades because I don't know which release this will land on, so I don't know which config file to change in the `DNN Platform/Website/Install/Config` folder

bdukes (approved):
I wouldn't bother removing the `web.config` settings on upgrades.

I agree that relying on the referrer is a bad idea, and I like the simplified logic.  I think this probably fixes more than it breaks, but I am concerned that we're losing some desired functionality.  

@daguiler do you have access to see the details of DNN-19925, which was the reason this referrer check was introduced? c00fc10671700bf3df4e4cb201294c6281678da9

daguiler:
Yes.
DNN-19925 is about `/login` not redirecting, so #2040 is essentially adding redirect support for `/login`.
In that ticket, this other PR is mentioned as the cause of the issue: #1969

valadas:
We need to run some tests on this and get some community feedback regarding:

- Localization
- Login module shown on all pages
- Selecting login page in the persona bar dropdown values
- Home page requiring login

PR #4264, opened 2020-10-29
Remove Localization HTTP module from config

bdukes:
It'll get added by `09.08.00.config`, having it already in there will create a duplicate

Thanks @daguiler for pointing out this issue with #4251!


valadas (approved):
Looks good to me and we need this the test the localization PR properly...

mitchelsellers (approved):

PR #4265, opened 2020-10-30
Fix Localization Issues

donker:
During testing we discovered that the thread locale was not set correctly when switching languages. This PR addresses that issue.

### DIFF
--- a/DNN Platform/HttpModules/Membership/MembershipModule.cs
+++ b/DNN Platform/HttpModules/Membership/MembershipModule.cs
@@ -8,6 +8,7 @@ namespace DotNetNuke.HttpModules.Membership
     using System.Linq;
     using System.Security.Principal;
     using System.Text.RegularExpressions;
+    using System.Threading;
     using System.Web;
     using System.Web.Security;
 
@@ -38,6 +39,19 @@ namespace DotNetNuke.HttpModules.Membership
 
         private static readonly Regex NameRegex = new Regex(@"\w+[\\]+(?=)", RegexOptions.Compiled);
 
+        private static bool IsActiveDirectoryAuthHeaderPresent()
+        {
+            var auth = HttpContext.Current.Request.Headers.Get("Authorization");
+            if (!string.IsNullOrEmpty(auth))
+            {
+                if (auth.StartsWith("Negotiate"))
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
         private static string _cultureCode;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


bdukes:
I've rebased these changes on the release branch

bdukes (approved):
Looks good to me, but will definitely need some good testing

DNN Platform/HttpModules/Membership/MembershipModule.cs
```diff
@@ -268,5 +249,25 @@ private void OnPreSendRequestHeaders(object sender, EventArgs e)
                 }
             }
         }
+
+        private void Context_PreRequestHandlerExecute(object sender, EventArgs e)
```
mitchelsellers, line 253: I assume this was moved here due to the new sequence requirements for the localization process?  Might we give this method a bit better name for the purpose, although the name is correct as to what event it responds to, the WHY this is important isn't here and with no comments I'm concerned that a few years down the road we will all forget again

mitchelsellers (changes requested):

valadas:
Strange the pipeline failed because of Dnn.ClientDependency, I gonna try to run it again to see if that is a fluke...

valadas:
Looks like it's failing because MyGet is down...

valadas:
Looks like MyGet is back up, firing up a new build

DNN Platform/HttpModules/Membership/MembershipModule.cs
donker: The problem I'm trying to solve is that membership (AuthenticateRequest) must run before this. PreRequestHandlerExecute is the last event before main processing begins, so I moved this code to run at that time so we know who the logged in user is at this point.

donker: https://docs.microsoft.com/en-us/troubleshoot/aspnet/http-modules-handlers

donker: BTW, I think the naming is not all that bad given that it mimics what we do for event handlers. The original name had the same convention.

mitchelsellers: I guess my point here is, yes the method name is “consistent”. But this was found after hours of debugging and there is nothing here that says WHY we do this. 

Renaming the method, or adding comments to save future self from pain was my point here.

donker: It's a good idea to comment. I'll add some.

donker:
> 
> 
> I've rebased these changes on the release branch

I updated my local repo with your changes but now we have 1000+ changes ... not sure how to resolve

valadas (changes requested):
Looks like this was rebased on develop instead of the release branch... I will try to fix that if I can...

valadas:
Yeah, I am not sure how to fix that @donker @bdukes can we fix this one or we redo a new PR, I am here to help if I can do anyhting, but I am not sure how to fix this other than a cherry pick on a new PR...

bdukes:
I've cherry-picked from your merge commit, @donker, and then rebased again on the release branch

valadas (approved):
Awesome, I downloaded this CI build and ran a few tests, looks fine to me in the context of a visitor, a regular user, an admin and a host both for WebAPI and WebForms.

Thanks guys.

mitchelsellers (approved):

PR #4266, opened 2020-11-01
Adds support for CORS configuration in AuthHandlers

valadas:
JWT is a popular authentication system these days for Apps. However many on-device calls get blocked by CORS in various WebViews.

One common fix Dnn users implement is to add Access-Control-Allow-Origin: "*" in their web.config. However, this opens up everything and one may only want to have this on let's say JWT login and some custom API endpoints.

For the api endpoints that totally up to the developer but for JWT login, there was no way to override the behavior only there without rebuilding it.

This PR bring CORS configuration options to the AuthHandlers.

Example config:

```
<add name="JWTAuth" type="Dnn.AuthServices.Jwt.Auth.JwtAuthMessageHandler, Dnn.AuthServices.Jwt" enabled="true" defaultInclude="false" forceSSL="true" accessControlAllowOrigins="ionic://localhost;http://localhost;https://localhost;http://loc.localtest.me;https://loc.localtest.me" accessControlAllowMethods="GET, POST" accessControlAllowHeaders="*" />
```

The would override those headers only for that auth handler. Note that you since we can have only a single value for Access-Control-Allow-Origin, the code will check the allowed list and bounce back the Origin header value from the request only if it is in that allowed list. I made it a semi-colon separated list in the web.config. For the other, they already support a coma delimited list so they get to be normal plain strings.

Due to the nature of these changes, I am targeting this at least for 9.9.0


david-poindexter (approved):
@valadas this really is excellent work and it will help a lot with 3rd party integrations, mobile apps, etc.  This has really been a pain point for quite some time and the workarounds just didn't feel right (proxies, allow all origins, etc.).  It is very rare that you can limit the origin to just one due to having multiple environments (local dev via browser, PWAs, hybrid apps in web views, etc.).  Now with the tighter constraints on hybrid apps, no longer allowing UIWebView, newer web views like WKWebView enforce CORS.  So, it is important to be able to configure JWT and custom endpoints appropriately.  Fantastic that we'll have a way to do this now!

To other reviewers, I was on a co-coding session with @valadas while he developed this and I also helped with testing.  It will be really great to get a few other eyes on this PR to make sure it is done right.

While it semantically would probably go in a `Future: Minor` release (e.g., `9.9.0`), I am thinking we may be fine getting this into a `Future: Patch` release (e.g., `9.8.1`).  Thoughts?

mitchelsellers:
This is a great addition, however, we are still implementing an All or Nothing solution for JWT. 

Is there a method we can enhance the JWT attribute to support config of this per APi?  

The risk of collision or changes accidentally could be real here.

valadas:
Hmm, I thought this would apply only to the code that lives in the actual handler... I'll have to test. The goal here was to provide this for the login/logout/renew parts and leave the API ends up to the developer using something like the attribute shown in about the middle of this article https://docs.microsoft.com/en-us/aspnet/web-api/overview/security/enabling-cross-origin-requests-in-web-api

But I will test and see if this apply to those or not before we merge this...

valadas:
And yeah, for the risk of collision, I would love to get some good testing on that for sure, that's why I believe it would go in a minor release to be on the safe side.

mitchelsellers:
Let me review this in a bit more detail as well

thienvc:
Great, I'm dealing with single sign-on (SSO) issues for multi-language sites where each language is a different domain, for example: global.dnndev.me (en-Gb); de.dnndev.de; ... because when I switch to another language, I have to log in again a bit annoying.
Also seems to have a small error here, when the un-approved user and Logger.IsTraceEnabled = False is still passed.
[https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs](url)
`if (!valid && Logger.IsTraceEnabled)
            {
                Logger.Trace("Inactive user status: " + status);
                return null;
            }`

DNN Platform/DotNetNuke.Web/Api/Internal/ServicesRoutingManager.cs
```diff
@@ -173,8 +197,26 @@ private static void RegisterAuthenticationHandlers()
 
                 try
                 {
-                    var type = Reflection.CreateType(handlerEntry.ClassName, false);
-                    var handler = Activator.CreateInstance(type, handlerEntry.DefaultInclude, handlerEntry.ForceSsl) as AuthMessageHandlerBase;
+                    var type = Reflection.CreateType(handlerEntry.ClassName, false);
+                    AuthMessageHandlerBase handler;
+                    if (type.GetConstructors().Any(c => c.GetParameters().Count() == 5))
```
bdukes, line 202:
```suggestion
                    if (type.GetConstructors().Any(c => c.GetParameters().Length == 5))
```

DNN Platform/Tests/DotNetNuke.Tests.Web/Api/BasicAuthMessageHandlerTests.cs
```diff
@@ -75,6 +75,47 @@ public void ResponseWithNullRequestReturnsUnauthorized()
...
             Assert.AreEqual("realm=\"DNNAPI\"", response.Headers.WwwAuthenticate.First().Parameter);
+        }
+
+        [Test]
+        public void DoesNotChangeAllowedOriginIfNotConfigured()
+        {
+            // Arrange
+            var response = new HttpResponseMessage(HttpStatusCode.OK) { RequestMessage = null };
+            response.Headers.Add("Access-Control-Allow-Origin", "*");
+            var handler = new BasicAuthMessageHandler(true, false, "", "", null);
```
bdukes, line 86: The default for the origins argument would be an empty array instead of `null`, right?
```suggestion
            var handler = new BasicAuthMessageHandler(true, false, "", "", Array.Empty<string>());
```

bdukes (approved):
LGTM

bdukes:
@thienvc that's a good catch, can you submit that changes as a PR?

mitchelsellers:
> Great, I'm dealing with single sign-on (SSO) issues for multi-language sites where each language is a different domain, for example: global.dnndev.me (en-Gb); de.dnndev.de; ... because when I switch to another language, I have to log in again a bit annoying.
> Also seems to have a small error here, when the un-approved user and Logger.IsTraceEnabled = False is still passed.
> https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
> `if (!valid && Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); return null; }`
> Maybe to fix it:
> `if (!valid) { if(Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); } return null; }`

The SSO part of this requires the cookie domain in the web.config to be changed.

mitchelsellers (changes requested):
Before this is merged I would like to discuss this at the approvers meeting.

thienvc:
> @thienvc that's a good catch, can you submit that changes as a PR?

Sorry, i can't

thienvc:
> > Tuyệt vời, tôi đang giải quyết các vấn đề đăng nhập một lần (SSO) cho các trang web đa ngôn ngữ trong đó mỗi ngôn ngữ là một miền khác nhau, ví dụ: global.dnndev.me (en-Gb); de.dnndev.de; ... vì khi chuyển sang ngôn ngữ khác thì phải đăng nhập lại hơi phiền. 
> > Cũng có vẻ như có một lỗi nhỏ ở đây, khi người dùng chưa được phê duyệt và Logger.IsTraceEnabled = False vẫn được thông qua. 
> > https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
> > `if (!valid && Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); return null; }` 
> > Có thể sửa lỗi:
> > `if (!valid) { if(Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); } return null; }`
> 
> Phần SSO của điều này yêu cầu miền cookie trong web.config phải được thay đổi.



> > Great, I'm dealing with single sign-on (SSO) issues for multi-language sites where each language is a different domain, for example: global.dnndev.me (en-Gb); de.dnndev.de; ... because when I switch to another language, I have to log in again a bit annoying.
> > Also seems to have a small error here, when the un-approved user and Logger.IsTraceEnabled = False is still passed.
> > https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
> > `if (!valid && Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); return null; }`
> > Maybe to fix it:
> > `if (!valid) { if(Logger.IsTraceEnabled) { Logger.Trace("Inactive user status: " + status); } return null; }`
> 
> The SSO part of this requires the cookie domain in the web.config to be changed.

The solution I am thinking of does not involve cookies, I want to take jwt and apply in this direction:
https://github.com/Aralink/ssojwt
Using iframe and js postMessage, the session cookie is returned via the webapi: https://github.com/DmitryFillo/AspNet.WebApi.CookiesPassthrough

bdukes:
@thienvc I'd be happy to walk you through the PR process.  For a change like this (a small change in a single file), it should be pretty easy to create a PR via the GitHub website.  You can also look at [our contributing guidelines](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/CONTRIBUTING.md) for details on the approach.

Go to the file you want to edit, [`JwtController.cs`](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs), and click the edit pencil icon in the top right.  You can make your change in the editor, and then at the bottom it will give you a place to commit the change.  It will create your forked repository if it doesn't exist already (which it looks like [it does](https://github.com/thienvc/Dnn.Platform)), and then create a new branch for your change, and finally create a pull request using that branch.

Let me know if you run into any issues, and I can help you work through them.  If that seems like too much work, I can make the change for you, but I'd love to make sure you get credit for the change and experience walking through the change process.  Hope it helps!

bdukes:
@thienvc I've also created a separate issue to discuss the JWT SSO

thienvc:
> JWT is a popular authentication system these days for Apps. However many on-device calls get blocked by CORS in various WebViews.
> 
> One common fix Dnn users implement is to add Access-Control-Allow-Origin: "*" in their web.config. However, this opens up everything and one may only want to have this on let's say JWT login and some custom API endpoints.
> 
> For the api endpoints that totally up to the developer but for JWT login, there was no way to override the behavior only there without rebuilding it.
> 
> This PR bring CORS configuration options to the AuthHandlers.
> 
> Example config:
> 
> ```
> <add name="JWTAuth" type="Dnn.AuthServices.Jwt.Auth.JwtAuthMessageHandler, Dnn.AuthServices.Jwt" enabled="true" defaultInclude="false" forceSSL="true" accessControlAllowOrigins="ionic://localhost;http://localhost;https://localhost;http://loc.localtest.me;https://loc.localtest.me" accessControlAllowMethods="GET, POST" accessControlAllowHeaders="*" />
> ```
> 
> The would override those headers only for that auth handler. Note that you since we can have only a single value for Access-Control-Allow-Origin, the code will check the allowed list and bounce back the Origin header value from the request only if it is in that allowed list. I made it a semi-colon separated list in the web.config. For the other, they already support a coma delimited list so they get to be normal plain strings.
> 
> Due to the nature of these changes, I am targeting this at least for 9.9.0

@valadas i am mailed to security@..., I apologize for accidentally making it public on github PR for underestimating this bug :(

valadas:
I am trying to think what would be better to avoid this beeing a global solution for all JWT endpoints....

a: check which API endpoint is called and only do the new behaviour when we are on login/logout/renew
b: Introduce a JWT config node in the web config where we can put more stuff just specific to JWT endpoints...

mitchelsellers:
Part of me thinks we need a way to define this via the Attribute or other method.

valadas:
Well, as explained before, on actual api endpoints there is a package to define that but the main issue is a 3rd party would have to recompile the JWT provider because users won't be able to login without it in there. So the goal here would be to make it configurable for those login/logout/renew endpoints without needing recompilation (such as in the web.config)...

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
```diff
@@ -2,55 +2,102 @@
...
-        /// </summary>
+        /// Gets a value indicating whether to bypass the anti forgery token.
+        /// </summary>
         public virtual bool BypassAntiForgeryToken => false;
-
+
         /// <summary>
-        /// Gets a value indicating whether this handler should be included by default on all API endpoints.
-        /// </summary>
+        /// Gets a value indicating whether this hander in included in all routes.
```
bdukes, line 78:
```suggestion
        /// Gets a value indicating whether this handler is included in all routes.
```

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
```diff
@@ -60,6 +107,11 @@ protected AuthMessageHandlerBase(bool includeByDefault, bool forceSsl)
         /// <returns>null normally, if a response is returned all inbound processing is terminated and the resposne is returned.</returns>
         public virtual HttpResponseMessage OnInboundRequest(HttpRequestMessage request, CancellationToken cancellationToken)
         {
+            if (request.Headers.Contains("Origin"))
+            {
+                this.origin = request.Headers.GetValues("Origin").FirstOrDefault();
```
bdukes, line 112: Is there one instance of this `DelegatingHandler` per request?  Do we know it's safe to store this into in the object instance, or could it be reused by multiple requests?

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
```diff
@@ -91,24 +183,24 @@ protected static bool IsXmlHttpRequest(HttpRequestMessage request)
...
         }
-
+
         /// <summary>
-        /// Sets the current principal for the request.
+        /// Sets the current principal.
         /// </summary>
-        /// <param name="principal">The principal to set.</param>
-        /// <param name="request">The current request.</param>
+        /// <param name="principal">The pricipal to set.</param>
```
bdukes, line 190:
```suggestion
        /// <param name="principal">The principal to set.</param>
```

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
```diff
@@ -91,24 +183,24 @@ protected static bool IsXmlHttpRequest(HttpRequestMessage request)
...
         protected static void SetCurrentPrincipal(IPrincipal principal, HttpRequestMessage request)
         {
             Thread.CurrentPrincipal = principal;
             request.GetHttpContext().User = principal;
         }
-
+
         /// <summary>
-        /// Asynchronously sends a response.
+        /// Sends an HTTP response asynchronously.
```
bdukes, line 199:
```suggestion
        /// Sends an HTTP request asynchronously.
```

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
valadas: Excellent question, this will need to be re-taught and well tested. I kinda got busy and could not get back to this. So I put it on hold but this still needs some work... I'll try to revisit this again this weekend or some such...

DNN Platform/DotNetNuke.Web/Api/Auth/AuthMessageHandlerBase.cs
valadas: This returns an HttpResponseMessage no ?

bdukes: It sends a request (from the parameters) and returns a response (right?)

valadas: Hmm, I'd say it handles a request by sending a response for it... Or am I confused...

bdukes: This overrides and ultimately delegates to [`HttpMessageHandler.SendAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpmessagehandler.sendasync), and that method's docs say
>Send an HTTP request as an asynchronous operation.

valadas: Ok, makes sense, thanks.

valadas:
Ok, I got a nice plan forward that is different than this one. An attribute that would read from config and only apply to those enpoints that have that attribute... I am closing this PR for now, will first do stylecop fixes on the Auth Providers and DotnetNuke.Web and then revisit this so the PR is cleaner after the fixes...

PR #4267, opened 2020-11-01
Fix the web.config errors for the new Localization module

donker:
Unfortunately the XmlMerge in DNN does not allow a collision check when doing any other action than "update". This prompted two Localization modules to be present in the web.config for new installs which makes the site unuseable. This PR changes the action to be "update". Note that it means we don't control where in the modules pipeline the module sits. But this has become a moot point since we changed the event order for this module.

In the future we should adjust the XmlMerge class to cater for collision control for the other actions, IMHO.

### DIFF
--- a/DNN Platform/Website/Install/Config/09.08.00.config
+++ b/DNN Platform/Website/Install/Config/09.08.00.config
@@ -1,6 +1,6 @@
 <configuration>
   <nodes configfile="web.config">
-    <node path="/configuration/system.webServer/modules/add[@name='UrlRewrite']" action="insertafter">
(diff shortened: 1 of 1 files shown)


bdukes:
#4264 removed the localization module from the config to start, so there shouldn't be duplicates now.

bdukes:
And I think we still want it to run as early as possible.

mitchelsellers:
I agree with @bdukes here

valadas:
Yeah I thought it was fixed, at least I don't have duplicates on a clean install from a build on the release branch... @donker was this tested after Brian's PR got merged ?

donker:
When I tested, Brian's PR hadn't been merged yet. Now it should be fine.

PR #4277, opened 2020-11-05
Upgrade optional packages when they've already been installed

donker:
DNN is distributed with a number of optional packages like Jwt Auth. If we make a fix to these packages they should be upgraded together with DNN.

This PR will install optional packages (i.e. .resources files) if they have been installed previously. It also reverts an earlier change where we made Jwt install by default to work around this issue.


mitchelsellers (approved):

valadas (approved):
Tested clean install and upgrade, works perfectly. Thank you!

PR #4281, opened 2020-11-06
Implement EnablePopups switch in portal settings PB module

donker:
This PR brings back the "EnablePopups" setting to the UI. It will appear under "More" on the Portal Settings PB module.


mitchelsellers (approved):

valadas (approved):
Awesome, looks good to me, a well needed setting IMO

image

hismightiness:
Why couldn't we just have disabled them and removed the settings everywhere entirely?  😢 😢 😢 😆 🤣

valadas:
I would not be against defaulting this to false on new installs...

PR #4284, opened 2020-11-07
Fixed an issue where new users would have no preferred locale.

valadas:
The user preferred locale is a required field in user profiles. Some change made it that in 9.8.0 this property is null. This PR reverts the behavior so the preferred locale is set on new users.

Basically the locale was already set but before initializing the profile and initializing the profile would reset it, this PR just moves one line


david-poindexter (approved):
Thanks @valadas 🎉

donker (approved):

PR #4285, opened 2020-11-07
Fixed an issue where all emails where missing the body

valadas:
While working on the Feedback module the messages where missing their body, then I tried creating new users and those emails where also missing the body. Looks like the CoreMailProvider was missing setting the body to the emails.

This PR solves that.


david-poindexter (approved):
Great catch - thanks @valadas 🎉

mitchelsellers (approved):

PR #4299, opened 2020-11-13
Roll back PR #4258 as it broke the data provider

donker:
PR #4258 caused the platform to no longer be able to connect to the database. This PR rolls back that change.


bdukes:
Sorry about that, I've created #4302 to fix the issue introduced (without having to roll back the included fix).

mitchelsellers (approved):

PR #4312, opened 2020-11-16
Ensure properties in DTO objects use backing fields

donker:
DTO objects should not have expression-bodied properties as .net serialization does not keep the thread locale and reverts to en-US. This caused the user profile editing to not respect the current locale.

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/Dto/UserBasicDto.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Users/Dto/UserBasicDto.cs
@@ -33,10 +33,11 @@ namespace Dnn.PersonaBar.Users.Components.Dto
             this.RequestsRemoval = user.RequestsRemoval;
             this.IsSuperUser = user.IsSuperUser;
             this.IsAdmin = user.Roles.Contains(this.PortalSettings.AdministratorRoleName);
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):
Great catch on this one!

We should most likely document this somewhere, as the expression-bodied syntax is a commonly used pattern

valadas (approved):
Nice catch, another thing to watch for with Stylecop and DTOs is a stylecop warning where IEnumerables (Arrays, Lists and such) should not have a `get`, this is a wrong warning in the context of a DTO and in this case the warning whould be disabled for the DTO.

bdukes (approved):

bdukes:
Just to clarify, we're really talking about computed properties, or properties without backing fields.  "Expression-bodied" refers to properties using the `Prop => value` syntax instead of the `Prop { get { return value; } }` syntax.  Both of those examples are computed at runtime and will cause an issue during serialization, but only the first is using an expression body.

donker:
> 
> 
> Great catch on this one!
> 
> We should most likely document this somewhere, as the expression-bodied syntax is a commonly used pattern

Not sure where to document. But I think the rule should be: if it is a DTO, then it should not branch to other code during serialization period. Even if you think it's "safe", I think the practice should be: do not do this.

valadas:
Yeah, it would be hard to document in a fashion that would be obvious... I guess we just need to watch for this in code reviews...

PR #4315, opened 2020-11-17
Update bug report template after 9.8.0 release

bdukes:


valadas:

Issue #4307, opened 2020-11-15
Response wrong language cookie


thienvc:
https://github.com/dnnsoftware/Dnn.Platform/blob/6e1385a4a9603ac37a1aa347fdf9edefdf77991e/DNN%20Platform/HttpModules/Membership/MembershipModule.cs#L160-L163
Point of error of ThreadCulture in Web API #3545

thienvc:
When authenticated user is setted PreferredLocale, should have only set the ThreadUICulture = user.Profile.PreferredLocale but i don't know why set the ThreadCulture = user.Profile.PreferredLocale as well

valadas:
@donker indicated that it was decided a long time ago that ThreadCulture and ThreadUICulture should always be both set to the same. Honestly I don't even know the difference myself. You have an input on this @donker ?

thienvc:
ThreadCulture is for Backend: for example select data where lang = 'en-US',
ThreadUICulture is for FrontEnd, after user login some personalization for the language can be reserved for that user but only in the external interface (Localization.GetString)
A specific example is when a website has 3 languages: en-US, en-VN, fr-FR
When the user is on the french language page fr.dnndev.me but if the user is logged in and userprofile.preferedlocale = en-US then the interface after logging in is still in English, but PortalSettings and website content is still the "content of french"

thienvc:
[https://github.com/dnnsoftware/Dnn.Platform/blob/6e1385a4a9603ac37a1aa347fdf9edefdf77991e/DNN%20Platform/Library/Services/Localization/Localization.cs#L1619-L1628](url)
`        /// <summary>
        /// Gets a value indicating whether allows users to select their own UI culture.
        /// When set to false (default) framework will allways same culture for both
        /// CurrentCulture (content) and CurrentUICulture (interface).
        /// </summary>`
[https://github.com/dnnsoftware/Dnn.Platform/blob/d7f7bcc53b9b89d091467b47874053e5fa158c76/DNN%20Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs#L48-L54](url)
The summary of this setting says it all @valadas

donker:
There are several problems in supporting differing Thread locales from Thread UI locales.

1. The user is unaware. I.e. we had requests to unify this back in the day because people didn't understand why X was in language Y when they switched languages in the UI. It has never been a good and clear experience.
[...]

Issue #3545, opened 2020-02-05
Thread locale doesn't pervade to data backend using WebAPI


donker:
# Description of bug
When going through WebAPI the thread locale is set correctly when I get to the handler. But once I retrieve data through DAL2 I find that in those methods the thread locale has been reset to en-US (system default).

## Steps to reproduce
1. Create new site and enable multiple languages in your portal.
2. Install test module and instantiate on a page: [https://github.com/donker/testmodule/releases/tag/0.0.1](https://github.com/donker/testmodule/releases/tag/0.0.1)
3. Switch to a non en-US/non portal default locale

There are two tables listing tabs. The object that is being used retrieves the thread locale when being created from data retrieval. The first table is done directly in MVC. The second is done through WebAPI.

## Current behavior
The first table shows the correct environment locale, the second does not in the last column.

## Expected behavior
Both tables should show the same.

## Affected version
* [ ] 10.0.0 alpha build
* [x] 9.5.0 alpha build
* [x] 9.4.4 latest supported release

donker:
[TestModule_00.00.01_Install.zip](https://github.com/dnnsoftware/Dnn.Platform/files/4162198/TestModule_00.00.01_Install.zip)

donker:
My suggestion is to lump this in with the thread locale issue we discovered when we wanted to go to async. The best solution I can think of right now is to add the locale to the request items like we do with portal settings and then draw them out again in a similar manner.

leedavi:
I'm getting the en-US always being used if a url has paramater.

The url may be fr-fr, but the data is en-us.

This makes DNN unworkable for multiple languages.

leedavi:
My issues with DNN9.4 langauges seem to have been solved in DNN9.5.  

Maybe this bug was fixed in the DNN9.5 release, but I'm unsure if this issue is related to the problems we had.

Maybe Peter has a better idea if this has been solved.

donker:
Closing for now. I believe it is no longer relevant. It had to do with the way DI was implemented which has since been changed.

thienvc:
This problem still appears on DNN 9.8, has someone found out and fixed this problem?
This is a serious problem for multilingual websites.

thienvc:
I found has many response language cookie many times:

Date: Sun, 15 Nov 2020 12:28:30 GMT
Set-Cookie: app_IsMobile=False; path=/; HttpOnly
[...]

PR #4316, opened 2020-11-18
Fix wrong Language Culture in WebAPI

thienvc:
`Localization.SetLanguage(user.Profile.PreferredLocale) `=> Wrong ThreadCulture in WebAPI
This error affects the multilingual web, and allows the user to select the display language, and the user has specified the language-specific PreferedLocale in the profile.

Fix: Response wrong language cookie #4307, Fix: Thread locale doesn't pervade to data backend using WebAPI #3545 

## Summary

Fix wrong Language Culture in WebAPI


donker:
As this changes behavior we need to table this until a major release

valadas:
As discussed in the approvers meeting:

We need to come back to those issues for Dnn10 but we first need to really decide and document how everything will work with developer documentation, etc. We do agree though on the principle that the user preferred culture should be used. We are closing this PR and coming back to a more complete solution for this.

PR #4321, opened 2020-11-20
Fixes AddContentItem Audit Trail for CreatedByUserId & LastModifiedByUserId

### LINKED ISSUE #4002: ContentController.AddContentItem Does Not Update CreatedByUserId/LastModifiedByUserId
## Description of bug
When building a custom DNN module that uses the `ContentController` to write to the `ContentItems` table, it will not create the correct records for `CreatedByUserId` or `LastModifiedByUserId` when using the `AddContentItem` API. 

## Steps to reproduce
Create a custom module to write a new content item to the database.
```c#
var contentItem = new ContentItem
{
    ModuleID = ActiveModule.ModuleID,
    Content = "Hello World",
    ContentTypeId = contentType.ContentTypeId // this is retrieved in omitted code
};

var contentController = new ContentController();
var contentItemId = contentController.AddContentItem(contentItem);``
```
When using the statement `contentController.AddContentItem(contentItem)` executes it does not insert the correct UserID into the new database row.

## Current behavior
When a new `ContentItem` is added via the `ContentController` it does not insert the correct UserId into the following columns
- **CreatedByUserId**
- **LastModifiedByUserId**

Instead it inserts the value `-1`

image

## Expected behavior
When using this API I would expect that the UserId would insert correctly. This creates an audit issue since it is impossible to know who was the first user to create the content.

## Workaround
If the code Adds the content and then updates it immediately, the `LastModifiedByUserId` will display correctly. This creates additional code for the module that shouldn't need to exist.

## Affected version
* [x] 09.07.00 release candidate 2

## Affected browser
N/A

## Reproduction Module
[ContentItemsBug_1.0.0.zip](https://github.com/dnnsoftware/Dnn.Platform/files/5109247/ContentItemsBug_1.0.0.zip)

I have created a reproduction module that allows you to confirm the problem. This module is a simple MVC module that inserts "Hello World" in to the `ContentItems` table every time the page is loaded. The results are then printed to the view, which helps debug the problem. 

1. Install module
2. Add it to a page
3. Navigate to the page

Once you have done that you can open SQL Management Studio or the SQL console in DNN and use the following script to view the data
```sql
select  
	ci.ContentItemID,
	ci.Content,
	LEN(ci.Content) CountLength,
	ci.ContentTypeID,
	ct.ContentType,
	ci.ModuleID,
	ci.CreatedByUserID,
	cru.Username CreatedByUsername,
	ci.CreatedOnDate,
	ci.LastModifiedByUserID,
	modu.Username LastModifiedByUsername,
	ci.LastModifiedOnDate
from ContentItems ci 
	join ContentTypes ct on ci.ContentTypeID = ct.ContentTypeID
	left join Users modu on ci.LastModifiedByUserID = modu.UserID
	left join Users cru on ci.CreatedByUserID = cru.UserID	
where ct.ContentType = 'ContentItemsBug'
```

ahoefling:
Fixes #4002 

## High Level Summary
Fixes audit trail for adding new content to the DNN database. This now properly assigns the CreatedByUserID & LastModifiedByUserID if it is not explicitly set on the `ContentItem`.

## Technical Summary
When adding Content to the DNN database using the `ContentController.AddContentItem(ContentItem item)` API it would not set the `CreatedByUserId` or `LastModifiedByUserId` correctly. In this API DNN invokes a Stored Procedure and uses the same value for both fields, it then sets `LastModifiedByUserId` after the stored procedure is invoked. This is why we would get some odd behavior where `CreatedByUserId` was -1 and `LastModifiedByUserId` would be 1. 

I have outlined the new rules in this table below
| New/existing rules | ContentItem.CreatedByUserID | CurrentUser().UserID | Resulting CreatedByUserID & LastModifiedByUser ID |
|------------------|----------------------------------|-------------------------|-----------------------------------------------------------|
| new | -1 (not set)                               | 5                                  | 5                                                                                   |
| existing | 5                                               | 5                                  | 5                                                                                   |
| existing | 2                                               | 5                                   | 5                                                                                   |

## Additional Changes
I needed to update usage of the `CBO` object in the `UpdateContentItemsCache()` method. This code change uses the instance object instead of the static API which allows us to properly test the changes to this API. I looked up what the static API was doing and made sure to use the correct instance API

### DIFF
--- a/DNN Platform/Library/Entities/Content/ContentController.cs
+++ b/DNN Platform/Library/Entities/Content/ContentController.cs
@@ -46,7 +46,7 @@ namespace DotNetNuke.Entities.Content
             Requires.NotNull("contentItem", contentItem);
             var currentUser = UserController.Instance.GetCurrentUserInfo();
             var createdByUserId = currentUser.UserID;
-            if (contentItem.CreatedByUserID != currentUser.UserID)
+            if (contentItem.CreatedByUserID >= 0 && contentItem.CreatedByUserID != currentUser.UserID)
             {
                 createdByUserId = contentItem.CreatedByUserID;
             }
--- a/DNN Platform/Library/Entities/Content/ContentController.cs
+++ b/DNN Platform/Library/Entities/Content/ContentController.cs
@@ -272,10 +272,11 @@ namespace DotNetNuke.Entities.Content
             DataCache.RemoveCache(GetContentItemCacheKey(contentItem.ContentItemId)); // remove first to synch web-farm servers
             if (readdItem)
             {
-                CBO.GetCachedObject<ContentItem>(
+                CBO.Instance.GetCachedObject<ContentItem>(
                     new CacheItemArgs(
                     GetContentItemCacheKey(contentItem.ContentItemId),
-                    DataCache.ContentItemsCacheTimeOut, DataCache.ContentItemsCachePriority), c => contentItem);
+                    DataCache.ContentItemsCacheTimeOut, DataCache.ContentItemsCachePriority), c => contentItem,
+                    false);
             }
         }
 
--- a/DNN Platform/Tests/DotNetNuke.Tests.Content/ContentControllerTests.cs
+++ b/DNN Platform/Tests/DotNetNuke.Tests.Content/ContentControllerTests.cs
@@ -8,7 +8,7 @@ namespace DotNetNuke.Tests.Content
     using System.Collections.Generic;
     using System.Collections.Specialized;
     using System.Linq;
-
+    using System.Web;
     using DotNetNuke.Abstractions;
     using DotNetNuke.Abstractions.Application;
     using DotNetNuke.Common;
@@ -18,6 +18,7 @@ namespace DotNetNuke.Tests.Content
     using DotNetNuke.Entities.Content;
     using DotNetNuke.Entities.Content.Data;
     using DotNetNuke.Entities.Controllers;
+    using DotNetNuke.Entities.Users;
     using DotNetNuke.Services.Cache;
     using DotNetNuke.Services.Search.Entities;
     using DotNetNuke.Tests.Content.Mocks;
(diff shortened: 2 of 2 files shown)


bdukes (approved):

ahoefling:
Not sure about the build failure here, it is referencing issues in code that seems to be unrelated `ModulePipeline`. Could we queue the build again?

bdukes:
New build: https://dev.azure.com/dotnet/DNN/_build/results?buildId=42094

ahoefling:
I am seeing the build script fail locally because it is attempting to use the .NET 5 cli
image

bdukes:
https://devblogs.microsoft.com/nuget/getting-started-with-nuget-5-8/
>Because NuGet 5.8 is the first release to support .NET 5, we consider it to be a known issue that your builds can break when using it with other tooling versions. You may run into issues with restoring your projects & solutions and then building them using mismatched versions of tooling.
>
>One such issue you may run into is [NETSDK1005](https://developercommunity.visualstudio.com/content/problem/1248649/error-netsdk1005-assets-file-projectassetsjson-doe.html). This happens when you restore packages using an earlier NuGet tooling version & building your projects & solutions with a newer tooling version of Visual Studio, dotnet CLI, or MSBuild.

bdukes:
@ahoefling I think if you delete `nuget.exe` from your tools dir, it will automatically download the newer version which should work.

ahoefling:
@bdukes I am still seeing build errors after removing the entire tools dir and letting the build re-generate it.

I can't pinpoint the exact problem here, but I think it is related to the release of .NET 5 as my visual studio is updated and msbuild errors are citing .NET 5 cli build paths. This only happens when using the cake build script, when I build through visual studio I don't see any problems. I think the hosted agents are having similar problem but the error message is different which has me confused.

Not sure if there is any action I need to take as these appear to be unrelated to this PR. Please let me know if anyone needs me to take action on the code here.

mitchelsellers (approved):

PR #4322, opened 2020-11-20
Use latest Cake.Frosting

bdukes:
[Cake.Frosting](https://cakebuild.net/docs/running-builds/runners/cake-frosting) is a build runner for Cake which uses a .NET console application instead of `.cake` files.  I've done a quick conversion to see if getting the latest and greatest helps fix our current build issue.


valadas:
Oh, I am excited with this :)

valadas:
Ok, I just fired up a new build... I don't get this, I was working on this issue for hours yesterday and today. For some reason gitversion was not just spitting out pure json but had text about that. I tried a bunch of stuff without success, then I reset everything and it just all started working fine. I am clueless.

valadas:
Well, I tried a couple of things that came to mind and to google mind. I am not sure what the solution is here...

valadas:
Well, it's building now with the .Net version change not using RC2 but just 5.0.100 :)

There is an issue running the tests, but this is a good step forward

valadas:
Ok, so the last commit in ae00bba resolves the issue with the unit tests not running.
It consists of two changes.
1. The path was not updated to use the new version of the Nunit runner (the main problem)
2. Before getting into that I thought it was caused by the EnsureAllProjectsBuilt task so I removed it to test...

Now about point 2 : It really speeds up the build times! My guess is it fired up a rebuild even if the projects where already built. I believe we should just assume that:
1. In CI we always build before running the tests anyway
2. Locally people probably use the built-in Visual Studio tests explorer or if they use this task and it fails, they would immediately know that they need to build first by the nature of the error message they would get.

Any objections to this change @bdukes @donker ?

valadas:
Oh and one more thing... Any objections including this project in the solution. If we do make this the startup project we now have our F5 solution to build 😄 And if we agree on that then I can add some build targets in the dropdown menu in Visual Studio. We could have stuff like:
- Package
- Deploy
- CreateSettings (this now being a console app, we can use Console.ReadLine() to implement some sort of guided wizard too...

donker (approved):
I verified the local build scripts still work

azure-pipelines.yml
```diff
@@ -58,14 +61,14 @@ steps:
   inputs:
     targetType: filePath
     filePath: ./build.ps1
-    arguments: '-Target $(CakeTarget) -ScriptArgs ''--CkBranch="$(Branch.CkEditor)"'',''--CdfBranch="$(Branch.CdfProvider)"'''
+    arguments: '--target=$(CakeTarget) --verbosity=diagnostic --CkBranch="$(Branch.CkEditor)" --CdfBranch="$(Branch.CdfProvider)"'
```
bdukes, line 64: Do we want to default to `--verbosity=diagnostic`?   If so, add it to the first Cake call, also

azure-pipelines.yml
```diff
@@ -1,9 +1,11 @@
 pool:
-  name: Hosted VS2017
+  vmImage: 'windows-latest'
 
 variables:
   - name: "Build.ArtifactStagingDirectory"
     value: "Artifacts/"
+  - name: "buildPlatform"
+    value: "x64"
```
bdukes, line 8: We don't need this anymore, right?

azure-pipelines.yml
valadas: I would vote no, logging takes time and the build would be faster without it. We can always temporarily add it if we have build issues... IMO

azure-pipelines.yml
valadas: I think not, It was part of the suggestions to fix why it was not building. Now if we remove it and it fails, then we put it back 😄

azure-pipelines.yml
bdukes: I moved it to a build-time variable, and defaulted to normal

valadas: Also CKEditor is now part of this repo and we no longer pull from a branch, so that could go away. I think that variable was removed from the build scripts already so passing it actually does nothing anymore

bdukes:
Once this build is done, good to merge 👍🏻

azure-pipelines.yml
bdukes: I don't see any usage of `CkBranch` or `CdfBranch`.  @donker do you know where we'd expect those to be used?  Are we still pulling CDF dynamically during build/packaging?

valadas:
Hmm, a test failed, is that a fluke or does it ring a bell...

bdukes:
Doesn't ring a bell, I've requeued it

bdukes:
Hmm… [Same test failed in both runs](https://dev.azure.com/dotnet/DNN/_build/results?buildId=42525&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=380b8dbc-2323-5a9d-d01f-cf0280cc79aa&l=388)

```
  Failed SharedDictionaryTests.TwoDictsShareALockWriteTest [51 ms]
  Error Message:
   An unexpected exception type was thrown
Expected: System.Threading.LockRecursionException
 but was: System.ObjectDisposedException : Cannot access a disposed object.
  Stack Trace:
  at System.Threading.ReaderWriterLockSlim.TryEnterWriteLockCore(TimeoutTracker timeout)
at System.Threading.ReaderWriterLockSlim.TryEnterWriteLock(TimeoutTracker timeout)
at System.Threading.ReaderWriterLockSlim.TryEnterWriteLock(TimeSpan timeout)
at DotNetNuke.Collections.Internal.ReaderWriterLockStrategy.GetWriteLock(TimeSpan timeout) in D:\a\1\s\DNN Platform\Library\Collections\ReaderWriterLockStrategy.cs:line 143
at DotNetNuke.Collections.Internal.SharedDictionary`2.GetWriteLock(TimeSpan timeOut) in D:\a\1\s\DNN Platform\Library\Collections\SharedDictionary.cs:line 262
at DotNetNuke.Collections.Internal.SharedDictionary`2.GetWriteLock() in D:\a\1\s\DNN Platform\Library\Collections\SharedDictionary.cs:line 251
at DotNetNuke.Tests.Core.Collections.SharedDictionaryTests.TwoDictsShareALockWriteTest() in D:\a\1\s\DNN Platform\Tests\DotNetNuke.Tests.Core\Collections\SharedDictionaryTests.cs:line 157
```

bdukes:
Worked fine on a totally new run `¯\_(ツ)_/¯`

valadas:
Cool, yeah some tests are a bit flaky. Merging then...

PR #4326, opened 2020-11-22
Fix build error

bdukes:
I think this missing reference is causing the build error we started seeing on Friday, though I'm not sure why it wasn't required before.


ahoefling (approved):
I took a brief look at this other day and I think this change will solve the problem. Technically `DotNetNuke.Library` should provide the `DotNetNuke.DependencyInjection` package reference, but it appears recently the build agents are accepting that anymore. It makes perfect sense for this project to have a reference to the `DotNetNuke.DependencyInjection` project like you did.

🙏 Thanks for submitting this PR, I've noticed several PRs started failing recently

mitchelsellers (approved):

valadas (approved):
Looks good to me

PR #4331, opened 2020-11-25
This is a test to make sure the new pipeline is the only one that runs

valadas:
Please do not merge this PR, it is just a test to confirm that only the new pipeline runs on PRs


valadas:
Cool, works fine... closing.

PR #4347, opened 2020-12-08
Removed "Running Default" check from Default.aspx

### LINKED ISSUE #4346: Remove Outdated Default Checks
## Description of problem
DNN Platform in the past has provided a method to check for "default credentials" due to prior limitations in the platform about automatically creating users without forcing a password change.

DNN Platform has forced password changes for a number of years and these checks add complexity to the code and do not provide valuable long-term benefits and should be removed.

## Description of solution
To follow the proper process the API check should be marked obsolete and the rest of the code removed.

We could leave this, but it adds extra work on every page load etc.

mitchelsellers:
Fixes #4346 by removing the Running Default check from default.aspx and marked enum members and validation methods as obsolete.


bdukes (approved):

valadas (approved):
Looks good to me

PR #4354, opened 2020-12-09
Issue-4353 - Connectors use black and white icons when they're in disconnected state which is against branding rules

### LINKED ISSUE #4353: Connectors use black and white icons when they're in disconnected state which is against branding rules
## Description of bug
Connectors use black and white icons when they're in disconnected state which is against branding rules. For example; Google has the following rules for Analytics logo branding which explicitly forbids changing the colors of their icons ( https://developers.google.com/analytics/terms/branding-policy#logo-usage ):
image

## Steps to reproduce
1. Log in with a host user
2. Navigate to PB > SETTINGS > Connectors

## Current behavior
Icons are black and white.

## Expected behavior
Icons should be colorful.

## Screenshots
image

## Additional context
Icons are suggested to be shown in their original colors. When a Connector gets connected, it'll already show a green tick next to the icon as in the following screenshot so that this change won't create any confusion about whether a Connector is connected or not too.
image

## Affected version

* [x] 10.00.00 alpha build
* [ ] 09.08.00 latest supported release

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #4353

## Summary

This simple CSS change will ensure that the colorful icons will be used both when the Connectors are connected and disconnected.

Demo: https://drive.google.com/file/d/1N_O47unL7H5UV0X4zZKEN8yvs9YnOmi6/view?usp=sharing

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Connectors/css/Connectors.css
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Connectors/css/Connectors.css
@@ -36,9 +36,8 @@ table#connectionstbl span.socialnetwork-favicon {
     width: 38px;
     height: 37px;
     background-repeat: no-repeat;
-    background-position: 0 0;
-    position: relative;
     background-position: 0 -37px;
+    position: relative;
 }
 
 table#connectionstbl span.socialnetwork-favicon div {
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Connectors/css/Connectors.css
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Connectors/css/Connectors.css
@@ -38,6 +38,7 @@ table#connectionstbl span.socialnetwork-favicon {
     background-repeat: no-repeat;
     background-position: 0 0;
     position: relative;
+    background-position: 0 -37px;
 }
 
 table#connectionstbl span.socialnetwork-favicon div {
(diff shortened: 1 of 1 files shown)


bdukes:
https://github.com/dnnsoftware/Dnn.Platform/blob/6b98a4064471151eba62712a4100b36b10f53f72/DNN%20Platform/Connectors/Azure/Images/Azure.png also needs to be adjusted to remove the black-and-white version, right?

berkarslan-xo:
> https://github.com/dnnsoftware/Dnn.Platform/blob/6b98a4064471151eba62712a4100b36b10f53f72/DNN%20Platform/Connectors/Azure/Images/Azure.png also needs to be adjusted to remove the black-and-white version, right?

It will work fine even with the current icons since the background-position will take -37px. However, if you suggest to remove the black and white parts from all of the icons, the CSS file and all of the images (both in Platform and Evoq) should be updated of course.

mitchelsellers (changes requested):
I am going to go ahead and say that I do NOT approve of this.  This appears to be a specific change for Evoq, and I would agree with the comment on the issue that grayscale for disconnected is an industry standard as well.

david-poindexter (commented):
It is common practice to show greyscale for inactive and colored for active.  This does not violate the brand guidelines since there is a greyscale version in the brand identity guide document at http://www.dnnsoftware.com/portals/0/sitefiles/dnn_corp_brand_guidelines.pdf.  Thanks.

mitchelsellers:
Per the feedback and standards of UI/UX feedback, we are closing this issue at this time.  

If there is a specific claim from a vendor regarding this pattern it should be discussed and addressed at that time.

Issue #4362, opened 2020-12-15
Denying the `ADD` permission for a role on an asset makes it invisible to users on CKE Editor.


mikebigun:
## Description of bug
Denying the `ADD` permission for a role on an asset makes it invisible to users on CKE Editor.

## Steps to reproduce
List the precise steps to reproduce the bug:
1. Go to `Assets` -> Site Folder, by default it is `My Website`
2. Create new Folder, name it `Test`
3. Go to edit `Test` folder -> `Permissions` tab and deny `ADD` permission for `All Users` role
4. Ensure `BROWSE FOLDER` and `VIEW FOLDER` permissions are granted for `All Users`
5. Save the changes
6. Go to edit mode of `Home` page
7. Click on any HTML module
8. Go to `Advanced Editor`
9. Click on `Link`
10. Click on `Browse server` to browse the assets from the server

## Current behavior
When you click on `Browse server`, the application creates the tree of a folders that the current user has `READ` permission. Despite of that, the `Test` folder is missing there. 

## Expected behavior
We have denied `ADD` permission for the `Test` folder in reproduction steps. Along with that `BROWSE FOLDER` and  `VIEW FOLDER` were granted. In this case I expect to see `Test` folder listed in CKEditor folder tree.

## Screenshots

![image](https://user-images.githubusercontent.com/22524011/102221677-1a098400-3eeb-11eb-845a-e362b53654e1.png)

## Affected version

* [ ] 10.00.00 alpha build
* [ ] 09.08.00 latest supported release

PR #4364, opened 2020-12-15
DNN-42147 Denying the `ADD` permission for a role on an asset makes it invisible to users on CKE Editor.

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/4362

The rootcause of this problem is in the following line of a `GetFoldersByPermissions` stored procedure:

image

The subquery above is trying to get all the folders for the specific user/role for which access is denied (`allowAccess=0`). The subquery returns list of folder ID's.

Problem is that `allowAccess` flag relates to a concrete folder permission (READ, WRITE, ADD...) and not to an entire folder. This means, if at least one permission is denied so we return entire folder and consider it as denied. This is the issue. Due to that it is not getting appeared on UI.
Instead, we must return concrete denied **folder permission** for the specific user/role. 

See the changes I did highlighted in yellow below. FolderID is replaced by PermissionID.

image

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.08.01.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.08.01.SqlDataProvider
@@ -1,0 +1,90 @@
+﻿
+/****************************************************************
+ * SPROC: GetFoldersByPermission DNN-42147
+ ****************************************************************/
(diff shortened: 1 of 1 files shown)


bdukes (approved):
Makes sense to me

mitchelsellers (approved):
This change makes sesne, I do believe we need to regression test this in other areas to ensure we don't have any unintended behaviors.

PR #4365, opened 2020-12-15
Try to fix flaky test

bdukes:
We have a unit test that started failing about 2/3 of the time, which is causing us to manually re-run builds.

This changes that test to be more explicit.


mitchelsellers (approved):

valadas (approved):
Looks good to me.

PR #4366, opened 2020-12-15
Warnings as Errors for test projects

bdukes:
This turns off the documentation rules for StyleCop and turns on Warnings as Errors for all test projects

### DIFF
--- a/DNN Platform/Tests/.editorconfig
+++ b/DNN Platform/Tests/.editorconfig
@@ -151,7 +151,7 @@ csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
 
 # SA1600: Elements should be documented
-dotnet_diagnostic.SA1600.severity = suggestion
+dotnet_diagnostic.SA1600.severity = none
 
 # SA1309: Field names should not begin with underscore
 dotnet_diagnostic.SA1309.severity = suggestion
--- a/DNN Platform/Tests/DNN.Integration.Test.Framework/DNN.Integration.Test.Framework.csproj
+++ b/DNN Platform/Tests/DNN.Integration.Test.Framework/DNN.Integration.Test.Framework.csproj
@@ -12,6 +12,7 @@
     <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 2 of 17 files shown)


mitchelsellers (approved):

valadas (approved):
Providing it builds after the rebase, this looks good to me.

daguiler:
Hi @bdukes
the solution doesn't build anymore in Visual Studio bacause of this. Am I missing something?

bdukes:
@daguiler it builds fine for me.  What errors are you getting?  Are you on the latest `develop`?

daguiler:
I am on latest `develop`, and I'm getting errors because of all the broken rules in the test projects.
Are you not getting the same?

image

bdukes:
No, those should be suggestions rather than warnings because of the configuration in https://github.com/dnnsoftware/Dnn.Platform/blob/7c3d891dc8a1e60666cbe104e9c6b877821a3593/DNN%20Platform/Tests/.editorconfig#L153-L226

bdukes:
What version of Visual Studio are you using?  I bet it's a fairly new feature to read severity for those rules through `.editorconfig`.

daguiler:
mm.. I'm still on 2017 Enterprise

daguiler:
According to [this article](https://dotnetthoughts.net/editorconfig-support-in-visual-studio-2017/), VS 2017 supports `.editorconfig` files. I'll dig a bit more.

bdukes:
The [Analyzer configuration documentation](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#configure-severity-levels) says
>Starting in Visual Studio 2019 version 16.3, you can configure the severity of analyzer rules, or diagnostics, in an [EditorConfig file](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#set-rule-severity-in-an-editorconfig-file), from the [light bulb menu](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#set-rule-severity-from-the-light-bulb-menu), and from the error list.

daguiler:
I had an update available for my 2017 version (from 15.9.23 to 15.9.30), but unfortunately it didn't help with this issue.
So I guess I will have to undo this change in my local environment from now on every time I want to build the solution :-(

bdukes:
If you'd like to submit a change which would work for VS 2017, we'd happily accept it.  The best option would be, of course, to resolve the warnings, but that's rather time consuming, which is why we haven't done it 😄 

A partial option would be to add a new `stylecop.json` file for tests which loosens the strictness of the rules (specifically, you can disable the documentation rules), but fix the other issues which can't be controlled via `stylecop.json`.

The other option is to introduce a `.ruleset` file which disables the rules and reference that from the test projects.

PR #4367, opened 2020-12-16
Add support for schema script in the script installer

donker:
I should start by saying I'm not 100% convinced we should do this, so I'd like a second opinion.

During module install the regular xx.yy.zz.sqldataprovider files are executed and also "upgrade.sqldataprovider". The latter's "version" is used to set the install version of the module. So in your manifest you really need to make sure you label this correctly, or you'll upset the install process.

### Critique of current situation

I have three issues with the above:

1. I think it's a bit of a misnomer. "Upgrade" suggests it only runs on upgrades, but in fact it runs on first install as well. "Schema" would have been better IMO.
2. The installedVersion is influenced by the upgrade script
3. You can only have one

I am not 100% sure what the upgrade script logic was meant to support but my guess was that developers could opt to use this instead of the regular scripts. I'm not sure why we'd encourage that.

This PR introduces support for a new type of install script. Namely one that begins with "schema.". So you could do "schema.SalDataProvider" but also "Schema.Views.SqlDataProvider" and "Schema.Sprocs.SqlDataProvider". So you can have multiple of these in your install. And these would run on every single install. Note they'd run in the order in which they appear in the manifest!

### Why

The biggest reason is so module developers can better version their SQL schema. If you have one file with these elements, you can easily see when you changed a SPROC or a VIEW. You'd do table and data changes in the regular files, but SPROCS, VIEWS, FUNCS etc can be written so they can be refreshed on every install. And your schema script would diff very well in source control.


DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -187,6 +202,19 @@ public override void Install()
                 {
                     bSuccess = this.InstallScriptFile(this.UpgradeScript);
                     installedVersion = this.UpgradeScript.Version;
+                }
```
mitchelsellers, line 205: Following the [documentation ](https://dnndocs.com/content/tutorials/extensions/dnn-manifest-schema/index.html#the-dnn-manifest-schema)to maintain expectations, I think it would make sense to do things in this order.

* Install
* Schema
* Upgrade

Thoughts?

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -271,6 +299,10 @@ protected override void ProcessFile(InstallFile file, XPathNavigator nav)
                 {
                     this._upgradeScript = file;
                 }
+                else if (file.Name.StartsWith("schema.", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    this._schemaScripts.Add(file);
+                }
                 else if (type.Equals("install", StringComparison.InvariantCultureIgnoreCase))
```
mitchelsellers, line 306: While we are in here, can we cleanup some of the logic to remove duplication and to make this more clear.

For example for the first conditional, make it check for either name starting with "install" or a type of install.

Also, while we are here is it appropriate that "the no match" is uninstall?  That seems a bit dangerous of a default.  I realize outside the scope of this a bit, but if we are editing it would be good to do this this way.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -187,6 +202,19 @@ public override void Install()
                 {
                     bSuccess = this.InstallScriptFile(this.UpgradeScript);
                     installedVersion = this.UpgradeScript.Version;
+                }
+
+                // Next process SchemaScripts - these script always runs if present
```
bdukes, line 207: I would think we would want these to run before versioned scripts.  But, I guess, regardless, I'm having a hard time seeing how you'd handle data migration if you're using these.  I guess you'd use schema scripts for views, functions, and procedures, but stick to versioned files for tables?  So, if that's the case, running schema scripts last makes sense to me.  (So, I don't think there's anything to change, but leaving this comment in case anyone wants to comment)

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -271,6 +299,10 @@ protected override void ProcessFile(InstallFile file, XPathNavigator nav)
                 {
                     this._upgradeScript = file;
                 }
+                else if (file.Name.StartsWith("schema.", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    this._schemaScripts.Add(file);
```
bdukes, line 304: Your description said that these would run in alphabetical order, but it looks like this would be in source order.  I think source order is the right choice, I just wanted to point out that it's different than what you described, unless I'm missing something.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -24,7 +24,8 @@ public class ScriptInstaller : FileInstaller
         private readonly SortedList<Version, InstallFile> _installScripts = new SortedList<Version, InstallFile>();
         private readonly SortedList<Version, InstallFile> _unInstallScripts = new SortedList<Version, InstallFile>();
         private InstallFile _installScript;
-        private InstallFile _upgradeScript;
+        private InstallFile _upgradeScript;
+        private readonly List<InstallFile> _schemaScripts = new List<InstallFile>();
```
mitchelsellers, line 28: GIven that this is a non-sorted generic list, the order of execution would match the order if was input in the manifest.  Not saying that is wrong, but given the others are sorted by a systematic process should we sort here as well?

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
bdukes: I don't have experience using the install or upgrade scripts, so I may be missing some context on how these actually get used.  That said, it seems like, regardless of whether the schema scripts run before or after the versioned scripts, that the upgrade script is designed to run last.  Again, I may be missing something, but I would think we'd want the upgrade script to run after these schema scripts.

mitchelsellers (dismissed):
Overall, I think this is a good change, as this allows someone to actually source control an entire extension following known standards.  (Such using Red Gate's SQL Source Control) or otherwise. 

However, I think there is a few areas that we could clean-up to make this implementation better, and also while we are in here cleanup unexpected behaviors.  I'm fine taking some of these unexpected behaviors to another PR but this just brought them to light.

bdukes (commented):
I like this idea. Having an easier way to process and source control individual SQL components would be great.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
bdukes: I think the default "sort" of source order is the least surprising, unless we require some new attribute to indicate ordering.

mitchelsellers: I'm ok with that as well, just calling out the difference vs the PR documentation which noted A-Z sorting

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
mitchelsellers: I typically used the versioned scripts myself thus far, but I believe that the "install" script concept would work similar to that of how it works inside of DNN platform as a whole.

Install - This is used to setup the "base" of the installation.  It is processed first, and once completed the version that you provide with it is used to determine which additional scripts (if included) are used for processing.

Versioned Scripts - These then execute, as we expect, version-by-version starting with the next version after the "install" version.

Schema - This, as I see it would be a "replacement" of sorts for a lot of what might get done in the current Install or Versioned scripts.  You can do whatever you want here, but you have to realize that this is executed on every (initial) installation of your extension.

Now, what is actually done in here could be interesting....

For example, if using EF inside of a module, you could use an Idempotent script flag to generate your model and use the single Schema file.  This would work for all changes, including incremental schema changes.

Another example would be more limited where you put all SP's etc in there that you drop/create every time.  But table migrations would need to be done otherwise.

My biggest thoughts are that this is good, but I'm worried that the current naming conventions and black magic automatic assumptions might make this more complicated.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Yeah, what @bdukes said. I have thought about adding ordering, but it would probably needlessly complicate matters. If a developer is particular about order, then they'll need to manage their manifest.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: I use install scripts from time to time. But mostly the versioned scripts, like @mitchelsellers .

About the purpose: as I point out, I see them being used in parts of the schema that can/need to be recreated. Take views: if you've changed a table "Foo" and added a column "Bar" and you have a view that includes "SELECT * FROM Foo" then Bar will not be present unless you recreate the view. 

I regularly deal with this as my pattern is to have a view in 99% of the tables I create and in code I use an object that inherits from the base object of the table. That way I can do CRUD operations easy where I expect the view for reading and the table (base object) for writing. Anyway, it only works if I meticulously keep track of the views as well.

So in terms of run order: I'd want the versioned scripts to run first. They do the most invasive stuff like work on data and change tables. I'm not sure about the upgrade scripts as I don't use those but I'd leave them as is just after. The schema scripts are really just a layer on top of the data that my code interacts with. Hence I'd keep it until last.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: You're right. They run in the order they appear in the manifest.

sleupold:
IMO this is a great enhancement of handling SQL scripts in DNN installer and I would prefer to even entend it, making it easier to understand and structure code, if needed. 
One subject is the name "schema" is misleading. The "database schema" are the tables (and on top, its indexes and constraints/relations), while views and procedures are database application code on top. While keeping "install", "uninstall" and "upgrade" file names as legacy support and version numbered scripts for install and schema increments, I would prefer adding naming prefixes "index..", "relation.", "view." and "sproc." and their equivalents "unindex", "unrelation", "unview" and "unsproc" as well as "data." with optionally an appropriate name following (like suggested for "schema." by Peter.
The benefit would be the option to uninstall code prior to an version upgrade and install it in current version afterwards.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
sleupold: In general, I'd agree and we need to make sure to be backwards compatible under all conditions. 
Besdes see my previous comment.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
sleupold: Versionsed scripts need to be kept for table modifications (schema updates), everything else can be dropped and recreated using the other scripts. This way, they must be modified once only (no complex creation of rollup scripts anymore) and views and sprocs may be edited "in place" in these files - making file versions and upgrade diffs much easier to follow for devs.

sleupold: Versionsed scripts need to be kept for table modifications (schema updates), everything else can be dropped and recreated using the other scripts. This way, they must be modified once only (no complex creation of rollup scripts anymore) and views and sprocs may be edited "in place" in these files - making file versions and upgrade diffs much easier to folllow for devs.

sleupold: This will also allow using advanced features like indexed views, which may support performance dramatically.

donker:
> IMO this is a great enhancement of handling SQL scripts in DNN installer and I would prefer to even entend it, making it easier to understand and structure code, if needed.
> One subject is the name "schema" is misleading. The "database schema" are the tables (and on top, its indexes and constraints/relations), while views and procedures are database application code on top. While keeping "install", "uninstall" and "upgrade" file names as legacy support and version numbered scripts for install and schema increments, I would prefer adding naming prefixes "index..", "relation.", "view." and "sproc." and their equivalents "unindex", "unrelation", "unview" and "unsproc" as well as "data." with optionally an appropriate name following (like suggested for "schema." by Peter.
> The benefit would be the option to uninstall code prior to an version upgrade and install it in current version afterwards.

I agree that "schema" is too broad as it includes the tables. But in the interest of keeping code as lean and clean as possible I went for the option of "schema.xyz". It will be up to the developer to devise their own naming for this. The only criteria is that the name starts with "schema".

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: You're assuming that the upgrade script makes no changes to the schema and only to data. This is fair enough but will probably need to be documented somewhere.

donker: BTW, there is no issue of backwards compatibility regarding the run order and what is in documentation. This is a new feature. So backwards compatible means no schema scripts are there.

sleupold:
> I agree that "schema" is too broad as it includes the tables. 
it doesn't just include the tables, the name excludes the code elements (sprocs, views)

sleupold:
For large extensions (and the code), IMO would be preferable to have the option to split the install script (with the table schema) into multiple files, as well as scripts for code AND scripts for uninstall.
Our current best practice of preceding each CREATE PROCEDURE with an optional DROP is not beneficial for supporting e.g. indexed views, as we need to remove all procedures and views prior to schema changes of the underlying tables. That's why I had the idea to split it up into uninstall scripts for code, constraints and indexes (besides the tables, which should only be deleted, when uninstalling an extension) - this way we could execute these uninstall scripts prior to any schema (table) upgrade script, next upgrade the schema (tables) and finally re-create the code objects with current version.
From my experience, dropping and re-creating views (even indexed views), functions and procedures doesn't take long to execute -  that might be different for schema updates (cleanup rows, adding and calculating columns, re-creating indexes for tables, if containing millions of records.

donker:
I added support for a "preupgrade" script in the same way as there is an "upgrade" script. The latter is a single script that runs **after** versioned upgrade scripts are run. The new upgrade script would run **before** and versioned scripts and only if this is not a new install. The goal is to allow uninstalling elements of the schema that get put back in the "upgrade" script. IMO this would complete support for a wide variety of scenarios.

To recap. The order of scripts is now:

1. Install (only one allowed) OR (if exists) Preupgrade (only one allowed)
2. Versioned scripts (many allowed run in order of version)
3. Upgrade (only one allowed)
4. Schema (many allowed, run in order of entry in manifest)

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Actually the type install is a bit of a bucket since it includes everything except uninstall. In hindsight the entire solution could have depended on correct type labeling. But at some point it all got mixed up.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -164,9 +193,17 @@ public override void Install()
                         bSuccess = this.InstallScriptFile(this.InstallScript);
                         installedVersion = this.InstallScript.Version;
                     }
-                }
-
-                // Then process remain Install/Upgrade Scripts
+                }
+                else
```
bdukes, line 197: I was thinking that `Upgrade` scripts didn't run on first install, but they do, so I think `PreUpgrade` script also run on first install.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -267,18 +317,30 @@ protected override void ProcessFile(InstallFile file, XPathNavigator nav)
...
                 {
                     // These are the Install/Upgrade scripts
                     this.InstallScripts[file.Version] = file;
                 }
-                else
+                else if (file.Name.StartsWith("uninstall.", StringComparison.InvariantCultureIgnoreCase)
+                    || type.Equals("uninstall", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    // These are the Uninstall scripts
+                    // This is the Uninstall script
```
bdukes, line 343: It looks like multiple uninstall scripts are supported
```suggestion
                    // These are the Uninstall scripts
```

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -267,18 +317,30 @@ protected override void ProcessFile(InstallFile file, XPathNavigator nav)
...
+                {
+                    this._schemaScripts.Add(file);
+                }
                 else if (type.Equals("install", StringComparison.InvariantCultureIgnoreCase))
                 {
                     // These are the Install/Upgrade scripts
                     this.InstallScripts[file.Version] = file;
                 }
-                else
+                else if (file.Name.StartsWith("uninstall.", StringComparison.InvariantCultureIgnoreCase)
```
bdukes, line 340: The condition on [line 360](https://github.com/dnnsoftware/Dnn.Platform/blob/303a8346cf937915d5c38dcb5a1d91948b281add/DNN%20Platform/Library/Services/Installer/Installers/ScriptInstaller.cs#L360) needs to be similarly adjusted.  It seems like that condition is superfluous, since we're also checking if the file is in `this.UnInstallScripts`, so maybe the `if` on line 360 could be simplied to just be `if (this.IsValidScript(fileExtension))`

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -267,18 +317,30 @@ protected override void ProcessFile(InstallFile file, XPathNavigator nav)
...
                     this._installScript = file;
                 }
-                else if (file.Name.StartsWith("upgrade.", StringComparison.InvariantCultureIgnoreCase))
+                else if (file.Name.StartsWith("preupgrade.", StringComparison.InvariantCultureIgnoreCase)
+                    || type.Equals("preupgrade", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    this._preUpgradeScript = file;
+                }
+                else if (file.Name.StartsWith("upgrade.", StringComparison.InvariantCultureIgnoreCase)
+                    || type.Equals("postupgrade", StringComparison.InvariantCultureIgnoreCase))
```
bdukes, line 326: Should this type just be `"upgrade"`?  I would expect it to run last if it's `"postupgrade"`.

bdukes (commented):
This looks pretty good.  I am still wondering about `schema` as the name for these new scripts, which makes me wonder if we could just add the `preupgrade` script type, and allow multiple `preupgrade` and `upgrade` scripts.  Would this be a simpler way to enable this paradigm?

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: This is a problem IMHO. The upgrade scripts *should* IMHO adhere to their naming and run on upgrades, not install. From what I understand the use case to be, it would serve to prepare the existing data/schema for an upgrade. So in this case ripping out SPROCs, views etc. Obviously you would not want to run this on first install as there is nothing to rip out or you're ripping out stuff you just installed using the install script.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: My bad. It is.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Correct. And while we're at it, why is the check IsValidScript here and not in the logic which parses out the scripts in the first place?

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Given that we have "preupgrade", "upgrade" becomes confusing. Currently "upgrade" is post-upgrade. I.e. after the versioned upgrade scripts have run.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -132,6 +147,20 @@ protected InstallFile UpgradeScript
...
+        /// </summary>
+        /// <value>An InstallFile.</value>
+        /// -----------------------------------------------------------------------------
+        protected List<InstallFile> SchemaScripts
+        {
+            get
+            {
+                return this._schemaScripts;
+            }
+        }
```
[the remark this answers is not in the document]
donker: It would be, but then I'd like to improve on the naming. "Upgrade" should be "post upgrade" to make it clear when this runs.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
bdukes: I agree in principle, but that seems like a breaking change.  Do we want to hold this change until v10? /cc @mitchelsellers

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -120,44 +119,31 @@ protected ProviderConfiguration ProviderConfiguration
...
         /// -----------------------------------------------------------------------------
         /// <summary>
-        /// Gets the Upgrade Script (if present).
+        /// Gets a list of Post-Upgrade Scripts (if present) - these scripts will always run after
+        /// and versioned upgrade scripts and also after initial install.
         /// </summary>
-        /// <value>An InstallFile.</value>
+        /// <value>A list of <see cref="InstallFile"/> instances.</value>
         /// -----------------------------------------------------------------------------
-        protected InstallFile UpgradeScript
```
bdukes: Do we need to keep this around for backwards compatibility?  I'm not sure what the use case for extensing `ScriptInstaller` would be, but it's certainly possible…  Maybe have this return the first item in `PostUpgradeScripts`? 

```c#
protected InstallFile UpgradeScript => this.PostUpgradeScripts[0];
```

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -120,44 +119,31 @@ protected ProviderConfiguration ProviderConfiguration
...
         /// <summary>
-        /// Gets the Pre-Upgrade Script (if present) - this script will always run before
+        /// Gets a list of Pre-Upgrade Scripts (if present) - these scripts will always run before
         /// any upgrade scripts but not upon initial installation.
         /// </summary>
-        /// <value>An InstallFile.</value>
+        /// <value>A list of <see cref="InstallFile"/> instances.</value>
         /// -----------------------------------------------------------------------------
-        protected InstallFile PreUpgradeScript
+        protected List<InstallFile> PreUpgradeScripts
```
bdukes, line 127: While we're defining new properties, let's [expose interfaces instead of concrete implementations](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1002)
```suggestion
        protected IList<InstallFile> PreUpgradeScripts
```

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -120,44 +119,31 @@ protected ProviderConfiguration ProviderConfiguration
...
         /// <summary>
-        /// Gets the Upgrade Script (if present).
+        /// Gets a list of Post-Upgrade Scripts (if present) - these scripts will always run after
+        /// and versioned upgrade scripts and also after initial install.
         /// </summary>
-        /// <value>An InstallFile.</value>
+        /// <value>A list of <see cref="InstallFile"/> instances.</value>
         /// -----------------------------------------------------------------------------
-        protected InstallFile UpgradeScript
+        protected List<InstallFile> PostUpgradeScripts
```
bdukes, line 142:
```suggestion
        protected IList<InstallFile> PostUpgradeScripts
```

bdukes (approved):

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -1,369 +1,412 @@
...
+                            }
+                        }
+                    }
+                }
+
+                // Next process Post-UpgradeScripts - this script always runs if present
+                foreach (InstallFile file in this.PostUpgradeScripts)
+                {
+                    bSuccess = this.InstallScriptFile(file);
+                    if (file.Version != new Version(0, 0, 0))
```
bdukes, line 216: What's the purpose of requiring the file to have a version?  Are you wanting to check `installedVersion != new Version(0, 0, 0)`, i.e. it is an upgrade, not a new install?

mitchelsellers:
In reviewing this, I think that this is a great change, but I think there are a few additional places that we can better solidify expectations and processes, and I think one of the behaviors is a breaking change per @bdukes expectations.  

I believe that this meets the expectations of a Major feature, with a breaking change, and I believe we should target this to 10.x for that reason.

In addition to the currently outstanding questions, I think if we make this process more complicated we should look at the actual installer process and document or report exceptions for manifests that have improper information.  (For example there has been a long-standing thought that you could use <version> on the installer scripts, and it doesn't seem that is really the case, or at least it wouldn't impact things as expected.

In further consideration, and for consistency I think that Install/Upgrade scripts should have a similar behavior to that of the DNN Platform itself if we can, where the version number would be part of the equation.

Lastly, we have a "Type" in the manifest for these scripts, but looking at the code it is actually ignored due to convention (Un-install for example).  I think we might want to consider reviewing this for a more wholistic solution

donker:
> In addition to the currently outstanding questions, I think if we make this process more complicated we should look at the actual installer process and document or report exceptions for manifests that have improper information. (For example there has been a long-standing thought that you could use on the installer scripts, and it doesn't seem that is really the case, or at least it wouldn't impact things as expected.

I'm not sure how it makes life more complicated as it adds more options for developers, not asking them to jump through more hoops.

> In further consideration, and for consistency I think that Install/Upgrade scripts should have a similar behavior to that of the DNN Platform itself if we can, where the version number would be part of the equation.

Not quite sure what you're getting at as we use version nrs from the install (to set the current install version) and (versioned) upgrade scripts (same as before and for sorting) are currently taken into account.

> Lastly, we have a "Type" in the manifest for these scripts, but looking at the code it is actually ignored due to convention (Un-install for example). I think we might want to consider reviewing this for a more wholistic solution

As you'll see in the code I have begun using this element in the proposed code. I agree this should play a more prominent role in the future solution

mitchelsellers:
Sorry this process is a bit of a mess, and my initial comment wasn't quite correct.

My point here is that if we are adding more options "complication", or more areas that developers can hurt themselves, I think we need to make the process more reliable to avoid making catastrophic changes accidentally

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -1,369 +1,424 @@
...
-            return bSuccess;
-        }
-    }
-}
+        protected override void UnInstallFile(InstallFile scriptFile)
+        {
+            // Process the file if it is an UnInstall Script
+            if (this.UnInstallScripts.ContainsValue(scriptFile))
+            {
+                if (scriptFile.Name.StartsWith("uninstall.", StringComparison.InvariantCultureIgnoreCase))
```
bdukes, line 351: Do we still want this check?  Since we're already checking if it's in `UnInstallScripts` above, we shouldn't need to validate, and we've started supporting `type="uninstall"` without the naming convention (right?)

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
```diff
@@ -1,369 +1,424 @@
...
+                {
+                    // we couldn't determine the file type
+                    this.Log.AddFailure(string.Format(Util.SQL_Manifest_BadFile, file.Name));
+                    return;
+                }
+            }
+            else
+            {
+                // bad file
+                this.Log.AddFailure(Util.SQL_Manifest_Error);
```
bdukes, line 337: Should this failure reference the file name, too?

donker: The file is potentially null in the logic above

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: My guess is that it is still there for legacy's sake. If someone didn't label the file with a type uninstall it would still get through. But I'm all in favor of deprecating these quirks and focusing on the types instead.

donker: We can deprecate the old behavior in a major version IMO

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: The old upgrade files could bump the version nr. I'm not sure if this is (a) something that would actually work as these run after the upgrade scripts and (2) something that would even be useful. In a next major upgrade we could remove this IMO.

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Done

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: The preupgrade scripts are new, so there is no change here, only additional functionality

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Implemented

DNN Platform/Library/Services/Installer/Installers/ScriptInstaller.cs
donker: Done

bdukes (approved):

mitchelsellers (approved):

PR #4372, opened 2020-12-16
Enable warnings as errors for 3rd party auth providers

bdukes:
This enabled warnings as errors for all four external auth providers (Twitter, Facebook, Google, and Live)


valadas (approved):
Looks good to me

donker (approved):

PR #4373, opened 2020-12-16
Clean up warnings in DnnExportImport

bdukes:
594 obsolete and documentation warnings remain


mitchelsellers (approved):

DNN Platform/Modules/DnnExportImport/Components/Dto/TriCheckedState.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace Dnn.ExportImport.Components.Dto
+{
+    /// <summary>Three check states.</summary>
+    public enum TriCheckedState
+    {
+#if false
```
valadas, line 9: Hmm, I am not familiar with this pattern, I am fine leaving it as is, but anyone understands this purpose ? `#if false` to me looks like that code will never get hit no ?

DNN Platform/Modules/DnnExportImport/Components/Services/AssetsExportService.cs
```diff
@@ -1,674 +1,757 @@
...
+                        createdBy,
+                        file.Sha1Hash,
+                        DateUtils.GetDatabaseLocalTime(),
+                        file.Title,
+                        file.Description,
+                        file.StartDate,
+                        file.EndDate ?? Null.NullDate,
+                        file.EnablePublishPeriod,
+                        Null.NullInteger);
+                        ////file.ContentItemId ?? Null.NullInteger);--If we keep it we will see FK_PK relationship errors.
```
valadas, line 687: Should we just delete this line ?

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
```diff
@@ -2,2239 +2,2280 @@
...
+                    return;
+                }
+
+                this.UpdateTabChangers(localTab.TabID, createdBy, modifiedBy);
+
+                // this is not saved upon updating the tab
+                localTab.IsVisible = otherTab.IsVisible;
+                EntitiesController.Instance.SetTabSpecificData(localTab.TabID, false, localTab.IsVisible);
+
+                // tabController.UpdateTab(localTab); // to clear cache
```
valadas, line 319: Should we remove commented out code ?

valadas (changes requested):
Just one type and a couple of interrogations, otherwise, happy with this.

DNN Platform/Modules/DnnExportImport/Components/Dto/TriCheckedState.cs
bdukes: Correct, I think it's just a fancy way of commenting out the code.  it's just there to show what the previous version of the type was.  We should be fine to remove it, if we want.

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
bdukes: We can, doesn't matter much to me either way.  For both of them you've called out here, there's at least a comment indicating the thinking somewhat, which might protect against re-making the same mistake.

donker (approved):

DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
valadas: Ok, makes sense

valadas (approved):

PR #4375, opened 2020-12-16
Enable warnings as errors for JWT

bdukes:


mitchelsellers (approved):

valadas (changes requested):
I would prefer a solution where we don't override deprecated usages, so we actually give the good example specially if we enable "Treat warning as errors". As discussed in slack I gonna try and do another PR to add the functionality so 3rd parties actually have a way and not be forced to implement stuff we declared deprecated.

bdukes:
Just to be clear, if we added a feature to replace `PortalSettings.PortalAlias`, that would need to go in 9.9.0.  Do we want to hold this PR until then, or do we want to merge this as-is, and then adjust it when the better solution is available?

valadas (approved):
Yeah, I am ok merging this as it is.
I am just not in love with both bypassing the warning and enabling "Treat warnings as errors", but yeah, we can improve that soon I guess and then remove this

PR #4393, opened 2020-12-30
Fix Issue #3466 - Incorrect Redirect Response Codes

### LINKED ISSUE #3466: Page redirect always returns 301 (permanent) regardless of setting
## Description of bug
Page redirect always returns 301 (permanent) regardless of setting.

## Steps to reproduce
1. Create a new page
2. Redirect to an existing page (i.e. Home), DO NOT check "Permanent Redirect"
2. Load the page
3. Observe that the page is redirected using a 301 (permanent) redirect

## Current behavior
The permanent redirect setting is not observed.

## Expected behavior
The permanent redirect setting should be observed; in this case, the page should be redirected using a temporary redirect (302 or 307).

## Screenshots
image

image

## Additional context
I suspect this may be related to Advanced URL Management. Changing `DNNFriendlyUrl` to `urlFormat="humanfriendly"` (and accessing PageName.aspx) returns a 302 or 307 redirect as expected.

My primary dev system is Windows 10 running IIS 10, FYI.

## Affected version

* [X] 9.4.4
* [X] 8.0.4
* [X] 7.4.2

## Affected browser

* [X] Chrome
* [X] Firefox

mikesmeltzer:
Fixes #3466

While using the default Advanced URL format setting on the DNNFriendlyUrl config, redirects would always return 301 Permanent response codes even when the user had it set as a Temporary redirect from the Persona Bar.

This PR removes: 

1. the evaluation of a static setting (ForwardExternalUrlsType) that always forces both Internal Page and External URL redirects to be permanent without evaluating the user defined setting. This setting is hardcoded and not read from the DB.

This PR adds: 

1. logic to check if the user has defined the page as permanent/redirect for both Internal DNN Page redirects and External URL redirects
2. A new redirect reason so that the response reason matches the response code.

### DIFF
--- a/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
@@ -1023,22 +1023,18 @@ namespace DotNetNuke.Entities.Urls
                                     if (redirectUrl != null)
                                     {
                                         doRedirect = true;
-                                        if (settings.ForwardExternalUrlsType == DNNPageForwardType.Redirect301)
+                                        if (tab.PermanentRedirect)
                                         {
                                             result.Action = ActionType.Redirect301;
-                                            result.Reason = RedirectReason.Tab_External_Url;
                                         }
-                                        else if (settings.ForwardExternalUrlsType == DNNPageForwardType.Redirect302)
+                                        else
                                         {
                                             result.Action = ActionType.Redirect302;
-                                            result.Reason = RedirectReason.Tab_External_Url;
                                         }
                                     }
 
                                     break;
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):
This change looks good and seems to match the expected flow of the tab specific setting.  Will be sure to add many additional test cases during the release process for this

bdukes (approved):
Looks right to me

mikesmeltzer:
Is it possible for this to be included in 9.8.1 or will it need to wait till 9.8.2 (or whatever the next release is) since 9.8.1 RC has been released?

mitchelsellers:
@mikesmeltzer Given that the RC has already been published, this will be in the next release.

PR #4396, opened 2021-01-04
Add back method which was removed in #3969

bdukes:
## Summary
The public method `Mail.ConvertToText` was accidentally removed in #3969, this PR adds it back.

### DIFF
--- a/DNN Platform/Library/Services/Mail/Mail.cs
+++ b/DNN Platform/Library/Services/Mail/Mail.cs
@@ -24,6 +24,13 @@ namespace DotNetNuke.Services.Mail
 
     public class Mail
     {
+        public static string ConvertToText(string sHTML)
+        {
+            var formattedHtml = HtmlUtils.FormatText(sHTML, true);
+            var styleLessHtml = HtmlUtils.RemoveInlineStyle(formattedHtml);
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):
All for this because it shouldn’t have been removed. 

But going forward. Should we commit to having this method?

bdukes:
I think it's a helpful method (we discovered it was missing because @donker was using it).  However, we are duplicating the functionality (#3969 moved this to be a `private` method of `CoreMailProvider`).  It would be nice to adjust one of the call sites to use the other (which, in this case, would probably mean adjusting `CoreMailProvider` to call `Mail.ConvertToText`).

This isn't functionality I would plan to directly use, so I can't comment on whether it's still a fit solution, so perhaps there's an alternative we should promote instead.

david-poindexter:
@bdukes shouldn't this be targeted to the 9.8.1 release branch?

bdukes:
Not necessarily.  We can decide that's what we want to do.  However, this is not a regression introduced by 9.8.1 (it was introduced in 9.8.0), so it doesn't meet the standard criteria for hot-fix adding to an active RC.

donker (approved):
Thanks!

PR #4139, opened 2020-09-28
Hide new members that differ only by case

bdukes:
Addresses [CA1708](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1708)
>Identifiers for namespaces, types, members, and parameters cannot
>differ only by case because languages that target the common language
>runtime are not required to be case-sensitive. For example, Visual
>Basic is a widely used case-insensitive language.

Fixes #4138

PR #4397, opened 2021-01-05
Remove member with case-sensitive naming conflict

bdukes:
`PortalInfo.PortalId` was introduced in DNN 9.7.2, but should have been made private (i.e. should have been an explicit interface implementation), in order to avoid breaking case insensitive languages (e.g. VB.NET)

I'm not sure this is the right solution, since it will be breaking to anyone using the new member for the last three releases (whereas with #4139 we resolved an issue introduced in 9.7.2 in 9.7.3).


mitchelsellers (approved):
Although a breaking change, I think the usage should be limited, but I think the consistency is important so I'm inclined to approve.

valadas (approved):
Looks good to me

PR #4399, opened 2021-01-05
Admin Logs - Fixed Log Settings errors

### LINKED ISSUE #4398: Admin Logs - Add Log Setting returns 401 for admins
## Description of bug
`Add Log Setting` option in Admin Logs returns 401 for admins.

## Steps to reproduce
1. Open MANAGE > Admin Logs > Log Settings
2. Click on "Add Log Setting"
3. Enable the `Logging` switch, set `Log Type` to "Application Shutting Down"
4. Click on "Save"

## Current behavior
A "success" notification pops up saying "The log setting could not be added. Please try again later."

## Expected behavior
Log setting should be added without issues, but in case an issue occurs, it should be displayed as an error, not as a successful operation.

## Screenshots
image

## Error information
Following error is shown in the browser's console:
```
Failed to load resource: the server responded with a status of 401 (Unauthorized)
```
and occurs when PB issues a POST request to /API/PersonaBar/AdminLogs/AddLogSetting

## Additional context
It seems this behavior was introduced by this PR: https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions/pull/231 three years ago.
Looking in the original ticket, it seems some QA test was expecting Admin users not to be able to add Admin Log settings and therefore this condition was added, although it doens't make any sense to me.
Anyone have a clue why admins should not be able to add admin log settings?? and whether this is a mistake or it is expected behavior?

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.08.01 release candidate
* [X] 09.08.00 latest supported release
and all the way down to 9.2.0

## Affected browser
All browsers.

daguiler:
Fixes #4398

## Summary
This pull request reverts 4d1f352e8ca076ab9092362d9bc2da367dadc4e6 and changes the Admin Logs panel to show error notifications when Log Settings operations fail (instead of success notifications)

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/AdminLogsController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/AdminLogsController.cs
@@ -386,11 +386,6 @@ namespace Dnn.PersonaBar.AdminLogs.Services
         {
             try
             {
-                var isAdmin = this.UserInfo.Roles.Contains(this.PortalSettings.AdministratorRoleName);
-                if (isAdmin)
-                {
-                    return this.Request.CreateResponse(HttpStatusCode.Unauthorized);
-                }
                 request.LogTypePortalID = this.UserInfo.IsSuperUser ? request.LogTypePortalID : this.PortalId.ToString();
 
                 var logTypeConfigInfo = JObject.FromObject(request).ToObject<LogTypeConfigInfo>();
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/AdminLogsController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/AdminLogsController.cs
@@ -386,6 +386,11 @@ namespace Dnn.PersonaBar.AdminLogs.Services
         {
             try
             {
+                var isAdmin = this.UserInfo.Roles.Contains(this.PortalSettings.AdministratorRoleName);
+                if (isAdmin)
+                {
+                    return this.Request.CreateResponse(HttpStatusCode.Unauthorized);
+                }
                 request.LogTypePortalID = this.UserInfo.IsSuperUser ? request.LogTypePortalID : this.PortalId.ToString();
 
                 var logTypeConfigInfo = JObject.FromObject(request).ToObject<LogTypeConfigInfo>();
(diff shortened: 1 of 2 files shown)


bdukes:
@dnnsoftware/approvers do you need to be a superuser to manage log settings?  It looks like there's already logic so that a non-superuser can only manage settings for their current portal.  Anyone have a reason why this restriction would need to stay in place?

sleupold:
site specific settings should be able to be managed by portal admins (and superusers), however, all system events (relevent for all sites) should be managed by superusers only.

daguiler:
Actually, the `Log Settings` tab is not even displayed if you are not a super user.

mitchelsellers:
I would agree with @sleupold on this one, and I am not 100% sure that we have the proper distinction in being able to do this within the configuration settings.

For example, an Admin should not be able to turn off logging of user logins etc.

daguiler:
The issue that was reported to us is that admins/superusers are not allowed to add log settings. And given that this issue has been in place for more than 3 years, I think it's fair to assume that nobody is using this feature at all.
So my suggestion is to just remove this restriction (current changes in this PR) and then create a new issue describing the optimal behavior of this part of DNN.
Otherwise this will become a design change, and it would go out of scope of my bugfix.

mitchelsellers:
I don't think we can just remove the restriction, as that changes the security profile quite a bit..  Let me chat with the others about this one.

valadas:
I agree that this should be reserved to hosts only, admins should view the logs but not be able to add log types for instance...

daguiler:
Ok, I agree it would have changed the security profile.
I now pushed a commit that doens't affect the security profile except by allowing super users to add log settings.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/AdminLogsController.cs
```diff
@@ -386,12 +386,16 @@ public HttpResponseMessage AddLogSetting([FromBody] UpdateLogSettingsRequest req
...
-                    return this.Request.CreateResponse(HttpStatusCode.Unauthorized);
+                if (!this.UserInfo.IsSuperUser)
+                {
+                    var isAdmin = this.UserInfo.Roles.Contains(this.PortalSettings.AdministratorRoleName);
+                    if (isAdmin)
+                    {
+                        return this.Request.CreateResponse(HttpStatusCode.Unauthorized);
+                    }
+
+                    request.LogTypePortalID = this.PortalId.ToString();
```
bdukes, line 397: I think we just want to completely deny access unless you're a superuser.  I don't think there's a point in checking against the administrator role anymore.

```suggestion
                    return this.Request.CreateResponse(HttpStatusCode.Unauthorized);
```

daguiler:
Thanks @bdukes
It seems Github didn't honor the encoding and/or the new line characters, so it looks like a huge change now. But if you enable the `Hide whitespace changes` checkbox, it will look fine.

bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #4401, opened 2021-01-06
Merges latest changes from 9.8.1 into develop

valadas:
In this case just the versions and to keep history in sync.


bdukes (approved):

donker (approved):

valadas:
Oh, thanks, I totally forgot to merge that one...

PR #4406, opened 2021-01-08
Updates bug report template as per 9.8.1 release

bdukes:


valadas (approved):

valadas:

PR #4418, opened 2021-01-15
Resolved issue with lowercase URL's and account verifications

### LINKED ISSUE #2862: Verification link not working with lowered url casing
## Description of bug
Don't know if this is already solved in a specific version but I came accross this, so I thought I would at least add it here...

When you set up DNN to allow for 'Verified' registrations, people get an e-mail with a link with a verification code, so they can 'verify' their account. The verification code consists of lower and uppercase characters.
However, DNN has a setting under SEO > Url Management > General settings, which is called
'Convert URLs to Lowercase' 

So this will convert a verification code of let's say 'Aa4Gh3bJ' to 'aa4gh3bj', and then the verification code is not valid anymore, and people cannot login with their newly created account.

## Current result
People cannot verify their account with the verification code

## Expected result
Even in the case where the Convert URLs to lowercase is set ON, people should still be able to use the verification code in the URL

mitchelsellers:
## Summary
Resolved #2862 by introducing a new default exclusion for lowercase URL's to prevent lowercasing the URL's that contain the verification code.

This DOES not change anything for users on upgrade that has custom values for the "prevent lowercase" setting due to risk of manipulating the value they may have.

### DIFF
--- a/DNN Platform/Library/Entities/Urls/FriendlyUrlSettings.cs
+++ b/DNN Platform/Library/Entities/Urls/FriendlyUrlSettings.cs
@@ -405,7 +405,7 @@ namespace DotNetNuke.Entities.Urls
 
         public string ForceLowerCaseRegex
         {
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me

donker (approved):

PR #4383, opened 2020-12-17
Feature/4101 update ck editor version

skamphuis:
Closes #4101 

## Summary
I updated CKEditor to version 4.15.1 and made the required/appropriate changes for that:
- moved DNN icons and images out of CKEditor folder
- changed "broken" javascript in EditHTML.ascx for resizing the editor after window.resize
- updated version numbers for CKEditor in code (mainly for file paths)

PR #4427, opened 2021-01-21
Feature/fix cke link popup

### LINKED ISSUE #4426: Link popup not working in updated CKEditor
I ran into an issue with a fresh develop build, caused by my PR #4383 .

Opening the link popup when no element was present in the editor, caused a javascript error. The problem is in the dnnpages plugin.

I have the issue fixed already and will submit a PR right after this issue.

skamphuis:
Closes #4426

## Summary
I found an issue with the link popup in the update CKEditor provider, after my PR #4383 was merged.

Opening the link popup of the editor, when there is no content in the editor, results in a javascript error.
This was caused by the dnnpages plugin.

It turns out that the dnnpages plugin is a copy of the link plugin, with some changes to facilite the portal-page dropdown. So to fix this is issue, I compared the existing dnnpages plugin.js to the original plugin.js of the link plugin of CKE 4.5.3 (the old one) and implemented the same changes in a fresh copy of the link plugin of CKE 4.15.

Also I found that jquery.ckeditor.adapter.js is now in CKE itself, so the extra js file isn't needed anymore.


donker (approved):

bdukes (approved):
LGTM

PR #4430, opened 2021-01-22
Run unit tests in parallel

bdukes:
Cake currently runs each test assembly in its own test process and these processes are not run in parallel.  This PR adjusts the build to pass all test assemblies to the test runner at once.


mitchelsellers (approved):

valadas (approved):
Love it

PR #4431, opened 2021-01-23
removed languages not provided in dnnpages plugin

skamphuis:
In the upgraded dnnpages-plugin for CKEditor, I left the full list of langues. Since only 3 (en,de,pl) language files are actually supported currently, the others result in an error.

This simple removal provides the same solution we had before.

I skipped creating an issue this time, since it's not in a RC yet, anyway. If needed, I'm happy to create one, of course.


valadas:
If you can guide me to what is needed to fix languages I could fix fr or maybe some others...

valadas:
Oh, I should have looked before asking 😄 I added french...

donker:
Can you add Dutch to this, @skamphuis ? Since we support that language quite well ...

skamphuis:
Sure, will do that tomorrow. I had that, to be honest, but found it a bit "selfish" and was thinking about a solution to make all of the editor provider localizable.

donker (approved):
Looks good to me

mtrutledge (approved):
Looks good

PR #4432, opened 2021-01-25
Generic module event

donker:
## Summary
This PR adds a generic event for modules. It includes an example in the HTML module.

## Background
In a scenario where we'd want to allow a more loose coupling between DNN extensions, having a generic handler allows us to send events between modules. It would also allow simple messagebus-type applications. The sample shows how the HTML module would implement this. As the UpdateModule event only fires when the **Settings** of a module are updated, it doesn't tell us when the actual content of a module is updated. Now, with this simple change, the HTML module can signal that it's content has changed and any other component can leverage this if it needs to.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/Actions/IModuleEventHandler.cs
+++ b/DNN Platform/Library/Entities/Modules/Actions/IModuleEventHandler.cs
@@ -13,5 +13,7 @@ namespace DotNetNuke.Entities.Modules.Actions
         void ModuleRemoved(object sender, ModuleEventArgs args);
 
         void ModuleDeleted(object sender, ModuleEventArgs args);
+
+        void ModuleGenericEvent(object sender, ModuleGenericEventArgs args);
     }
 }
--- a/DNN Platform/Library/Entities/Modules/Actions/ModuleGenericEventArgs.cs
+++ b/DNN Platform/Library/Entities/Modules/Actions/ModuleGenericEventArgs.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Entities.Modules.Actions
+{
+    /// <summary>
+    /// Event arguments for a generic module event.
+    /// </summary>
+    public class ModuleGenericEventArgs : ModuleEventArgs
+    {
+        /// <summary>
+        /// Gets or sets the name of the event.
+        /// </summary>
+        public string EventName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the payload for the event.
+        /// </summary>
+        public object Payload { get; set; }
+    }
+}
--- a/DNN Platform/Modules/HTML/EditHtml.ascx.cs
+++ b/DNN Platform/Modules/HTML/EditHtml.ascx.cs
@@ -270,6 +270,7 @@ namespace DotNetNuke.Modules.Html
                         }
 
                         this._htmlTextController.UpdateHtmlText(htmlContent, this._htmlTextController.GetMaximumVersionHistory(this.PortalId));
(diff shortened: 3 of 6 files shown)


DNN Platform/Library/Entities/Modules/Actions/IModuleEventHandler.cs
```diff
@@ -12,6 +12,8 @@ public interface IModuleEventHandler
 
         void ModuleRemoved(object sender, ModuleEventArgs args);
 
-        void ModuleDeleted(object sender, ModuleEventArgs args);
+        void ModuleDeleted(object sender, ModuleEventArgs args);
+
+        void ModuleGenericEvent(object sender, ModuleGenericEventArgs args);
```
bdukes, line 17: Adding a new member to an interface (that we expect third parties to implement) is a breaking change, unfortunately…

DNN Platform/Library/Entities/Modules/Actions/ModuleGenericEventArgs.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the name of the event.
+        /// </summary>
+        public string EventName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the payload for the event.
+        /// </summary>
+        public object Payload { get; set; }
```
bdukes, line 20: Should we make this a generic class?

DNN Platform/Modules/HTML/EditHtml.ascx.cs
```diff
@@ -270,6 +270,7 @@ protected void OnSaveClick(object sender, EventArgs e)
                         }
 
                         this._htmlTextController.UpdateHtmlText(htmlContent, this._htmlTextController.GetMaximumVersionHistory(this.PortalId));
+                        Entities.EventManager.Instance.OnModuleGenericEvent(new Entities.Modules.Actions.ModuleGenericEventArgs { Module = this.ModuleConfiguration, EventName = "Update", Payload = htmlContent });
```
bdukes, line 273: I would think we'd want to set an example of having a unique event name, e.g. `DNN_HTML_Update`

DNN Platform/Library/Entities/Modules/Actions/ModuleGenericEventArgs.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Entities.Modules.Actions
+{
+    /// <summary>
+    /// Event arguments for a generic module event.
+    /// </summary>
+    public class ModuleGenericEventArgs : ModuleEventArgs
```
bdukes, line 10: Would we want to introduce a separate generic event for things that aren't related to a specific module?

bdukes (commented):
I think we may want to create an RFC to work out a list of use cases we do and don't want to support.

DNN Platform/Modules/HTML/EditHtml.ascx.cs
donker: Given that the module info is sent along we already know the context. So I can't see the point of asking module developers to adhere to a central naming scheme. Every module developer can create their own little world of conventions, IMHO.

DNN Platform/Library/Entities/Modules/Actions/ModuleGenericEventArgs.cs
donker: Probably. But it's good to have one for modules regardless of that effort, IMHO, given that we pass in the moduleinfo which is very rich in providing context.

mitchelsellers:
I believe that this needs a bit more consideration before we accept, as I echo all of @bdukes comments and have a few more.  Possibly discuss this @ the next approvers meeting?

donker:
After discussion the best avenue is to implement a more powerful and standardized approach with a library like MediatR. The main drawback of the solution proposed in this PR is that any consumer would need to filter for the module and event it is waiting for. This is obviously not very efficient and poses risks if this is scaled up.

Closing this for now.

PR #4433, opened 2021-01-25
Upgrade Cake to 1.0.0-RC3

bdukes:
We're currently using RC1 of Cake 1.0.0 and RC2 was recently released.


bdukes:
Looks like this is running into an issue with https://github.com/cake-build/cake/issues/3038, so this PR can hang out until next Cake release

bdukes:
I did adjust this in a way to work around the RC2 issue, so this does appear to be working if we want to merge.  No hurry.

donker (approved):

bdukes:
I've updated to the Cake 1.0.0-RC3 release, which is working fine now.

valadas (approved):
Looks good, thanks!

Issue #4434, opened 2021-01-25
Unable to upload files when there's an unexpected line break at allowable file extensions


mikebigun:
## Description of bug
Using new line separators in the allowed extensions list invalidates extension value itself. 

## Steps to reproduce

1. Go to `SETTINGS` - >`Security` -> `More` -> `More security settings`
2. Add a line break somewhere in between extensions
3. Save the changes

## Current behavior
The extension value which is moved to a new line is getting saved along with a new line separator `\n` in front of it. Due to this, the value of extension becomes invalid and application behaves incorrectly when needed to check extension whitelist. 

## Expected behavior
New line separators should be removed prior to saving the list to db.

## Screenshots

![image](https://user-images.githubusercontent.com/22524011/105744393-bccf0c80-5f45-11eb-99eb-3b92b63a8f7d.png)

## Affected version

* [ ] 10.00.00 alpha build
* [ ] 09.08.01 latest supported release

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge (Classic)
* [ ] Microsoft Edge Chromium

PR #4435, opened 2021-01-25
DNN-46668 Unable to upload files when there's an unexpected line brea…

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/4434

Fix summary:
Whitespaces for file extensions are trimmed.
UTs: Updated.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/EscapedString.cs
+++ b/DNN Platform/Library/Common/Utilities/EscapedString.cs
@@ -62,10 +62,35 @@ namespace DotNetNuke.Common.Utilities
         /// Takes an escaped string and splits it into an IEnumerable of seperate strings.
         /// </summary>
         /// <param name="combinedString">The string to seperate.</param>
+        /// <param name="trimWhitespaces">Trims whitespaces.</param>
+        /// <returns>IEnumerable of all the seperated strings.</returns>
+        /// <remarks>The escape character is '\', the seperator char is ','.</remarks>
+        public static IEnumerable<string> Seperate(string combinedString, bool trimWhitespaces)
+        {
+            return Seperate(combinedString, DefaultSeperator, trimWhitespaces);
+        }
+
+        /// <summary>
+        /// Takes an escaped string and splits it into an IEnumerable of seperate strings.
+        /// </summary>
+        /// <param name="combinedString">The string to seperate.</param>
+        /// <param name="separator">The character on which to split.</param>
+        /// <returns>IEnumerable of all the seperated strings.</returns>
+        /// <remarks>The escape character is '\', the seperator char is ','.</remarks>
+        public static IEnumerable<string> Seperate(string combinedString, char separator)
+        {
+            return Seperate(combinedString, separator, false);
+        }
+
+        /// <summary>
+        /// Takes an escaped string and splits it into an IEnumerable of seperate strings.
+        /// </summary>
+        /// <param name="combinedString">The string to seperate.</param>
         /// <param name="seperator">The character on which to split.</param>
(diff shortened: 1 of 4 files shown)


DNN Platform/Library/Common/Utilities/EscapedString.cs
```diff
@@ -74,7 +74,7 @@ public static IEnumerable<string> Seperate(string combinedString, char seperator
                 return result;
             }
 
-            var segments = combinedString.Split(new[] { seperator });
+            var segments = combinedString.RemoveNewLineSeparators().Split(new[] { seperator });
```
bdukes, line 77: Since this is a public method, I don't think we can alter this behavior without adding another argument to control whether or not to include whitespace.

Secondly, I think it would be more straightforward to call `Trim` on each segment, rather than do a `Regex.Replace` on the whole thing.  This would account for more whitespace characters than the hardcoded list, and would allow whitespace in the list values, just not around the values.  That seems like a better choice for a general-purpose utility method like this.

DNN Platform/Library/Common/Utilities/StringExtensions.cs
```diff
@@ -65,6 +65,21 @@ public static string ReplaceIgnoreCase(this string source, string oldValue, stri
...
+        /// <param name="source">The source.</param>
+        /// <returns>The string without new line separators.</returns>
+        public static string RemoveNewLineSeparators(this string source)
+        {
+            if (string.IsNullOrWhiteSpace(source))
+            {
+                return source;
+            }
+
+            return Regex.Replace(source, @"\t|\n|\r", string.Empty);
```
bdukes, line 82: If we're going to include tabs (`\t`) then let's not call the method `RemoveNewLineSeparators`.

DNN Platform/Library/Common/Utilities/StringExtensions.cs
```diff
@@ -65,6 +65,21 @@ public static string ReplaceIgnoreCase(this string source, string oldValue, stri
...
+        /// <summary>
+        /// Removes new line separators.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <returns>The string without new line separators.</returns>
+        public static string RemoveNewLineSeparators(this string source)
+        {
+            if (string.IsNullOrWhiteSpace(source))
+            {
+                return source;
```
bdukes, line 79: I would think we'd want a method named `RemoveNewLineSeparators` to still act on a string that is only white-space.  Should this check be `IsNullOrEmpty`?

bdukes (commented):
I definitely agree that we should be trimming whitespace that's in the file extension list, and I'm looking forward to coming up with a solution that works for everyone. Thanks!

DNN Platform/Library/Common/Utilities/EscapedString.cs
mikebigun: @bdukes thanks for this valuable input. You have mentioned important point regarding whitespace characters in the list values. Initial idea was to remove all occurrences of whitespace characters, so I had to to apply Regex.Replace to an entire string  (i.e. there is `jpeg` but user accidentally put the line break and we got `jpe\ng`).

Do you think we should keep those characters inside and trim them around only?

mikebigun: I agree with all your suggestions @bdukes Here is another solution.
Argument to trim whitespaces is added along with a multiple overloaded methods to ensure we do not introduce breaking changes into existing solution. Files extensions are trimmed now while other consumers of this `Separate` utility method works as before without changes.

bdukes (approved):
Great, thanks for the quick response!

donker (approved):

PR #4440, opened 2021-01-28
Use parameters instead of variables in pipeline

bdukes:
Currently the YAML pipeline for Azure DevOps documents the required variables in comments, but they're actually configured separately with the pipeline.  Using [parameters](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script) instead allows us to restrict the available options and declare defaults in the YAML document, instead of having them defined separately.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,11 +4,31 @@ pool:
 variables:
   - name: "Build.ArtifactStagingDirectory"
     value: "Artifacts/"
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

bdukes:
Here's what the Run Pipeline dialog looks like with this change:
Azure DevOps run dialog with parameters and open select list

valadas (approved):
Awesome

PR #4441, opened 2021-01-29
sitemappriority for new page set to 0.5

### LINKED ISSUE #3081: Sitemap priority 0 when adding a single page
@schotman commented on [Fri Mar 01 2019](https://github.com/dnnsoftware/Dnn.Platform/issues/2602)

## Description of bug
When using the add a new page function in the pages dialog the SiteMapPriority is not set in the dialog and defaults to 0 in the database

## Steps to reproduce
1. Go to pages 
2. Click on add page
3. add pagename and
4. click the add page button

## Current result
SiteMapPriority = 0

## Expected result
SiteMapPriority defaults to 0.5 (as it is when you add multiple pages)

## Screenshots
seems clear

## Error log
no error

## Affected version
* [x] 9.2.2

## Affected browser
not dependant of browser

* [x] Chrome
* [x] Firefox

---

@sleupold commented on [Fri Mar 01 2019](https://github.com/dnnsoftware/Dnn.Platform/issues/2602#issuecomment-468612435)

IMO this should be fixed in DNN API code, not the database default value.

---

@Timo-Breumelhof commented on [Fri Mar 01 2019](https://github.com/dnnsoftware/Dnn.Platform/issues/2602#issuecomment-468627463)

@skamphuis 

---

@Timo-Breumelhof commented on [Fri Mar 01 2019](https://github.com/dnnsoftware/Dnn.Platform/issues/2602#issuecomment-468627882)

workaround:

UPDATE TABS SET SiteMapPriority = 0.5 WHERE TABID IN (SELECT   TabID
FROM     Tabs
WHERE   (PortalID = 0) AND (SiteMapPriority = 0) and (IsSystem = 0)) 

---

@schotman commented on [Fri Mar 01 2019](https://github.com/dnnsoftware/Dnn.Platform/issues/2602#issuecomment-468642932)

> IMO this should be fixed in DNN API code, not the database default value.

Yes, it should

skamphuis:
Fixes #3081 

## Summary
I added the default value for sitemapPriority on the client, since several other "controls" get their ddefault values there too, and since I didn't really find a better place for it.
I concidered PagesController.GetDefaultSettings but that didn't seem right either.

This change adds the default value to the property that is being used for the value of the dropdown. When the page is saved, this value is posted to the server too and correctly saved.

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/services/pageService.js
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/services/pageService.js
@@ -92,6 +92,7 @@ const PageService = function () {
                 page.parentId = parentPage && typeof parentPage !== "function" && parentPage.id || page.parentId;
                 page.iconFile = null;
                 page.iconFileLarge = null;
+                page.sitemapPriority = 0.5;
                 return page;
             });
     };


bdukes:
Is the history of this issue that 0.5 was the default before the Persona Bar, and then we lost that default when the page settings was moved to the Persona Bar?

We probably also need to add a script to add 0.5 as the value for all pages missing a value, right?

I had thought that there was a setting somewhere to control the default value, but didn't find it after a quick search. Does anyone remember that being configurable, or did I make that up?

skamphuis:
> Is the history of this issue that 0.5 was the default before the Persona Bar, and then we lost that default when the page settings was moved to the Persona Bar?

Yes, that is indeed what happened. In the server side code you will find the default too, if you search for "0.5F", which is probably why it does work when adding multiple pages.
> 
> We probably also need to add a script to add 0.5 as the value for all pages missing a value, right?

I don't think that's possible, because we can't see which pages have been set to 0 intentionally, and which because it was the default.
> 
> I had thought that there was a setting somewhere to control the default value, but didn't find it after a quick search. Does anyone remember that being configurable, or did I make that up?

I'm not aware of that setting. I looked at what needs to be done to create such a setting, but I think that would be a nice enhancement and not so much a bug fix.

donker (approved):

valadas:
I have put this on hold to review further.

If memory serves good we have a site level default priority setting that had a problems with the value being a double in one place and an int somewhere else. I kinda remember there was a PR to fix that issue but I am on mobile now. At any rate if that is the case, we should default to that value and not 0.5 IMO

valadas:
Well, apparently memory did not serve good, the other PR must be just solving that value mismatch on one of those exiting fields, but we do not have a site-level default. So I am ok with this PR.

valadas (approved):
Looks good to me

PR #4444, opened 2021-01-31
Feature/2334 setting for minimum username length

### LINKED ISSUE #2334: Add User  - username validation problem
## Description of bug
When creating a username with less than 5 characters the validation message:
"The username specified is invalid. Please specify a valid username"

Looks like a good validation rule. 

But It's a problem when you upgraded from an older version and you already have users with username names that have less than 5 characters, You can't update these users anymore.

## Steps to reproduce
1. Go to 'Manage'- 'Users';
2. Click on 'Account Settings" of a user with a username less than 5 characters;
3. Click "Save".

## Current result
Validation message: "The username specified is invalid. Please specify a valid username"

## Expected result
There should be no validation on "old" users that have a username of less than 5 characters.

There should be a setting to enable/disable this validation.

Length of the minimal required characters for username should be a setting.

## Screenshots

image

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2

## Affected browser

* [x ] Chrome
* [x ] Firefox
* [ x] Safari
* [ x] Internet Explorer
* [x ] Edge

skamphuis:
Fixes #2334 

## Summary
This PR changes the built in minimum length of a username, to a setting under Security --> Member Accounts --> Registration Settings.
The default minimum length is now a constant in Globals.
Also, I changed the behavior of updating UserBasicDetails to only check (and change) a usrename when it has actually changed. This way, when the minimum username length is changed to a higher value, existing users with a shorter username can still be editted.
The username can of course only be changed to a valid length according to the new setting.

### DIFF
--- a/DNN Platform/Library/Common/Globals.cs
+++ b/DNN Platform/Library/Common/Globals.cs
@@ -197,6 +197,12 @@ namespace DotNetNuke.Common
         public const string glbUserNameRegEx = @"";
 
         /// <summary>
+        /// User Name default minimum length.
+        /// </summary>
+        /// <value></value>
+        public const int glbUserNameMinLength = 5;
+
+        /// <summary>
         /// Format of a script tag.
         /// </summary>
         /// <value><![CDATA[<script type=\"text/javascript\" src=\"{0}\" ></script>]]></value>
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -1813,7 +1813,15 @@ namespace DotNetNuke.Entities.Users
         public bool IsValidUserName(string userName)
         {
             char[] unallowedAscii = HostController.Instance.GetString("UsernameUnallowedCharacters", Globals.USERNAME_UNALLOWED_ASCII).ToCharArray();
-            return userName.Length >= 5 &&
+
+            // if the httpcontext is null, then use the default minimum length
+            var usernameMinLength = Globals.glbUserNameMinLength;
+            if (HttpContext.Current != null)
+            {
+                usernameMinLength = PortalController.GetPortalSettingAsInteger("Security_UserNameMinLength", this.PortalId, usernameMinLength);
+            }
+
+            return userName.Length >= usernameMinLength &&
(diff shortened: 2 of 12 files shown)


skamphuis:
(accidentally closed while looking around in the GitHub app 🤦‍♂️)

mitchelsellers (commented):
Overall, I think this is a positive change, however, I'm not sure of the user experience in how they will know how to clear the issue with the username.  As the message displayed to the user isn't going to help them understand what might be wrong.

skamphuis:
> 
> 
> Overall, I think this is a positive change, however, I'm not sure of the user experience in how they will know how to clear the issue with the username. As the message displayed to the user isn't going to help them understand what might be wrong.

Thanks. You're right the message didn't get any better. And I definitely agree that "Username is invalid" sounds a bit like "Computer says no". 
But the message didn't get any worse either :-).

I did have a quick look into the options to give the user a more specific message about what it is that's wrong with the username, but that's not all that easy because of the many parts that are involved. For starters, there is now one method that validates the username (Usercontroller.IsValidUserName). So that would either need to return something other than the boolean is does now, to indicate what was wrong with the username. Or we would need extra methods so the calling method can try and figure out itself what was wrong with it. Either way it's not ideal. But please do give me a hint if there's a better solution.

I figured the tooltip message could come in handy too, and then I noticed it still has the fixed length in it, so I'll change that in a later commit to this PR, so at least the label will show the correct minimum length.

skamphuis:
With my latest commit to this PR, the minimum length is also being checked client side. Final step will be to make the error message more specific than "Invalid username".

donker:
> Overall, I think this is a positive change, however, I'm not sure of the user experience in how they will know how to clear the issue with the username. As the message displayed to the user isn't going to help them understand what might be wrong.

I agree with Stefan. It is actually non-trivial to do this and the proposed change doesn't impact this. I.e. if you currently try to add a user with too short a username you also get the non-descript message.

donker (approved):
I ran a successful test of this. Built and tried adding a user with too short a username. It worked as advertised.

mitchelsellers (approved):

PR #4455, opened 2021-02-05
Add Logging to the Dependency Injection SafeGetTypes function

### LINKED ISSUE #4454: DependencyInjection is currently not logging loading errors
## Description of problem

When you implement a class with the IDnnStartup interface and are missing references in your package the module is not loaded as expected, however, there are no errors logged. There is currently a TODO in the SafeGetTypes function stating logging needs to be added.

## Description of solution
At the point of startup, the DependencyInjectionInitialize has a static ILog available. This just needs to be passed into the SafeGetTypes function.

## Affected Versions
All versions of DNN since 9.4.0

mtrutledge:
Fixes #4454

## Summary
Added logging to the SafeGetTypes function. The function would successfully load the types safely, without crashing, but would eat any exceptions that were thrown about missing references. I have added an `ILog` parameter to the function to print out the missing types.

### DIFF
--- a/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
+++ b/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
@@ -22,7 +22,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
         /// </summary>
         /// <param name="assembly">The assembly to retrieve all types from.</param>
         /// <returns>An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.</returns>
-        /// <remarks>This is obsolete because logging is not added. Please use the SafeGetTypes with the ILog parameter</remarks>
+        /// <remarks>This is obsolete because logging is not added. Please use the SafeGetTypes with the ILog parameter.</remarks>
         [Obsolete("Please use the SafeGetTypes with the ILog parameter.")]
         public static Type[] SafeGetTypes(this Assembly assembly)
         {
--- a/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
+++ b/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
@@ -23,7 +23,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
         /// <param name="assembly">The assembly to retrieve all types from.</param>
         /// <returns>An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.</returns>
         /// <remarks>This is obsolete because logging is not added. Please use the SafeGetTypes with the ILog parameter.</remarks>
-        [Obsolete("Please use the SafeGetTypes with the ILog parameter.")]
+        [Obsolete("Deprecated in DotNetNuke 9.9.0. Please use the SafeGetTypes overload with the ILog parameter. Scheduled removal in v11.0.0.")]
         public static Type[] SafeGetTypes(this Assembly assembly)
         {
             return assembly.SafeGetTypes(null);
--- a/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
+++ b/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
@@ -29,9 +29,7 @@ namespace DotNetNuke.DependencyInjection.Extensions
         /// <returns>
         /// An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.
         /// </returns>
-#pragma warning disable CS3001 // Argument type is not CLS-compliant
         public static Type[] SafeGetTypes(this Assembly assembly, ILog logger = null)
-#pragma warning restore CS3001 // Argument type is not CLS-compliant
         {
             Type[] types = null;
             try
--- a/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
+++ b/DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
@@ -15,11 +15,6 @@ namespace DotNetNuke.DependencyInjection.Extensions
     /// </summary>
     public static class TypeExtensions
     {
-        // There is no logging in this file by design as
-        // it would create a dependency on the Logging library
-        // and this library can't have any dependencies on other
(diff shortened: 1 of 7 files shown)


mitchelsellers (approved):
This looks awesome!

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
```diff
@@ -1,23 +1,25 @@
...
-        // it would create a dependency on the Logging library
-        // and this library can't have any dependencies on other
-        // DNN Libraries.
-
+    {
+        // There is no logging in this file by design as
+        // it would create a dependency on the Logging library
+        // and this library can't have any dependencies on other
+        // DNN Libraries.
+
```
valadas, line 22: @bdukes 

Do we:
🅰️ Remove this comment and accept having that dependency on DotNetNuke.Instrumentations
🅱️ There is a reason for not having that dependency and we need to think about another solution ?

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
```diff
@@ -26,10 +28,15 @@ public static class TypeExtensions
...
+        /// A optional <see cref="ILog"/> object. This will log any messages from the exception catching to the logs as an Error.
+        /// </param>
         /// <returns>
         /// An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.
-        /// </returns>
-        public static Type[] SafeGetTypes(this Assembly assembly)
+        /// </returns>
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
+        public static Type[] SafeGetTypes(this Assembly assembly, ILog logger = null)
+#pragma warning restore CS3001 // Argument type is not CLS-compliant
```
valadas, line 39: @mtrutledge What was not CLS compliant here. Since this class is public, it would be nice to be able to fix it so it is CLS compliant ?

valadas (changes requested):
Left 2 comments just to get other eyes on these concerns...

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
valadas: Could it be that it is because the DotNetNuke.Instrumentation library is simply not declared as CLSCompliant ? What that the issue

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
mtrutledge: I vote A.Ah missed that comment. There is no reason not to have a dependency there. The DI library was not referenced any where else and we already had a logger available. I can remove the comment.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
mtrutledge: I could not find a reason. I will check the DotNetNuke.Instrumentation dll and see if it needs the CLSCompliant added. if it does I will add it and update the PR.

mtrutledge: Yeah, the DotNetNuke.Instrumentation library was missing the CLS Compliant attribute. Added it and updated the PR.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
valadas: I agree, it's just not my expertise and I don't know (did not dig into it yet) who put the comment there, so would like to get input from @bdukes and/or @ahoefling  maybe on this...

mtrutledge: Also since there was already an ILog object in the DotNetNuke.DependencyInjection project that library already had a dependency on DotNetNuke.Instrumentation.

valadas: Oh, then should we adjust the .nuspec file to declare this dependency here https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Build/Tools/NuGet/DotNetNuke.DependencyInjection.nuspec @donker @mitchelsellers ?

valadas: Where is that other ILog @mtrutledge ?

mtrutledge: Line 19 of DependencyInjectionInitialize.cs

valadas: Awesome so my last concern is just about if we need to adjust the nuspec file @donker  ?

valadas (approved):
Ok, so I added the dependency in our nuspec file and I am ok with these changes until we have something different for logging.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
```diff
@@ -26,10 +23,13 @@ public static class TypeExtensions
         /// <param name="assembly">
         /// The assembly to retrieve all types from.
         /// </param>
+        /// <param name="logger">
+        /// A optional <see cref="ILog"/> object. This will log any messages from the exception catching to the logs as an Error.
+        /// </param>
         /// <returns>
         /// An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.
-        /// </returns>
-        public static Type[] SafeGetTypes(this Assembly assembly)
```
bdukes, line 32: This is a breaking change, can we use an overload instead of optional parameters?

mtrutledge: Yes, I will add an overload and remove the optional parameter.

mtrutledge: OK PR updated

bdukes:
:+1: Ready to merge (let's do a Squash Merge on this one) once the build is complete.

PR #4456, opened 2021-02-05
MailKit missing BouncyCastle.Crypto

mtrutledge:
When I was doing my PR for DependencyInjection logging the MailKit provider was logging an error saying it was missing the 1.8.8 version of BouncyCastle.Crypto. The NuGet package was referenced, however, the DLL was never being added to the bin folder on install. This PR fixes that issue.

### DIFF
--- a/DNN Platform/Library/DotNetNuke.Library.csproj
+++ b/DNN Platform/Library/DotNetNuke.Library.csproj
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-	<Deterministic>true</Deterministic>
+    <Deterministic>true</Deterministic>
   </PropertyGroup>
   <PropertyGroup>
     <ProjectType>Local</ProjectType>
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/DotNetNuke.Library.csproj
```diff
@@ -76,8 +76,8 @@
   </PropertyGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.Targets" />
   <ItemGroup>
-    <Reference Include="BouncyCastle.Crypto, Version=1.8.9.0, Culture=neutral, PublicKeyToken=0e99375e54769942, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Portable.BouncyCastle.1.8.9\lib\net40\BouncyCastle.Crypto.dll</HintPath>
+    <Reference Include="BouncyCastle.Crypto, Version=1.8.8.0, Culture=neutral, PublicKeyToken=0e99375e54769942, processorArchitecture=MSIL">
```
valadas, line 79: Any reason for lowering the version or is that a mistake ?

DNN Platform/Library/packages.config
```diff
@@ -11,7 +11,7 @@
   <package id="MimeKit" version="2.10.1" targetFramework="net472" />
   <package id="Newtonsoft.Json" version="10.0.3" targetFramework="net45" />
   <package id="PetaPoco.Compiled" version="6.0.415" targetFramework="net472" />
-  <package id="Portable.BouncyCastle" version="1.8.9" targetFramework="net472" />
+  <package id="Portable.BouncyCastle" version="1.8.8" targetFramework="net472" />
```
valadas, line 14: Any reason for lowering the version or is that a mistake ?

valadas (changes requested):
Not sure if there is a reason to lower the version of if it was a mistake ?

DNN Platform/Library/DotNetNuke.Library.csproj
mtrutledge: The version that MailKit and MimeKit were depending on was 1.8.8. With 1.8.9 in there .NET Kept throwing errors trying to load the dll. Wasnt until I went to 1.8.8 that the loading errors stopped.

DNN Platform/Library/packages.config
mtrutledge: The version that MailKit and MimeKit were depending on was 1.8.8. With 1.8.9 in there .NET Kept throwing errors trying to load the dll. Wasnt until I went to 1.8.8 that the loading errors stopped.

DNN Platform/Library/DotNetNuke.Library.csproj
valadas: Interesting... Ok, makes sense then...

DNN Platform/Library/packages.config
valadas: Interesting... Ok, makes sense then...

valadas (approved):

mitchelsellers (approved):

PR #4459, opened 2021-02-07
Migrate from moment to dayjs for Sites.Web

david-poindexter:
## Summary
This is related to #3875 and is for the `Sites.Web` project.  
* `moment` replaced with `dayjs`
* ClientSide project name changed from `boilerplate` (wtf) to `sites`


david-poindexter:
Please hold on merging this because I need to correct the version numbers.

valadas (approved):
Looks good to me other than the version nit which does not matter really as the build will override it but if you can just revert the version number it would look nicer...

david-poindexter:
> Looks good to me other than the version nit which does not matter really as the build will override it but if you can just revert the version number it would look nicer...

@valadas the version numbers have now been corrected.  I am not clear as to how they ever got changed in the first place.  🤷🏼‍♂️

mitchelsellers (approved):

PR #4462, opened 2021-02-07
Remove moment package from Roles.Web

david-poindexter:
`moment` is not used in this project, so to avoid confusion, the package has been removed. This resolves the need identified in #3875 for migrating to `dayjs` as well.


valadas (approved):
Awesome

mitchelsellers (approved):

PR #4464, opened 2021-02-07
Remove moment package from AdminLogs.Web

david-poindexter:
## Summary
Remove `moment` package from `AdminLogs.Web` because it is not used.  This PR relates to #3875.


valadas (approved):
Awesome

mitchelsellers (approved):

PR #4465, opened 2021-02-07
Remove moment package from Prompt.Web

david-poindexter:
## Summary
Remove `moment` package from `Prompt.Web` since it is not used.  This PR relates to #3875


valadas (approved):
Excellent

mitchelsellers (approved):

Issue #3375, opened 2019-12-05
Add Security policy to develop branch


david-poindexter:
## Description of documentation
The security policy is missing from the new `develop`branch.

## Resolution
Add SECURITY.md to the root of project.

## Additional context
This seemed to have been overlooked in the setup of the new default develop branch.

PR #4466, opened 2021-02-08
Migrate moment to dayjs for Dnn.React.Common

david-poindexter:
## Summary
Migrate `moment` to `dayjs` for `Dnn.React.Common` - relates to #3375

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Dnn.React.Common/.storybook/utils.js
+++ b/Dnn.AdminExperience/ClientSide/Dnn.React.Common/.storybook/utils.js
@@ -184,7 +184,9 @@ function formatDate(dateValue, longformat) {
         return "-";
     }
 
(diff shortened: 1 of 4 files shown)


valadas (approved):
Looks good... 

For v.10 I think it would be a good idea to investigate the other `externals` that are unused, this could seriously reduce the common bundle size.

The goal I think was that modules would use the ones from the common bundle, but most of the time it looks like that is not even the case.

Anyway, just a note, nothing that has to do with this PR.

mitchelsellers (approved):

david-poindexter:
> Looks good... 
> 
> 
> 
> For v.10 I think it would be a good idea to investigate the other `externals` that are unused, this could seriously reduce the common bundle size.
> 
> 
> 
> The goal I think was that modules would use the ones from the common bundle, but most of the time it looks like that is not even the case.
> 
> 
> 
> Anyway, just a note, nothing that has to do with this PR.

Agreed 💯

PR #4474, opened 2021-02-09
Add explicit values to all enums

### LINKED ISSUE #4289: Not Explicitly Specifying Enumeration Values Can Lead to Unexpected Results
## Description of Issue/Bug
First, let me say I don't know if this should be considered a bug or not. Maybe it is just a suggestion for a change in the code that code prevent or minimize future issues.

Enumerations such as DotNetNuke.Services.Mail.MessageType do not have explicitly declared values. This can lead to unexpected results for third-party modules that were complied against one version of an enum if the order or the number of enum values changes in the current version.

## Real World Example
We had developed a custom registration module for a client. We and the client were aware that there would be a dependency on DNN and that future DNN updates could break functionality. The custom registration code was built when the client was running on DNN 9.3.2. Later the client needed to upgrade DNN to 9.6.2. The upgrade was tested (not well enough however) and no issues were found and was deployed to production. Soon, we started receiving issues of user not being able to login. In most cases the result was their accounts not being verified. I started digging into the issue and finally realized the issue. Somewhere between 9.3.2 and 9.6.2 the enum values for DotNetNuke.Services.Mail.MessageType had changed. The value 'UserRegistrationVerified' which had been 5 was now 6. This meant when the custom registration code called "Mail.SendMail(user, MessageType.UserRegistrationVerified, portalSettings)" it was sending a 5 which in the new enum mapped to 'UserRegistrationPublic'. Thus, the user was not getting the email with the verification link but the public registration email instead which did not include the verification link.

## Suggested Solution
If enums in DNN were to explicitly declare the numeric value and there was an attempt to maintain these values whenever possible this would eliminate or minimize the issue we ran into. I can imagine a scenario where a third-party developer is trying to target multiple versions of DNN where this issue would become even more of a problem. Explicitly declaring the enum numeric values would seem to be a practice that DNN enums should employ.

`namespace DotNetNuke.Services.Mail
{
    public enum MessageType
    {
        PasswordReminder = 0,
        ProfileUpdated = 1,
        UserRegistrationAdmin = 2,
        UserRegistrationPrivate = 3,
        UserRegistrationPrivateNoApprovalRequired = 4,
        UserRegistrationPublic = 5,
        UserRegistrationVerified = 6,
        UserUpdatedOwnPassword = 7,
        PasswordUpdated = 8,
        PasswordReminderUserIsNotApproved = 9,
        UserAuthorized = 10,
        UserUnAuthorized = 11
    }
}`

If there is another solution or something else we should have done differently please let me know.

Thanks

bdukes:
Fixes #4289


mitchelsellers (approved):

valadas (approved):
Looks good to me

PR #4478, opened 2021-02-10
Update Compress Images GitHub Actions workflow

bdukes:
The Compress Images workflow isn't catching most of the images being
committed. This is, in part, because it doesn't run with permissions to
edit PRs on forks. This updates the configuration to open a new PR when
a commit is merged, if there are images that can be compressed. It also
adds a scheduled run, in case something gets missed.

Based on https://github.com/calibreapp/image-actions#combined-workflow

### DIFF
--- a/.github/workflows/image-actions.yml
+++ b/.github/workflows/image-actions.yml
@@ -1,7 +1,23 @@
 name: Compress images
-on: pull_request
+on:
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good

donker (approved):

PR #4480, opened 2021-02-10
Optimize images

bdukes:
We have new images in DNN 9.9.0 that didn't get optimized (see #4478 for a related fix to the workflow). This PR is just a rebase of the automatic commit from the release branch with those optimizations.


sleupold:
I doubt, compressing editor icons and smileys will have a big size benefit - will this really improve performance?

david-poindexter (approved):
Thanks @bdukes

mitchelsellers (approved):
This looks good. @sleupold the reduction here is > 50% on average per file, considering multiple images are loaded there is a good gain.  Additionally, as part of our continuous improvement efforts we should be optimizing everything.

PR #4482, opened 2021-02-10
Fixes #4481 AzureFolderProvider uploading same file with different case

### LINKED ISSUE #4481: AzureFolderProvider - Issue re-uploading same name file but with different case.
## Description of bug
If you upload a file say DNNINTERNALS.pdf, then upload dnninternals.pdf, DNN will correctly ask you if you want to replace the current file, and if you choose to replace it, everything is fine in DNN, however because Azure is case sensitive DNN is leaving the original file in Azure.

## Steps to reproduce
1. (Using Azure Folder Provider.  Upload 190701-F-JL169-1028D.JPG
2. Upload 190701-F-JL169-1028D.jpg (with lower case extension)
image
3. Click "Replace" to replace the existing file
4. Observe what is left in Azure, both files exist
image

## Current behavior
DNN is not removing the replaces file because of Azure's case sensitivity

## Expected behavior
DNN should remove the file that is being replaced

* [ ] 10.00.00 alpha build
* [ ] 09.09.00 release candidate
* [X ] 09.08.01 latest supported release

msant7:
Fixes #4481

## Summary
Override UpdateFile in the AzureFolderProvider, look to see if the original filename is different from the new one.  If it is, then delete it.

### DIFF
--- a/DNN Platform/Providers/FolderProviders/AzureFolderProvider/AzureFolderProvider.cs
+++ b/DNN Platform/Providers/FolderProviders/AzureFolderProvider/AzureFolderProvider.cs
@@ -270,11 +270,18 @@ namespace DotNetNuke.Providers.FolderProviders.AzureFolderProvider
             this.ClearCache(folderMapping.FolderMappingID);
         }
 
+        /// <summary>
+        /// Updates a file in Azure folder provider.
+        /// </summary>
+        /// <remarks>
+        /// Azure is case sensitive.  If you update dnninternals.pdf with DNNINTERNALS.pdf, to DNN it's the same
+        /// so it just re-uploads it, causing both files to exist in Azure. This azure specific method deletes the
+        /// old existing duplicate that only differs in case as part of the update.
+        /// </remarks>
         public override void UpdateFile(IFolderInfo folder, string fileName, Stream content)
         {
-            // Azure is case sensitive.  If you update dnninternals.pdf with DNNINTERNALS.pdf, to DNN it's the same
-            // so it just re-uploads it, causing both files to exist in Azure.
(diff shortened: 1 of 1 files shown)


DNN Platform/Providers/FolderProviders/AzureFolderProvider/AzureFolderProvider.cs
```diff
@@ -270,6 +270,21 @@ protected override void MoveFolderInternal(FolderMappingInfo folderMapping, stri
             this.ClearCache(folderMapping.FolderMappingID);
         }
 
+        public override void UpdateFile(IFolderInfo folder, string fileName, Stream content)
+        {
+            // Azure is case sensitive.  If you update dnninternals.pdf with DNNINTERNALS.pdf, to DNN it's the same
+            // so it just re-uploads it, causing both files to exist in Azure.
+            IFileInfo originalFile = FileManager.Instance.GetFile(folder, fileName);
```
valadas, line 277:
```suggestion

        /// <summary>
        /// Updates a file in Azure folder provider.
        /// </summary>
        /// <remarks>
        /// Azure is case sensitive.  If you update dnninternals.pdf with DNNINTERNALS.pdf, to DNN it's the same
        /// so it just re-uploads it, causing both files to exist in Azure. This azure specific method deletes the
        /// old existing duplicate that only differs in case as part of the update.
        /// </remarks>
        public override void UpdateFile(IFolderInfo folder, string fileName, Stream content)
        {
            IFileInfo originalFile = FileManager.Instance.GetFile(folder, fileName);

```

valadas (changes requested):
Can we just update the comment to a method documentation like so or similar

donker (approved):

mitchelsellers (approved):

PR #4484, opened 2021-02-11
Fixed DDRMenu to handle duplicate page names

### LINKED ISSUE #4483: DDR Menu doesn't handle duplicate page names correctly
## Description of bug

## Steps to reproduce
1) Create following page hierarchy:
- Home
- ⤷ DDRMenu Test
- DDRMenu Test
- Some Other Page
- And Yet Another Page

the key here is to have 2 pages with identical names (at different branches of the hierarchy of course)

2) Change the DDR Menu tag in Portals\_default\Skins\Xcillion\Home.ascx to look like this:
```
<dnn:MENU ID="MENU" MenuStyle="Menus/MainMenu" runat="server" NodeSelector="*" IncludeNodes="Home, DDRMenu Test"></dnn:MENU>
```
(notice we are including only "Home" and "DDRMenu Test")

3) Refresh the home page
## Current behavior
The menu shows the "Home" link, and only 1 of the "DDRMenu Test" pages:

image

## Expected behavior
The menu should display both instances of the duplicate page name:

image

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.09.00 release candidate
* [X] 09.08.01 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

daguiler:
Fixes #4483

## Summary
The bug is located [here](https://github.com/dnnsoftware/Dnn.Platform/blob/df519597b3c45f56f56defa1936314f6612ee644/DNN%20Platform/Modules/DDRMenu/MenuBase.cs#L274) (irrelevant parts omitted for brevity):

image

The `FilterNodes` function cycles through the names specified in the "IncludeNodes" attribute (available in the `nodeTextStrings` variable). For each of these names, it invokes the `FindByNameOrId` function of the root node of the menu. The idea is to populate the `filteredNodes` list with all the nodes that we need to preserve. Later on, the `RemoveAll(<predicate>)` function is called, which takes care of removing all items of the menu that are not present in this list.

The problem is that `FindByNameOrId` just returns _the first_ node that meets the criteria. When several pages share a common name, only the page that is found first survives this purge.

The solution I analyzed first was changing `FindByNameOrId` to return a list of all matching nodes, and not just the first node that satisfies the condition. However, this function is a public function referenced from several places. Changing its signature would be a breaking change.

So I chose to create a new function for this purpose (`FindAllByNameOrId`) and modify `FilterNodes` to call this new function instead.


DNN Platform/Modules/DDRMenu/MenuNode.cs
```diff
@@ -326,6 +326,31 @@ public MenuNode FindByNameOrId(string tabNameOrId)
...
+                found.AddRange(child.FindAllByNameOrId(tabNameOrId));
+            }
+
+            if (tabNameOrId.Equals(this.Text, StringComparison.InvariantCultureIgnoreCase)
+                || tabNameOrId == this.TabId.ToString())
+            {
+                found.Add(this);
+            }
+
+            return found;
```
bdukes, line 351: I think this fix works, thanks! What do you think about making this an iterator method instead of creating a bunch of (mostly empty) lists?

```suggestion
            foreach (var found in this.Children.SelectMany(child => child.FindAllByNameOrId(tabNameOrId)))
            {
                yield return found;
            }

            if (tabNameOrId.Equals(this.Text, StringComparison.InvariantCultureIgnoreCase)
                || tabNameOrId == this.TabId.ToString())
            {
                yield return this;
            }
```

bdukes (approved):

DNN Platform/Modules/DDRMenu/MenuNode.cs
daguiler: That's of course a better solution. I was too lazy and just copied the original function. Accepting.

Timo-Breumelhof:
Cool, thanks for the fix :-)

valadas:
It would be a bit of a breaking change for some sites. I assigned this to Dnn10 for that reason.

PR #4488, opened 2021-02-12
Remove samples folder from CKEditor

bdukes:
The CKEditor upgrade includes a `samples` folder that wasn't in the previous version. Among other things, it includes `php` files which get flagged by the Security Analyzer. These shouldn't be included.


bdukes:
@skamphuis I didn't see any usages of this `samples` folder, does that seem right to you that we can remove it altogether?

valadas (approved):
Good idea

skamphuis:
Yes, that does seem right. I obviously just unpacked the editor there.

Another thing I though of the other day: should we remove the old one in the installer package?

mitchelsellers (approved):

bdukes:
@skamphuis yes, removing the old files in the installer makes sense to me

PR #4208, opened 2020-10-18
Removed Module Creator from distribution

valadas:
- Removes the Module Creator admin module from the distribution as per it's move to https://github.com/DNNCommunity/Dnn.ModuleCreator/releases/tag/v9.8.1
- Also reduces build warnings by about 46

Closes #3620

PR #4495, opened 2021-02-15
Remove references to Module Creator

bdukes:
As of DNN 9.8.1, Module Creator is not included in the distribution of DNN (see #4208)

However, there are other Module Creator references that were left after that removal.  There were still references in the database script, so the module would still be created on install, even though the files didn't exist.=
Error message from trying to add Module Creator to a page

Also, the _Develop_ module action was pointing to the module creator.  This PR moves it to another View Source control that is shipped with the Platform (presumably what was used before the Module Creator).
Screenshot of old View Source control


david-poindexter (approved):
Great catch @bdukes 🎉

mitchelsellers (approved):

PR #4500, opened 2021-02-17
Azure page css

### LINKED ISSUE #4499: Page Style Sheet can not be used with AzureFolderProvider
## Description of bug
If the user uploads a CSS style sheet in Site Assets using a Folder Provider such as the AzureFolderProvider, then tries to use that stylesheet on a page using the built-in DNN Page Stylesheet setting, DNN will not include the stylesheet because it doesn't exist on the local filesystem.

## Steps to reproduce
1. Upload a css file to an AzureFolderProvider folder.
image
2. Add a dnn page and set the Page Stylesheet to the file uploaded
image
3. Save the Page

## Current behavior
Since the css file doesn't exist in the local filesystem, it's not included.

## Expected behavior
If the css file is on the local file system, it should be included as it currently does through the ClientResourceManager
If the css file is through a folder provider that generates an external http[s]:// url, the css file should be included in the page header as a `<link>`  example `<link href="https://azurefp.blob.core.windows.net/forkofdnn/Cloud%202/test.css" rel="stylesheet" type="text/css" />`

## Affected version
* [ ] 10.00.00 alpha build
* [ ] 09.09.00 release candidate
* [X] 09.08.01 latest supported release

msant7:
Fixes #4499

## Summary
1. To support Page Stylesheets stored in AzureFolderProvider (and others).  Attempt to get the Page Stylesheet from the FolderProvider first.  If it exists with an external url, then include the css in a page header `link`.  If not, include through ClientResourceMananger (current way)
2. Add CSS contenttype to FileContentTypeManager.cs and Database Lists of type 'ContentTypes'.  If this doesn't exist .css will be uploaded as 'application/octet-stream' and browsers won't use it as a css file.

### DIFF
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -632,8 +632,29 @@ namespace DotNetNuke.Framework
             // register the custom stylesheet of current page
             if (this.PortalSettings.ActiveTab.TabSettings.ContainsKey("CustomStylesheet") && !string.IsNullOrEmpty(this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString()))
             {
-                var customStylesheet = Path.Combine(this.PortalSettings.HomeDirectory, this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString());
-                ClientResourceManager.RegisterStyleSheet(this, customStylesheet);
+                var customStylesheet = this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString();
+                var cssIncluded = false;
+                
+                // Try and go through the FolderProvider first, if it gives back an external url, include it
+                // as a <link> in the Page Header
+                IFileInfo fi = GetPageStylesheetFileInfo();
+                if (fi != null)
+                {
+                    string url = FileManager.Instance.GetUrl(fi);
+                    if (url.StartsWith("http://", StringComparison.InvariantCultureIgnoreCase) || url.StartsWith("https://", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        string cssLink = string.Format("<link href=\"{0}\" rel=\"stylesheet\" type=\"text/css\" />", url);
+                        this.Page.Header.Controls.Add(new LiteralControl(cssLink));
+                        cssIncluded = true;
+                    }
+                }
+
+                // Include it as DNN always has, as a relative path through ClientResourceManager
+                if (!cssIncluded)
+                {
+                    var customStylesheetPath = Path.Combine(this.PortalSettings.HomeDirectory, customStylesheet);
+                    ClientResourceManager.RegisterStyleSheet(this, customStylesheetPath);
+                }
             }
 
             // Cookie Consent
@@ -736,6 +757,22 @@ namespace DotNetNuke.Framework
         private IFileInfo GetBackgroundFileInfoCallBack(CacheItemArgs itemArgs)
         {
             return FileManager.Instance.GetFile(this.PortalSettings.PortalId, this.PortalSettings.BackgroundFile);
+        }
+
+        private IFileInfo GetPageStylesheetFileInfo()
+        {
+            string cacheKey = string.Format(Common.Utilities.DataCache.PortalCacheKey, this.PortalSettings.PortalId, "PageStylesheet" + this.PortalSettings.ActiveTab.TabID);
+            var file = CBO.GetCachedObject<Services.FileSystem.FileInfo>(
+                new CacheItemArgs(cacheKey, Common.Utilities.DataCache.PortalCacheTimeOut, Common.Utilities.DataCache.PortalCachePriority),
+                this.GetPageStylesheetInfoCallBack);
+
+            return file;
+        }
+
+        private IFileInfo GetPageStylesheetInfoCallBack(CacheItemArgs itemArgs)
+        {
+            var customStylesheet = this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString();
+            return FileManager.Instance.GetFile(this.PortalSettings.PortalId, customStylesheet);
         }
     }
 }
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -632,28 +632,17 @@ namespace DotNetNuke.Framework
             // register the custom stylesheet of current page
             if (this.PortalSettings.ActiveTab.TabSettings.ContainsKey("CustomStylesheet") && !string.IsNullOrEmpty(this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString()))
             {
-                var customStylesheet = this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString();
-                var cssIncluded = false;
-                
-                // Try and go through the FolderProvider first, if it gives back an external url, include it
-                // as a <link> in the Page Header
-                IFileInfo fi = GetPageStylesheetFileInfo();
-                if (fi != null)
-                {
-                    string url = FileManager.Instance.GetUrl(fi);
-                    if (url.StartsWith("http://", StringComparison.InvariantCultureIgnoreCase) || url.StartsWith("https://", StringComparison.InvariantCultureIgnoreCase))
-                    {
-                        string cssLink = string.Format("<link href=\"{0}\" rel=\"stylesheet\" type=\"text/css\" />", url);
-                        this.Page.Header.Controls.Add(new LiteralControl(cssLink));
-                        cssIncluded = true;
-                    }
-                }
+                var styleSheet = this.PortalSettings.ActiveTab.TabSettings["CustomStylesheet"].ToString();
 
-                // Include it as DNN always has, as a relative path through ClientResourceManager
-                if (!cssIncluded)
+                // Try and go through the FolderProvider first
+                var stylesheetFile = GetPageStylesheetFileInfo(styleSheet);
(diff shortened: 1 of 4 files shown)


DNN Platform/Website/Default.aspx.cs
```diff
@@ -632,8 +632,29 @@ private void InitializePage()
...
+                        cssIncluded = true;
+                    }
+                }
+
+                // Include it as DNN always has, as a relative path through ClientResourceManager
+                if (!cssIncluded)
+                {
+                    var customStylesheetPath = Path.Combine(this.PortalSettings.HomeDirectory, customStylesheet);
+                    ClientResourceManager.RegisterStyleSheet(this, customStylesheetPath);
+                }
```
bdukes, line 657: `RegisterStyleSheet` can handle external URLs, so I don't think we need different code paths here
```suggestion
                var stylesheetFile = GetPageStylesheetFileInfo();
                if (stylesheetFile != null)
                {
                    ClientResourceManager.RegisterStyleSheet(this, FileManager.Instance.GetUrl(stylesheetFile));
                }
```

DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.09.01.SqlDataProvider
```diff
@@ -0,0 +1,15 @@
...
+/*****              SqlDataProvider                     *****/
+/*****                                                  *****/
+/*****                                                  *****/
+/***** Note: To manually execute this script you must   *****/
+/*****       perform a search and replace operation     *****/
+/*****       for {databaseOwner} and {objectQualifier}  *****/
+/*****                                                  *****/
+/************************************************************/
+
+-- Add CSS contnet type, this will help when a Page Style Sheet is uploaded to a folder provider such as AzureFolderProvider.
```
bdukes, line 11:
```suggestion
-- Add CSS content type, this will help when a Page Style Sheet is uploaded to a folder provider such as AzureFolderProvider.
```

DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.09.01.SqlDataProvider
```diff
@@ -0,0 +1,15 @@
...
+IF NOT EXISTS (SELECT * FROM {databaseOwner}{objectQualifier}Lists WHERE ListName='ContentTypes' AND Value = 'css')
+BEGIN
+    INSERT INTO {databaseOwner}[{objectQualifier}Lists] ( [ListName], [Value], [Text], [ParentID], [Level], [SortOrder], [DefinitionID], [Description], [PortalID], [SystemList], [CreatedByUserID], [CreatedOnDate], [LastModifiedByUserID], [LastModifiedOnDate]) VALUES (N'ContentTypes', N'css', N'text/css', 0, 0, 0, -1, N'', -1, 1, -1, '2020-02-17 09:13:51.647', -1, '2020-02-17 09:13:51.647')
```
bdukes, line 14: Was there are reason to use a hardcoded date? I see that the previous script did that, too:
https://github.com/dnnsoftware/Dnn.Platform/blob/714db32390f800ea2336bb0fd2d54fc589256e6b/DNN%20Platform/Website/Providers/DataProviders/SqlDataProvider/07.03.02.SqlDataProvider#L117-L167

Even more weird is that when I look at the `ContentTypes` list in a site I manage, all of the `CreatedOnDate` fields are set to `2000-01-01T00:00:00` and the last modified fields are `NULL`…

Any @dnnsoftware/approvers know what the deal is?  I can't imagine it makes much difference, but it certainly is confusing.

```suggestion
    INSERT INTO {databaseOwner}[{objectQualifier}Lists] ( [ListName], [Value], [Text], [ParentID], [Level], [SortOrder], [DefinitionID], [Description], [PortalID], [SystemList], [CreatedByUserID], [CreatedOnDate], [LastModifiedByUserID], [LastModifiedOnDate]) VALUES (N'ContentTypes', N'css', N'text/css', 0, 0, 0, -1, N'', -1, 1, -1, GETDATE(), -1, GETDATE())
```

bdukes (commented):
Thanks!

DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.09.01.SqlDataProvider
msant7: Yes, I was just following what was done before, no issue changing it to getdate()

DNN Platform/Website/Default.aspx.cs
msant7: Thanks Brian! I didn't know RegisterStyleSheet could handle absolute urls.  My only concern is that while any CSS that is referenced "Should" be in the dnn file system.  I image there could be a case where it's not in the DNN filesystem especially since the css field can contain anything the user enters and the DNN filesystem isn't synced.  However if you feel it's best to make the assumption that the css file is in the DNN filesystem, I'd be happy to make the change.  Just want to make sure.

DNN Platform/Website/Default.aspx.cs
```diff
@@ -737,5 +758,21 @@ private IFileInfo GetBackgroundFileInfoCallBack(CacheItemArgs itemArgs)
         {
             return FileManager.Instance.GetFile(this.PortalSettings.PortalId, this.PortalSettings.BackgroundFile);
         }
+
+        private IFileInfo GetPageStylesheetFileInfo()
+        {
+            string cacheKey = string.Format(Common.Utilities.DataCache.PortalCacheKey, this.PortalSettings.PortalId, "PageStylesheet" + this.PortalSettings.ActiveTab.TabID);
```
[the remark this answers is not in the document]
msant7: Thanks Brian, you are correct, that would make more sense especially if the same stylesheet is used on multiple pages.

DNN Platform/Website/Default.aspx.cs
msant7: Testing out ClientResourceManager.RegisterStyleSheet(), it doesn't seem to take absolute paths.

Message:'https:/folderprovider.blob.core.windows.net/dnnfork/Cloud%202/test.css' is not a valid virtual path.

StackTrace:

InnerMessage:'https:/folderprovider.blob.core.windows.net/dnnfork/Cloud%202/test.css' is not a valid virtual path.

InnerStackTrace:

   at System.Web.Util.UrlPath.CheckValidVirtualPath(String path)
   at System.Web.Util.UrlPath.Combine(String appPath, String basepath, String relative)
   at System.Web.VirtualPath.Combine(VirtualPath relativePath)
   at System.Web.HttpRequest.MapPath(VirtualPath virtualPath, VirtualPath baseVirtualDir, Boolean allowCrossAppMapping)
   at System.Web.HttpServerUtility.MapPath(String path)
   at DotNetNuke.Web.Client.ClientResourceManagement.ClientResourceManager.RegisterStyleSheet(Page page, String filePath, Int32 priority, String provider, String name, String version) in C:\dnnsites\dnnfork\DNN Platform\DotNetNuke.Web.Client\ClientResourceManager.cs:line 303
   at DotNetNuke.Web.Client.ClientResourceManagement.ClientResourceManager.RegisterStyleSheet(Page page, String filePath, Int32 priority, String provider) in C:\dnnsites\dnnfork\DNN Platform\DotNetNuke.Web.Client\ClientResourceManager.cs:line 282
   at DotNetNuke.Web.Client.ClientResourceManagement.ClientResourceManager.RegisterStyleSheet(Page page, String filePath) in C:\dnnsites\dnnfork\DNN Platform\DotNetNuke.Web.Client\ClientResourceManager.cs:line 248
   at DotNetNuke.Framework.DefaultPage.InitializePage() in C:\dnnsites\dnnfork\DNN Platform\Website\Default.aspx.cs:line 644
   at DotNetNuke.Framework.DefaultPage.OnInit(EventArgs e) in C:\dnnsites\dnnfork\DNN Platform\Website\Default.aspx.cs:line 187
   at System.Web.UI.Control.<InitRecursiveAsync>d__225.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Web.Util.WithinCancellableCallbackTaskAwaitable.WithinCancellableCallbackTaskAwaiter.GetResult()
   at System.Web.UI.Page.<ProcessRequestMainAsync>d__523.MoveNext()

bdukes: You're right about the file potentially being external, I've added an `else` block to my suggestion.

The error you've hit is line 303 here, which is inside the `if (filePath.Contains(".css?"))` block.  I think we need to add some better error handling in there, because the addition of a query string to the URL shouldn't mean that it has to be a valid virtual path.
https://github.com/dnnsoftware/Dnn.Platform/blob/edb0facc4253c93a4ef3bdd974f599506de56258/DNN%20Platform/DotNetNuke.Web.Client/ClientResourceManager.cs#L296-L323

We could adjust that code like this:
```
if (!Uri.TryCreate(filePath, UriKind.Absolute, out _) && filePath.Contains(".css?"))
```

msant7: Thanks Brian, I have refactored/applied the changes.  Let me know... thanks

DNN Platform/Website/Default.aspx.cs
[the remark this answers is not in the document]
msant7: cache key updated

bdukes (approved):
Looks great, thanks!

mitchelsellers (approved):

PR #4510, opened 2021-02-20
Auto-generates packages checksums

valadas:
This PR auto-generates a checksums.md file together with the artifacts that we can use to copy/paste into the release notes upon release.


bdukes (approved):

david-poindexter (approved):
Looks good - thanks @valadas 🎉

PR #4527, opened 2021-02-26
Make sure the content is decoded before being passed to tokenization providers.

### LINKED ISSUE #4526: Content isn't decoded prior to calling the tokenization provider.
## Description of problem
In the onLoad method of the HTMLModule.ascx, the content is brought straight from the database and then passed to the token provider for tokenization and then decoded after the process is complete. This is an issue since providers can use parameterization for their tokens and it puts the burden of checking for encoding and then decoding on the token provider, which can not know how many times the content has been encoded. Given the example of [Namespace:TokenName(Param1="Value")], when saving this in an HTML module, the param value delimiters will be encoded once. If a user is further modifying the content without touching the token, upon saving, the parameter value delimiter will be encoded a second time.
For example, in MyTokens, tokens can have parameters and those parameters can have their value wrapped in quotes ( [my:token(param1="abc")] ) and because the value is not yet decoded when the content tokenization is applied this makes it cumbersome to parse tokens. In the past (DNN pre-9.7.2) MyTokens was applying some patches to the HTML module to make this work correctly, but as part of the migration to TokenProviders we wanted to get rid of the patches altogether.

## Description of solution
The solution for this issue would be to make sure that the content is decoded before being delivered to the tokenization providers, not after being tokenized. I will provide a pull request later.

rhaiamz:
Fixes #4526

### DIFF
--- a/DNN Platform/Modules/HTML/Components/HtmlTextController.cs
+++ b/DNN Platform/Modules/HTML/Components/HtmlTextController.cs
@@ -70,6 +70,9 @@ namespace DotNetNuke.Modules.Html
         /// <returns></returns>
         public static string FormatHtmlText(int moduleId, string content, HtmlModuleSettings settings, PortalSettings portalSettings, Page page)
         {
+            // Html decode content
+            content = HttpUtility.HtmlDecode(content);
+
             // token replace
             if (settings.ReplaceTokens)
             {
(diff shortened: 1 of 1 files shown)


bdukes (approved):
This seems like a good fix to me

mitchelsellers (approved):

PR #4531, opened 2021-02-26
Enable URL integration tests to run

bdukes:
## Summary
The `DotNetNuke.Tests.Urls` project was in a state where the tests did not run. These integration tests load the site in memory to run various tests. One of the prerequisites to run these tests is to have a DNN database that is setup, and that database's connection string needs to be configured in `DNN Platform/Tests/App.config` (it does _not_ appear that the `siteUrl` setting is used for this set of tests, so only the database is required).

The configuration of the components in memory was incomplete, which resulted in the tests failing to run. As well, there were some missing dependencies that also caused the tests to fail. I've adjusted the configuration code to reuse the dependency injection container, which appears to have resolved the majority of the issues.

Please note, not all of the tests _pass_ at the moment, but they do run without throwing exceptions. `DotNetNuke.Tests.Urls` contains three test classes, `ExtensionUrlProviderControllerTests`, `FriendlyUrlTests`, and `UrlRewriteTests`. The first two have been passing without issue for me, but `UrlRewriteTests` has about 15% of the tests failing. All of the failing tests that I've seen fail have to do with requests that give a 200 response when a 301 is expected. I don't know if this is a legitimate drift in functionality since these tests have been run, or if there's an issue with the integration where the fake request isn't coming across is the expected way.

This PR does _not_ enable these tests in CI (we would need to setup a database during the CI process to do this), so the failing tests shouldn't cause an immediate issue. Ultimately, we should find a way to be running these tests on a regular basis, and especially during the RC period (though there may be some of these tests we don't want to keep).


valadas:
Oh, this is awesome, I can't review right now but I am very interested in this and looking at the failing tests and maybe add some more, this would help tremendously to pinpoint some oddities we are having with urls and localization :)

@bdukes is there any special local setup to do or the tests just run in Visual Studio Tests Explorer ?

bdukes:
@valadas so long as the connection string in `App.config` is valid, you can run the tests from Visual Studio like any other test.

valadas (approved):
Well, it looks good to me in the browser, but I tried different options in the App.Config to make it actually run because this interests me to help with many url issues fixes.

If ever you feel like showing me @bdukes please slack me up, I must be doing something wrong...

bdukes:
These same changes also enable the tests in the `DotNetNuke.Tests.Integration`, you just also need to set the `siteUrl` key in the tests `App.Config` file.

mitchelsellers (approved):

PR #4543, opened 2021-03-05
Enable Sourcelink

bdukes:
## Summary
This enables [SourceLink](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) for our NuGet packages and makes a couple of other adjustments to the build (e.g. using separate `snupkg` symbols NuGet packages instead of embedding symbols in the main package)


donker (approved):
Awesome new feature

mitchelsellers (approved):
Awesome, thanks for adding this in!

PR #4547, opened 2021-03-06
fixed redirect to primary alias when PortalAliasMapping is set to redirect

### LINKED ISSUE #4306: Portal Alias not enforcing Redirect PortalAliasMapping setting on multi-portal/multi-alias sites
## Description of bug
On a multi-portal site with multiple aliases, DNN does not redirect to the primary URL when the PortalAliasMapping is set to Redirect. I have the same domains with and without the www prefix, and I am trying to redirect all traffic to the www one but it does not redirect. I have verified this on two separate DNN v9.8 sites.

I have verified that the aliases are setup properly in the PortalAlias table, the primary domain's IsPrimary column is set to 1, and the primary alias also matches the DefaultPortalAlias in the PortalSettings table.

## Steps to reproduce
Create a multi-portal DNN installation and set multiple aliases for each portal. Set one domain with the www prefix and another without the prefix.

## Current behavior
After setting the aliases, load the www and non-www aliases and notice there is no redirect.

## Expected behavior
When the PortalAliasMapping is set to Redirect, it should always redirect to the primary alias.

## Additional context
This may or may not be related to issue [https://github.com/dnnsoftware/Dnn.Platform/pull/4158](https://github.com/dnnsoftware/Dnn.Platform/pull/4158)

This also may be problematic from an SEO perspective since there is no canonical tag and sites could get indexed using multiple aliases.

## Affected version
* [X] 09.08.00 latest supported release

## Affected browser
* [X] Any

skamphuis:
Fixes #4306 

If PortalAliasMapping is set to redirect, and the request is not for the primary alias, user will be redirected

## Summary
In advancedUrlRewriter.IsPortalAliasIncorrect there was a check to allow redirection to an alias when forceAlias was used in the querystring.
I added to that check that if the request is not for a primary alias, and PortalAliasMapping is set to redirect, the user will also be redirected to the primary alias

### DIFF
--- a/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
+++ b/DNN Platform/Library/Entities/Urls/AdvancedUrlRewriter.cs
@@ -255,7 +255,12 @@ namespace DotNetNuke.Entities.Urls
             {
                 var portalAliases = PortalAliasController.Instance.GetPortalAliasesByPortalId(result.PortalId).ToList();
 
-                if (queryStringCol != null && queryStringCol["forceAlias"] != null && queryStringCol["forceAlias"] != "true")
+                // if we're not on the primary alias, and portalaliasmapping is set to redirect, we might need to be redirected
+                var redirectToPrimary = !result.PortalAlias.IsPrimary && result.PortalAliasMapping == PortalSettings.PortalAliasMapping.Redirect;
+
+                // forceAlias used in querystring?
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):
This appears to be a good change and makes sense with the expected behavior.  We want to be sure that we add this to the final testing for the next release @valadas

valadas (approved):
Looks good to me

PR #4556, opened 2021-03-09
Bring back update notification

donker:
The current framework update service is broken. The proposed solution retrieves the last 5 releases from Github and checks the version of the first non-prerelease/RC version against the installed version. The result is cached and is shown to host users as a clear blue button prompting to upgrade. The button opens a new window with the release on Github visible.

Screenshot 2021-03-09 at 23 07 41


Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/Services/ServerSummaryController.cs
```diff
@@ -83,72 +85,48 @@ private bool GetVisibleSetting(string settingName)
...
+            }
+
+            return CBO.GetCachedObject<string>(new CacheItemArgs("DnnUpdateUrl"), this.RetrieveUpdateUrl);
+        }
+
+        private string RetrieveUpdateUrl(CacheItemArgs args)
+        {
+            try
+            {
+                var latestReleases = Globals.GetJsonObject<List<DTO.GithubLatestReleaseDTO>>("https://api.github.com/repos/dnnsoftware/dnn.platform/releases?per_page=5");
```
valadas, line 104: Sould the url come from a config file maybe ?

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/main.js
```diff
@@ -770,7 +770,7 @@ require(['jquery', 'knockout', 'moment', '../util', '../sf', '../config', './../
                         };
 
                         viewModel.updateText = ko.computed(function() {
-                            return viewModel.updateType() === 2 ? util.resx.PersonaBar.CriticalUpdate : util.resx.PersonaBar.NormalUpdate;
```
valadas, line 773: @mitchelsellers we used to have "Update" and "Critical Update", is that scenario still a supported use case ?

valadas (commented):
I left 2 small comments, everything looks fine to me, the comments are mainly for discussion, other than discussion that I am fine with this change...

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/Services/ServerSummaryController.cs
donker: Although I am in favor of it, the truth is that people will never configure this. I.e. if we move to a different location, a new release and upgrade would hold the new values, but the older ones never will. So IMHO it is trying to solve something which we both see as a potential issue, but doesn't really solve it.

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/main.js
donker: The code still exists, but in our current procedure I can't see (1) how we'd do it in Github and (2) if it is really useful/accurate/effective. Most of our upgrades are critical. But still few people upgrade ...

bdukes: Wouldn't normal vs. critical be based on security bulletins? We already have code in the Security extension of the Persona Bar to read that data. https://github.com/dnnsoftware/Dnn.Platform/blob/3e0d9c7c12e08f62ab458304aecdbc42b540d893/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SecurityController.cs#L608-L685

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/Services/ServerSummaryController.cs
bdukes: Are there concerns about rate limiting if we're accessing the API without a token?

mitchelsellers (changes requested):

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/main.js
valadas: yeah, that's what I was talking about just did not know where it came from in code...

thabaum:
For the API rate limits, would a solution for this issue be a service created on a hosted site that caches hourly this information that is needed to connect to the link information provided by GitHub for the update?

mitchelsellers:
@thabaum We have a solution ;)

mitchelsellers (approved):

valadas:
@donker this looks good to me but I think we need to sync up with the changes in the service itself right? Let's discuss this in next meeting if you are available...

donker:
> @donker this looks good to me but I think we need to sync up with the changes in the service itself right? Let's discuss this in next meeting if you are available...

Yes, but I'm assuming that is the easy part given we control everything

mitchelsellers:
The UPdate Service changes have been deployed, this is good to go

PR #4554, opened 2021-03-08
Apply StyleCop to Build project

bdukes:
## Summary
This PR adds StyleCop to the Cake Build project and turns on Warnings as Errors.

PR #4565, opened 2021-03-19
Update Build.csproj

donker:
Dnn.CakeUtils has just been bumped to 2.0.0 to fix a number of issues with DNN's build process


bdukes:
I've included this change in #4554, thanks!

PR #4575, opened 2021-03-30
Add package for MailKit

### LINKED ISSUE #4563: Create DotNetNuke Library Package for MailKit
## Description of problem
A new library package should be created to contain the MailKit, MimeKit, and associated references so they can be installed during installation, rather than included during the build process. 

Additionally, the Cake build script should also exclude these files from the Upgrade package

## Description of solution
Creating a package will prevent upgrade issues for those that have existing versions of these dependencies.

## Additional context
This solution was discussed during the TAG meeting as a solution for the current upgrade issues impacting those with certain users of third-party modules

bdukes:
Fixes #4563 

I haven't had a chance to test that this does the right thing, waiting for the generated packages to check.

Also, I'd like to adjust how we're generating these to read the versions and update the manifest magically, but haven't gotten to that yet.

### DIFF
--- a/Build/Tasks/OtherPackages.cs
+++ b/Build/Tasks/OtherPackages.cs
@@ -17,6 +17,7 @@ namespace DotNetNuke.Build.Tasks
 
     /// <summary>A cake task to include other 3rd party packages.</summary>
     [Dependency(typeof(PackageNewtonsoft))]
+    [Dependency(typeof(PackageMailKit))]
     public sealed class OtherPackages : FrostingTask<Context>
     {
         /// <inheritdoc/>
--- a/Build/Tasks/PackageMailKit.cs
+++ b/Build/Tasks/PackageMailKit.cs
@@ -1,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Build.Tasks
+{
+    using System;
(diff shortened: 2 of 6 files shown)


mitchelsellers:
@bdukes I think the only other thing we need to do here is also exclude these files from the created install/upgrade packages correct?

bdukes:
@mitchelsellers they're excluded from the packages in `Build/Tasks/packaging.json`

bdukes:
Alright, I think this is ready for review. The manifest version and assembly versions are automatically set during packaging. I've done an install and an upgrade from 9.8.1 to 9.9.1 and verified that the `MailKitMailProvider` works on both sites.

One thing to note is that I did _not_ get a `MailKitMailProvider` entry added to the `web.config` upon upgrade, I had to add it manually. I don't think that's related to these changes, but perhaps another issue we need to investigate.

valadas:
[x] v9.9.1 Clean install: Emails work and mailkit is available but not set a default
[x] v9.9.0 pre-configures for MailKit, stays configured with MailKit and still works
[ ] v9.1.1 upgrade emails keep working but MailKit is not available in the web.config providers. What is the plan for this ? merge this PR and do another one for that ?

Other than that, it looks good to me.

valadas (approved):
Looks good to me if we handle the upgrade adding of the provider in the web.config separatelly...

valadas:
I think #4582 will fix the web.config part of this, waiting for the build to confirm that...

mitchelsellers (approved):

PR #4576, opened 2021-03-31
Use binary logger instead of console logger for speedier builds

bdukes:
See https://msbuildlog.com/ for the log viewer and other details on binary logs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,5 @@ yarn-error.log
 /Build/bin/
 /.dotnet/
 /Build/Tools/
+
+*.binlog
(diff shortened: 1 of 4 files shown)


bdukes:
This PR turns off the console logger during CI and produces a binary log file (which gets included in the Artifacts). It then reads the log file and reports warnings all at once. This appears to save at least a couple of minutes of build time. The binary log also has more complete details if there's an issue to investigate (the downside being the friction of needing a separate tool to interrogate it).

mitchelsellers (approved):
I like this, clean and faster builds

valadas (approved):
Nice!

PR #4577, opened 2021-04-01
Add initial dependabot config

bdukes:
With [the announcement of General Availability for Dependabot version updates](https://github.blog/changelog/2021-03-31-dependabot-version-updates-are-now-generally-available/), I thought it would be good to setup a basic config. I added npm and NuGet at monthly for now. We can also disable them altogether, or set them up for a different branch.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me!

valadas:
This will be fun though 😄  For reviewers we need to be very careful on the React stuff, there will be a lot! And we need to run some decent testing... But I'll help where I can for sure.

mitchelsellers (approved):

PR #4604, opened 2021-04-11
Fixed an issue when joining by RSVP would set the end date to today

### LINKED ISSUE #2543: Joining by RSVP sets the end date to today
## Description of bug
When users join a role using an rsvp code, their role assignment end date is the same day.

## Steps to reproduce
1. Create a new role group
2. Create a new role in that group
3. Set an rsvp code and copy the link
4. Logout
5. Register a new account
6. Navigate to the rsvp link
7. Logout
8. Login with host account
9. Go to the Roles module
10. Choose the role group and expand the list of roles

## Current result
The user is listed but not counted (since it is already expired expired) and you can see the end date is now

* [x] 9.3.0
Did not have time to try older releases, so not sure.

chapmasc:
## Summary
It looked like the effective date and expiration date were switched; so I switched them back so the effective date was set to the current date; instead of the expiration date being set to today, thus having the user not added to the role when the rsvp code is entered.  This alone did not fix it though.  I'm unsure what the other if statement was for; but I just set to to set the value to Null.NullData as the ExpiryDate will always be < DateTime.Now since it is not defaulted to Null.  Hopefully someone else more familiar with this knows what this was supposed to do.  Was it meant for EffectiveDate too?

Fixes #2543

### DIFF
--- a/DNN Platform/Library/Security/Roles/RoleController.cs
+++ b/DNN Platform/Library/Security/Roles/RoleController.cs
@@ -507,8 +507,8 @@ namespace DotNetNuke.Security.Roles
             else
             {
                 int UserRoleId = -1;
-                DateTime ExpiryDate = DateTime.Now;
-                DateTime EffectiveDate = Null.NullDate;
+                DateTime EffectiveDate = DateTime.Now;
+                DateTime ExpiryDate = Null.NullDate;
                 bool IsTrialUsed = false;
                 int Period = 0;
                 string Frequency = string.Empty;
@@ -542,7 +542,7 @@ namespace DotNetNuke.Security.Roles
 
                 if (ExpiryDate < DateTime.Now)
                 {
-                    ExpiryDate = DateTime.Now;
+                    ExpiryDate = Null.NullDate;
                 }
 
                 if (Period == Null.NullInteger)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=4604) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=4604) before we can accept your contribution.<br/><br/>:x: chapmasc [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=4604)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=4604) it.</sub>

chapmasc:
I'm not sure how to add the labels it seems to want.  I read the contributors markdown file and did not see any info in there about labels.  I could have just missed it to though.  Also, I made the fix for the version I'm currently using, but I would imagine the fix would merge to the other branches without any issues since the defect still exist in the latest version.

valadas (changes requested):
Thanks for your contribution.

Ok, so just for context Null.NullDate in DNN is hardcoded to return DateTime.MinValue, then the data provider checks for that value and sends NULL to the actual database. This is historical because it was written before we had nullable types in the language.

Soooo, the first change looks good to me but let's look into some situations since this code can get called from many other places than the RSVP feature:

1. If the user gets a new role, a role the user did not have. Then this looks fine to me.
2. If the user gets a role he already has (which is handled in between your 2 changes, then this won't work the same. If the user had it's original role already expired, then the second change will be true and that role would loose it's expired date and get set to Null.NullDate.

I am thinking this may not be the place to correct this issue if it only happens with the RSVP link because it works fine in other ways to get a role and there is nothing here that is specific to RSVP usage...

As for the labels and milestone, those get set by maintainers, they are read-only by contributors.

valadas:
I think the root of the issue here is that this code here uses `UpdateUserRole` instead of calling `AddUserRole`:

https://github.com/dnnsoftware/Dnn.Platform/blob/9f83285a15d23203cbaad72d62add864ab5b8c7f/DNN%20Platform/HttpModules/Membership/MembershipModule.cs#L137-L147

chapmasc:
Thank you for the through review as well as the additional information.  I see two ways to handle this; the first is to change:

                if (ExpiryDate < DateTime.Now)
                {
                    ExpiryDate = DateTime.Now;
                }

to be inside the

                if (userRole != null)
                {
                    UserRoleId = userRole.UserRoleID;
                    EffectiveDate = userRole.EffectiveDate;
                    ExpiryDate = userRole.ExpiryDate;
                    IsTrialUsed = userRole.IsTrialUsed;
                }

This way it handles your 2nd note (If the user gets a role he already has (which is handled in between your 2 changes, then this won't work the same. If the user had it's original role already expired, then the second change will be true and that role would loose it's expired date and get set to Null.NullDate.) since that is the only thing that would change the ExpiryDate variable.

So I figure the end result would look like 

                if (userRole != null)
                {
                    UserRoleId = userRole.UserRoleID;
                    EffectiveDate = userRole.EffectiveDate;
                    if(userRole.ExpiryDate < DateTime.Now)
                    {
                        ExpiryDate = DateTime.Now
                    }
                    else
                    {
                        ExpiryDate = userRole.ExpiryDate;
                    }
                    IsTrialUsed = userRole.IsTrialUsed;
                }

That being said, if you are calling to update a role; I'm not sure why you would want to change the expiration date from the originally set expiration date and set it to today; the end result still has it expired and now it just looks like it expired today instead of being an old expiration date.

In regards to your 2nd post in which you believe the issue is the RSVP code logic should be calling AddUserRole instead of UpdateUserRole.  The only scenario I can think of is, what if a user enters a RSVP code for a role they already are subscribed to?  In looking at the code

                foreach (RoleInfo objRole in RoleController.Instance.GetRoles(this.PortalSettings.PortalId))
                {
                    if (objRole.RSVPCode == code)
                    {
                        RoleController.Instance.UpdateUserRole(this.PortalId, this.UserInfo.UserID, objRole.RoleID, RoleStatus.Approved, false, false);
                        rsvpCodeExists = true;

                        // Raise SubscriptionUpdated Event
                        this.OnSubscriptionUpdated(new SubscriptionUpdatedEventArgs(false, objRole.RoleName));
                    }
                }

In the AddUserRole function, I'm not sure if the third line below is the check to make sure the role doesn't already exist or not.  I believe it is though, and if so, then it should not cause an error if trying to add a role that a user already belongs to.  I still question what scenario there is in which you would want the ExpiryDate set to DateTime.Now() instead of just leaving it with the current expired ExpiryDate as the end result is still the same - it's expired.

            UserInfo user = UserController.GetUserById(portalId, userId);
            UserRoleInfo userRole = this.GetUserRole(portalId, userId, roleId);
            if (userRole == null)
            {
                // Create new UserRole
                userRole = new UserRoleInfo
                {
                    UserID = userId,
                    RoleID = roleId,
                    PortalID = portalId,
                    Status = status,
                    IsOwner = isOwner,
                    EffectiveDate = effectiveDate,
                    ExpiryDate = expiryDate,
                };
                provider.AddUserToRole(portalId, user, userRole);
                EventLogController.Instance.AddLog(userRole, PortalController.Instance.GetCurrentPortalSettings(), UserController.Instance.GetCurrentUserInfo().UserID, string.Empty, EventLogController.EventLogType.USER_ROLE_CREATED);
            }
            else
            {
                userRole.Status = status;
                userRole.IsOwner = isOwner;
                userRole.EffectiveDate = effectiveDate;
                userRole.ExpiryDate = expiryDate;
                provider.UpdateUserRole(userRole);
                EventLogController.Instance.AddLog(userRole, PortalController.Instance.GetCurrentPortalSettings(), UserController.Instance.GetCurrentUserInfo().UserID, string.Empty, EventLogController.EventLogType.USER_ROLE_UPDATED);
            }

I'm good re-doing the change with either of those two options.  please let me know what your preference is and I'll cancel this pull request, make the changes and resubmit a new pull request.

valadas:
>  I still question what scenario there is in which you would want the ExpiryDate set to DateTime.Now() instead of just leaving it with the current expired ExpiryDate as the end result is still the same - it's expired.

I honestly don't see a scenario off the cuffs where that would be an issue, but I think calling AddUserRole instead of modifying the existing UpdateUserRole is less risky to cause issue with something we may not be thinking about and since this is a public API we probably have 3rd-party extensions using it. So I am fine either way with a slight preference for the least invasive approach :)

@bdukes @mitchelsellers @david-poindexter @donker any opinions on this ?

bdukes:
I think there's definitely some surprising behavior in `UpdateUserRole`, but making changes there could have a widespread effect on folks who've been working around those issues for years. If we're going to make major changes in the behavior, I would prefer that we do it on a new interface and deprecate this implementation.

I think @valadas' suggestion of switching to `AddUserRole` is the simplest approach here.

chapmasc:
Thank you both for the time in reviewing this.  I'll work on making that change today after work and will also test it to confirm if it resolves the issue.  If testing goes well, should I cancel this pull request and submit a new one, or is there some way to change this pull request so the comments are still there and it doesn't possibly go to someone else on the team to review?

valadas:
> or is there some way to change this pull request so the comments are still there and it doesn't possibly go to someone else on the team to review

Whatever is easier for you, you can push other commits to your branch to revert this and do the other way and it will update on 
this PR. Or you can submit a new PR and we should be able to keep up, it's a small review team :) Or you can link to this PR by doing a "pound" sing and this pr number too, even if it is closed. Like #4604

chapmasc:
Shoot.  Thank you for the reminder, I forgot.  Yes, I will add a reminder
in my phone this time.  I will hopefully have time to make the change and
test in the next two weeks.

Thanks,
Scott

sent via phone

mitchelsellers:
Due to inactivity on the requested changes we are closing this PR.  We welcome future submissions with the changes addressed.

puresystems:
Hi, did this issue ever get addressed? We have come across this issue in v9.10.1.

valadas:
@puresystems unfortunatelly not, we suggested this should be probably be handled differently and the PR never got adjusted for it, so it is still an issue as far as I know...

PR #4608, opened 2021-04-14
Produce snupkg format for NuGet

bdukes:
## Summary
We're currently producing ["legacy" symbols packages](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages) for NuGet, which is causing issues with automated pushing of NuGet packages (the ordering matters of whether you push the non-symbols or symbols package first and we're getting a conflict error).

This PR tells NuGet to produce the [newer symbols package format, `*.snupkg`](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg), which should resolve that issue.

I've had unrelated issues getting my build to finish locally, so this change still needs confirmation from the PR build.


valadas (approved):
Looks good to me

mitchelsellers (approved):

PR #4619, opened 2021-04-21
Fix MailKitMailProvider using incorrect priority

### LINKED ISSUE #4618: MailKitMailProvider sends messages with low/non-urgent priority by default
>Just changed to using MailKit in DNN 9.9.1. Have now noticed all mail messages sent from the site are sent with low priority. Is there a way to set it to normal? If there isn't I can open an enchantment to address. Has anyone else noticed this?image

_Originally posted by @james7342 in https://github.com/dnnsoftware/Dnn.Platform/discussions/4617_

>You're right, there's a mismatch here. Here's where MailKitMailProvider sets the priority:
https://github.com/dnnsoftware/Dnn.Platform/blob/2e68b5d00bb8d551ef87a557356263edfb6b510a/DNN%20Platform/Library/Services/Mail/MailKitMailProvider.cs#L198

>We're converting from `System.Net.Mail.MailPriority` to `MimeKit.MessagePriority` which order their values differently (`Normal`, `Low`, `High` vs. `NonUrgent`, `Normal`, `Urgent`).

Update: We're actually converting from `DotNetNuke.Services.Mail.MailPriority` but the issue is the same.

bdukes:
Fixes #4618

### DIFF
--- a/DNN Platform/Library/Services/Mail/MailKitMailProvider.cs
+++ b/DNN Platform/Library/Services/Mail/MailKitMailProvider.cs
@@ -195,7 +195,7 @@ namespace DotNetNuke.Services.Mail
                 mailMessage.ReplyTo.AddRange(InternetAddressList.Parse(mailInfo.ReplyTo));
             }
 
-            mailMessage.Priority = (MessagePriority)mailInfo.Priority;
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

valadas (approved):
Thank you sir, looks good to me.

PR #4641, opened 2021-04-30
add email to password sent message for error tracing, fixes #4605

### LINKED ISSUE #4605: Improve logged details for password sent failure event
## Description of problem
Our site uses the standard DNN Authentication module. After a recent upgrade to Evoq 09.06.04, we began experiencing some issues related to user logins.

The most puzzling of those is the PASSWORD_SENT_FAILURE event, which appears to occur in the DNN Platform code.

- The message suggests possible success but the user does not receive the email.
- No corresponding entries are found in the DNN exception log.
- The SMTP server is confirmed to be working and has no additional information corresponding to the logged timestamps.
- If the user's email is known (sometimes they contact support), we're able to look at their account details, but a comparison of those vs. another account (which has no problems logging in) reveals nothing. It's not a simple issue of the email being entered incorrectly.

## Description of solution
Include, at minimum, the email address that was used in the request, but anything that would indicate the root cause would helpful.

## Screenshots
DNN password sent failure

## Affected browser
All, probably.

jcmrva:
#4605

## Summary
Includes the email address or username in the password reset message for better visibility into PASSWORD_SENT_FAILURE events.

### DIFF
--- a/DNN Platform/Website/admin/Security/App_LocalResources/SendPassword.ascx.resx
+++ b/DNN Platform/Website/admin/Security/App_LocalResources/SendPassword.ascx.resx
@@ -136,7 +136,7 @@
     <value>This account has been locked out after too many unsuccessful login attempts.  Please wait 10 minutes before trying to login again.  If you have forgotten your password, please try the Password Reminder option before contacting an Administrator.</value>
   </data>
   <data name="PasswordSent.Text" xml:space="preserve">
-    <value>If the details entered were correct, you should receive an email message shortly with a link to reset your password.</value>
+    <value>If the details entered for {0} were correct, you should receive an email message shortly with a link to reset your password.</value>
   </data>
   <data name="EnterUsername.Text" xml:space="preserve">
     <value>Enter your user name.</value>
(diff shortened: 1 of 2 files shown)


bdukes (changes requested):
For security reasons we can't have different message formats based on whether the username matches or doesn't. But I think it would be okay to include the username/email in all of the messages, whether it matched or not.

I would prefer to add a placeholder to the default message and use string.Format rather than just tacking it on to the end, so that administrators can have control over the formatting of that message (and have the option to include or exclude the extra information)

mitchelsellers:
@jcmrva Do you plan on making adjustments to this PR based on the feedback above?

jcmrva:
@mitchelsellers It's been my intention to. Is there a target date for the 9.9.2 milestone (which would help me prioritize it)?

mitchelsellers:
2-3 weeks

mitchelsellers:
@jcmrva Any update on possible changes?

jcmrva:
It seems like the change the last commit will take care of it. Should any of the other messages be updated as well?

bdukes (approved):
LGTM, thanks!

bdukes:
FYI, we don't get notified on pushes, so we weren't aware you'd already addressed this. We appreciate your quick responsiveness 👍🏻

jcmrva:
My bad, I should've added a comment.

valadas (approved):
Looks good to me, thanks!

Issue #4647, opened 2021-05-03
Add async support to MVC controllers


bdukes:
## Description of problem
An MVC controller which implements [`IAsyncController`](https://docs.microsoft.com/en-us/dotnet/api/system.web.mvc.async.iasynccontroller) or an action method which returns `Task<ActionResult>` aren't respected and no asynchronous processing happens. This is a standard supported feature of MVC, and now that DNN supports async processing in the main view process, this should be able to be enabled.

## Description of solution
Update the processing of MVC controllers and results to support asynchronicity.

## Description of alternatives considered
n/a

## Additional context
Currently, a `Task` returning action will be interpreted as a `ContentResult` and the text `System.Threading.Tasks.Task[System.Web.Mvc.ActionResult]` will be displayed.

dimarobert:
@bdukes Here is a rough POC for the Async MVC Controller execution. I would like to get the discussion going again and see if it is something that is still wanted for the platform.
https://github.com/dnnsoftware/Dnn.Platform/compare/develop...dimarobert:async-mvc-actions?expand=1

How the async implementation in WebForms works:
1. The normal lifecycle of the page/controls is executed from OnPreInit to OnPreRender (see [docs](https://learn.microsoft.com/en-us/previous-versions/aspnet/ms178472(v=vs.100))) in a synchronous manner.
    - The `Page.RegisterAsyncTask(new PageAsyncTask(this.MethodAsync))` can be called in any of the events to queue up async tasks.
    - All Controls collections get frozen (read-only) after the OnPreRender.
2. The queued-up tasks are now asynchronously executed in the order they were added throughout the events.
    - Did not test, but we most likely cannot modify any Controls collections in any of the async tasks. So, if this kind of manipulation is needed after the execution of the module it might not be possible anymore and needs to be rethought.
3. The remaining page events are executed synchronously starting with OnPreRenderComplete.
    - No other async tasks can be registered as they will not get executed anymore.

Changes will be limited to the MvcHostControl so no other types of modules should be affected, but because of the lifecycle changes, there might be side effects for some Container, Pane, and Skin implementations if they want to access stuff in the IModuleControl or the ModuleRequestResult in their OnInit or in any other event after that (OnLoad, OnPreRender). That is because the execution of the module needs to be moved to an async task registered in the OnInit of MvcHostControl which means that the actual execution will happen in step 2 above, after all those sync events.
[...]

PR #4648, opened 2021-05-03
Allow async processing of controller actions

bdukes:
For #4647

**This does not currently work.** When processing the action the web server never responds.

## Summary
I wanted to get some eyeballs on this and see if anyone had any input into how the code handles the possibility of async action. There are some APIs which are currently synchronous but intended to be overridable, so I can't switch them to async completely. One alternative I've considered is to create a new async type, e.g. `AsyncModuleApplication`, and switch on the runtime type higher up.

Also, if anyone sees anything that might fix the reason the response is never completing, that'd be very helpful also :-)

And, apparently, you'll need to view [the diff without whitespace changes](https://github.com/dnnsoftware/Dnn.Platform/pull/4648/files?diff=split&w=1) 🤷🏻‍♂️


bdukes:
Wasn't able to address this, closing for now.

PR #4670, opened 2021-05-20
Bugfix/issue 4669

### LINKED ISSUE #4669: Page gets redirected, if the page URL is the same as portal alias with a different domain
## Description of bug
If the page name is the same as the portal name in a different domain, the page can't be accessed since it redirects to the other portal when tried to be accessed.

## Steps to reproduce
1. Set a DNN site with the following domain: "**domain**.dnndev.me"
2. Create a page without a parent page and name it "test1" (URL to access the page should be "**domain**.dnndev.me/test1")
3. Create a new portal using the same page name as the site name ("test1"). The URL should be "dnndev.me/test1".
4. Try to access the test1 page again.

## Current behavior
The request gets redirected to the new portal's home page.

## Expected behavior
The "test1" page of the "**domain**.dnndev.me" site should be opened. 

## Screenshots
Reproduction video: https://drive.google.com/file/d/1syAoLyUFDKJumYP5ImqX7bkTJeaXP1y1/view

## Affected version

* [X] 10.00.00 alpha build

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #4669

## Summary
The portal alias check is moved just after to the tab check control. This way, the tabs get searched first and if it can't find one, the portal alias gets searched.

**Note: Writing UT now and that's why the PR is in draft status.**


sleupold:
I disagree that a page URL should override a sub-website with same URL.
Is there any specific use case you are having in mind?

sleupold:
note: changing the order of checking for page path and sub-website seems to me to be a breaking change.

berkarslan-xo:
Asking for additional confirmations that this can't be implemented before closing the issue. I updated the reproduction steps to make it easier to understand the issue in #4669 .

bdukes:
I've provided some feedback on the issue. I think the bug can be fixed, but this approach doesn't seem like the right approach.

https://github.com/dnnsoftware/Dnn.Platform/issues/4669#issuecomment-845995580

mitchelsellers:
Closing this until conversation on the actual issue have been resolved.

PR #4678, opened 2021-05-23
Clarified text on module anchor feature

valadas:
The feature does not create a hyperlink (which would have an href parameter), when an `<a>` tag has no href, it is called an anchor (hence the `<a>`). Also the phrasing was a bit unclear and this new way to phrase it explains the feature a little better.


david-poindexter (approved):
Good catch! Thanks @valadas 🎉

mitchelsellers (approved):

PR #4144, opened 2020-10-02
export after changing a module pane in a page and importing in anothe…

tauqeer-haider:
…r site doesn't reflect pane change in view mode.

Fixes #4143 

## Summary

Pane change is now also going through adding new module and deleting the existing one, instead of just updating module properties. In update path, update to version is explicitly disabled. Fix is tested on Platform version and found no side effects.

Issue #4143, opened 2020-10-02
Export a page after changing a module pane and importing in another site doesn't reflect pane change in view mode


tauqeer-haider:
## Description of bug
Exporting a page after changing a module pane and importing in another site doesn't reflect pane change in view mode.

## Steps to reproduce
0. Evoq 9.6.x installed (producing in evoq consistently)
1. Create a new page on Site A
2. Add an html module to any pane
3. Export that page to Site B
4. Confirm export on Site B (everything looks good)
5. Move the module on Site A to a new pane and publish page.
6. Export the page to Site B

## Current behavior
On Site B, after the import, the module has not moved to the new pane.
When entering edit mode the module is shown in the new pane, however there is no publish button.

## Expected behavior
The imported page should show the module in the correct (modified) pane in view mode as well. Edit/publish should be possible.

## Screenshots
If applicable, provide screenshots to help explain the bug.

## Error information
Provide any error information (console errors, error logs, etc.) related to this bug.

## Additional context
The module's "PaneName" attribute gets updated in the TabModules table, but not in the "TabVersionDetails" table.

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.07.02 release candidate
* [x] 09.07.01 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge (Classic)
* [x] Microsoft Edge Chromium

PR #4685, opened 2021-05-26
Issue-4684 - Exporting a page after changing a module's order within the same pane doesn't reflect in imported site

### LINKED ISSUE #4684: Exporting a page after changing a module's order within the same pane doesn't reflect in imported site
## Description of bug
This bug is very similar to #4143 but the only difference is that #4143 handles the case when a module is moved to another pane (e.g. from HEADERPANE to CONTENTPANE) whereas it doesn't handle the case when there are multiple modules within the same pane and their locations get changed internally (without moving to another pane).

## Steps to reproduce
1. Create a new page on Site A
2. Add 2 html modules to HEADERPANE (let's say they appear in this order: "Test1" & "Test2")
3. Export the page and import it to Site B
4. Confirm the page in Site B looks good
5. Move the top module on Site A just below the other module within the same pane (so that the modules are in this order now: "Test2" & "Test1").
6. Export the page and import it again to Site B

## Current behavior
The order of the modules don't change on Site B.

## Expected behavior
The order of the modules in Site B should be the same as the Site A.

## Affected version

* [x] 10.00.00 alpha build

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #4684 

## Summary
The bug was very similar to #4143 and the code to check the order of modules is added on top of the fix provided for #4143's PR (#4144).

### DIFF
--- a/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
+++ b/DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs
@@ -983,7 +983,7 @@ namespace Dnn.ExportImport.Components.Services
                         {
                             var localExpModule = localExportModules.FirstOrDefault(
                                 m => m.ModuleID == local.ModuleID && m.FriendlyName == local.ModuleDefinition.FriendlyName);
-                            var hasPaneChanged = local.PaneName != other.PaneName;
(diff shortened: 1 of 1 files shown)


bdukes (approved):
LGTM

valadas (approved):
great

PR #4694, opened 2021-05-31
Commit linter auto-fixes across JS and JSX files

david-poindexter:
## Summary
Some dependency bot, somewhere along the way, turned on auto-fix. This is simply to commit those auto-fixes by the linter. This will help prevent huge diffs on every local build. Yay!


valadas (approved):
Looked at each file in detail, it's mostly whitespace changes, some missing semi-colons, harmonizing strings to use double quotes (which I think makes sense in a C#/JS project like this). My main concern was some changes from `var` to `let` but inspected them all and they make sense to me.

This is hard to review in the browser, if anyone else decides to go into details of this one, I recommend using the github visual studio extension to review.

donker (approved):

PR #4695, opened 2021-05-31
Bring back update notification

donker:
The current framework update service is broken. The proposed solution retrieves the last 5 releases from Github and checks the version of the first non-prerelease/RC version against the installed version. The result is cached and is shown to host users as a clear blue button prompting to upgrade. The button opens a new window with the release on Github visible.

Screenshot 2021-03-09 at 23 07 41


david-poindexter:
Referencing #4556

david-poindexter (approved):
Looks good to me - thanks @donker for all the hard work on this - it's a great improvement!

david-poindexter (changes requested):
Okay, I finally got this in a test environment and see something is off.  This is showing there is an update available, but I am running a version greater than the current version.  See below.

image

valadas (approved):
Looks good to me,
a vey welcome back feature, thank you sir!

valadas (dismissed):
Oh, just saw the other comment, well except that issue, it looks good to me :)

mitchelsellers:
Looking at this, 9.9.2 is not a known version since it doesn't exist on GitHub, so it is assumed to be an upgrade available.

I think we have 2 options.

1 Update the update service to return the current version if the current version isn't known.
2 Leave this and let it be as once we release an RC we should be fine

mitchelsellers (approved):

bdukes (approved):

mitchelsellers:
This PR itself is working as desired, the Update Service needs to be updated yet, which will address the pending behavior.

PR #4703, opened 2021-06-04
Fix for #4672: Cache-Setting 'NoCaching' not saved

### LINKED ISSUE #4672: Cache-Setting 'NoCaching' not saved
## Description of bug
It is not possible to turn Caching off (set Caching to 'NoCaching') in the Server-Settings of the PersonaBar (DNN GUI). 

## Steps to reproduce
1. Go to 'Settings > Server-Settings > Performance'
2. Click on 'Cache-Settings' dropdown
3. Select 'NoCache' and 'Save' settings
4. Reload page (or restart app to be sure) or check database: the setting is not stored. Value in database should be '0'. This only happens when selecting 'NoCache", all other options of the dropdown box are stored correctly. 

## Current behavior
When selecting 'NoCache' value is not set in the database. Changing setting to 'Light/Moderate/Heavy Caching' works fine. 

## Expected behavior
Caching should be turned off. 
SettingValue of SettingName 'PerformanceSetting' in Table HostSettings should be set to '0' (zero). 

## Screenshots
Screenshot of the dropdown attached.
image

## Error information

## Additional context

## Affected version
* [X] 09.09.00 
* [X] 09.09.01 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Microsoft Edge Chromium

manojattal:
## Summary
Fixed onchange event handler for 'Cache-Settings' dropdown 

UI was not sending the request in the correct format to UpdatePerformanceSettings() API, when "No Caching" value is selected.

The root cause being in the Performance component's onChangeField(). The value selected in the 'Cache-Settings' dropdown was not validated correctly.

Fixed code in the if clause. When "No Caching" is selected, "event.value" is "0". The updated code makes sure that if function evaluates to true when caching value is "0". So API request sends correct value for "CacheSetting" parameter.

Fixes #4672

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
+++ b/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
@@ -72,7 +72,7 @@ class Performance extends Component {
         let value = event;
         if (event && event.value !== undefined) {
             value = event.value;
-        } else if (event && event.target && event.target.value) {
+        } else if (event && event.target && event.target.value !== undefined) {
             value = event.target.value;
         }
 
--- a/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
+++ b/Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
@@ -70,7 +70,7 @@ class Performance extends Component {
 
     onChangeField(key, event) {
         let value = event;
-        if (event && event.value) {
+        if (event && event.value !== undefined) {
             value = event.value;
         } else if (event && event.target && event.target.value) {
             value = event.target.value;


Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
```diff
@@ -70,7 +70,7 @@ class Performance extends Component {
 
     onChangeField(key, event) {
         let value = event;
-        if (event && event.value) {
+        if (event && event.value !== undefined) {
             value = event.value;
         } else if (event && event.target && event.target.value) {
```
bdukes, line 75: Do we know the scenario when this `if` fires? Should we also be doing `!== undefined` for `event.target.value`?

manojattal: @bdukes `onChangeField()` is bound to multiple inputs on Performance.jsx. I triggered value change for all the inputs on the page. There was no any case when ` if (event && event.target && event.target.value) `got fired.

mitchelsellers (approved):

Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
valadas: My guess is that it is for browser compatibility 🤷 if you could also add that condition in the elseif it would be nice, if not then I guess it is fine for browsers most admins use.

valadas (approved):

Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Performance.jsx
manojattal: @valadas Please let me know in case changes of "else if" are blocking the merge.

bdukes: @manojattal please add the same check to the `else if` branch. Thanks!

manojattal: @bdukes I have added the check. Please let me know if anything required. Thanks!

PR #4708, opened 2021-06-11
Fixed an issue where UserInfo.UserName was wrong with "Use email as username" setting turned on

donker:
When an email goes out that uses the [user:username] property, it doesn't properly fill in the username if the site is set to use email as username. This PR fixes that.

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserInfo.cs
+++ b/DNN Platform/Library/Entities/Users/UserInfo.cs
@@ -421,7 +421,12 @@ namespace DotNetNuke.Entities.Users
                         return PropertyAccess.ContentLocked;
                     }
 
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

valadas (approved):
Looks good to me.

valadas:
Do we need to wait 9.10.0 for this or we consider it's actually a bugfix? I personally would take it into 9.2.2...

mitchelsellers:
Considering the ship has sailed and the next release will be 9.10.0 based on what has already  been merged we are good.  (I'll ping you on Skype too)

valadas:
Ok, cool cool

PR #4710, opened 2021-06-16
Issue 3703 - Removal of WebSlice Option

### LINKED ISSUE #3703: Removal of the WebSlice option

joshgk00:
Fixes #3703

## Summary
The changes proposed here are to remove the WebSlice options from the Module Settings UI as well as the underlying tables, since it's no longer supported. 

## How to Test
- Do a clean install of DNN Platform and verify that the install completes without any errors.
- Upgrade an existing instance of DNN Platform and verify that the upgrade process completes without any errors
- Verify that the "Is a WebSlice" options have been removed from the Module Settings > Page Settings > Basic Settings section

Before this fix:
- Is a WebSlice options existed in Module Settings
image
- Views, Tables, Stored Procedures and Constraints referenced the IsWebSlice column in the TabModules table
   - TabModules Table
      image
   - TabModules Contraints
      image
   - TabModules Indexes
      image
   -  vw_TabModules
      image
   - SP: AddTabModule
     image
   - SP: UpdateTabModule
     image

After this fix:
- No options exist for the WebSlice Options in the Module Settings UI
image

- The Tables, Views, SPs, Indexes, and Constraints have been updated/removed


mitchelsellers:
@valadas SHould this be retargeted?

valadas:
Well, in an earlier meeting and on slack we decide we needed to recreate future/10 from develop because there have been just way to many conflicts with all the build refactoring and stylecop fixes, so the idea was to have this target develop until we can decide what we want to cherry pick from the current future/10 branch and actually do it. From the moment we do it, we should rebase it weekly or some such to keep up with the conflicts.

Back then I did not have the knowledge I have now about rebases but now I think I could handle keeping up weekly or bi-weekly or some such, so maybe we put this in next week agenda ?

valadas:
Replaced by #4711

PR #4714, opened 2021-06-21
Issue-4673 - SQL Console raises error for variable declarations with "@" character

### LINKED ISSUE #4673: SQL Console raises error for variable declarations with "@" character
## Description of bug
Executing queries in SQL Console raises an error since the "@" character is a special character in PetaPoco to pass arguments to the queries. The queries have to contain double "@" characters for escaping purposes in order for the PetaPoco library to not raise any errors.

## Steps to reproduce
1. From the article https://dnnsupport.dnnsoftware.com/hc/en-us/articles/360011483340 , copy the query in Step 3.1 in the Solution section and paste it into the SQL Console. Or just use a simple query text like "declare @test INT = 0"
2. Run the query.

## Current behavior
It returns an error - "System.ArgumentException: Parameter '@parameterNameHere' specified but none of the passed arguments have a property with this name."

## Expected behavior
The query should work with single "@" character.

## Screenshots
image

## Additional context
I need confirmation from the community to fix this behavior by replacing all of the passed "@" characters with "@@" so that the queries can work just like in SSMS (SQL Server Management Studio).

## Affected version

* [x] 10.00.00 alpha build

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #4673 
## Summary
"@" character is escaped since it has a special meaning in PetaPoco (it's used for passing arguments). The character is only escaped if the "args" parameter doesn't have any value. If the "args" is not empty, PetaPoco will continue handling the parameter replacements as before.


bdukes (commented):
This change seems fine to me, thanks for the implementation @berkarslan-xo. I was expecting that the fix would be done in the SQL component of the Persona Bar, rather than in the PetaPoco base API. I'm not sure that I have a scenario where this is a problem, it's just a wider-ranging change than I expected.

Could there be issues if a developer is using this API and already worked around the escaping issue, e.g. `ExecuteReader("SELECT * FROM Users WHERE Email = '" + email.Replace("@", "@@") + "'")`? This is obviously bad code and I can't think of a scenario where "good" code would cause a problem. @berkarslan-xo also [mentioned](https://github.com/dnnsoftware/Dnn.Platform/issues/4673#issuecomment-846137091) that this issue was introduced after 9.2.1, so I'm less worried about backwards compatibility (I would see this more as fixing a bug than an enhancement).

berkarslan-xo:
Hi @bdukes , the reason why I made the change in the PetaPoco side was that if for example PetaPoco is wanted to be changed with something else in the future and if I made the change in the Persona Bar side, replacing "@" characters might most probably be a problem for the new library. Since the problem is strictly related to PetaPoco, I thought that adding this code to a PetaPoco related place might make more sense since Persona Bar isn't (& shouldn't be) aware that PetaPoco is used for handling the SQLs.

bdukes (approved):

mitchelsellers (approved):

PR #4730, opened 2021-07-04
Added support for absolute and relative URLs

### LINKED ISSUE #3111: Tab Redirect Url improved hinting/placeholder
## Description of Problem
I would like to request for comment/consideration a small improvement to the Page Settings (tab settings) interface to give better help/information to admin users.

I often see an issue where admin users will attempt to setup a page redirect and encounter issues because they do not realize that the Redirect URL expected is a full path complete URL vs being able to accept a local path / incomplete URL. Because of the error messages seen, it can be hard for them to troubleshoot if they don't fully understand what has happened in their page redirect.

Take the example of an admin who attempts to set the Page Setings (tab settings) URL Redirect for a page originally called "Dictionary"

They want to redirect from //Dictionary to //Glossary so in the URL field, they enter
//Glossary
and are expecting the redirect to happen as 
https://www.domainname.com/Glossary
they often enter this type of value into the HTML module when working with anchor links and among other reasons have learned to enter //Glossary so that it loads correctly whether https or http for their domain name.

however, what is saved in the database and displays in the Page Settings the next time it is loaded is a value of
http:////Glossary

image

This then performs a redirect loading either a browser's error message of "cannot locate domain", OR in older DNN versions, an error message like this
```
Exception:
Invalid URI: The hostname could not be parsed.
Stack Trace:
at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind) at System.Uri.CreateUri(Uri baseUri, String relativeUri, Boolean dontEscape) at System.Uri..ctor(Uri baseUri, String relativeUri) at System.Web.HttpResponse.ConvertToFullyQualifiedRedirectUrlIfRequired(String url) at System.Web.HttpResponse.Redirect(String url, Boolean endResponse, Boolean permanent) at System.Web.HttpResponse.RedirectPermanent(String url) at DotNetNuke.Entities.Urls.AdvancedUrlRewriter.CheckForTabExternalForwardOrRedirect(HttpContext context, UrlAction& result, HttpResponse response, FriendlyUrlSettings settings, Guid parentTraceId) at DotNetNuke.Entities.Urls.AdvancedUrlRewriter.ProcessRequest(HttpContext context, Uri requestUri, Boolean useFriendlyUrls, UrlAction result, FriendlyUrlSettings settings, Boolean allowSettingsChange, Guid parentTraceId)
Administrators
You can see this exception because the customErrors attribute in the web.config is set to 'off'. Change this value to 'on' or 'RemoteOnly' to show Error Handling
The error handling would have shown this page : Default.aspx?TabId=412
```
This can be difficult for an admin to troubleshoot if they do not understand what has happened with the redriect attempt. The information within the DNN Event Log is perfect and gives good information regarding the issue, but they may not see the error.

## Proposed Solution Option 1
I would like to propose that we improve the help text and the hinting/placeholder in the field to give better information to the user regarding what is expected for the field

image

Suggested Help Text: Redirect to a URL Address. Please enter a full-path URL complete with http/https.
Suggested Field Hinting: Enter complete path URL

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0
* [x] 8.x
* [x] 7.x

david-poindexter:
Resolves #3111

## Summary
URL page types now support absolute and relative URLs.  Furthermore, better handling has been added for URLs that begin with `//` or have not including a leading `/` (e.g., `PageName`, `PageName#AnchorName`).

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
@@ -1226,7 +1226,8 @@ namespace Dnn.PersonaBar.Pages.Components
 
             if (url.StartsWith("//"))
             {
-                return Globals.AddHTTP(url.Remove(0, 2));
+                return url;
+
             }
 
             if (url.ToLower().StartsWith("http"))
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
@@ -1230,7 +1230,8 @@ namespace Dnn.PersonaBar.Pages.Components
 
             }
 
-            if (url.ToLower().StartsWith("http"))
+            if (url.IndexOf("://") != -1)
+
             {
                 return Globals.AddHTTP(url);
             }
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
@@ -1236,10 +1236,6 @@ namespace Dnn.PersonaBar.Pages.Components
                 return Globals.AddHTTP(url);
             }
 
-            if (!url.StartsWith("/"))
-            {
-                return "/" + url;
-            }
 
             return url;
         }
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
@@ -1219,7 +1219,27 @@ namespace Dnn.PersonaBar.Pages.Components
                 return null;
             }
 
-            return url.ToLower() == "http://" ? "" : Globals.AddHTTP(url);
+            if (url.ToLower() == "http://")
+            {
+                return string.Empty;
+            }
+
+            if (url.StartsWith("//"))
+            {
+                return Globals.AddHTTP(url.Remove(0, 2));
+            }
+
+            if (url.ToLower().StartsWith("http"))
+            {
+                return Globals.AddHTTP(url);
+            }
+
+            if (!url.StartsWith("/"))
+            {
+                return "/" + url;
+            }
+
+            return url;
         }
 
         private static Func<RolePermission, bool> NoLocked()
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Pages/App_LocalResources/Pages.resx
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Pages/App_LocalResources/Pages.resx
@@ -382,10 +382,10 @@
     <value>Redirect to an existing page on your site.</value>
   </data>
   <data name="ExternalUrl.Text" xml:space="preserve">
-    <value>External Url</value>
+    <value>URL</value>
   </data>
   <data name="ExternalUrlTooltip.Text" xml:space="preserve">
-    <value>Redirect to an External URL Resource.</value>
+    <value>Redirect to an absolute or relative URL.</value>
   </data>
   <data name="PermanentRedirect.Text" xml:space="preserve">
     <value>Permanent Redirect</value>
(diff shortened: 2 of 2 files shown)


Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
```diff
@@ -1217,9 +1217,29 @@ private static string GetExternalUrlRedirection(string url)
...
+            }
+
+            if (url.ToLower() == "http://")
+            {
+                return string.Empty;
+            }
+
+            if (url.StartsWith("//"))
+            {
+                return Globals.AddHTTP(url.Remove(0, 2));
```
valadas, line 1229:
```suggestion
                return url;

```

`//google.com` would be a perfectly fine url (technically a Protocol-relative link or PRL), see https://en.wikipedia.org/wiki/URL#Protocol-relative_URLs
So I suggest we keep this "un-manipulated"

In the original issue @moorecreative was describing this as a relative URL but that would be a single slack as in `/MyPage` and not `//MyPage`.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
```diff
@@ -1217,9 +1217,29 @@ private static string GetExternalUrlRedirection(string url)
...
+            }
+
+            if (url.ToLower().StartsWith("http"))
```
valadas, line 1232:
```suggestion
            if (url.IndexOf("://") != -1)

```

And this will cover other protocols like `ftp://someserver.com/folder/file.zip`

valadas (changes requested):
Left some suggestions

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
david-poindexter: Thank you @valadas - I am testing this change now.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
david-poindexter: Excellent point @valadas - I am testing this change now.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
```diff
@@ -1,1485 +1,1507 @@
...
+            if (url.IndexOf("://") != -1)
+
+            {
+                return Globals.AddHTTP(url);
+            }
+
+            if (!url.StartsWith("/"))
+            {
+                return "/" + url;
+            }
```
valadas, line 1242:
```suggestion
```

And one more thing that comes to mind is we should not automatically add the initial `/` because it would break other types of valid urls such as `tel:+1234567890`

valadas (changes requested):

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
david-poindexter: Great point - testing now.

valadas (approved):
Thank you sir and sorry for all the "edge cases" :)

david-poindexter:
> Thank you sir and sorry for all the "edge cases" :)

No problem at all - it is a better PR now as a result.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs
```diff
@@ -1,1485 +1,1503 @@
...
+        }
+
+        private static string GetExternalUrlRedirection(string url)
+        {
+            if (url == null)
+            {
+                return null;
+            }
+
+            if (url.ToLower() == "http://")
```
mitchelsellers, line 1222: If we are making changes here, should we possibly consider defaulting to https:// as the default?

david-poindexter: No, this is checking to see if "http://" was injected in due to old data.  This is not setting anything.

mitchelsellers (approved):

PR #4743, opened 2021-07-13
Fix some conflicts with ASP.NET dependencies

bdukes:
## Summary
The original PR #4716 missed a few pieces.

1) Need to include the WebMatrix DLLs in the new AspNetWebPages package, they were also updated.
2) Need to include DotNetNuke.Web.Mvc in the AspNetMvc package, since it relies on a particular version of MVC (another option would be adjusting the version range on the binding redirect, but that would only fix subsequent upgrades).
3) Move the HTTP module for MVC into the AspNetMvc package, since the pieces it relies on aren't included until after install.

### DIFF
--- a/Build/Tasks/packaging.json
+++ b/Build/Tasks/packaging.json
@@ -37,6 +37,8 @@
     "/bin/System.Web.WebPages.dll",
     "/bin/System.Web.WebPages.Deployment.dll",
     "/bin/System.Web.WebPages.Razor.dll",
+    "/bin/WebMatrix.Data.dll",
+    "/bin/WebMatrix.WebData.dll",
     "/Install/Module/DNNCE_Website.Deprecated_*_Install.zip"
   ],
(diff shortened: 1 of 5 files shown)


valadas:
Thank you soooo much for looking into this!

I will test a clean install and upgrade after it builds.
Just one little thing, where the removals of newlines in the comments in release.config and development.config intentional ?

bdukes:
No, the whitespace changes were done automatically by my editor, I've reverted those changes.

david-poindexter:
I'll be on my computer again in a bit - I can also do a test as soon as the build finishes.

bdukes:
I successfully installed the package and successfully upgraded from 9.9.1

david-poindexter:
Awesome @bdukes - I just downloaded and will test as well.

david-poindexter (approved):
Woo hoo - great success!  Thank you @bdukes 🎉

valadas (approved):
Awesome, a clean install worked fine for me too.

PR #4755, opened 2021-07-21
Add workflow to auto-open PRs to v10 branch

bdukes:
## Summary
This GitHub Actions workflow should open a PR to the `release/10.0.0` branch whenever any PR is merged into the `develop` branch.

## Considerations
The `release/10.0.0` branch should be created before this is merged.

I'm not sure if there's a good way to test this without truly enabling it.

This copies the labels from the original PR. Do we want that, or do we want these v10 merges to always be labelled _Type: Maintenance_?

### DIFF
--- a/.github/workflows/open-merged-pr-to-future.yml
+++ b/.github/workflows/open-merged-pr-to-future.yml
@@ -7,7 +7,7 @@ on:
       - develop
 
 jobs:
-  open-merged-pr-to-ten-branch:
+  open-merged-pr-to-release-branch:
     runs-on: ubuntu-latest
     if: ${{ github.event.pull_request.merged }}
     steps:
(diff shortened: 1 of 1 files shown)


valadas:
Will this leave a ton of v10/xxxx branches ?

bdukes:
Is there an option to delete the branch when merging if it's in the same repo? I know [there's a setting to do that automatically](https://docs.github.com/en/github/administering-a-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches) but I don't remember if it's an option manually. We can always add another workflow to delete them if it's a problem.

valadas:
I know on my repos, I usually just turn that setting on, we can always restore those branches after the fact when needed and it does not do it to protected branches.

valadas (approved):

valadas:

PR #4773, opened 2021-08-03
Issue-4772 - Wrong PortalId is getting set when there are multiple sites existing

### LINKED ISSUE #4772: Wrong PortalId is getting set when there are multiple sites existing
## Description of bug
When there are multiple sites, SiteSettings tab doesn't set the correct portal id on the initial page load. This causes problems on Evoq side, one can watch [this](https://drive.google.com/file/d/1m-1fKyDFRhlmSH5T6HY2FU6_DkDhIyys/view?usp=sharing) video about it. There might be some possible errors on Dnn Platform functionality as well but couldn't find one yet.

## Steps to reproduce
For Evoq:
1. Add a new site
2. Open SiteSettings>Search>Crawling
3. Add an Excluded Directory
4. Refresh the page

For Dnn Platform:
1. Add a new site
2. Open the newly added site
3. Open SiteSettings

## Current behavior
For Evoq: The Excluded Directory can't be seen.
For Dnn Platform: Wrong portal id is being used as can be seen from the following screenshot:
image

## Expected behavior
For Evoq: The Excluded Directory should be seen.
For Dnn Platform: Correct portal id should be set to the state object.

## Affected version

* [x] 10.00.00 alpha build

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

berkarslan-xo:
Fixes #4772 

## Summary

Currently used portal id is getting set correctly now. As a fallback, the old logic is preserved too.


valadas (approved):
Tested it out and works great, a most welcome bug fix!

mitchelsellers (approved):

ggiordan:
I think there may still be an issue here.  We happen to use the DNN.Events module and this exposes the error then the API is called.

If you try to load or create an event, it loads the FCK editor.  As soon as the FCK editor loads, the user gets logged out.

Watching the network trace, it can be seen that this is done by the call to: `https://<host>/API/CKEditorProvider/CloudServices/GetToken`

In chasing this down, in the call to Authenticate in (MembershipModule.cs:99) request, this code returns the portal settings for portal 0, not the current portal:
`
            // Obtain PortalSettings from Current Context
            PortalSettings portalSettings = PortalController.Instance.GetCurrentPortalSettings();
`

So unless the logged in user is a user in portal 0, this results in a NULL being returned:
`var user = UserController.GetCachedUser(portalSettings.PortalId, context.User.Identity.Name);`

To see this happen do the following:
1) have a user who is not a member of portalid = 0
2) log that user in
3) from the current tab or another tab in your brower hit the URL: https://<host>/API/CKEditorProvider/CloudServices/GetToken

At this point the user is logged out... web server will return with a 302 and a delete the session login cookie

if you grab the process in the debugger, you will see that portalSettings.PortalId == 0, where is was correct for all the other calls.

Another way to recreate this is to have a calendar on the page, try to add a new event to the calendar (where the FCK editor gets loaded) you will find that the user gets logged out.

If you change the Event module to not use the FCK editor, the problem is hidden because it does not call the API.

PR #4774, opened 2021-08-04
Added lint task to all node projects 

valadas:
This adds a `lint` task into all projects with autofix enabled. This will make it easier to auto-fix eslint warnings if we change the configs or while updating eslint dependency in batch using lerna or yarn workspaces. Unfortunately (well fortunately) there were no autofixable warnings to resolve after this change which was my main goal with this PR, anyway, it's still good to have I believe for the future.

Also, this PR includes the commits in #4770 so if it is merged first, this becomes easier to review. If this one gets merged first than dependabot should close #4770 automatically too.


bdukes (approved):

bdukes:
With #4770 merged, I've rebased this PR

mitchelsellers (approved):

PR #4778, opened 2021-08-09
Use Yarn 3

bdukes:
## Summary
The project is currently using Yarn 1. I decided to try to migrate to Yarn 3 and see if we can get any benefits from it. Along the way I resolved some issues that we probably want to backport to Yarn 1 if we decide we don't want to go to Yarn 3 at this point.

### Benefits
I think the main benefit we get is better performance and also a better experience when manually managing packages.

### Downsides
The first downside that I see is that Dependabot only support Yarn 1. The other is that there are some differences in how things are run which can cause compatibility issues, but it doesn't look like any remain after I resolved a couple I ran into (and those may not have been 100% Yarn-related anyway).

### Further Reading
* [Yarn 2 announcement](https://dev.to/arcanis/introducing-yarn-2-4eh1)
* [Yarn 3 announcement](https://dev.to/arcanis/yarn-3-0-performances-esbuild-better-patches-e07)


valadas (approved):
Looks good to me and I tested the build and saw all modules loaded fine.

mitchelsellers (approved):

PR #4780, opened 2021-08-12
Updates CDF to v1.9.10

valadas:
This PR updates the client dependency framework to use the Nuget published one v1.9.10 and removes MyGet dependency.

I had to time to test it out right away so sending the PR to see a CI build and will test it out after dinner...


valadas:

PR #4786, opened 2021-08-18
Avoid HTML module using Globals.DependencyProvider

bdukes:
## Summary
Based on [a question on the forums](https://dnncommunity.org/forums/aft/2588) I thought it might be helpful to remove some of the `internal` usages of `Globals.DependencyProvider`, since 3rd party modules can't copy that pattern.


mitchelsellers (approved):
This is a great change!

valadas (approved):
Looks good to me

PR #4788, opened 2021-08-19
Resolve serialization issue

### LINKED ISSUE #3592: AuthenticationConfig RedisCachingProvider serialization error
## Description of problem
There is a caching error during login when running an installation using Azure Redis cache provider which expects all cache items to be serializable.  This is a DNN installation using the provider plugin available here: https://github.com/davidjrh/dnn.rediscachingprovider

DotNetNuke.Providers.RedisCachingProvider.RedisCachingProvider - Error while trying to store in cache the key DNN_Authentication.DNN_0 (Object type: DotNetNuke.Services.Authentication.AuthenticationConfig): System.Runtime.Serialization.SerializationException: Type 'Microsoft.Extensions.DependencyInjection.ServiceProvider' in Assembly 'Microsoft.Extensions.DependencyInjection, Version=2.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' is not marked as serializable.

## Steps to reproduce

1. Ensure you have a functional DNN installation using Azure Redis Cache Provider 
2. Verify you're moving data into your cache via your Azure portal
2. Restart your DNN App Pool
3. Login
4. View the log4net file for the error as shown above

## Additional Context
It may be related to the AuthenticationConfigBase protected IServiceProvider DependencyProvider property.  Initial thought was to test adding the NonSerialized attribute but that won't work with protected members.  This error is showing in an App Service install but it is likely that it affects any installation that tries to implement the Redis Caching Provider plugin or Redis in general.  It's possible this is an installation specific issue.  This is the only DNN error I see when reviewing the logs for redis related issues.

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

bdukes:
## Summary
`AuthenticationConfigBase` is marked as `Serializable` but has an `IServiceProvider` property which cannot be serialized (it's a type defined outside of DNN and is not marked as `Serializable`). This causes an issue when using a caching provider that uses serialization, e.g. https://github.com/davidjrh/dnn.rediscachingprovider

This PR ~adds the `NonSerialized` attribute to the property and~ makes it retrieve the `IServiceProvider` dynamically (so that it always has a valid value, even after going through serialization).

Fixes #3592

### DIFF
--- a/DNN Platform/Library/Services/Authentication/AuthenticationConfigBase.cs
+++ b/DNN Platform/Library/Services/Authentication/AuthenticationConfigBase.cs
@@ -34,7 +34,6 @@ namespace DotNetNuke.Services.Authentication
         public int PortalID { get; set; }
 
         /// <summary>Gets the Dependency Provider to resolve registered services with the container.</summary>
-        [NonSerialized]
         protected IServiceProvider DependencyProvider => Globals.DependencyProvider;
     }
 }
--- a/DNN Platform/Library/Services/Authentication/AuthenticationConfigBase.cs
+++ b/DNN Platform/Library/Services/Authentication/AuthenticationConfigBase.cs
@@ -17,18 +17,13 @@ namespace DotNetNuke.Services.Authentication
     [Serializable]
     public abstract class AuthenticationConfigBase
     {
-        /// <summary>
-        /// Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.
-        /// </summary>
+        /// <summary>Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.</summary>
         public AuthenticationConfigBase()
         {
-            this.DependencyProvider = Globals.DependencyProvider;
         }
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.
-        /// </summary>
-        /// <param name="portalID"></param>
+        /// <summary>Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.</summary>
+        /// <param name="portalID">The portal ID.</param>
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):
This looks good to me and should resolve the issue

ggiordan:
@bdukes thank you so much!!

ggiordan:
@bdukes I'm confused, I'm looking at the code for the file you changed in the main repo and it looks different that your branch.  I see this:
```
 [Serializable]
    public abstract class AuthenticationConfigBase
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.
        /// </summary>
        public AuthenticationConfigBase()
        {
            this.DependencyProvider = Globals.DependencyProvider;
        }

        /// <summary>
        /// Initializes a new instance of the <see cref="AuthenticationConfigBase"/> class.
        /// </summary>
        /// <param name="portalID"></param>
        protected AuthenticationConfigBase(int portalID)
            : this()
        {
            this.PortalID = portalID;
        }

        [Browsable(false)]
        public int PortalID { get; set; }

        /// <summary>
        /// Gets the Dependency Provider to resolve registered
        /// services with the container.
        /// </summary>
        /// <value>
        /// The Dependency Service.
        /// </value>
        protected IServiceProvider DependencyProvider { get; }
    }
```
The code you changed had an [unserializable] decoration.  Also, the getter here is different.
Am I looking in the right place?
The reason I ask is that if this is a good fix, I need to patch my running system.

Thanks!!

Jeff

bdukes:
I had to remove the `NotSerialized` attribute because it only applies to fields, not properties.  The change to the property is all that you should need.

valadas (approved):
Looks good

PR #4790, opened 2021-08-20
IE11 corrections fixes #4787

### LINKED ISSUE #4787: Styling missing for personabar in IE11
## Description of bug
Because of the use of CSS variables the personabar looks pretty bad. I do encourage the use of modern tech, this issue could be easily have been prevented.

## Steps to reproduce
open de site as an admin, you'll see

## Current behavior
styling issues, most obvious, missing backgrounds

## Expected behavior
What we had...

## Affected version
version 09.10.00 

## Affected browser
* [ ] Internet Explorer 11
ie11-issue-personabar

schotman:
Closes #4787

## Summary
added IE specific styling by means of a media query

### DIFF
--- a/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/editBar/css/main.css
+++ b/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/editBar/css/main.css
@@ -195,3 +195,10 @@ body {
 #editBarOptionsExtensionPoint {
     margin-left: 7px;
 }
+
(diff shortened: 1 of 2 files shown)


valadas (approved):
I am ok with this in a 9.x.x release, not sure yet what our stance will be in the future for IE11 support vs going lean. But here it does not impact other browsers and the added lines are very minimal, so I am ok with this.

mitchelsellers (approved):
I agree with @valadas on this item, thanks for the contributions!

PR #4811, opened 2021-09-03
Allow multiple DesktopModules per PackageID

bdukes:
By allowing a DesktopModule to have an explicit FriendlyName and
Description set by the install manifest, instead of using the Package
FriendlyName and Description, allows multiple DesktopModules per Package

Fixes #2887
Reimplementation of #2887 and #2886


david-poindexter (approved):
Looks good - thanks! 🎉

mitchelsellers (approved):

valadas (approved):
Great!

PR #4815, opened 2021-09-04
Marked nuget dependency from DotNetNuke.Web on Microsoft.AspNet.WebApi.Core

valadas:
I am not 100% if what I am doing is correct, but when upgrading a module to depend on DNN 9.10.1, the build fails saying that DotNetNuke.Web uses System.Web.Http v5.2.7 but the project references v5.2.3. I usually just update the dependency manually, but though we could declare it so it updates it automatically.

I am no nuget expert, so not sure if this is the right way to do it, looking at the nuget experts (you know who you are) :P


bdukes (approved):
LGTM

mitchelsellers (approved):

PR #4826, opened 2021-09-13
Added documentation about yarn updates

valadas:
Our build does not work using old version of yarn if yarn is installed globally


valadas:
Hmm, that diff is wrong, I will resubmit...

PR #4827, opened 2021-09-13
Added documentation about yarn updates

valadas:
Our current build does not work if the developer has an old version of yarn installed globally.

I don't know why but the diff looks very strange...


bdukes (approved):

david-poindexter (approved):
I think GitHub must be having some styling issues.  😉 
This looks good though.  Thanks for the contribution @valadas 🎉

PR #4828, opened 2021-09-14
Fixed an issue where it was impossible to edit html on child portals

### LINKED ISSUE #4791: DNN9.10.0 Calling API/CKEditorProvider/CloudServices/GetToken results in user being logged out
## Description of bug
If a user in a portal other than the root portal calls API/CKEditorProvider/CloudServices/GetToken, the user will be logged out.  The user must not be a member of the root portal and from what I can see must not be a super user.

I think this is a different manifestation of the same problem:
[https://github.com/dnnsoftware/Dnn.Platform/pull/4773](https://github.com/dnnsoftware/Dnn.Platform/pull/4773)
In this case, fixing the javascript on the site doesn't help (it's an API call)

## Steps to reproduce
1. Create a new user in a portal that  is not the root portal.  this user must not exist in the root portal 
2. Log the user into the portal
3. Access url `https://<host>/API/CKEditorProvider/CloudServices/GetToken`
4. The response will be a 302 and the user is logged out (the cookie is wiped)

## Current behavior
The response will be a 302 and the user is logged out (the cookie is wiped)

## Expected behavior
A 200 response that does not log the user out

## Screenshots
Capture

## Error information
No errors other than the 302 response

## Additional context
In chasing this down, in the call to Authenticate in (MembershipModule.cs:99) request, this code returns the portal settings for portal 0, not the current portal:
```
// Obtain PortalSettings from Current Context
PortalSettings portalSettings = PortalController.Instance.GetCurrentPortalSettings();

```
So unless the logged in user is a user in portal 0, this results in a NULL being returned:
`var user = UserController.GetCachedUser(portalSettings.PortalId, context.User.Identity.Name);`

To see this happen do the following:

have a user who is not a member of portalid = 0
log that user in
from the current tab or another tab in your brower hit the URL: https://<host>/API/CKEditorProvider/CloudServices/GetToken
At this point the user is logged out... web server will return with a 302 and a delete the session login cookie

if you grab the process in the debugger, you will see that portalSettings.PortalId == 0 for the API call, but it's correct for all the other activity loading the page.

## Affected version

* [ ] 10.00.00 alpha build
* [x] 09.10.00 latest supported release

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer 11
* [x] Microsoft Edge Chromium

valadas:
Fixed an issue where it was impossible to edit html on child portals.
Closes #4791

I have also cleaned up the file a bit
- Resolved all stylecop warnings
- Removed unhelpful comments on private methods
- Renamed some variables and methods for better code clarity
- Updated some comments on public methods
- Removed commented out code

For an easier review, the bug is the first commit and the other commits are cleanup.

### DIFF
--- a/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web/EditorControl.cs
+++ b/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web/EditorControl.cs
@@ -33,6 +33,7 @@ namespace DNNConnect.CKEditorProvider.Web
     using DotNetNuke.Entities.Modules;
     using DotNetNuke.Entities.Portals;
     using DotNetNuke.Entities.Users;
+    using DotNetNuke.Framework;
     using DotNetNuke.Framework.JavaScriptLibraries;
     using DotNetNuke.Framework.Providers;
     using DotNetNuke.Security;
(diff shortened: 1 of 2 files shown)


DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web/EditorControl.cs
```diff
@@ -426,12 +399,13 @@ public NameValueCollection Settings
                 }
 
                 // cloudservices variables need to be set regardless
-                this.settings.Add("cloudServices_tokenUrl", "/API/CKEditorProvider/CloudServices/GetToken");
+                var tokenUrl = ServicesFramework.GetServiceFrameworkRoot() + "API/CKEditorProvider/CloudServices/GetToken";
+                this.settings.Add("cloudServices_tokenUrl", tokenUrl);
```
valadas, line 403: Meh, my commits got mixed up, the end result is still fine but this is basically the fix

valadas: And thanks to @bdukes for the hint on the simpler solution than I had in head 😄

DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web/EditorControl.cs
```diff
@@ -729,14 +692,10 @@ public static Control FindModuleInstance(Control editorControl)
...
-        /// The has rendered text area.
+        /// Checks if the text area was rendered.
         /// </summary>
-        /// <param name="control">
-        /// The control.
-        /// </param>
-        /// <returns>
-        /// Returns if it has rendered text area.
-        /// </returns>
+        /// <param name="control">The control to ckeck.</param>
```
bdukes, line 697:
```suggestion
        /// <param name="control">The control to check.</param>
```

DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web/EditorControl.cs
```diff
@@ -804,7 +757,7 @@ public bool LoadPostData(string postDataKey, NameValueCollection postCollection)
         }
 
         /// <summary>
-        /// The raise post data changed event.
+        /// Taises post data changed event.
```
bdukes, line 760:
```suggestion
        /// Raises the post data changed event.
```

bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #4831, opened 2021-09-17
Removed module specific styling for a default DNN Button

### LINKED ISSUE #4830: Core Messaging add Attachement button alignment issue
## Description of problem
While styling a the Core messaging popup, I noticed that one of the buttons has an incorrect height.
(you need to allow attachments for this to show up)
I have overridden the DNN button styling in my Theme and noticed this button to have an incorrect height.
AFAIK on the whole site this is the only button that does not apply the correct styling:
image
(the left button)
This is caused by this line in module.css: 
https://github.com/dnnsoftware/Dnn.Platform/blob/bf50bee95a499f6fa5d689ec2818152fec5b5208/DNN%20Platform/Modules/CoreMessaging/module.css#L452

That styling isn't needed as the styling of DNN buttons is already defined elsewhere IMO.
I don't know why one would want to overrule the styling of just that button only for this module TBH.

I noticed that also with the default skin, the alignment is off.
image

## Description of solution
Remove the height and padding as the styling for the DNN buttons is already defined elsewhere.

## Screenshots
After this "fix" the popup looks like this in the default Theme:

image

Timo-Breumelhof:
Removed module specific styling for a default DNN Button in the Core Messaging module

fixes #4830

### DIFF
--- a/DNN Platform/Modules/CoreMessaging/module.css
+++ b/DNN Platform/Modules/CoreMessaging/module.css
@@ -449,7 +449,7 @@ button, input[type="button"], input[type="reset"], input[type="submit"], .dnnPri
         padding-top: 18px;
     }
 	.composeMessageDialog .fileUploadArea{ display: inline-block;vertical-align: top;}
-    .composeMessageDialog .dnnTertiaryAction{ margin: 0 10px 0 0;padding: 6px 6px;height: 32px;}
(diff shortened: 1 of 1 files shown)


bdukes (approved):
LGTM

valadas (approved):
I am fine with it... It may affect some themes/customizations of people that tried to fix this other ways, but it won't break the interwebs either.

PR #4838, opened 2021-09-21
Remove unused RestSharp reference

bdukes:
Supersedes #4781


mitchelsellers (approved):
This is awesome, one less dependency.

bdukes:
We never actually shipped the RestSharp assembly, so it wasn't a runtime dependency, just something referenced in the build

valadas (approved):
Awesome

PR #4851, opened 2021-10-03
Resolved an issue where the wrong RedirectAfter tabs were showing for localized sites.

### LINKED ISSUE #4232: SiteSettings of type "RedirectAfter" are not saved correctly in a multi-lingual environment
## Description of bug
Changing the Settings>Site Settings>RedirectAfterLogout in the default language works as expected. But fails in any other language of the portal.

## Steps to reproduce
0. Make sure you have at least 2 languages active in your site, and you are in the default language part of the site
1. Go to Settings>Site Settings>RedirectAfterLogout 
2. Select any page from the dropdown
3. Scroll down and click 'Save'
4. The TabId of the selected page is saved correctly in table PortalSettings with SettingName='Redirect_AfterLogin', with the correct CultureCode
5. Now scroll to the top of the Settings>Site Settings page and click on another language in the language selector (top-right corner)
6. repeat step 1,2 and 3

## Current behavior
The TabId of the selected page is NOT saved correctly in table PortalSettings with SettingName='Redirect_AfterLogin'.  CultureCode in not take into account.

The applies to all settings of this type:
RedirectAfterLoginTabId 
RedirectAfterLoginTabName
RedirectAfterLogoutTabId
RedirectAfterLogoutTabName 
RedirectAfterRegistrationTabId 
RedirectAfterRegistrationTabName 

## Expected behavior
A new record should have been created in the PortalSettings, with the correct value and the correct Culture.

## Additional context
Initial code analysis shows that to problem starts in GetDefaultPagesSettings() on lines 312/313
`    var portal = PortalController.Instance.GetPortal(pid, cultureCode);`
`    var portalSettings = new PortalSettings(portal);`
Here the PortalInfo variable 'portal' is correctly created (with the cultureInfo) and passed through to PortalSettings() constructor.
But clearly, down the line, the cultureInfo is not take into account when needed.

## Affected version
* [ X] 09.08.00 release candidate
* [ X] 09.07.02 latest supported release

### LINKED ISSUE #4260: Login/logout redirection tabs are defaulting to 0 instead of -1
## Description of bug
After login/logout redirection settings were migrated to Site Settings (#3845), there is an issue when saving.
When other settings in the panel are changed but `Redirect After Login/Logout` are left untouched, their default values are set to 0, instead of -1. This causes DNN to navigate to tab ID 0 after logout, causing a 404:

http://platform-9-7-1.dnndev.me/tabid/0/Default.aspx

Notice the tabid/0 part of the URL.

Upon further investigation, I noticed that these values are `null` in the front-end, and converted to `0` in the back-end when cast to integers.

## Steps to reproduce
1. Clean DNN 9.7.0 or above
3. Open SETTINGS > Site Settings > Site Behavior > DEFAULT PAGES
4. Click on the **User Profile Page** dropdown (the dropdown expands and the Save button becomes enabled)
6. Click Save
7. Close the Persona Bar
8. Log out

## Current behavior
DNN redirects to `/tabid/0/Default.aspx` after logout, causing a 404.

## Expected behavior
DNN should redirect to the home page.

## Screenshots
image

## Error information
N/A

## Additional context
N/A

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.08.00 release candidate
* [X] 09.07.02 latest supported release
* [X] 09.07.01 latest supported release
* [X] 09.07.00 latest supported release
* [ ] 09.06.02 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

valadas:
The settings for the current portal alias local were returned to the UI instead of the requested language picked in the Page Settings editor. This caused an issue where you would get let's say the English tabs for a french site setting if the last page you happened to visit was English instead of french. The user would then save any setting and all the RedirectAfters would save the wrong thing for that locale.

Closes #4232
Closes #4260

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs
@@ -310,7 +310,11 @@ namespace Dnn.PersonaBar.SiteSettings.Services
                 }
 
                 var portal = PortalController.Instance.GetPortal(pid, cultureCode);
-                var portalSettings = new PortalSettings(portal);
+                var localizedPortalSettings = PortalController.Instance.GetPortalSettings(portal.PortalID, cultureCode);
+
+                int redirectAfterLoginTabId = int.TryParse(localizedPortalSettings["Redirect_AfterLogin"], out redirectAfterLoginTabId) ? redirectAfterLoginTabId : -1;
+                int redirectAfterLogoutTabId = int.TryParse(localizedPortalSettings["Redirect_AfterLogout"], out redirectAfterLogoutTabId) ? redirectAfterLogoutTabId : -1;
+                int redirectAfterRegistrationTabId = int.TryParse(localizedPortalSettings["Redirect_AfterRegistration"], out redirectAfterRegistrationTabId) ? redirectAfterRegistrationTabId : -1;
 
                 return this.Request.CreateResponse(HttpStatusCode.OK, new
                 {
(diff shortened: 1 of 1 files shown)


valadas:
For reviewers, the first commit is the fix and the second commit cleans this file a bit to reduce about 200 warnings.

mitchelsellers (approved):

mitchelsellers:
Pulled this one locally seems to fix the issue on my site that was multi language. Do we think this fixes it for others too?

valadas:
> Do we think this fixes it for others too?

This kinda fixes two things
1. the default value from the getter could have been null, now it would be -1
2. on localized sites it would return the locale for the portal alias instead of the one for the currently selected language in the editor

So I do believe this should fix both problems.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs
```diff
@@ -394,10 +398,11 @@ public HttpResponseMessage UpdateDefaultPagesSettings(UpdateDefaultPagesSettings
                 portalInfo.TermsTabId = this.ValidateTabId(request.TermsTabId, pid);
                 portalInfo.PrivacyTabId = this.ValidateTabId(request.PrivacyTabId, pid);
                 PortalController.Instance.UpdatePortalInfo(portalInfo);
-
-                PortalController.UpdatePortalSetting(pid, "Redirect_AfterLogin", this.ValidateTabId(request.RedirectAfterLoginTabId, pid).ToString(), false, cultureCode);
-                PortalController.UpdatePortalSetting(pid, "Redirect_AfterLogout", this.ValidateTabId(request.RedirectAfterLogoutTabId, pid).ToString(), false, cultureCode);
-                PortalController.UpdatePortalSetting(pid, "Redirect_AfterRegistration", this.ValidateTabId(request.RedirectAfterRegistrationTabId, pid).ToString(), false, cultureCode);
+
+                var portalSettings = PortalController.Instance.GetCurrentSettings();
```
david-poindexter, line 412: I think maybe this line was added by mistake @valadas - can you confirm?

david-poindexter (changes requested):
Awesome find and fix @valadas 🎉

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs
valadas: Good catch, fixed

david-poindexter (approved):

bdukes (approved):

PR #4857, opened 2021-10-04
Removed unused Fare nuget package

valadas:
Supercedes #4855


valadas:
Let's see if all tests pass, but I think this reference was simply unused...

bdukes:
TestStack.Dossier depends on AutoFixture which depends on Fare. It's only used in the integration tests project, so we wouldn't get a test failure from the CI build.

https://github.com/dnnsoftware/Dnn.Platform/blob/857cf0dd9e7acf77f7f749fa9ff0fd6932d94c02/DNN%20Platform/Tests/DotNetNuke.Tests.Integration/Executers/Builders/TabPermissionsBuilder.cs#L10-L12

valadas:
Oh I see, ok, let's close this then

PR #4907, opened 2021-11-09
Fix #4906: cache the roles array in the user info object.

### LINKED ISSUE #4906: UserInfo.Roles object need to be cached to optimize the performance
## Description of problem
UserInfo.Roles value will be calculated from Social.Roles every time when trying to access it, and it will create a new copy of the array data in memory. this property is used in PortalSecurity.IsInRole method, if this method is called frequency, there will be lots of garbage generated by it and hit the server performance.

## Description of solution
We can cache the roles object so that it will be generated only once, and then can be called from the cached object directly to improve the performance.

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

zyhfish:
Fix #4906.

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserInfo.cs
+++ b/DNN Platform/Library/Entities/Users/UserInfo.cs
@@ -40,6 +40,7 @@ namespace DotNetNuke.Entities.Users
         private string _administratorRoleName;
         private UserMembership _membership;
         private UserProfile _profile;
+        private string[] _roles;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="UserInfo"/> class.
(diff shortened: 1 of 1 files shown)


valadas (approved):
Makes sense to me

mitchelsellers (approved):

abner-diaz:
Hi -- I work for the client that request this patch as a result of our internal load testing.  I think the wording _cache_ is not as accurate it just being a private member in the class.

The `UserInfo` class is `CacheLevel.notCacheable`, so it's lifecycle seems limited to a request.   Also, the code pattern used in the `Roles` get accessor was taken from the `Membership` get accessor.  So any issues with cache refresh or stale data would also occur there (since that is also likely to change).  That is unlikely because the `UserInfo` class does not seem targeted for caching.

zyhfish:
> How does this cache get cleared if the roles were to change?

Hi @bdukes , the Roles info is calculated from Social.Roles property that is saved in a private field as well, so it should be good enough to store in a private field for it.

valadas:
Sounds good to me... @bdukes you are fine with merging this ?

PR #4925, opened 2021-12-01
Fixed SEO tab not updating correctly after deleting a URL

### LINKED ISSUE #4924: SEO tab in Pages panel doesn't update correctly after deleting a URL
## Description of bug
After deleting a page URL in the SEO tab, the view doesn't update correctly.
Navigating to another tab and then back to S.E.O. forces the view to update correctly.

## Steps to reproduce
1. Open PB > CONTENT > Pages
2. Click on **[EXPAND ALL]**
3. Click on **My Profile** page
4. Click on **Advanced** > **S.E.O.**. There is only one URL in the table: **/Activity-Feed/My-Profile Active (200) Automatic**.
5. Click on **+Add URL**
6. Type **test** in the URL Path field (rendered value will be **/test**)
7. Click on **Save**. The new URL is created (replacing the original URL) and 2 additional 301 automatic URLs are created.
8. Click on the delete icon next to the **/test** URL
9. Confirm the delete operation

## Current behavior
The **/test** URL is deleted but the 2 automatic URLs remain.

## Expected behavior
All 3 URLs should be gone and the original **Active (200) Automatic** URL should be reinstated.

## Screenshots
1) Before creating the /test URL
image

2) After creating the /test URL
image

3) After deleting the /test URL (notice the other 2 URLs remain)
image

4) After navigating to APPEARANCE and then back to S.E.O. (the view shows updated now)
image

## Error information
N/A

## Additional context
N/A

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.10.02 release candidate
* [X] 09.10.01 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer 11
* [X] Microsoft Edge Chromium

daguiler:
Fixes #4924

## Summary
The S.E.O. view doesn't update correctly after changing the state because all automatic URLs are returned by the backend with the same `id` of -1. And this `id` property is used as the React component key, which needs to be unique.
This is a typical pitfall with React.

My fix consists of using the array index to generate a negative sequence so that the first automatic URL gets -1, the second -2 and so on.

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/Table.jsx
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/Table.jsx
@@ -15,9 +15,9 @@ class Table extends Component {
 
         const { siteAliases, primaryAliasId, onSave, onCancel, onDelete, onChange, editedUrl,
             pageHasParent, editingUrl, onOpenEditForm } = this.props;
-        return pageUrls.map(url => {
+        return pageUrls.map((url, index) => {
             return <UrlRow 
-                key={url.id}
+                key={url.id === -1 ? -1 - index : url.id}
                 url={url}
                 editedUrl={editedUrl}
                 onOpenEditForm={onOpenEditForm}


mitchelsellers (approved):
This appears to improve the behavior, thank you

bdukes (approved):

valadas (commented):
awesome

PR #4926, opened 2021-12-01
Adjust release merge workflow

bdukes:
The [workflow triggered](https://github.com/dnnsoftware/Dnn.Platform/runs/4382542237) by the merge/rebase of #4925 failed, presumably because it was accessing the `merge_commit_sha` but the PR was closed via rebase. I've updated the workflow to just use `sha`, which still points to the rebased commit, and [tested on my repo](https://github.com/bdukes/Dnn.Platform/pull/4).

### DIFF
--- a/.github/workflows/open-merged-pr-to-future-9.11.0.yml
+++ b/.github/workflows/open-merged-pr-to-future-9.11.0.yml
@@ -7,7 +7,7 @@ on:
       - develop
 
 jobs:
-  open-merged-pr-to-ten-branch:
+  open-merged-pr-to-release-branch:
     runs-on: ubuntu-latest
     if: ${{ github.event.pull_request.merged }}
     steps:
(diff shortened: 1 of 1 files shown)


valadas (approved):

valadas:

bdukes:
@valadas @mitchelsellers it looks like there's still a permissions issue (I guess on my test repo opening a PR for my own repo didn't trigger this scenario). I think we may need to add a secret with an access token that has permission to create branches, and use that instead of `GITHUB_TOKEN`.

valadas:
I don't know, I have done it with GITHUB_TOKEN for the updating of versions one...

bdukes:
I saw someone saying that GitHub won't allow editing workflow files with `GITHUB_TOKEN`, so maybe it's just that `develop` currently has some changes to the workflow files…

bdukes:
I guess it just successfully created #4934, so maybe it has to do with who created the PR it's reacting to?

PR #4930, opened 2021-12-02
No module message should be able to contain any scripts

donker:
In the framework we can push a message to the UI which appears in a colored box. This message should not be able to contain any scripts.


bdukes (approved):

david-poindexter (approved):
Thanks @donker

mitchelsellers:
@valadas this should be noted explicitly in the documentation as this is a breaking change for developers as the API that is used to call/display these messages has not historically had filtering on it, this will most likely result in changes in behavior for some implementations.

PR #4932, opened 2021-12-02
HTML encode message input

bdukes:
## Summary
Content added to message needs to be encoded to ensure proper display


david-poindexter (approved):
Thanks @bdukes 🎉

mitchelsellers (approved):

PR #4959, opened 2021-12-10
Make check for RequestVerificationToken case insensitive as per RFC 7230 and RFC 7540.

### LINKED ISSUE #4958: MVC ValidateAntiForgeryTokenAttribute does case sensitive checks.
## Description of bug
The logic that checks for the AntiForgery header/form keys does case sensitive checks, but as per [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) and [RFC 7540](https://datatracker.ietf.org/doc/html/RFC7540#section-8.1.2) they are case-insensitive.

The issue reproduces (for me at least, in Edge Chromium 96 and Chrome) when the requests are done with http2. When that happens the headers are sent all lowercase. In http 1.1 looks like they are sent as set in JavaScript, so the issue does not reproduce.

## Steps to reproduce
1. Setup a http2 instance in IIS (needs https binding)
2. Create an MVC view that supports POST and has the ValidateAntiForgeryToken Attribute.
3. Make an XHR request to the MVC view URL (with $.ServicesFramework headers set)
4. 401 will be thrown.

## Current behavior
Request fails with 401.

## Expected behavior
Request should succeed.

## Screenshots

## Error information

## Additional context

## Affected version

* [ ] 10.00.00 alpha build
* [x] 09.10.02 release

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [x] Microsoft Edge Chromium

dimarobert:
Fixes #4958

## Summary
Properly check for RequestVerificationToken and __RequestVerificationToken in ValidateAntiForgeryTokenAttribute.
The AllKeys property is a plain string[] and Contains is the System.Linq.Enumerable.Contains<T>() which by default does a case-sensitive check.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionFilters/ValidateAntiForgeryTokenAttribute.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionFilters/ValidateAntiForgeryTokenAttribute.cs
@@ -9,10 +9,6 @@ namespace DotNetNuke.Web.Mvc.Framework.ActionFilters
     using System.Linq;
     using System.Threading;
     using System.Web;
-    using System.Web.Mvc;
-
-    using DotNetNuke.Entities.Portals;
-    using DotNetNuke.Entities.Users;
     using DotNetNuke.Web.Mvc.Common;
 
     public class ValidateAntiForgeryTokenAttribute : AuthorizeAttributeBase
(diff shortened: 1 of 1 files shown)


bdukes (approved):
LGTM, thanks!

mitchelsellers (approved):
This looks good to me, and MIGHT be a root cause for some persistent request verification errors that have been reported over the eyears.

PR #4956, opened 2021-12-08
Journal: Website Url for journal link posts not being saved to ItemData in Journal database - Fixes #3308 

thabaum:
## Summary

Fixes #3308 

`url` string was being checked if it had `//` so any address with it would get denied.  

I changed the statement that did NOT allow urls with `//` to allow  `HTTPS://`, `HTTP://` and `www.`

image

Sure seems to show up the links now in the database and when testing.

I am seeing the effects of my changes after rebuilding the project.  This is also my first attempt in long time to do a PR.  I found it easiest since only 1 file to copy and paste the changed file contents until I get used to github branches again.

I changed the not allow to allow urls with `HTTPS://`, `HTTP://` and `www.`

I will try to make these also create a link for the address typed in the message.  I hope this is acceptable.  I also changed a few names to have UppercaseFormat and was going to document what I could but I need to learn how AND I wanted to do some dependency injection fixes on that same file.

If I do before this gets applied, if it does, I will resubmit a new PR with all the changes made.

PR #4960, opened 2021-12-12
Journal Module: Updates PrepareUrl() to use secured links - Fixes #2314

### LINKED ISSUE #2314: Journal creates website link thumbnail images in posts that are not secured when SSL Enforced or Enabled on a page.
## Description of bug
Journal Creates website link images that are not secured when SSL Enforced or Enabled

## Steps to reproduce
1. Go to journal activity page that has SSL Enforced/Enabled
2. Create a post with a link inside so it creates an image of the website.
3. The image is displayed with a linke starting with HTTP instead of HTTPS causing SSL certificate warning in browsers

## Current result
Unsecured images being displayed on secured pages from journal posts with website links

## Expected result
links to the image thumbnails created from the website link be displayed using same SSL state as the page.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

thabaum:
Fixes #2314 and should be included with fix #4956 

image

Should be included with fix #4956 

## Summary

Updates PrepareUrl() method in the Utilities.cs file of the Journal module to use HTTPS instead of HTTP.

I can make a switch according to portal SSLEnabled setting to use the old way.  Since us small time community members are probably the only ones using this, not sure if it matters unless under an HTTP Load balancer potentially.

The reason I mention the above is because http://www.msn.com will be recreated as https://www.msn.com.  In todays world this is how it should act.  However if you want it to stay HTTP:// we will need to adjust things more.

Here are my test results with the last Url being copied and pasted from the link in journal post below (test this with fix #4956):
image

### DIFF
--- a/DNN Platform/Modules/Journal/Components/Utilities.cs
+++ b/DNN Platform/Modules/Journal/Components/Utilities.cs
@@ -95,7 +95,7 @@ namespace DotNetNuke.Modules.Journal.Components
             sText = HttpUtility.HtmlEncode(sText);
             return sText;
         }
-
+        
         public static bool AreFriends(UserInfo profileUser, UserInfo currentUser)
         {
             var friendsRelationShip = RelationshipController.Instance.GetFriendRelationship(profileUser, currentUser);
@@ -154,35 +154,31 @@ namespace DotNetNuke.Modules.Journal.Components
             url = url.Trim();
             if (!url.StartsWith("http"))
             {
-                url = "https://" + url;
+                url = "http://" + url;
             }
 
-            if (url.Contains("http://"))
+            if (url.Contains("https://"))
             {
-                url = url.Replace("http://", "https://");
+                url = url.Replace("https://", "http://");
             }
 
             if (url.Contains("http://http://"))
             {
-                url = url.Replace("http://http://", "https://");
+                url = url.Replace("http://http://", "http://");
             }
 
-            if (url.Contains("https://https://"))
+            if (!(url.IndexOf("http://") == 0))
             {
-                url = url.Replace("https://https://", "https://");
-            }
-
-            if (!(url.IndexOf("https://") == 0))
-            {
-                url = "https://" + url;
+                url = "http://" + url;
             }
 
             Uri objURI = null;
 
             objURI = new Uri(url);
+
             return url;
         }
-        
+
         internal static LinkInfo GetLinkData(string URL)
         {
             string sPage = GetPageFromURL(ref URL, string.Empty, string.Empty);
@@ -195,6 +191,16 @@ namespace DotNetNuke.Modules.Journal.Components
             string sTitle = string.Empty;
             string sDescription = string.Empty;
             string sImage = string.Empty;
+
+            if (URL.Contains("http://"))
+            {
+                URL = URL.Replace("http://", "//");
+            }
+
+            if (URL.Contains("https://"))
+            {
+                URL = URL.Replace("https://", "//");
+            }
 
             link.URL = URL;
             link.Images = new List<ImageInfo>();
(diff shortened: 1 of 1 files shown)


mitchelsellers (changes requested):
This solution will fix things for those that are SSL Secured, however, it will not allow operation for those that are not SSL secured.

The preference for implementation would be to use // rather than including the protocol at all which should be agnostic to HTTP or https

thabaum:
I believe what you may be trying to say is pull off the `https://` and `http://` and just have the 'domain.com/pic.jpg'

am I correct?

This PR makes it so the images are secured.  They are all forced to HTTP currently.

mitchelsellers:
@thabaum Close, the suggestion is to change them all to

`//domain.com/pic.jpg`

Defaulting everything to https breaks ALL sites that don't have HTTPS, defaulting to HTTP breaks SOME sites with SSL.  Using the // syntax allows the protocol of the page being loaded to dictate https or HTTP

thabaum:
@mitchelsellers gotcha, I will do some testing and play with the `//` only being applied.  Fix the PR as needed.  Thank you.

thabaum:
@mitchelsellers looks like `//` adjustment in this area breaks the link functionality.  I will have to dig deeper into the sources to make this change.  Everything is free to encrypt now with LetsEncrypt with most all sites in today's world should serve HTTPS more so than not.  A user enters an HTTPS address and it is saved HTTP currently creating all kinds of images that are not secured on a secured page.

I will play with this some more in the near future to see where I can actually remove the current `HTTP:` leaving just the `//` while saved to JSON in database.

thabaum:
@mitchelsellers links all convert to `//` when stored in the database.  This PR plus #4956 need to be applied together... please review.

DNN Platform/Modules/Journal/Components/Utilities.cs
```diff
@@ -1,355 +1,374 @@
...
+                    if (subM.Groups[4].Value.Equals("OG:TITLE", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        link.Title = subM.Groups[9].Value;
+                    }
+
+                    if (subM.Groups[4].Value.Equals("OG:IMAGE", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        sImage = subM.Groups[9].Value;
+                        ImageInfo img = new ImageInfo();
+                        if (sImage.Contains("http://"))
```
mitchelsellers, line 239: If the image path has a set protocol, I really don't think we should be changing it.  If no protocol, fine add the // to assume site protocol, but blindly changing to match site protocol isn't a great idea

mitchelsellers:
This PR has not had any activity in the past 3 months, if you would like to re-submit with the review comments resolved please do so.

DNN Platform/Modules/Journal/Components/Utilities.cs
thabaum: > @thabaum Close, the suggestion is to change them all to
> 
> `//domain.com/pic.jpg`
> 
> Defaulting everything to https breaks ALL sites that don't have HTTPS, defaulting to HTTP breaks SOME sites with SSL. Using the // syntax allows the protocol of the page being loaded to dictate https or HTTP

@mitchelsellers 

I figured the website serving the image would set this upon request.  If the requested web server is enabled to serve HTTPS it would serve it that way or serve it HTTP.  DNN would just ask for '//' and allow the servers to figure out which is best upon request.

so your saying only the www.domainrequest.com images should contain the '//' however I understood 'ALL' should have '//' which would then allow any images requested to come as secured or insecured depending on the page requesting the image.  What I am trying to resolve is unsecured images warning on pages that allow only secured content.

I don't believe allowing images in 'http://' would fix this issue.   The problem is unsecured content on pages (images).  At the time the url is "prepared" it saves this url to the database.  Right now it ONLY saves 'http://' which is way off from being a secured website approach.

What do you recommend?

thabaum: if else statements for each variation of a url link to prepare for?  As in a url without 'http' just add '//' and rules that contain 'http' save with the http and urls with 'https' save with https I suppose maybe the solution.  But for my sites I really want only HTTPS images served or no images to avoid the page looking insecured.

So maybe a switch in the panel to dictate the behavior of how these image urls are saved.  Default like it is or allow server protocol to decide.  I just see the current version as not being secured at all serving content via HTTP and so it makes a site page look scary flashing warnings in browsers.

PR #4975, opened 2022-01-04
Ensure Resource Manager installs by default

### LINKED ISSUE #4944: The Resource Manger package needs to be converted from a .resources to a .zip

donker:
Change the extension of the install file to zip so it will install by default. 
Resolves #4944

### DIFF
--- a/Build/Tasks/CreateUpgrade.cs
+++ b/Build/Tasks/CreateUpgrade.cs
@@ -25,6 +25,7 @@ namespace DotNetNuke.Build.Tasks
         public override void Run(Context context)
         {
             this.RenameResourcesFor98xUpgrades(context);
+            this.RenameResourcesFor910xUpgrades(context);
             context.CreateDirectory(context.ArtifactsFolder);
             var excludes = new string[context.PackagingPatterns.InstallExclude.Length + context.PackagingPatterns.UpgradeExclude.Length];
             context.PackagingPatterns.InstallExclude.CopyTo(excludes, 0);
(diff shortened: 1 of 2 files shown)


mitchelsellers (commented):
This change should ONLY be for clean installations. I believe this will impact both

david-poindexter:
@donker I'll go ahead and edit your description, but take a look at [this article](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) for future reference on the specific keywords that will ensure the associated issue will be auto-closed from a PR merge.

donker:
> This change should ONLY be for clean installations. I believe this will impact both

Our build processes are not designed for this. We can examine if this is possible to change or we can accept that upgraded installations have the new module but only have it replace the old one optionally.

bdukes:
We're already renaming `zip` to `resources` for the Telerik packages in the Upgrade zip:
https://github.com/dnnsoftware/Dnn.Platform/blob/29abc2ee481aab4c97ea44e100211eda946c7a02/Build/Tasks/CreateUpgrade.cs#L40-L61

sleupold:
getting the new module installed even on upgrades is not a problem IMO. Admins still need to replace the module on the admin page and uninstall DAM (and Telerik).

Build/Tasks/CreateUpgrade.cs
```diff
@@ -25,6 +25,7 @@ public sealed class CreateUpgrade : FrostingTask<Context>
         public override void Run(Context context)
         {
             this.RenameResourcesFor98xUpgrades(context);
+            this.RenameResourcesFor910xUpgrades(context);
```
mitchelsellers, line 28: Considering that this code is only running for the current version releases, and is internal to the build process.

Could we combine the existing method and this method to something such as `RenameResourcesForUpgrades` which is more descriptive of the action necessary and leave the `[obsolete]` tag?

mitchelsellers (changes requested):

Build/Tasks/CreateUpgrade.cs
donker: Done

bdukes: I think @mitchelsellers was saying to leave the `Obsolete` attribute (we still want to remove this method in DNN 10), otherwise this looks good to me 👍🏻

valadas: Yeah, I would also prefer to keep the obsolete so we don't forget to remove this.

donker: Done

bdukes (approved):

mitchelsellers (approved):

PR #5001, opened 2022-02-01
Purge expired auth cookies

### LINKED ISSUE #4982: Purge expired AuthCookies
## Description of problem

It appears AuthCookies stay in the database forever, leading some users to see millions of records.

## Description of solution

There is a [DeleteExpiredCookies on AuthCookieController](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Security/Cookies/AuthCookieController.cs#L58) but that doesn't appear to be used anywhere. It seems obvious we should call it from a scheduled task or similar.

donker:
## Summary
Add a scheduled task to purge AuthCookies from the DB that have expired.

Fixes #4982

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.11.00.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.11.00.SqlDataProvider
@@ -25,7 +25,7 @@ WHERE SettingName = 'SSLSetup'
 GO
 
 
-/*    Add scheduled task to delete soft deleted users       */
+/*    Add scheduled task to delete expired auth cookies       */
 /************************************************************/
 
 IF NOT EXISTS(SELECT ScheduleID FROM {databaseOwner}[{objectQualifier}Schedule] WHERE TypeFullName = 'DotNetNuke.Security.Cookies.PurgeAuthCookiesTask, DOTNETNUKE')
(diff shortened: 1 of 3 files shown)


sleupold (commented):
sproc: Using TOP should include an "Order By" statement.

donker:
> sproc: Using TOP should include an "Order By" statement.

Not for DELETE TOP.

valadas (approved):
Looks good to me, seen a ton of delete examples that don't have an OrderBy

DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.11.00.SqlDataProvider
```diff
@@ -24,6 +24,35 @@ WHERE SettingName = 'SSLSetup'
  AND SettingValue = 'True'
 GO
 
+
+/*    Add scheduled task to delete soft deleted users       */
```
bdukes, line 28:
```suggestion
/*    Add scheduled task to delete expired auth cookies       */
```

bdukes (approved):

mitchelsellers (approved):

sleupold:
for large tables, it might be useful to add an index on "ExpiresOn" column and use "ORDER BY ExpiresOn DESC"

WillStrohl:
Nice one! :)

PR #5004, opened 2022-02-04
Setup the localization for partial MVC view requests.

### LINKED ISSUE #5003: CurrentCulture/CurrentUICulture are not set when rendering a partial MVC view.
## Description of bug
When a MVC view is called using it's endpoint (e.g. `/DesktopModules/MVC/<ModuleFolder>/<Controller>/<Action>`) the `Localization.SetThreadCultures` method is never called to set CurrentCulture and CurrentUICulture properly.

It is however set in all other situations (MVC module rendering through Default.aspx, WebAPI).

I found two ways to fix this:
1. Add an ActionFilter in the GlobalFilters that sets up localization ([link](https://stackoverflow.com/a/11726897))
2. Use the existing DnnMvcHandler to setup localization.

Please let me know how do we want to do this and I will create a Pull Request.

## Steps to reproduce
1. Create a MVC view and register it.
2. Call it with the URL scheme `/DesktopModules/MVC/<ModuleFolder>/<Controller>/<Action>`.
3. The `CultureInfo.CurrentCulture` will always be `en-US` and the `CultureInfo.CurrentUICulture` will be `en` (no region).

## Affected version
* [ ] 10.00.00 alpha build

## Affected browser
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

dimarobert:
Fixes #5003

## Summary
Sets up the thread culture info for partially rendered MVC views, similar to how it is done for normal page requests (WebForms and MVC) and WebAPI requests.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcHandler.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/DnnMvcHandler.cs
@@ -10,7 +10,9 @@ namespace DotNetNuke.Web.Mvc
     using System.Web.SessionState;
 
     using DotNetNuke.ComponentModel;
+    using DotNetNuke.Entities.Portals;
     using DotNetNuke.HttpModules.Membership;
+    using DotNetNuke.Services.Localization;
     using DotNetNuke.UI.Modules;
     using DotNetNuke.Web.Mvc.Common;
     using DotNetNuke.Web.Mvc.Framework.Modules;
@@ -66,6 +68,7 @@ namespace DotNetNuke.Web.Mvc
 
         void IHttpHandler.ProcessRequest(HttpContext httpContext)
         {
+            this.SetThreadCulture();
             MembershipModule.AuthenticateRequest(this.RequestContext.HttpContext, allowUnknownExtensions: true);
             this.ProcessRequest(httpContext);
         }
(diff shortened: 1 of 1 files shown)


bdukes (approved):

bdukes:
It looks to me like this failed because it ran on an agent with windows 2022 instead of 2019.

>D:\a\1\s\Dnn.AdminExperience\Dnn.PersonaBar.Extensions\Dnn.PersonaBar.Extensions.csproj(755,3): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v11.0\WebApplications\Microsoft.WebApplication.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [D:\a\1\s\Dnn.AdminExperience\Dnn.PersonaBar.Extensions\Dnn.PersonaBar.Extensions.csproj]

The line it's pointing to is
https://github.com/dnnsoftware/Dnn.Platform/blob/edf47830ae896a0f3fa862195d9848b9a7a8e79c/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Dnn.PersonaBar.Extensions.csproj#L755

This is an error during the clean target, not even the rebuild target.

One solution is to revert to windows-2019 instead of using windows-latest. We can also look at upgrading Cake to 2.x, which is supposed to know more about the newer version of Visual Studio (which, I assume, is pre-installed on the agent).

valadas (approved):
Awesome

PR #5007, opened 2022-02-08
Use windows-2019 image for CI build

bdukes:
The 2022 image is currently causing issues, presumably we need to upgrade to Cake 2.x

See https://github.com/dnnsoftware/Dnn.Platform/pull/5004#issuecomment-1030281326 for details


valadas (approved):

valadas:

PR #5020, opened 2022-02-17
Fixed an issue where WebFormsMvp initial version mismatched with the config

valadas:
A clean install would fail due the the mismatched version in the web.config vs the file in the bin folder.

Upon upgrade the DLL file in the bin folder is not included and the package should take care of upgrading the one that is there.


bdukes (approved):

valadas:
I pushed an additional commit that removes it from a clean install too and tested it is not needed to run the install.

david-poindexter (approved):

david-poindexter:
🌮

PR #5028, opened 2022-02-27
Fix versions of dependencies in Nuget spec files

donker:
Since version 9.10.0 DNN moved up the dependencies on Microsoft's libraries. This should be reflected in the Nuspec file or else your module will not build. This also means that 9.10.0, 9.10.1 and 9.10.2 nuget packages for DNN will not work to build MVC style modules.


bdukes (approved):

mitchelsellers (approved):
Approving this. I’m also going to look and see if this can be managed automatically like it is of other project types for the future.

Issue #5047, opened 2022-03-15
DNN Platform\Website\admin\Skins\Privacy.ascx.cs & Terms.ascx.cs 


Andy9999:
By default the links to privacy and terms pages are getting a rel="nofollow" added from the CS code. There are mixed opinions on having it set to nofollow (pro and contra). Some say having it set to nofollow could look kind of shady to Google.

To change this, one currently has to recompile DNN and use a custom build.

To improve this I would suggest to remove it from the CS code and put the rel="nofollow" into the ascx files instead, where one could customize the behavior without recompiling.

I could submit this change and would be glad not having to make custom builds anymore.

Best regards & happy DNNing
Andy9999

bdukes:
Yes, this would be a welcome change. We would want the skin object to still default to `rel=nofollow` if there's nothing specified, but allow overriding that.

Andy9999:
Thank you for the feedback! I will work on this issue (and the other 3 ones) on the side line so it may take a while.

Best regards & happy DNNing
Andy9999

PR #5054, opened 2022-03-20
Allows optionally removing rel from privacy and terms

Andy9999:
Fix for issue #5047.

This will allow to optionally remove the rel attribute from privacy and terms links. Default is still "rel=nofollow" but it could now either be changed or removed entirely from the ascx file.

Unfortunately the asp.net control doesn't have a rel attribute, so we have to add it ourselves.

Added Rel attribute to Privacy and Terms class and added rel="nofollow" to privacy.ascx and terms.ascx skin file.

### DIFF
--- a/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
@@ -60,6 +60,7 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
 
                 this.hypPrivacy.NavigateUrl = this.PortalSettings.PrivacyTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Privacy") : this._navigationManager.NavigateURL(this.PortalSettings.PrivacyTabId);
+                this.hypPrivacy.Attributes["rel"] = this.Rel;
             }
             catch (Exception exc)
             {
--- a/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
@@ -31,6 +31,8 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
+        public string Rel { get; set; }
+
         protected override void OnInit(EventArgs e)
         {
             base.OnInit(e);
@@ -58,7 +60,6 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
 
                 this.hypPrivacy.NavigateUrl = this.PortalSettings.PrivacyTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Privacy") : this._navigationManager.NavigateURL(this.PortalSettings.PrivacyTabId);
-                this.hypPrivacy.Attributes["rel"] = "nofollow";
             }
             catch (Exception exc)
             {
--- a/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
@@ -31,7 +31,7 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
-        public string Rel { get; set; } = "nofollow"
+        public string Rel { get; set; } = "nofollow";
 
         protected override void OnInit(EventArgs e)
         {
--- a/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Privacy.ascx.cs
@@ -31,7 +31,7 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
-        public string Rel { get; set; }
+        public string Rel { get; set; } = "nofollow"
 
         protected override void OnInit(EventArgs e)
         {
--- a/DNN Platform/Website/admin/Skins/Terms.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Terms.ascx.cs
@@ -60,7 +60,7 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
 
                 this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
-                this.hypPrivacy.Attributes["rel"] = this.Rel;
+                this.hypTerms.Attributes["rel"] = this.Rel;
             }
             catch (Exception exc)
             {
--- a/DNN Platform/Website/admin/Skins/Terms.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Terms.ascx.cs
@@ -31,6 +31,8 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
+        public string Rel { get; set; }
+
         protected override void OnInit(EventArgs e)
         {
             base.OnInit(e);
@@ -58,8 +60,6 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
 
                 this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
-
-                this.hypTerms.Attributes["rel"] = "nofollow";
             }
             catch (Exception exc)
             {
--- a/DNN Platform/Website/admin/Skins/Terms.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Terms.ascx.cs
@@ -31,7 +31,7 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
-        public string Rel { get; set; } = "nofollow"
+        public string Rel { get; set; } = "nofollow";
 
         protected override void OnInit(EventArgs e)
         {
--- a/DNN Platform/Website/admin/Skins/Terms.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Terms.ascx.cs
@@ -60,6 +60,7 @@ namespace DotNetNuke.UI.Skins.Controls
                 }
 
                 this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
+                this.hypPrivacy.Attributes["rel"] = this.Rel;
             }
             catch (Exception exc)
             {
--- a/DNN Platform/Website/admin/Skins/Terms.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/Terms.ascx.cs
@@ -31,7 +31,7 @@ namespace DotNetNuke.UI.Skins.Controls
 
         public string CssClass { get; set; }
 
-        public string Rel { get; set; }
+        public string Rel { get; set; } = "nofollow"
 
         protected override void OnInit(EventArgs e)
         {
--- a/DNN Platform/Website/admin/Skins/privacy.ascx
+++ b/DNN Platform/Website/admin/Skins/privacy.ascx
@@ -1,2 +1,2 @@
 <%@ Control Language="C#" AutoEventWireup="false" Inherits="DotNetNuke.UI.Skins.Controls.Privacy" Codebehind="Privacy.ascx.cs" %>
-<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
+<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False" rel="nofollow"></asp:hyperlink>
--- a/DNN Platform/Website/admin/Skins/privacy.ascx
+++ b/DNN Platform/Website/admin/Skins/privacy.ascx
@@ -1,2 +1,2 @@
 <%@ Control Language="C#" AutoEventWireup="false" Inherits="DotNetNuke.UI.Skins.Controls.Privacy" Codebehind="Privacy.ascx.cs" %>
-<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False" rel="nofollow"></asp:hyperlink>
+<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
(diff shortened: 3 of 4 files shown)


valadas (changes requested):
Unless I am misunderstanding something, this will not make it configurable, it will always be rel="nofollow" this way...

valadas:
Also, we use the PR titles mainly to write the release notes, so I changed the title to explain what the PR does.

Andy9999:
> Unless I am misunderstanding something, this will not make it configurable, it will always be rel="nofollow" this way...

Yes, correct. This is the current behavior. The difference is that it could now be removed from the ascx file if desired, without recompiling.

Andy9999:
> Also, we use the PR titles mainly to write the release notes, so I changed the title to explain what the PR does.

OK. Thank you, will do it properly the next time.

valadas:
>Yes, correct. This is the current behavior. The difference is that it could now be removed from the ascx file if desired, without recompiling.

But it's not a best practice to edit core files anyway, I was thinking it could be made an option in a parameter like you did but it looks to me like the parameter is not being used this way.

Andy9999:
> > Yes, correct. This is the current behavior. The difference is that it could now be removed from the ascx file if desired, without recompiling.
> 
> But it's not a best practice to edit core files anyway, I was thinking it could be made an option in a parameter like you did but it looks to me like the parameter is not being used this way.

The reason why rel ended up in the cs file in the first place is because Microsoft's asp hyperlink doesn't have a rel attribute, so it cannot be set from the markup usually. Attempting to set it from the ascx produces an error. I found this workaround on stackexchange for example.

The (in my opinion) better alternative was to add the missing attribute to the class(es). It works. I tested it yesterday. With the change in place, it will still render the rel as nofollow. When removing it from the ascx t will disappear entirely. It looks a bit strange because the classes member Rel is not used anywhere, but it will still get rendered (magically).

DNN Platform/Website/admin/Skins/Privacy.ascx.cs
```diff
@@ -29,7 +29,9 @@ public Privacy()
 
         public string Text { get; set; }
 
-        public string CssClass { get; set; }
+        public string CssClass { get; set; }
+
+        public string Rel { get; set; }
```
bdukes, line 34:
```suggestion
        public string Rel { get; set; } = "nofollow"
```

DNN Platform/Website/admin/Skins/Privacy.ascx.cs
```diff
@@ -58,7 +60,6 @@ protected override void OnLoad(EventArgs e)
                 }
 
                 this.hypPrivacy.NavigateUrl = this.PortalSettings.PrivacyTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Privacy") : this._navigationManager.NavigateURL(this.PortalSettings.PrivacyTabId);
```
bdukes, line 62:
```suggestion
                this.hypPrivacy.NavigateUrl = this.PortalSettings.PrivacyTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Privacy") : this._navigationManager.NavigateURL(this.PortalSettings.PrivacyTabId);
                this.hypPrivacy.Attribute["rel"] = this.Rel;
```

DNN Platform/Website/admin/Skins/Terms.ascx.cs
```diff
@@ -30,6 +30,8 @@ public Terms()
         public string Text { get; set; }
 
         public string CssClass { get; set; }
+
+        public string Rel { get; set; }
```
bdukes, line 34:
```suggestion
        public string Rel { get; set; } = "nofollow"
```

DNN Platform/Website/admin/Skins/Terms.ascx.cs
```diff
@@ -58,8 +60,6 @@ protected override void OnLoad(EventArgs e)
                 }
 
                 this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
```
bdukes, line 62:
```suggestion
                this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
                this.hypTerms.Attributes["rel"] = this.Rel;
```

DNN Platform/Website/admin/Skins/privacy.ascx
```diff
@@ -1,2 +1,2 @@
 <%@ Control Language="C#" AutoEventWireup="false" Inherits="DotNetNuke.UI.Skins.Controls.Privacy" Codebehind="Privacy.ascx.cs" %>
-<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
\ No newline at end of file
+<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False" rel="nofollow"></asp:hyperlink>
```
bdukes, line 2:
```suggestion
<asp:hyperlink id="hypPrivacy" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
```

DNN Platform/Website/admin/Skins/terms.ascx
```diff
@@ -1,2 +1,2 @@
 <%@ Control Language="C#" AutoEventWireup="false" Inherits="DotNetNuke.UI.Skins.Controls.Terms" Codebehind="Terms.ascx.cs" %>
-<asp:hyperlink id="hypTerms" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
\ No newline at end of file
+<asp:hyperlink id="hypTerms" runat="server" cssclass="SkinObject" enableviewstate="False" rel="nofollow"></asp:hyperlink>
```
bdukes, line 2:
```suggestion
<asp:hyperlink id="hypTerms" runat="server" cssclass="SkinObject" enableviewstate="False"></asp:hyperlink>
```

bdukes (commented):
As @valadas said, our goal would be to allow theme authors to override this without making any changes to core files. I think this approach with my edits should work. Thanks!

valadas:
The way I see it, the login one has it hardcoded in the view and it cannot be customized (without editing the core files, which we do not recommend).

Andy9999:
OK. Interesting to see the devs perspective. Am more in the user perspective here. :)

I made a couple of such changes and merge them during each update. From my perspective this is totally acceptable. And of course I prefer changing some (core) markup files over having to recompile DNN.

Andy9999:
> > Rel is now initialized twice, once from the rel attribute in the HTML markup and once from the code behind file. But the markup will win
> 
> If you accept the suggestions @bdukes made above by clicking the "Commit suggestion" on each, then it will have a default and theme developers will have a setting on the skin object to override that default. The class will change it at runtime accordingly thus no longer needing to modify the core ascx file, it will be simply a theme implementation that does not get overridden upon upgrades.

Oh....wasn't aware of that.

Andy9999:
> @Andy9999 the intention of the Platform is that you should not have any changes which get overwritten by an upgrade. Any changes that you want which would be overwritten by an upgrade should use an extension point which DNN does not overwrite (or a new extension point if one doesn't exist, e.g. adding a `Rel` property to the `Privacy` theme object control)
> 
> ASP.NET web controls can have their `Attributes` collection set via the `ascx` markup if an attribute in the markup doesn't match an existing property on the control's class. So the `asp:HyperLink` control doesn't need any special support to set `rel="nofollow"`, it just renders it directly as an attribute on the rendered `<a>` tag. We want the `Privacy` control to allow a theme developer to set `<dnn:Privacy runat="server" rel=""/>` (which sets the `Rel` property defined on the class), which then sets the `"rel"` attribute on the `asp:HyperLink`, resulting in `rel=""` being rendered).

Thanks....maybe the exception I saw, when adding rel to the markup without that change, was because the code behind attempted to add an attribute that already exists in the markup/skin. I didn't investigate the crash any further.

Andy9999:
Seems I also missed out that Login.Ascx or User.Ascx use hyperlink directly whereas Privacy and Terms have their own classes.

Andy9999:
To sum it up and if I got your feedback right, you would prefer to have a setting somewhere, e.g. in PortalSettings to turn it on and off?

Then....no core files would have to be touched. Right?

DNN Platform/Website/admin/Skins/Privacy.ascx.cs
```diff
@@ -1,73 +1,74 @@
...
+                {
+                    this.hypPrivacy.Text = this.Text;
+                }
+                else
+                {
+                    this.hypPrivacy.Text = Localization.GetString("Privacy", Localization.GetResourceFile(this, MyFileName));
+                }
+
+                this.hypPrivacy.NavigateUrl = this.PortalSettings.PrivacyTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Privacy") : this._navigationManager.NavigateURL(this.PortalSettings.PrivacyTabId);
+            }
```
bdukes, line 63:
```suggestion
                this.hypPrivacy.Attributes["rel"] = this.Rel;
            }
```

DNN Platform/Website/admin/Skins/Terms.ascx.cs
```diff
@@ -1,74 +1,74 @@
...
+                {
+                    this.hypTerms.Text = this.Text;
+                }
+                else
+                {
+                    this.hypTerms.Text = Localization.GetString("Terms", Localization.GetResourceFile(this, MyFileName));
+                }
+
+                this.hypTerms.NavigateUrl = this.PortalSettings.TermsTabId == Null.NullInteger ? this._navigationManager.NavigateURL(this.PortalSettings.ActiveTab.TabID, "Terms") : this._navigationManager.NavigateURL(this.PortalSettings.TermsTabId);
+            }
```
bdukes, line 63:
```suggestion
                this.hypPrivacy.Attributes["rel"] = this.Rel;
            }
```

bdukes:
I don't think we would need anything added to `PortalSettings`. Wherever the privacy and terms controls are used (i.e. in the theme) is the place where `rel` could be set, which wouldn't require any core changes, just theme changes.

bdukes (approved):
This looks good to me now, thanks!

Andy9999:
Shall I add the semicolons?

valadas (approved):
Awesome, thanks guys, this now also looks good to me, merging.

Andy9999:
I love to contribute by adding semicolons. ;) That's easy and I do it the whole day anyways.

valadas:
@bdukes I am a bit confused about the build error, how do we find the problem nowadays, is the only way with the binlog with the viewer ?

Andy9999:
My last update (semicolons) doesn't have the Green verified thingy [https://github.com/dnnsoftware/Dnn.Platform/pull/5054/commits](https://github.com/dnnsoftware/Dnn.Platform/pull/5054/commits)

bdukes:
@valadas 😕it should be posting build errors as errors in DevOps, but apparently that isn't working, so viewing the binlog is the only way to get more details

bdukes:
Sorry, my typo. Fix building now.

Andy9999:
Very agile!

Andy9999:
Would probably be easier if we were in the same room or in a chat somehow.

bdukes:
If you really don't want `rel=""` you can set the value to `null`:

```
<dnn:Terms runat="server" ID="TheTerms" />

<script runat="server">
protected override void OnLoad(EventArgs e)
{
  TheTerms.Rel = null;
}
</script>
```

Andy9999:
I think the whole change could now be undone, because that was possible before. :)

bdukes:
We can add a `if (!string.IsNullOrWhiteSpace(this.Rel) { this.hypTerms["rel"] = this.Rel; }` wrapper around the defining of the attribute if you want to make a follow-up contribution.

It is now possible to change or remove the `rel="nofollow"` attribute from the link without making any changes to the core files, but only adjusting your theme. This is a valuable contribution, otherwise we wouldn't have accepted it. I'm sorry if there was miscommunication on what our goals were.

Andy9999:
> We can add a `if (!string.IsNullOrWhiteSpace(this.Rel) { this.hypTerms["rel"] = this.Rel; }` wrapper around the defining of the attribute if you want to make a follow-up contribution.
> 
> It is now possible to change or remove the `rel="nofollow"` attribute from the link without making any changes to the core files, but only adjusting your theme. This is a valuable contribution, otherwise we wouldn't have accepted it. I'm sorry if there was miscommunication on what our goals were.

I guess this won't help due to the order of execution.

1) Rel is initialized to nofollow
2) OnLoad is execute and add the attribute to the hyperlink (nofollow)
3) Then Rel is set to whatever is in the markup code without any effect

bdukes:
Are you trying to set the `Rel` property on the `<dnn:Terms>` control in your theme file?

Andy9999:
No, maybe that's the culprit! I did it in the Terms.Ascx file, the file you told me to not change. :)

Andy9999:
I slowly start to understand what we are doing here.

Andy9999:
Wow! Works perfectly if including your last suggestion. Now it matches the issue title again. Can I still submit it or is it too late now?

Andy9999:
Just submitted the change and will change the skin files. The way it should be done! Thanks Daniel and Brian.

Andy9999:
New issue??

bdukes:
No, you can associate it with the same issue. Thanks!

bdukes:
You will have to open a new PR, but it'll be associated to the same issue #5047

Andy9999:
Like this? https://github.com/dnnsoftware/Dnn.Platform/pull/5057

bdukes:
That works 👍🏻

The one thing to do differently is instead of include #5047 in the PR title, in the description include "Fixes #5047" (or "For #5047" if it's not a complete fix). This will let GitHub automatically create a link between the PR and the issue.

Andy9999:
Thanks.

Andy9999:
Got it deployed on the production server now along with the other issue (search result category localization). Looks great now! I find nothing to complain about.

Somehow (with my limited understanding) it did not seem to be a big difference if rel would be changed in the Terms/Privacy markup files or the skin. I have to watch out when updating the skin as well and merge changes with the ones we did. We use the awesome EasyDNN themes. But I clearly see the advantage now. :)

Andy9999:
One final question.....would it make a big difference to initialize the new Rel attribute from the markup instead of having it in the code behind file? And...would this raise more awareness for the skin developers that there is now such an attribute that could be customized? Or would they consult the code behind anyways or get that info from the release notes?

Just a question for future issue of similar nature.

bdukes:
`<dnn:Terms runat="server" ID="TheTerms" Rel="" />` is the way you would initialize the property via markup. There isn't a place to do that initialization in the Terms control itself, it has to be in the codebehind.

We probably do need to document this new property by creating a page on https://docs.dnncommunity.org/content/tutorials/themes/theme-objects/index.html#breadcrumb-theme-object-example (e.g. like [the breadcrumb docs](https://docs.dnncommunity.org/content/tutorials/themes/theme-objects/breadcrumb/index.html))

bdukes:
Doc comments are always appreciated :-)

valadas:
@Andy9999  documentation about theme objects is also welcome if you have the time https://docs.dnncommunity.org/content/tutorials/themes/theme-objects/index.html

PR #5056, opened 2022-03-21
Fix not reporting errors or warnings from build

bdukes:
## Summary
Due to a bug I introduced in #5034 we stopped reporting any build errors or warnings from the build process (_only_ errors from the clean before the build were reported)


valadas (approved):
Awesome

valadas:

PR #5066, opened 2022-03-29
Fixes relative path issues in loading monaco editor

donker:
This has bugged me for a while. But on some of my sites I was unable to use the SQL Console. This is because the scripts are downloaded using an absolute path from the root of the domain. For virtual roots, this will mean those scripts do not load and the module will break. This PR fixes aims to fix that.


bdukes (approved):
I've verified that both editors load as expected when running in a subfolder

mitchelsellers (approved):

david-poindexter (commented):
Great catch @donker

PR #5074, opened 2022-04-04
Updated esprima checksum

daguiler:
I just built a clean develop branch and got the following yarn error:
```
➤ YN0018: │ esprima@https://github.com/substack/esprima.git#commit=0a7f8489a11b44b019ce168506f535f22d0be290: The remote archive doesn't match the expected checksum
```
I had to run the following to make it work
```
YARN_CHECKSUM_BEHAVIOR=update yarn
```
which essentially updated the checksum in the yarn.lock file.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10580,7 +10580,7 @@ __metadata:
   bin:
     esparse: ./bin/esparse.js
     esvalidate: ./bin/esvalidate.js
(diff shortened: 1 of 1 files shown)


valadas:
Looks like it did not like it...

daguiler:
Yeah :-) trying to figure out what went wrong

daguiler:
I don't know what's wrong. When I build it locally, I need the updated checksum or otherwise it fails for me. But then the CI pipeline doesn't like the new checksum. It seems to be the same issue reported here: https://github.com/yarnpkg/berry/issues/1142

Anyway, the build works for me locally with the updated checksum, so I'll go ahead and close this PR. If the issue surfaces again we can take a deeper look.

PR #5077, opened 2022-04-06
Update the existing check to report on the status of Telerik

### LINKED ISSUE #4947: Update the existing check to report on the status of Telerik.- Fail if Telerik Installed & Used - Give warning about 10.x upgrade issue- Fail if Telerik Installed & Not-Used - Give message about the removal- Pass if Telerik Not Installed

daguiler:
`Closes #4947`

## Summary
This pull request replaces the existing `CheckTelerikVulnerability` security check with the new `CheckTelerikPresence` security check. As defined in #4947, the new check will:
* PASS if Telerik is not installed
* FAIL if Telerik is installed
  * if installed but unused, the following message will be returned:
    > Telerik was found in this site, but no dependencies on Telerik were discovered. You should have a reasonable expectation of being safe to uninstall Telerik per the 9.8.0 instructions.
  * if installed and used, the following message will be returned, followed by the list of dependent assemblies:
    > Dependencies on Telerik were discovered in assemblies that will NOT be addressed by un-installing per the instructions with 9.8.0.

Wording and logic for Telerik dependency detection adapted from the [Dnn Telerik Identifier](https://github.com/IowaComputerGurus/DnnTelerikIdentifier).


daguiler:
I tested the following scenario:

1) Installed a vanilla site including these changes.
2) Verified that the security check FAILED, showing the "installed but unused" message.
3) Installed the [DNN.Events module](https://github.com/DNNCommunity/DNN.Events/releases/tag/07.00.07)
4) Verified that the security check still FAILED, but returned the "installed and used" message, showing only `dotnetnuke.modules.events.dll` in the list.
5) Uninstalled DNN.Events, and [all built-in Telerik dependencies](https://docs.dnncommunity.org/content/getting-started/setup/telerik-removal/index.html).
6) Verified that this time the security check PASSED.

mitchelsellers:
In my Telerik Identifier project - https://github.com/IowaComputerGurus/DnnTelerikIdentifier/blob/main/src/IowaComputerGurus.Dnn.TelerikIdentifier/Controllers/ContentController.cs I had to wrap the call to get referenced assemblies in a try-catch because of certain third-party modules that deployed native compiled dll's that couldn't be loaded by .NET for inspection.  

Do we believe we should do something similar here?

One example of those exceptions from a bug report we had

>Microsoft.DiaSymReader.Native.amd64.dll (Could not load file or assembly '1495800 bytes loaded from IowaComputerGurus.Dnn.TelerikIdentifier, Version=1.0.0.38009, Culture=neutral, PublicKeyToken=null' or one of its dependencies. An attempt was made to load a program with an incorrect format.)

valadas (approved):
I _think_ @mitchelsellers is talking about this pr right? https://github.com/IowaComputerGurus/DnnTelerikIdentifier/pull/1/files

Other than that, it looks awesome, thanks a lot for that contribution. Do you guys think we should add some note about Telerik being forcibly removed in a future version?

Other than @mitchelsellers concern, I am approving this.

daguiler:
Yes, it was a concious decision not to catch the exceptions because I noticed that exceptions are caught by the AuditChecks class, [here](https://github.com/dnnsoftware/Dnn.Platform/blob/68ff1caf44a5c0f37237152b8845bc27941fb6bc/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Security/AuditChecks.cs#L58). It seems exceptions that occur in the security checks are shown in the Notes panel.

But I agree, maybe it's better not to show these raw exception errors and wrap the code in a try...catch to compose a more user friendly error message and avoid disclosing server physical paths and cluttering the UI with technical details.
I'll do this tomorrow.

mitchelsellers:
Given the high percentage of DNN users that will see this, I think it should be swallowed and only noted.  (At least 3 popular third-party solutions include these types of assemblies.)

But as @valadas said, this is awesome otherwise

daguiler:
Ok, with the last set of commits I addressed the following:
* Improved error handling by logging exceptions internally and sending a generic message to the UI.
* Added a note about Telerik being forcibly removed in 10.x (the security check "reason")
* Added unit tests
* Updated the security check name

Here's a screenshot showing all of the above:

image

I think this is now ready for review.

bdukes (approved):

bdukes:
We want this to target the 9.11.0 branch, right?

valadas:
@bdukes I don't see a reason this could not be in a possible 9.10.3, do you have any concerns about it ?

bdukes:
It's not a bug fix and it's part of the larger effort for 9.11. No specific concerns

valadas:
Cool, works for me

valadas:
Oh wait it was tagged 9.11 but still targetted develop when I merged, sorry...
Hugh, is it work reverting and retargetting or we just let it go ?

bdukes:
I wouldn't worry about reverting

valadas:
Ok, I just changed the milestone, sorry for the confusion...

PR #5080, opened 2022-04-10
Add new Security Analyzer check for permissions on "Activity Feed" and "User Profile" pages

### LINKED ISSUE #4948: Add new Security Analyzer check for permissions on "Activity Feed" and "User Profile" pages
Add a new check for permissions on the "Activity Feed" and "User Profile" pages.- If User Registration = None or Private and Either of these pages are visible to all users, warn that user profile information is displayed publicly- For all other user registration values, show an informational message if user profiles are public

daguiler:
- Closes #4948

## Summary
This pull request adds this new `CheckUserProfilePage` class to check the visibility of the user profile page defined in Site Settings > Site Behavior > Default Pages > User Profile Page, according to the following criteria:

### Case 1: the selected user profile page is "Activity Feed"
* If the Activity Feed page cannot be found, or is deleted, it returns PASS.
* Otherwise, if the Activity Feed page is public, it returns ALERT.
* Otherwise, if My Profile cannot be found or is deleted, it returns PASS.
* Otherwise, if My Profile is public, it returns ALERT.
* Otherwise, it returns PASS.

### Case 2: the selected user profile page is not "Activity Feed"
* If the selected user profile page cannot be found, or is deleted, it returns PASS.
* Otherwise, if the selected user profile page is public, it returns ALERT.
* Otherwise, it returns PASS.

image

Also, a new base class for audit checks is introduced, and some refactoring to the existing `CheckTelerikPresence` class is done in order to avoid redundancy.

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Security/AuditChecks.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Security/AuditChecks.cs
@@ -36,6 +36,7 @@ namespace Dnn.PersonaBar.Security.Components
                 new CheckAllowableFileExtensions(),
                 new CheckHiddenSystemFiles(),
                 new CheckTelerikPresence(applicationStatusInfo),
+                new CheckUserProfilePage(),
             };
 
             if (Globals.NETFrameworkVersion <= new Version(4, 5, 1))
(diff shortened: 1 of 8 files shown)


mitchelsellers (approved):
This looks great to me!  I also updated the description so that the issue was linked for auto-close.

valadas (approved):
Awesome work, thanks a lot, looks good to me.

PR #5082, opened 2022-04-10
Removed two unused utility methods that depended on `momentjs`

david-poindexter:
## Summary
These two utility methods are unused and they depended on `momentjs`.  Instead of swapping out for `dayjs` I felt it best to remove the methods instead.  

This PR relates to #3875

### DIFF
--- a/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/editBar/scripts/util.js
+++ b/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/editBar/scripts/util.js
@@ -118,18 +118,6 @@ define(['jquery'], function ($) {
                     return trimmed;
                 },
 
-                deserializeCustomDate: function (str) {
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me, did not find anything calling that.

mitchelsellers (approved):
Approving, as internal usage isn't there.

Do we have any thoughts that external folks would have used this feature?  IE is this breaking

bdukes (approved):
I'm okay with removing it

david-poindexter:
> Approving, as internal usage isn't there.
> 
> Do we have any thoughts that external folks would have used this feature? IE is this breaking

I would be shocked if there is any external use of these rogue functions.

valadas:
I would be extremely surprised if any 3rd party even knows those functions are there + it's not a package or a library, I consider this internal code myself.

Issue #5038, opened 2022-03-08
Links to Portal File or Other URL are being Rendered as Absolute URLs when SSL is Enabled on the site.


jhaggerty-dnn:
## Description of bug
After adding an SSL to the DNN Site and when creating a hyperlink using the default HTML Editor the relative hyperlink is showing up as an absolute link. 

## Steps to reproduce
List the precise steps to reproduce the bug:
1. Install a SSL Certificate on the site and ensure your connected using "https."
2. Sign into DNN as Superuser and open the html editor
3.  Select text to be hyperlinked and click the hyperlink/link button
4.  Choose link type as "Portal file or other URL" and click the "Browse Server Button"
5. Select the "Page Link" Tab, then select another page within the site.  
6. Look at the "Choose URL Type" and you will see that both the Relative and Absolute URL have the Absolute URL.

## Current behavior
Links to internal site pages are being added as Absolute Links when editing over https but works fine over http.

## Expected behavior
Links to internal site pages will be added as Relative Links when editing over http and https.

## Screenshots
![NoSSL_Relative_Works](https://user-images.githubusercontent.com/58789910/157157237-2da4e128-798f-4f54-9ef7-044526ce896e.PNG)

![WithSSL_Relative_broken](https://user-images.githubusercontent.com/58789910/157157264-02ca2883-12b7-44c3-ab54-09262c5202bb.PNG)

## Error information
No Errors

## Additional context
Provide any additional context that may be helpful in understanding and/or resolving the bug.

## Affected version

* [ ] 10.00.00 alpha build
* [x ] 09.10.02 latest supported release

## Affected browser

* [x ] Chrome
* [ x] Firefox
* [ x] Safari
* [x ] Internet Explorer 11
* [x ] Microsoft Edge (Classic)
* [x ] Microsoft Edge Chromium

PR #5086, opened 2022-04-12
Ensure FileServerHandler returns an absolute URL which takes into account the SSL Offload.

### LINKED ISSUE #5085: Link click for files doesn't take into account the SSL Offload.
## Description of bug
When requesting a link to a file through the LinkClick/FileServerHandler a relative URL is returned, which doesn't take into account the   SSL offload.

## Steps to reproduce
1. Request a file link through LinkClick.
2. Open browser console. Click on the link.
3. File will not open automatically.
4. In the console you will notice a mixed content error. 

## Current behavior
Currently, the FileServerHandler simply returns a relative URL, which in some cases might not be enough in order for the browser to properly determine the protocol (HTTP/HTTPS).

## Expected behavior
FileServerHandler should return an absolute URL, while taking into account the SSL Offload.

## Screenshots

## Error information

## Additional context

## Affected version

* [ ] 10.00.00 alpha build

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

rhaiamz:
…

Fixes #5085

### DIFF
--- a/DNN Platform/Library/Services/FileSystem/FileServerHandler.cs
+++ b/DNN Platform/Library/Services/FileSystem/FileServerHandler.cs
@@ -187,7 +187,9 @@ namespace DotNetNuke.Services.FileSystem
                                     }
                                     else
                                     {
-                                        context.Response.Redirect(directUrl, /*endResponse*/ true);
+                                        var appUrl = new UriBuilder(Globals.AddHTTP(_portalSettings.PortalAlias.HTTPAlias));
+                                        appUrl.Path = appUrl.Path.TrimEnd('/').Append(directUrl, string.Empty);
+                                        context.Response.Redirect(appUrl.Uri.AbsoluteUri, /*endResponse*/ true);
                                     }
                                 }
                                 catch (PermissionsNotMetException)


mitchelsellers (commented):
Can we get confirmation that this doesn't break file downloads for those with Azure etc?  I believe we will need to be testing the result of the direct url for http or https as we shouldn't assume the path is always relative.

mitchelsellers:
Looking at the upstream implementation I believe that this fix should NOT be implemented here.  Upstream the individual provider is called (https://github.com/dnnsoftware/Dnn.Platform/blob/1b6eaacdf6a1d3b6447c99394de3d05dd7c4547c/DNN%20Platform/Library/Services/FileSystem/FileManager.cs#L723)

This means that the "directUrl" that is shown here might not always be a relative path.  The AzureFolderProvider for example is a https:// path, this change would break that provider.

mtrutledge:
This should be done at the Folder Provider level like @mitchelsellers says.

rhaiamz:
I made some changes to the code to ensure that only relative URLs get switched.
While I do understand the concerns for the initial commit, moving the logic upstream might introduce more confusion such as #5038, so in this case, I think leaving the logic at the FileServerHandler level is the best choice since this will leave absolute URLs alone.

mitchelsellers:
@dnnsoftware/approvers I'm still in the belief that we should NOT be touching URL's that could come from providers that we don't know.  The risk of breaking something isn't right

rhaiamz:
@mitchelsellers  Was there any discussion with the approvers regarding this PR? Any updates?

mitchelsellers:
@rhaiamz My apologies for not getting the update on this ticket as I should have, shortly after this comment I was out of the office for about 4 weeks and this didn't make it back to the top of the list.

In our discussion, given the nature of the provider model, this change really NEEDS to be implemented at the provider level, not at this level as we cannot guarantee that we would not be manipulating things that should not be manipulated.

This change, if within the implementation for the standard folder provider, would be acceptable.

mitchelsellers:
Closing this PR for now, this can be re-submitted with the change targeting the proper provider implementation.

PR #5087, opened 2022-04-12
Allow the underlying FileInstaller to delete the file if needed.

### LINKED ISSUE #5025: AssemblyInstaller does not delete the file when UnRegister is specified.
## Description of bug
When using the UnRegister feature of the Assembly component to remove a module registered assembly it gets removed from database and web.config (binding), but it is not removed from disk. The issue looks to be in the [AssemblyInstaller.InstallFile](https://github.com/dnnsoftware/Dnn.Platform/blob/master/DNN%20Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs#L134) which calls [AssemblyInstallerI.DeleteFile()](https://github.com/dnnsoftware/Dnn.Platform/blob/master/DNN%20Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs#L98) which in turn calls [base.DeleteFile()](https://github.com/dnnsoftware/Dnn.Platform/blob/4141ee8b9f9d9707b02fda0bc6005ef521832702/DNN%20Platform/Library/Services/Installer/Installers/FileInstaller.cs#L279). Here, before deleting the file, the `this.DeleteFiles` boolean property is checked which is never set when in the installation process.

One thing that prevents a quick/clean fix for this is that the `DeleteFiles` property is at the File/AssemblyInstaller level while the UnRegister flag is at the InstallFile level, so to solve this we need to have some sort of try-finally block that sets and then reverts the `DeleteFiles` property with each assembly file that gets evaluated based on it's UnRegister flag.

## Steps to reproduce
1. Create a new version of a module which has in the manifest a new assembly record that was just marked for removal with the UnRegister flag. Make sure that the module is the only one / the last one that retains that assembly.
2. Update the module on an existing instance.
3. Check bin\ and see that the dll file is still present.

## Current behavior
The assembly file is still present.

## Expected behavior
The assembly file should be deleted when it is not retained by any other module.

## Affected version

* [ ] 10.00.00 alpha build

## Affected browser
_Not applicable._
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

dimarobert:
Fixes #5025.

## Summary

Ensures that the `FileInstaller.DeleteFiles` flag is set to `true` when UnRegister-ing an Assembly file in the AssemblyInstaller.

### DIFF
--- a/DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
+++ b/DNN Platform/Library/Services/Installer/Installers/AssemblyInstaller.cs
@@ -136,7 +136,16 @@ namespace DotNetNuke.Services.Installer.Installers
             bool bSuccess = true;
             if (file.Action == "UnRegister")
             {
-                this.DeleteFile(file);
+                var prevDeleteFiles = this.DeleteFiles;
+                try
+                {
+                    this.DeleteFiles = true;
+                    this.DeleteFile(file);
+                }
+                finally
+                {
+                    this.DeleteFiles = prevDeleteFiles;
+                }
             }
             else
             {


donker (approved):
I'm OK with this.

mitchelsellers (approved):
This fixes a feature that appears to have been non-functional for a long time.  We did discuss in the Approvers meeting today that there is a potential for a module that wasn't good with assembly registration to be broken by this fix.

PR #5090, opened 2022-04-16
Add folder cleanup support to CleanupInstaller

daguiler:
## Summary
The `CleanupInstaller` class supports a `glob` parameter that works well, but leaves an empty folder structure behind after deleting all files.
This pull request adds support for a new `<folder>` element in the package schema that allows for recursive folder deletion.

Usage:
```xml
<dotnetnuke type="Package" version="5.0">
    <packages>
        <package name="..." type="Provider" version="02.00.00">
            <friendlyName>...</friendlyName>
            ...
            <components>
                ...
                <component type="Cleanup">
                    <files>
                        ...
                    </files>
                    <folder>
                        <path>Providers/HtmlEditorProviders/MyProvider/1.0.0</path>
                    </folder>
                    <folder>
                        <path>Providers/HtmlEditorProviders/MyProvider/1.0.1</path>
                    </folder>
                    <folder>
                        <path>Providers/HtmlEditorProviders/MyProvider/1.0.2</path>
                    </folder>
                    ...
                </component>
            </components>
        </package>
    </packages>
</dotnetnuke>
```

The value of `path` is validated so that it can only reference valid paths inside the virtual directory.


DNN Platform/DotNetNuke.Abstractions/Utilities/IFileSystemUtils.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Abstractions.Utilities
+{
+    using System;
+
+    /// <summary>
+    /// File system utility.
+    /// </summary>
+    public interface IFileSystemUtils
```
mitchelsellers, line 12: This creates a new public interface for the platform, is this in alignment with how we want to expose file-system API's?  

Part of me believes for the internal, that this should be an internal element to the insaller?

DNN Platform/Library/Services/Installer/Installers/CleanupInstaller.cs
```diff
@@ -102,6 +152,47 @@ public override void UnInstall()
...
+                normalized = Path.GetFullPath(path);
+            }
+            catch
+            {
+                return false;
+            }
+
+            var appPath = this.applicationStatusInfo.ApplicationMapPath;
+            var fullPath = Path.Combine(appPath, normalized);
+            if (!normalized.StartsWith(appPath, StringComparison.InvariantCultureIgnoreCase) || fullPath == appPath)
```
mitchelsellers, line 188: I have a feeling we need to be more protective with this validation test as it would be very easy to corrupt an installation with this.

Values such as

* /DesktopModules
* /Portals/0
* /Bin

Or otherwise, and since this is a recursive delete, that doesn't expect folders to be empty I think we need to prevent at least our critical directories from being cleaned?

DNN Platform/Tests/DotNetNuke.Tests.Core/Services/Installer/CleanupInstallerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        [TestCase(@"\")]
+        [TestCase(@"\\")]
+        [TestCase(@"\\fileshare")]
+        [TestCase(@"C:/")]
+        [TestCase(@"D:/inetpub/wwwroot")]
+        [TestCase(@"/")]
+        [TestCase(@"//")]
+        [TestCase(@"//fileshare")]
+        [TestCase(@"file://C:/Windows")]
+        [TestCase(@"http://dnndev.me")]
```
mitchelsellers, line 35: We should also test against path traversal here `..\` for example

mitchelsellers (changes requested):

DNN Platform/Tests/DotNetNuke.Tests.Core/Services/Installer/CleanupInstallerTests.cs
daguiler: Thanks. Fixed.
After I added this test I discovered some holes in the implementation, so I refactored the unit test class to multiply the test cases more easily.
The case of ".." is interesting because even though with the previous implementation it was not possible to go up in the hierarchy, it was allowing relative paths like "../wwwroot/dir". While this path could fall within the app root path, it discloses the folder structure outside the app. So now I am disallowing ".." entirely.

DNN Platform/DotNetNuke.Abstractions/Utilities/IFileSystemUtils.cs
daguiler: Agreed. I had added this abstraction just for the sake of unit testing, so I could have made it internal.
I have re-implemented this now, and I found some existing internal classes and interfaces in `DNN Platform\Library\Common\Internal`, so I moved this interface there.

DNN Platform/Library/Services/Installer/Installers/CleanupInstaller.cs
daguiler: Yes, good point, although critical files could still get deleted using the `glob` parameter.
Anyway, I have changed the original indiscriminate delete now with a new method that only deletes _empty_ directories. This avoids the risk of any files being deleted by CleanupInstaller.

mitchelsellers (approved):

valadas (approved):
Awesome work, thanks!

PR #5099, opened 2022-04-25
Update Upgrade Process to remove Telerik (WIP)

### LINKED ISSUE #4888: Update Upgrade Process to remove Telerik
The process to remove Telerik during DNN 9.11.0 upgrade should include the following actions

* Install Resource Manager
* Update all Digital Asset Manager modules to be Resource Manager modules
* Uninstall DigitalAssetsManagement (Digital Asset Management)
* Uninstall DotNetNuke.Web.Deprecated (DNN Deprecated Web Controls Library)
* Uninstall DotNetNuke.Website.Deprecated (DNN Deprecated Website Codebehind files)
* Uninstall DotNetNuke.Telerik.Web (DotNetNuke Telerik Web Components)
* Uninstall DNNSecurityHotFix20171 (DNN Security HotFix 2017.1)
* Remove SiteUrls.config entries for Telerik
* Remove packages from App_Data\ExtensionPackages for all above uninstalled packages

daguiler:
Closes #4888

## Summary
This PR adds a new module to uninstall the Telerik components.

This module can operate unattended, as part of the upgrade process if the user chooses to remove Telerik in the Security tab of the Upgrade Wizard, but it can also run interactively, if dropped on a page.

The module is installed during upgrade even if the user chooses not to remove Telerik. In this case, the user can drop the module on a page later on, and run it interactively.

The module removes Telerik by executing the steps documented in the [removal process](https://docs.dnncommunity.org/content/getting-started/setup/telerik-removal/index.html).

I'll paste some screenshots below in the comments.


daguiler:
The wording is the same as in the previous PR.

If Telerik is Installed but not used, you get this:

image

You need to accept the backup checkbox, and then the button gets enabled:

image

Removal report:

image

Removal report if previously executed:

image

daguiler:
> Is there a plan for integrating this into the upgrade process?

Yes, that was my intention, but I don't really know how to do that.
I have a commit already to make this a proper extension (add a manifest, add the Module.build target, license, etc.) and I tested it and it works fine, except the extension is left there as an "available" extension, meaning it doesn't get installed automatically. And also I wouldn't know how to "run" an extension automatically as part of the upgrade process, if that makes any sense and is at all possible. Maybe we can just ship this as an available extension and let users deploy it to a test page and run it manually?

bdukes:
If you add an [`eventMessage` component](https://github.com/dnnsoftware/Dnn.Platform/blob/37a767c6cf45f6c3813b225793e35875455f9989/DNN%20Platform/Providers/FolderProviders/FolderProviders.dnn#L72-L79) to the manifest and [implement `IUpgradeable`](https://github.com/dnnsoftware/Dnn.Platform/blob/37a767c6cf45f6c3813b225793e35875455f9989/DNN%20Platform/Providers/FolderProviders/Components/BusinessController.cs#L20-L35) you can run logic when the extension is installed. If that's the way we want to go, I would think we'd need some kind of setting that the `IUpgradeable` implementation would check so that it only runs the uninstall logic if it's an upgrade (and not if someone manually installs the module).

Alternatively, take the uninstall logic, move it into the installer, and run it directly as an install step, instead of inside of a separate module.

bdukes:
Yeah, I was thinking `Status` would probably work :+1:

daguiler:
image

daguiler:
image

daguiler:
All steps completed:

image

Then...

image

And then re-visiting the page...

image

daguiler:
So this is still a work in progress because the removal logic is not yet integrated to the upgrade process, but I think it's ready to get some test and feedback.

daguiler:
image

daguiler:
Hi @bdukes @valadas @david-poindexter @mitchelsellers 

I was able to implement `IUpgradeable` and hook the Telerik removal logic to the upgrade process.

As explained in code, this process is only triggered if the `telerikUninstallOption` host setting is present and has a value of `Y`. And this value is only set in the Security tab of the Upgrade wizard, so this ensures that the process will not get triggered in any other circumstance, like the Install Wizard or a manual installation of this extension.

One thing to note is that the extension is installed during the upgrade process, even if the user chooses _not_ to remove Telerik. This is not necessarily a bad thing IMO. The user may choose to _manually_ run the extension at a later time. Maybe we can mention this possibility in the Security tab.

Ok, please give it a try and let me know if you encounter any issues.

Thanks!

daguiler:
Hi @EPTamminga 

Telerik presence detection works by searching for `Telerik.Web.UI.dll` in the bin folder. Simple as that.
If found, we then look for other assemblies that reference this DLL.
If found, and such assemblies are not our own, then we say "we found Telerik in your system, but these other components depend on it, so we can't automatically uninstall Telerik for you".

Where I think we are falling a bit short is in that we are treating all Telerik version as equally risky, which might not be correct.

EPTamminga:
@daguiler I really am fan of removing (the DNN version of) Telerik and w have done that for several sites that we have upgraded, using the instructions in the doc.

However: we also have some sites for which we use a commercial updated version of Telerik, which is perfectly safe to do. Any warning that we have an insecure site would be incorrect (and problematic to some site managers).

Is there an option to check the Version of the Telerik DLL so that the messages of the removal module can be more specific?

mitchelsellers:
@EPTamminga and @valadas 

I think it is a bit of a slippery slope as even many sites that have “patched” Telerik didn’t actually go far enough forward to truly fix it. 

One thought is that we would catch these sites by having additional module references, we could check the version number against ours and modify the message. 

“Telerik version _____, was detected. This version is greater than the DNN distributed version of ______.  You are responsible for all security compliance.”

The problem with this, is that it leaves at last 4 holes in DNN. As removal of the Digital Asset Manager is mandatory for other reasons. So it doesn’t fully meet our security principles.

mitchelsellers:
@EPTamminga Ok, in your scenario, I think it would be safe to say that from a DNN perspective you are "ok".

I cannot divulge much more adhering to our security policy in this channel, but it is safe to assume that ANY DNN module that has a dependency on Telerik must be removed, DNN 9.8.0 and later that is ONLY the digital assets manager module.  So if you have removed DAM you are good.

As a side note, Telerik has had a LOT of security notices in the past 2-3 years so if you haven't updated things in quite a white I'd strongly suggest reviewing the notices.

@dnnsoftware/approvers 

For the scenario that @EPTamminga is mentioning above, do we think my proposed language might help?  Maybe add a note about the removal of Digital Asset Management regardless?

EPTamminga:
@mitchelsellers Thank you for your clarification, no need for more details on security issues here.

As a side note: the replacement for DAM, ie the Resource Manager, has its own merits, but has a very limited usability on site with a lot of folders and files (yes, I know that that is a known fact). We use the Evotiva UserFiles as the replacement for DAM in these cases.

mitchelsellers:
@EPTamminga There is active development to resolve those issues

EPTamminga:
> @EPTamminga There is active development to resolve those issues

Yes, I know.

david-poindexter:
I am a bit torn on this one as there are so many use cases with no clear cut answer to address all scenarios.  I do think at the end of the day though that we have to protect DNN Platform at the core first.  If we can help to alleviate pain for third party extensions then by all means we should.  However, in situations like these where there is no clear delineation for what is a truly "acceptable" use of the Telerik DLL (at least that is within our DNN Platform purview/knowledge), then we need to err on the side of caution.  We could perhaps place programmatic trust in the version, as @mitchelsellers alluded too, but this has a different set of risks, so I am not sure that is "the" answer.

At this point, I am leaning towards removal regardless but with some carefully worded messaging to potentially help third party extension developers understand ramifications.  I just wouldn't want to promote opting out for obvious reasons so wording here is crucial.

Thoughts?

daguiler:
I just want to say that getting a red cross on each build because of the missing label is demoralizing 😃

daguiler:
oh it's not the label now, it's because it's on hold

david-poindexter:
> I just want to say that getting a red cross on each build because of the missing label is demoralizing 😃

Yeah, it was put `On Hold` to prevent accidental merging prior to the resolution of a chosen path.  Thanks for the patience.

EPTamminga:
@david-poindexter: My opinion is that DNNPlatfporm should take care of its own problems and not do anything with code of others.

So: if the Telerik DLL is the DNNPlatform version, than DNNPlatform should do what is must do. I

f the Telerik DLL is NOT the DNNPlatform version: create a warning but don't mess with the installation, that is not the responsibility of the platform.

Compare this with other types of extensions. If other (none platform) extensions have potential security issues because of the use of custom code or external (none platform) libraries, DNNPlatform is not responsible. It would be wrong if the platform starts messing with other peoples code.

david-poindexter:
@EPTamminga there is unfortunately no solid programmatic way to differentiate the two.  Comparing versions, as @mitchelsellers mentioned, has significant ramifications.

EPTamminga:
@david-poindexter The message that Mitch proposed, would be a workable solution, although the second part of the message (the warning) can be applied to any kind of none-Platform extension.

mitchelsellers:
The reason for the additional note @EPTamminga is that if you leave DNN stuff that is dependent upon Telerik (Digital Asset Management) you are leaving issues on the table that aren't just fixed by having a new Telerik version.

At the end of the day, due to how Telerik was initially included & supported, the usage of Telerik in DNN is riddled with some real negative connotations that are hard to workaround.

Simply doing nothing, MAY be ok, but it might not be, and I don't know that we can properly outline exactly all of those situations.  So, the more I think about this, I think we might need to get a bit more verbose and implement the following.

**If Nothing is found to use Telerik:**

Do as we are now, suggest the removal

**If Telerik is found, regardless of version**

Check to see if the "Digital Asset Management" is installed, of so, show the following message

> Your installation is utilizing Telerik, in the following assemblies, including the Digital Asset Manager module.  There are known security issues with the DNN Distributed version of Telerik and the usage of Telerik by DNN modules, such as Digital Asset Management. If you have otherwise complied with Telerik security patching via other means, you should at a minimum uninstall Digital Asset Manager and replace it with the newer Resource Manager module, however, if you have not made any security adjustments the modules listed above should be updated/replaced and then Telerik removed to ensure proper security of your installation.

If "Digital Asset Management" is not found

> Your installation is utilizing Telerik in the following assemblies, it does appear that you have removed a portion of the DNN Platform extensions that depended on Telerik.  We are unable to remove Telerik for you, however, please ensure that you have in fact properly patched Telerik and removed all DNN Platform distributed Telerik components per the guidance found here https://docs.dnncommunity.org/content/getting-started/setup/telerik-removal/index.html 

I don't necessarily like this, but its about the safest way we can provide true guidance?  @dnnsoftware/approvers any thoughts?

EPTamminga:
@mitchelsellers Well said. This gives a proper explanation.

daguiler:
I have made updates per the comments above.
Plus, I added the found Telerik version number and made tiny wording adjustments. By all means, I can change this back if needed.

It looks like this now:

Telerik installed and used, and DAM present:

image

Telerik installed and used, DAM not present:

image

daguiler:
Updated the Upgrade Wizard as well:

image

image

EPTamminga:
@daguiler What is the message if you have a up to date Telerik.dll?

mitchelsellers:
We cannot differentiate between “up to date” and not.

EPTamminga:
@mitchelsellers I meant: a Telerik DLL that was NOT provided by DNNPlatform and has a version that is LATER/HIGHER than a version that was provided in the platform.
A version provided in the Platform is wrong for sure.

donker:
> @mitchelsellers I meant: a Telerik DLL that was NOT provided by DNNPlatform and has a version that is LATER/HIGHER than a version that was provided in the platform. A version provided in the Platform is wrong for sure.

From my understanding the check is twofold: (1) just on dll name and (2) dependencies of other DLLs. So this will **NOT** pick up on "up to date" Telerik dlls if they have the same name. However ... if any component depends on that new Telerik, then it would be pciked up by the second step.

daguiler:
Do we need to adjust the Security Check?
Other than including the found Telerik version number, I made no other changes, and this table shows current logic:

image

Notice the `Unverified` wording (only seen in case of internal error). Is that sentence still true?
Also the last row is there just for completeness because there is no condition currently that leads to that outcome.

mitchelsellers:
@daguiler The check is totally fine!

Do we have an idea of next-steps here for this one.  Some initial testing was done and we did not see Telerik actually get removed in a few initial tests off of this branch.

mitchelsellers:
@daguiler Will you be able to finish this one up at all in the near future?  Anything we can help with?

mitchelsellers:
@daguiler Thanks for the update.  @dnnsoftware/approvers we might want to call a meeting to review next-steps to help get this one over the finish line

david-poindexter:
Best wishes in your new job @daguiler

bradignite:
Just dropping in to say an engineer from Ignite has picked this up and is doing a deep dive into the PR.

mitchelsellers:
This is replaced by #5166

PR #5118, opened 2022-05-09
Top level pages

### LINKED ISSUE #5103: Permission to Add Top-Level Pages
## Description of problem
Currently, only admins (or superusers) can create new top-level pages. We have many sites in which we would like to delegate the ability to create top-level pages to users without giving them full admin permissions to the site.

Without the "overriding" admin permission, [page creation permissions are based on whether the user has edit permissions to the parent page](https://github.com/dnnsoftware/Dnn.Platform/blob/c2615a7db254be5853d0b9db5578e400e978bd44/DNN%20Platform/Library/Security/Permissions/TabPermissionController.cs#L66-L74) (assuming the default permission provider is in place; otherwise, there may be additional permissions in play, but they are still based on the parent page).

## Description of solution
We propose adding a new `Permission` specifically for creating top-level pages.

There would be a permission grid added to the Persona Bar to manage this permission. This could be added to the Security page of the Persona Bar. Alternatively, we could introduce a top-level item to the tree view of the Pages page of the Persona Bar, and selecting that top-level item would allow managing these permissions.

* We could allow `TabPermission` rows with `TabID` set to `NULL`.
* Instead of a top-level page creation permission, we could introduce a "page admin" permission that gave full permissions to manage pages. However, there would be scenarios where we would not want these users to have this level of permission (perhaps they can create new pages but not edit pages we have setup).

## Additional context
We have a developer available to tackle this implementation if approved.

DanielBolef:
Fixes #5103 

~~This PR grants add page permissions for top level pages to users that have add page permissions for the home page. Also adjusts default permissions of new top level pages to match those of the home page so that users are less likely to create pages they cannot edit.~~


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5118) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5118) before we can accept your contribution.<br/><br/>:x: DanielBolef [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5118)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=5118) it.</sub>

sleupold:
I strictly disagree to creating an new table.
There is already a table for page permissions (TabPermissions) and no need for creating yet another one.

sleupold (changes requested):
I disagree with this solution, there shouldn't be created yet another table for specific page permissions. please use existing table TabPermissions according to the discussion of this feature.

bdukes:
@sleupold these permissions don't pertain to a specific tab, they pertain to all of the tabs in a portal, so we can't use the `TabPermission` table

bdukes:
This PR is ready for review. It adds the concept of `PortalPermissions` and introduces two permissions, _Page Admin_ and _Add Top Level Page_. In the default/base permission provider, a user with the _Page Admin_ permission for a site will have access to all permissions for every page of the site.

It also [adjusts the default permissions when adding a top-level page](https://github.com/dnnsoftware/Dnn.Platform/pull/5118/files#diff-b527fc12f2f771945033f62ed19b24060ecbd5cfefd889a14ef623be48839e1d) to be a copy of the home page (in an attempt to avoid scenarios where users are creating pages that they cannot access).

This PR does _not_ include frontend changes to manage these new permissions. At the present time they are managed only through the database. Let us know if the existence of a management UI is a blocker.

/cc @DanielBolef

donker (approved):
Great enhancement IMHO

mitchelsellers (approved):

valadas:
We just have one problem here, this was slated for 9.11.0 but got merged into develop.
It is probably not a big deal as we may release 9.11.0 soonish but if we need to do a 9.10.3 release, we will have to revert/correct this...

bdukes:
@valadas #5207 reverted this PR on `develop` and #5209 reapplied this PR on `release/9.11.0`, so we should be good.

valadas:
Oh, sorry I missed that...

Issue #5120, opened 2022-05-11
ResetDevSite doesn't produce export-bundle.js


donker:
## Description of bug
When building your local dev site the resulting website does not include export-bundle.js or any of the other react projects and as a result the PersonaBar doesn't work.

## Analysis
From a Slack discussion about this issue:

I figured out why the build was not generating the yarn js products on his local dev site. Hold on to your seat, here we go:

- When DNN runs a build it passes through AEModule.build MSBuild script which calls “yarn run build”
- This fires off lerna from the root package.json as “lerna run build --stream”
- This then goes to each JS project listed in that package.json and finds their package.json and runs the “build” script
- E.g. Sites.Web will have this “set NODE_ENV=production&&webpack -p”
- Now the webpack.config.js for that project is loaded and you’ll find it attempts to establish mode as follows: `const isProduction = process.env.NODE_ENV === “production”;`
- This is where things go wrong. It turns out this is not set to production by the command and now isProduction is false
- The output path looks at whether we are in production mode or not and if not then we see if we have a local settings file. If you are a DNN developer, you will have one. This will then be used for output. Meaning that a proper production output path is not used but the development path.

I am not sure how we got here. I also cannot explain why NODE_ENV does not persist to the webpack config. I also cannot explain what if anything changed. I did not experience any issues because I still had (old) local copies in place which were being used.

## Affected version
We can't establish how long this problem has persisted.

david-poindexter:
Closing per #5121 - thanks again @donker 🎉

PR #5121, opened 2022-05-11
Fix issue with undefined isProduction in Yarn scripts

donker:
Resolves issue #5120 

What it does is to replace the following:

```
module.exports = {
```

with 

```
module.exports = (env, argv) => {
  const isProduction = argv.mode === "production";
  return {
```

for all the webpack configurations where we use isProduction. The command line p option (i.e. `webpack -p`) will now properly set isProduction in the configuration.


valadas (approved):
Looks good to me, nice catch there.

bdukes (approved):

PR #5127, opened 2022-05-25
Remove dependency on SharpZipLib

### LINKED ISSUE #2735: Remove SharpZipLib
## Description of problem
SharpZipLib is an external (and modified) library that's included in DNN.  At the time of inclusion, it was the only functional option available to work with managing ZIP archives.  Since, .NET has included and iterated on a viable alternative that ships with .NET.  There's no longer any need to keep this library in the core since removal has a viable alternative and will also reduce the footprint of DNN overall.

## Description of solution
* Ensure Install/Upgrade does not require SharpZipLib
* Create a Package for distribution and upgrade to the latest version.  (To ensure installs have latest possitble)
* Do not include SharpZipLib DLL or license file in new installs.  
* Leave the SharpZipLib DLL in the Bin folder on upgrades for backwards compatibility with vendors that might be manually referencing the DLL.
* Update the wrapper code in the core to no longer use or reference SharpZipLib and use the .NET alternative instead.

None. 

## Screenshots
None.

## Additional context
None.

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0
* [x] 9.1
* [x] 9.1.1
* [x] 9.2
* [x] 9.2.1
* [x] 9.2.2
* [x] 9.3
* [x] 9.3.1
* [x] 9.3.2
* [x] 9.4
* [ ] 10.0 (suggested target for the change)

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

donker:
Fixes #2735

### DIFF
--- a/DNN Platform/Library/Common/Utilities/FileSystemExtensions.cs
+++ b/DNN Platform/Library/Common/Utilities/FileSystemExtensions.cs
@@ -5,6 +5,7 @@
 namespace DotNetNuke.Common.Utilities
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.IO.Compression;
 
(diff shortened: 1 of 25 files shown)


bdukes (approved):

ohine:
It's difficult for no reason to track changes when functions are completely removed and then added back in later with an additional obsolete tag. The solution that @bdukes started was much cleaner for the removal of SharpZipLib.

mitchelsellers (approved):
This looks good, and I did test this was working, however, @valadas I believe you had some feedback on this PR?  Can you confirm that you are also ok with this?  If so, please merge

valadas:
Nah, I am good with it, my concern was with build packaging, but that part is not distributed, so good to go.

PR #5129, opened 2022-05-31
Fixed 2 build issues

valadas:
AEModule.build had 2 globs that would return some of the same files and have them duplicated in resources.zip files, which was not a major issue but found it while troubleshooting another issue.

Also, WebFormsMVP had recently been made into a package but it was not added to the list of unversioned manifest and would get automatically versioned by the current DNN version upon build.


david-poindexter (approved):
Great catches @valadas - thank you!

PR #5152, opened 2022-06-22
Improve server screen UI

donker:
Improves the edit boxes UI


mtrutledge (commented):
Hey @donker, wouldn't we need to do the same changes on

/Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpAdminController.cs

that you did on 

/Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpHostController.cs

since the SMTP service on the javascript end checks if the user is an admin or a host and then calls a different end point?

donker:
> Hey @donker, wouldn't we need to do the same changes on
> 
> /Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpAdminController.cs
> 
> that you did on
> 
> /Dnn.PersonaBar.Extensions/Services/ServerSettingsSmtpHostController.cs
> 
> since the SMTP service on the javascript end checks if the user is an admin or a host and then calls a different end point?

Good catch. Thanks.

mtrutledge (approved):
This looks good to me. Thanks @donker

mitchelsellers (approved):

david-poindexter:
Good job @donker - thanks for the quick contribution on this!

PR #5185, opened 2022-07-05
Upgrade SharpZipLib from 0.86.0 to 1.3.3

bdukes:
## Summary
This PR upgrades SharpZipLib to the latest version.

[Potential breaking changes](https://github.com/icsharpcode/SharpZipLib/wiki/Release-History) include:
- The `ICSharpCode.SharpZipLib.Checksums` namespace has been renamed `ICSharpCode.SharpZipLib.Checksum`.


valadas (approved):
Awesome!

donker (approved):
Awesome

schotman:
Is this removal of SharpLib announced in advance? We now have 3rd party modules break because SharpLib is no longer there.

bdukes:
@schotman this PR was not intended to remove SharpZipLib from the installation, that was an oversight. Thanks for testing! I'll be submitting a PR to add the SharpZipLib library to the installation

schotman:
Thanks, Question: do we want to remove SharpLibZip in a future upgrade? if so we need to want 3rd party module devs

bdukes:
Yes, the APIs that depend on SharpZipLib have been marked as deprecated and will be removed in v11.0.0. Our current intention is that we would remove SharpZipLib at that time. Any 3rd party module could then start including SharpZipLib if they require it, or migrate to `System.IO.Compression` as we did in #5127

PR #5193, opened 2022-07-19
v9.11.0 Bump react-collapse from 4.0.3 to 5.1.0

bdukes:
Applying #4808 to release/9.11.0


mitchelsellers (approved):

valadas (changes requested):
I mentioned that #4808 would need some good UI testing but it was merged without anyone confirming..

I just gave that build a quick test and we have issues with that:
image

image

image

I _think_ we need to revert that until someone has time to adjust it's usage...

bdukes:
I'll look into fixing the usage

david-poindexter:
Good catch @valadas - thanks for the effort @bdukes 🎉

valadas (approved):
As discussed on Slack with @bdukes , it will be easier to merge this and then do the react-collapse usage fix. So merging this one now.

PR #5194, opened 2022-07-20
Replace `DigitalAssets` with `Resource Manager` in templates and remove `DigitalAssets` module

### LINKED ISSUE #4895: Existing Website Templates (Default & Blank) should be updated to remove Telerik, and the existing Digital Asset Manager, and replace those with Resource Manager
## Description of problem
Currently the `Digital Asset Manager` is placed on `Site Assets` and `Global Assets` by default.  These should be replaced with the new `Resource Manager` by default.

## Description of solution
The website templates need to be updated to implement this change.

n/a

## Screenshots
n/a

## Additional context
n/a

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

david-poindexter:
## Summary
Resolves #4895 

* Updated `Blank Website` template to replace `Digital Assets` module with `Resource Manager`
* Updated `Default Website` template to replace `Digital Assets` module with `Resource Manager`
* Updated `admin` template (though this is not used in the UI) to replace `Digital Assets` module with `Resource Manager`
* Removed `DigitalAssets` module


mitchelsellers:
Question - Should digital assets live on in a separate repo?  (I don't believe so, but just asking)

valadas:
> 

I don't think so either, we should not maintain a Telerik-based solution IMO and if someone really really wants to, they can grab it from the git history.

valadas (approved):
Looks good to me, thanks guys!

valadas:
Strange the build failed creating nuget packages but the artifact is fine, I'll try to re-fire the build

david-poindexter:
I believe the problem is in the bundle:

https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec#L33-L34

valadas:
> I believe the problem is in the bundle:
> 
> https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Build/Tools/NuGet/DotNetNuke.Bundle.nuspec#L33-L34

That makes sense, why do include the DAM in there? I think we could simply remove it from the nuspec file right ?

david-poindexter:
Yeah, I think so.  I'll test that now on my local branch.

david-poindexter:
Okay, this PR is ready for a second review.  CC @dnnsoftware/approvers

mitchelsellers (approved):
This looks good, thank you!

Issue #5114, opened 2022-05-05
Standardize jQuery Usage to Current Version


mitchelsellers:
## Description of bug
DNN Platform currently distributes multiple versions of jQuery and jQuery UI, including in the CKEditor Provider and a few other locations.

We need to ensure that all versions, including those loaded from a CDN are secure and up to latest versions.  Ideally using the standardized registration processes for jQuery.

Caution will be needed to ensure that we do not break other functions within the application.

## Steps to reproduce
Search the installation code base for used version of jQuery.

## Current behavior
Multiple versions of jQuery are used

## Expected behavior
A singular version of jQuery should be used

## Affected version
09.10.02 and Earlier

## Affected browser
N/A

PR #5205, opened 2022-07-27
Use Client Resources Manager for CKEditor Provider dialogs

bdukes:
For #5114

This PR adds Client Resource Manager support in the `Browser.aspx` and `Options.aspx` dialogs used by the CKEditor Provider and adjusts the JS and CSS of those dialogs to be loaded via the CRM.

I also adjusted the build script to resolve issues with builds not updating the website.

Finally, I made a couple of localization related changes, changing "Browser-Modus" to "Browser Mode" in the default resx file and extracting a common localization function.


valadas (approved):
Excellent!

donker (approved):
Awesome, thanks

PR #5213, opened 2022-07-29
React-collapse fixes

bdukes:
## Summary
This resolves issues introduced by #4808, where the react-collapse library switched to CSS-based animations.

It also removed support for pass-through props, so any usage of `style` or `className` had to be adjusted.
The `fixedHeight` prop is now "implicitly supported", so each of those props needed to be converted to CSS that targeted the child of the collapse element.
The `keepCollapsedContent` is now the default behavior. There's a separate component which unmounts the content when collapsed. I have not made use of that separate component, but didn't see any issues from places where previously the content had been unmounted by default.


valadas (approved):
supercalifragilisticexpialidocious

Excellent, I downloaded this build and went through the UI, it appears everything I tested was now in order. Thanks a lot!

donker (approved):
Awesome, thanks

PR #5223, opened 2022-08-02
Disable PR merge for 10.0.0 branch

bdukes:
Until we have a `release/10.0.0` this will fail every time


valadas (approved):

valadas:

PR #5229, opened 2022-08-06
Create User invalid request has a response which returns 500 #5228

### LINKED ISSUE #5228: Create User invalid request has a response which returns 500
## Description of problem
When you try to change any user password, invalid password set causes an error response has a status code of 500 which is system exception.
1. Go to Persona Bar>Users>Change Password
2. Enter an invalid password e.g. a small password (single character)
3. Observe the error and check the details in the network response (browser console)

Error response shouldn't have a status code of 500

## Description of solution
Change status code of the bad request. 

I set the response to 400 but 422 may be considered. 

## Screenshots
-

## Additional context
-

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

armaganpekatik:
Fixes #5228 

## Summary

When you try to change any user password, invalid password set causes an error response has a status code of 500 which is system exception.

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
@@ -70,13 +70,13 @@ namespace Dnn.PersonaBar.Users.Services
             }
             catch (Exception ex)
             {
+                Logger.Error(ex);
                 if (ex.GetType() == typeof(InvalidUserRegisterException))
                 {
                     return this.Request.CreateErrorResponse(HttpStatusCode.BadRequest, ex.Message);
                 }
                 else
                 {
-                    Logger.Error(ex);
                     return this.Request.CreateErrorResponse(HttpStatusCode.InternalServerError, ex.Message);
                 }
             }
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
@@ -70,8 +70,15 @@ namespace Dnn.PersonaBar.Users.Services
             }
             catch (Exception ex)
             {
-                Logger.Error(ex);
-                return this.Request.CreateErrorResponse(HttpStatusCode.InternalServerError, ex.Message);
+                if (ex.GetType() == typeof(InvalidPasswordException))
+                {
+                    return this.Request.CreateErrorResponse(HttpStatusCode.BadRequest, ex.Message);
+                }
+                else
+                {
+                    Logger.Error(ex);
+                    return this.Request.CreateErrorResponse(HttpStatusCode.InternalServerError, ex.Message);
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


mitchelsellers (changes requested):
Although I understand the desired goal here, this does introduce a process change, and  you might be masking an error that would have historically been logged.

For example, UserCreate can fail for password, duplicate email, or other reasons.  This process now assumes that all failures are invalid passwords which is not correct.  I think if any attempt to bypass logging etc should be adjusted.

Now, I could be convinced that ALL errors should be a 400 response, rather than a 500 though?

@dnnsoftware/approvers any thoughts?

david-poindexter:
I'm thinking we should be using `403 Forbidden` in this case, no?  I agree that `500 Internal Server Error` is not good.

armaganpekatik:
> Although I understand the desired goal here, this does introduce a process change, and you might be masking an error that would have historically been logged.
> 
> For example, UserCreate can fail for password, duplicate email, or other reasons. This process now assumes that all failures are invalid passwords which is not correct. I think if any attempt to bypass logging etc should be adjusted.
> 
> Now, I could be convinced that ALL errors should be a 400 response, rather than a 500 though?
> 
> @dnnsoftware/approvers any thoughts?

You are right for the naming, we can change InvalidPasswordException to InvalidUserRegistrationException. This was because I was thinking that I can define password error at the beginning. So the error is coming from https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Entities/Users/UserController.cs#L779 and all of them are user validation error. Please see last commit.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
```diff
@@ -69,9 +69,16 @@ public HttpResponseMessage CreateUser(CreateUserContract contract)
                     : null);
             }
             catch (Exception ex)
-            {
-                Logger.Error(ex);
-                return this.Request.CreateErrorResponse(HttpStatusCode.InternalServerError, ex.Message);
+            {
+                if (ex.GetType() == typeof(InvalidUserRegisterException))
+                {
+                    return this.Request.CreateErrorResponse(HttpStatusCode.BadRequest, ex.Message);
```
mitchelsellers, line 75: This is a nice differentation between "expected" and "unexpected", but I do not want to change the behavior of logging the failures under ALL circumstances.  If you move the `Logger.Error(ex)` to be called all of the time I can agree with this.

mitchelsellers (changes requested):

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UsersController.cs
armaganpekatik: Do you want to log invalid user register attempts? I can add this as well..

armaganpekatik: done

mitchelsellers (approved):
This looks acceptable to me

donker (approved):
LGTM

PR #5234, opened 2022-08-22
Ensure StandardFolderProvider always returns an absolute URL.

### LINKED ISSUE #5085: Link click for files doesn't take into account the SSL Offload.

rhaiamz:
Fixes #5085 

Includes the fixes suggested in PR #5086.

### DIFF
--- a/DNN Platform/Library/Entities/Portals/PortalSettings.cs
+++ b/DNN Platform/Library/Entities/Portals/PortalSettings.cs
@@ -169,7 +169,7 @@ namespace DotNetNuke.Entities.Portals
         }
 
         /// <inheritdoc/>
-        public string DefaultPortalAlias
+        public virtual string DefaultPortalAlias
         {
             get
             {
--- a/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
+++ b/DNN Platform/Library/Services/FileSystem/Providers/StandardFolderProvider.cs
@@ -219,10 +219,10 @@ namespace DotNetNuke.Services.FileSystem
                     file.LastModificationTime.GetHashCode().ToString(),
                     portalSettings.GUID.ToString());
 
-                return TestableGlobals.Instance.ResolveUrl(fullPath + "?ver=" + cachebusterToken);
+                return this.EnforceAbsoluteUrl(TestableGlobals.Instance.ResolveUrl(fullPath + "?ver=" + cachebusterToken), portalSettings);
             }
 
-            return TestableGlobals.Instance.ResolveUrl(fullPath);
+            return this.EnforceAbsoluteUrl(TestableGlobals.Instance.ResolveUrl(fullPath), portalSettings);
         }
 
         /// <inheritdoc/>
@@ -466,5 +466,18 @@ namespace DotNetNuke.Services.FileSystem
         {
             return PathUtils.Instance.GetRelativePath(folderMapping.PortalID, path);
         }
+
+        private string EnforceAbsoluteUrl(string uri, PortalSettings portalSettings)
+        {
+            var initialUri = new Uri(uri, UriKind.RelativeOrAbsolute);
+            if (!initialUri.IsAbsoluteUri)
+            {
+                var appUrl = new Uri(Globals.AddHTTP(portalSettings.DefaultPortalAlias));
+                return new Uri(appUrl, initialUri).AbsoluteUri;
+            }
+
+            return uri;
+        }
+
     }
 }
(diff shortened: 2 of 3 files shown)


mitchelsellers (commented):
Upon further consideration and thought, I'm not sure that this is an acceptable change at all, as it will have major breaking change implications to how files are managed within the platform.

Consider a few situations that are impacted by this change.

* With this change a request for a file link will always be an absolute path, as much as this "fixes" the SSL offload issue.  In a situation where the SSL Offload is a different URL this is expected, but for almost all other situations this is going to create a real issue.  For example if I want to move my site from production to a test environment, every single file link will be incorrect now as it is hard-coded with an absolute path.  With no way to edit
* With this change, settings in the Editor, or other locations for allowing, or disallowing relative paths will no longer function

It MIGHT be possible to override this, ONLY when SSL is enabled, with an offload, AND a different URL but still not 100% sure that is a proper solution.

@dnnsoftware/approvers thoughts on this one?

rhaiamz:
@mitchelsellers In this case, what are the next steps? Your current concerns are the reason why in my initial implementation, I placed it at a lower level, so the changes only happen at runtime.

bdukes:
I think this logic should only apply when there's an SSL offload header setup.

mitchelsellers:
I agree with @bdukes that only doing this logic if an offload url is set would be proper. 

That is a situation where content owners will expect environmental specific urls.

mitchelsellers:
There is a very, very unique situation where the SSL Offload is done using a DIFFERENT URL rather than the actual public/normal URL.

valadas:
> There is a very, very unique situation where the SSL Offload is done using a DIFFERENT URL rather than the actual public/normal URL.

Exactly the source of my confusion, being relative, it does not include the admin part in it right? So relative should work, unless I am missing something, I don't understand...

rhaiamz:
I will post some screenshots in order to bring some light to this issue. 
This is the setup made in the security tab
image
This is the link that will be clicked in order to open the file in the browser.
image
Which the browser will autocomplete with https as you can see in the below screenshot.
image
But after the request is made to the LinkClickController, the browser opens this tab for the returned URL
image

rhaiamz:
I will post a more extended explanation in the following hours.

rhaiamz:
We did a bit of digging on our end. The URL(line 175 in [FileServerHandler](https://github.com/dnnsoftware/Dnn.Platform/blob/3697c5344cef8d49214230f0cc2efcd9e93a00a8/DNN%20Platform/Library/Services/FileSystem/FileServerHandler.cs#L175)) that leaves the DNN code remains relative before it is passed to [Response.Redirect](https://referencesource.microsoft.com/#system.web/HttpResponse.cs,2372) at line 190 in [FileServerHandler](https://github.com/dnnsoftware/Dnn.Platform/blob/3697c5344cef8d49214230f0cc2efcd9e93a00a8/DNN%20Platform/Library/Services/FileSystem/FileServerHandler.cs#L190).

As you can see, at [line 2394](https://referencesource.microsoft.com/#system.web/HttpResponse.cs,2394) this will force the URL to be [converted](https://referencesource.microsoft.com/#system.web/HttpResponse.cs,3308) to an Absolute URL based on a configuration setting found in system.web/httpRuntime. This for DNN seems to default to true, possibly from the days when the [RFC2616](https://www.rfc-editor.org/rfc/rfc2616#section-14.30) which required the Location header to be an absolute URI. In 2014 it was replaced by [RFC7231](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2) which states that relative URIs in the Location header are now permitted. Upon setting the  "useFullyQualifiedRedirectUrl" to false, the URL remained relative in the Location header, and the request no longer changed from HTTPS to HTTP. 
While simply turning this option off would fix the issues mentioned, I still think that returning an absolute URL for the file would be a better option, as this would eliminate any confusion and possibility for such an issue to occur in the future.

bdukes:
Thanks @rhaiamz, that clarification helps a lot.

As @valadas pointed out, using a relative URL here seems most correct, so if we could preserve the relative URL, that would resolve the issue completely from my perspective.

I would suggest that the PR add a [config transform](https://github.com/dnnsoftware/Dnn.Platform/tree/develop/DNN%20Platform/Website/Install/Config) to set `useFullyQualifiedRedirectUrl` to `false`. The only issue noted in the [docs for that setting](https://docs.microsoft.com/en-us/dotnet/api/system.web.configuration.httpruntimesection.usefullyqualifiedredirecturl) states "Some browsers might have problems loading pages in cookieless sessions when this value is false."

My assumption is that "some browsers" includes old browsers we don't support, and we don't support cookieless sessions, either AFAIK, so this seems like a safe change to me.

valadas:
Thanks, that explanation helps me understand now

valadas:
As discussed, this PR cannot be merged this way as it would produce too many side-effects.

PR #5237, opened 2022-08-23
Add support for the RoleUpdated event.

### LINKED ISSUE #5236: Support for Role Update event.
## Description of problem
While trying to use the event system in DNN to watch for changes to Tabs, Modules, Roles, etc. I noticed that there is no event triggered for the update of a role.

## Description of solution
Implement a new event that DNN can trigger when a role is updated. This needs changes in the following interfaces:
1. IEventManager - to add a new method: `void OnRoleUpdated(RoleEventArgs args)`.
2. IRoleEventHandlers - to add the new event method: `void RoleUpdated(object sender, RoleEventArgs args)`.
    - as this would be a breaking change for all existing implementation of the handler, I would instead create a new extension interface `IRoleUpdateEventHandlers` and add the method there. Any implementation that require RoleUpdate event functionality will have to implement both interfaces.

## Affected browser
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

dimarobert:
Closes #5236

## Summary
Implements the event trigger for Role Update event.


DNN Platform/Library/Entities/EventManager.cs
```diff
@@ -102,6 +102,10 @@ public EventManager(IEventLogger eventLogger)
                 this.RoleDeleted += handler.Value.RoleDeleted;
                 this.RoleJoined += handler.Value.RoleJoined;
                 this.RoleLeft += handler.Value.RoleLeft;
+                if (handler.Value is IRoleUpdateEventHandlers updateHandler)
```
bdukes, line 105: I would think you'd want to be able to implement `IRoleUpdateEventHandlers` without also implementing `IRoleEventHandlers`. If you only want to handle the `RoleUpdated` event, it would be surprising to me as a developer that I need to implement the other interface, too.

DNN Platform/Library/Entities/IEventManager.cs
```diff
@@ -167,6 +167,12 @@ public interface IEventManager
         /// <param name="args">The event arguments.</param>
         void OnRoleCreated(RoleEventArgs args);
 
+        /// <summary>
+        /// Fires up when a role was updated.
+        /// </summary>
+        /// <param name="args">The event arguments.</param>
+        void OnRoleUpdated(RoleEventArgs args);
```
bdukes, line 174: This will be a breaking change for anyone that's implemented `IEventManager`. I don't expect that it's probably happened outside of tests. However, we need to think about a longer term strategy for adding to these interfaces. Do we need to have much smaller interfaces? Do we need to version interfaces? Do we need to use base classes instead of or in addition to interfaces? 🤔

DNN Platform/Library/Security/Roles/IRoleEventHandlers.cs
```diff
@@ -13,5 +13,13 @@ public interface IRoleEventHandlers
         void RoleJoined(object sender, RoleEventArgs args);
 
         void RoleLeft(object sender, RoleEventArgs args);
+    }
+
+    /// <summary>
+    /// The base interface <see cref="IRoleEventHandlers"/> has to be implemented by the class as well for the handler to be picked up.
+    /// </summary>
+    public interface IRoleUpdateEventHandlers
```
bdukes, line 21: I think our StyleCop rules require each type to be in its own file.

bdukes (commented):
Thanks!

bdukes:
@dnnsoftware/approvers I would think we'd have this target 9.11.0, should we go ahead an retarget this PR to the `release/9.11.0` branch?

DNN Platform/Library/Entities/EventManager.cs
dimarobert: To find the handler implementations, DNN uses `MefComposition` so an implementation would look something like this (from what I could figure out):
```cs
[Export(typeof(IRoleEventHandlers))]
[ExportMetadata("MessageType", "RoleEventHandlers")]
public class RoleEvents : IRoleEventHandlers {
    // ... implementation
}
```

If we go with `interface IRoleUpdateEventHandlers  : IRoleEventHandlers {...}` and then only implement that, I think that the final implementation would be something like this:

```cs
[Export(typeof(IRoleEventHandlers))]
[ExportMetadata("MessageType", "RoleEventHandlers")]
public class RoleEvents : IRoleUpdateEventHandlers {
    // ... implementation
}
```

Which, in my opinion, would be very hard to figure out that you need to implement `IRoleUpdateEventHandlers` but you still need to specify the export as the old interface `[Export(typeof(IRoleEventHandlers))]`, otherwise I would guess that the `EventHandlersContainer<IRoleEventHandlers>` [here](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Entities/EventManager.cs#L99) would not get it.

bdukes: I was thinking that the event manager would start using `EventHandlers<IRoleUpdateEventHandlers>` so that an implementation would be 

```
[Export(typeof(IRoleUpdateEventHandlers))]
[ExportMetadata("MessageType", "IRoleUpdateEventHandlers")]
public class RoleEvents : IRoleUpdateEventHandlers {
    // ... implementation
}
```

But I guess that makes it more difficult for the common scenario where you do want both.

We could support both scenarios (where `IRoleUpdateEventHandlers` is registered as `IRoleUpdateEventHandlers` or `IRoleUpdateEventHandlers` is registered as `IRoleEventHandlers`). I feel like that would be a more friendly dev experience. Open to discussion if others don't think that's worth it or makes it more complicated.

DNN Platform/Library/Entities/IEventManager.cs
valadas: I am not opposed to versioned interfaces myself, it would open up soooo much possibilities. I know it does not look pretty, but that is only for devs, and it would make our lives much easier to revamp some areas of the platform...

DNN Platform/Library/Entities/EventManager.cs
dimarobert: So what you are saying is that we should do both things: a foreach on `EventHandlersContainer<IRoleUpdateEventHandlers>` and a cast check to `IRoleUpdateEventHandlers` in the existing `EventHandlersContainer<IRoleEventHandlers>` foreach, so that we support both scenarios.

I have no problem with that, but I am not sure what should happen next. Will you make a decision in the next approvers meeting?

Same for the comment below with the smaller/versioned interfaces.

_As a side note/question for the versioned interfaces or base classes: it will help with adding new things to the interfaces without breaking changes, but what will happen when you want to remove obsolete API from the interface and stop using/supporting that API throughout DNN's code? Will that just go through the normal two versions obsolete then removal process as a breaking change? Will the versioned interfaces be supported only for a specific number of major versions?_

bdukes: Yeah, I think we'll probably need to have a discussion at least on the interface question next Tuesday. We'll let you know what we decide. Thanks!

dimarobert: @bdukes Have you had the chance to discuss about this?

bdukes: We did discuss this on Tuesday. This PR includes a breaking change on the `IEventManager` interface, and so will have to be held until at least version 10.0. We're planning to have a larger discussion in about a month around making these sorts of changes easier in the future.

bdukes (approved):

valadas (approved):
Approved, thanks!

PR #5116, opened 2022-05-09
upgrade jquery-ui version

armaganpekatik:
Fixes #5115
Closes #4816 

## Summary

DNN is using an old version of jquery-ui 1.12.1
Upgrade jquery-ui to newer version from 1.12.1 to latest 1.13.1

PR #5238, opened 2022-08-24
Fix wrong file name for jQuery UI

bdukes:
Fix issue in #5116 where file was renamed in one place but not another.


mitchelsellers (approved):

valadas (approved):
Excellent

PR #5239, opened 2022-08-24
Use JSON.parse over eval where appropriate

bdukes:
Replace unnecessary usages of `eval` when `JSON.parse` will do.


mitchelsellers (approved):

valadas (approved):
Awesome

PR #5241, opened 2022-08-30
Removes resource manager PB registration

valadas:
Initially we though about making the resource manager live directly in the Persona Bar in addition to be used as a module but we then decided to bring this idea to a step 2.
This PR removes that registration from the manifest for now.


mitchelsellers (approved):

donker (approved):
OK

bdukes (approved):
LGTM

PR #5242, opened 2022-08-31
Added support for HtmlAttributesAsString in html module views

valadas:
Added support for HtmlAttributesAsString in html module views.

Also makes the resource manager use this so it works with sub-portals, etc.


david-poindexter (approved):
This looks great to me.  I really appreciate the more understandable logic here too.  That was some confusing logic before.  Thank you @valadas 🎉 🌮

donker (approved):
LGTM

PR #5243, opened 2022-08-31
Fix relative path in RM

donker:
Fixes the path in the new RM js paths


valadas (approved):
Awesome, missed that part

david-poindexter (approved):
Good catch! 🎉

PR #5244, opened 2022-08-31
Improve HTML attribute support in client resource manager

bdukes:
## Summary
This PR improves on #5242. 

- Use dictionaries of attributes directly instead of requiring them to be serialized and deserialized as strings
- Add HTML attributes overloads for CSS
- Clean up some comments, typos, and other code


valadas (approved):
This is awesome, even cleaner.
I used empty dictionaries so it does not get confused and hit the wrong overload (multiple strings passing nulls), never thought about using the parameter name to resolve that ambiguity. Great idea, and love the ternaries for empty strings, much much simpler and easier to understand!

valadas:
@bdukes you think the build failure is a fluke

bdukes:
Yeah, I've re-run the build

david-poindexter (approved):

PR #5249, opened 2022-09-01
Use JSON.parse over eval where appropriate

valadas:
This reapplies this fix onto 9.11.0 independently in hopes to resolve the merge conflict we are having


valadas:

PR #5252, opened 2022-09-01
Ensure Resource Manager script files are included in package

### LINKED ISSUE #5251: No Site Assets and Global Assets after upgrade to DNN 9.11.00 (228)
## Description of bug
Upgraded from DNN 09.10.2 to 09.11.00 and now the Site/blobal Settings is empty.
Module is on the page but is empty

## Steps to reproduce
1. extract upgrade package;
2. ran the upgrade wizard (no issues);
3. open website
4. open site Assets page

## Current behavior
no file manager

## Expected behavior
there should be a file manager

## Screenshots
image

## Error information
image

## Affected version
* [ ] 09.11.00 latest supported release

## Affected browser
all

bdukes:
Fixes #5251

## Summary
The MSBuild script copies files from `ResourceManager.Web\dist\dnn-resource-manager` to `Scripts\dnn-resource-manager` and then expects those files to be includes in the resources zip, but the timing of operations means that the files for the resources zip get calculated because the copy.

This PR adjusts the timing of the copying (which I don't think actually does anything, but I don't want to spend more time trying to understand MSBuild). The second change is to explicitly include those new files into the `Resources` variable in MSBuild during the `UpdateFiles` target, so that they get included in the later `Package` target.

### DIFF
--- a/DNN Platform/Modules/ResourceManager/Module.build
+++ b/DNN Platform/Modules/ResourceManager/Module.build
@@ -15,10 +15,7 @@
     </PropertyGroup>
     <Import Project="$(BuildScriptsPath)\Package.Targets" />
     <Import Project="$(BuildScriptsPath)\Module.Build"/>
-    <ItemGroup>
-        <Scripts Include="$(MSBuildProjectDirectory)\Scripts\**\*.*" />
-    </ItemGroup>
-    <Target Name="UpdateFiles">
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

david-poindexter (approved):
This looks good to me - would love for @valadas to confirm.

valadas (approved):
Great catch! Thanks for this. I`ll merge this and do an RC2

PR #5264, opened 2022-09-03
Fixed an issue that prevent using resource-manager with CDF

### LINKED ISSUE #5262: DependencyHandler.axd may return wrong Content-Type (Resource Manager broken in 9.11.0 RC2)
## Description of bug
If _Client Resource Management_ is enabled (Compose files, minify CSS, minify JS), the new Resource Manager doesn't load. Some HTTP responses via DependencyHandler.axd have a wrong content-type, so the browser doesn't process the JavaScript/CSS.

## Steps to reproduce
1. Setup new DNN 9.11.0 RC2
2. Resource Manager (/Host/Global-Assets/portalid/0) will work
3. Enable _Client Resource Management_ (Compose files, minify CSS, minify JS)
4. Reload the page containing the Resource Manager.
5. Resource Manager is not shown anymore, browser console has error because of invalid Content-Type.

## Current behavior
Resource Manager is not shown anymore, once _Client Resource Management_ is enabled.
Some responses from DependencyHandler.axd have a wrong _Content-Type_ header (see screenshot). Browser-console has error:

`Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/css". Strict MIME type checking is enforced for module scripts per HTML spec.`

## Expected behavior
Resource Manager should be shown and work, also when _Client Resource Management_ is enabled.
Responses from DependencyHandler.axd must have the correct _Content-Type_ header, so browsers process the responses.

## Screenshots
image

## Error information

Request to GET /DependencyHandler.axd/3fbadd4c9a9a365925513c8fba5b61d5/39/p-f7e975c5.js (initiated via **script**-tag) returns http 200 with JavaScript code in the body, but with a wrong http header: Content-Type: **text/css**.

Error in Browser Console (Chrome 104):
`Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/css". Strict MIME type checking is enforced for module scripts per HTML spec.`

## Additional context
There are also other errors in the browser console, as soon as _Client Resource Management_ is enabled.

It affects new DNN Installation and also Upgrades from older DNN Version (e.g. 9.6.0)

## Affected version

* [x] 09.11.00 release candidate 2

## Affected browser

* [x] Chrome 104

valadas:
Closes #5262

It was a bit of a rabbit hole to go down to figure this one out. Initially I thought the CDF had an error serving javascript with a css mime type. But the underlying issue is that with Stencil.js there is a main script that registers individual components scripts in a different way depending on browser support and loads or not some polyfills accordingly (very efficient). This means that that "entry" file was served by the CDF but the file attempts to load it's "component details" files relative to that path. Unfortunatelly thanks to CDF, that relative path was the CDF handler itselft instead of the path to the other js files. And the CDF defaults to css mime type if it can't figure out the right one.

The Client Dependency Framework does not support javascript modules (because of script relative paths) and there are no plans to add such support according to their repository. Also with this modern way of lazy-loading and with modern http2 having compression and multiplexing, bundling those files is actually a performance hit and not recommended.

Which left us with a couple options:
1. Not take advantage of Js modules by bundling all js files (but this compile target was deprecated by Stencil for the reasons above), so it would have taken quite some time and incur a performance hit instead of a boost.
2. Not use CDF for this project, but this was broken if hosting in a virtual directory (IIS Application under an IIS site folder).

This PR implementes #2 but adds support for ApplicationPath to the existing Request SPA token. This new token support can also be very useful for other SPA modules.

So under normal hosting, this returns `/` but if the site is hosted in a dnn subdirectory it will return '/dnn' (or whataver the web path is for that web application.

I think this will cover all scenarios here.

### DIFF
--- a/DNN Platform/Library/UI/Modules/Html5/RequestPropertyAccess.cs
+++ b/DNN Platform/Library/UI/Modules/Html5/RequestPropertyAccess.cs
@@ -10,17 +10,20 @@ namespace DotNetNuke.UI.Modules.Html5
     using DotNetNuke.Entities.Users;
     using DotNetNuke.Services.Tokens;
 
+    /// <summary>
+    /// Replaces tokens related to the current http request.
+    /// </summary>
     public class RequestPropertyAccess : IPropertyAccess
     {
-        private readonly HttpRequest _request;
+        private readonly HttpRequest request;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RequestPropertyAccess"/> class.
         /// </summary>
-        /// <param name="request"></param>
+        /// <param name="request">The current http request.</param>
         public RequestPropertyAccess(HttpRequest request)
         {
-            this._request = request;
+            this.request = request;
         }
 
         /// <inheritdoc/>
@@ -35,7 +38,9 @@ namespace DotNetNuke.UI.Modules.Html5
             switch (propertyName.ToLowerInvariant())
             {
                 case "querystring":
-                    return this._request.QueryString.ToString();
+                    return this.request.QueryString.ToString();
+                case "applicationpath":
+                    return this.request.ApplicationPath;
             }
 
             propertyNotFound = true;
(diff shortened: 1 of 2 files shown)


david-poindexter (approved):
@valadas this is fantastic - nice work!  🎉

mitchelsellers (approved):

PR #5267, opened 2022-09-05
Added support for RelativeAppPath in Request token

### LINKED ISSUE #5265: Still an issue with the resourcemanager on 9.11.0 rc0003
## Description of bug
Issue with the new resourcemanager on fresh install on 9.11.0 rc0003 (dutch)

## Steps to reproduce
Install and go to the host or site files

## Current behavior
Module not showing, F12 (developer tools) gives an error  seems like a wrong path...
it tries to load the scripts from http://desktopmodules/blablabla...

image

## Expected behavior
Would like the resourcemanager to work :-)

## Affected version
* [x] 09.11.00 RC003 latest supported release

## Affected browser
all

valadas:
Closes #5265

The .Net Request.ApplicationPath behaves this way:
- For a normal install it returns `/`
- For a virtual directory install it retursn `/dnn` (or whatever virtual directory name)

Notice here that in one case the string ends with a / and in the other case not.

This behaviour makes it diffucult to use it in a multi-purpose token made for relative urls as in:
```
<script src="[Request.ApplicationPath]/DesktopModules" />
```
Would render:
- `/dnn/DesktopModules` for a virtual directory (OK)
- `//DesktopModules` for a normal hosting scenario (NOT OK)

This PR adds `RelativeAppPath` which is a variation of `ApplicationPath` that will always end with `/`.

Usage as follows makes it usable in both scenarios and consumers can expect to always have the ending `/` provided:
```
<script src="[Request.RelativeAppPath]DesktopModules" />
```

### DIFF
--- a/DNN Platform/Library/UI/Modules/Html5/RequestPropertyAccess.cs
+++ b/DNN Platform/Library/UI/Modules/Html5/RequestPropertyAccess.cs
@@ -41,6 +41,11 @@ namespace DotNetNuke.UI.Modules.Html5
                     return this.request.QueryString.ToString();
                 case "applicationpath":
                     return this.request.ApplicationPath;
+                case "relativeapppath":
+                    // RelativeAppPath is like ApplicationPath, but will always end with a forward slash (/)
+                    return this.request.ApplicationPath.EndsWith("/")
+                        ? this.request.ApplicationPath
+                        : $"{this.request.ApplicationPath}/";
             }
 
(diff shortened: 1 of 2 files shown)


david-poindexter (approved):
@valadas thank you so much for digging deep to find a solution that works for both IIS setup scenarios!  🎉 🌮

donker (approved):
Great. Thanks.

PR #5269, opened 2022-09-05
Fixed an issue where Telerik analysis would fail

### LINKED ISSUE #5260: Upgrade to v9.11.0-rc2 may fail at GetReferencedAssemblyNames()
## Description of bug
Upgrade wizard doesn't continue at 2nd step "Checking security aspects of this installation". There's an XHR to /Install/UpgradeWizard.aspx/GetSecurityTab which throws an exception:
`IOException: Could not load assembly '..\bin\runtimes\win-x64\native\imageflow.dll' at DotNetNuke.Maintenance.Telerik.TelerikUtils.GetReferencedAssemblyNames(String assemblyFilePath, AppDomain domain)`

The file path is correct, the file is there, but the DLL may not be readable by .NET. It seems to be a native (COM?) library from [Imageflow](https://www.nuget.org/packages/Imageflow.NativeRuntime.win-x86_64/).

## Steps to reproduce
1. Install DNN 9.6.0 with 2sxc 13 (or newer). It will place a DLL at \bin\runtimes\win-x64\native\imageflow.dll
2. Unzip the DNN Upgrade Package
3. Start the Upgrade Wizard
4. Enter host login credentials
5. Click _Next_
6. It will stuck on the next step "Checking security aspects of this installation"
7. Console will show the error details

## Expected behavior
Skip non readable Assemblies/DLLs in the Telerik checks (`AssemblyDependsOnTelerik()`) or inform the superuser which DLLs he should check/remove manually.

## Error information
XHR to `/Install/UpgradeWizard.aspx/GetSecurityTab` returns HTTP 500

```
System.IO.IOException: Could not load assembly '..\bin\runtimes\win-x64\native\imageflow.dll'
   at DotNetNuke.Maintenance.Telerik.TelerikUtils.GetReferencedAssemblyNames(String assemblyFilePath, AppDomain domain)
   at DotNetNuke.Maintenance.Telerik.TelerikUtils.AssemblyDependsOnTelerik(String path, AppDomain domain)
   at System.Linq.Enumerable.WhereArrayIterator`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at DotNetNuke.Maintenance.Telerik.TelerikUtils.GetAssembliesThatDependOnTelerik()
   at DotNetNuke.Services.Install.UpgradeWizard.GetSecurityTab(Dictionary`2 accountInfo)
```

## Affected version

* [x] 09.11.00 release candidate 2

## Affected browser

* [x] Chrome

valadas:
Closes #5260

Some dlls may be native code and fail when attemping to check if they have dependencies on Telerik.

This PR basically skips those files, if we can't detect if they depend on Telerik, then they can't possibly depend on Telerik.

### DIFF
--- a/DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
+++ b/DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
@@ -95,8 +95,16 @@ namespace DotNetNuke.Maintenance.Telerik
 
         private bool AssemblyDependsOnTelerik(string path, AppDomain domain)
         {
-            return this.GetReferencedAssemblyNames(path, domain)
-                .Any(assemblyName => assemblyName.StartsWith("Telerik", StringComparison.OrdinalIgnoreCase));
+            try
+            {
+                return this.GetReferencedAssemblyNames(path, domain)
+                    .Any(assemblyName => assemblyName.StartsWith("Telerik", StringComparison.OrdinalIgnoreCase));
+            }
+            catch (Exception)
+            {
+                // If we can't get the referenced assemblies, then it can't depend on Telerik.
+                return false;
+            }
         }
 
         private string[] DirectoryGetFiles(string path, string searchPattern, SearchOption searchOption) =>
--- a/DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
+++ b/DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
@@ -10,6 +10,7 @@ namespace DotNetNuke.Maintenance.Telerik
     using System.Linq;
 
     using DotNetNuke.Abstractions.Application;
+    using DotNetNuke.Instrumentation;
 
     /// <inheritdoc />
     public class TelerikUtils : ITelerikUtils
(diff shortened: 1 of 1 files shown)


david-poindexter (approved):
This seems to make sense to me.

valadas:
Basically, https://github.com/valadas/Dnn.Platform/blob/ae119fc1f3fefc072e5de1d4b373f193f3f0c729/DNN%20Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs#L115-L127 already throws an exception so we catch it here to prevent a failure and just accept it does not depend on Telerik

DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
```diff
@@ -1,128 +1,136 @@
...
+        {
+            try
+            {
+                return this.GetReferencedAssemblyNames(path, domain)
+                    .Any(assemblyName => assemblyName.StartsWith("Telerik", StringComparison.OrdinalIgnoreCase));
+            }
+            catch (Exception)
+            {
+                // If we can't get the referenced assemblies, then it can't depend on Telerik.
+                return false;
```
mitchelsellers, line 106: Ideally I would love to see this logged, so they can at least know. As it is not ALWAYS that it cannot be loaded. It could be file in use. Etc.

mitchelsellers (changes requested):

DNN Platform/DotNetNuke.Maintenance/Telerik/TelerikUtils.cs
valadas: Does https://github.com/dnnsoftware/Dnn.Platform/pull/5269/commits/0e70665f658cfa6ef3e62ea55b10b5bdebe8e509 work ? If we want to surface this to the UI, we may have to reachitect that...

mitchelsellers: That would be good enough for me!

mitchelsellers: My goal isn't to make it crazy, but for when things go sideways, we can direct them to the /portals/_default/logs folder

mitchelsellers (approved):

PR #5273, opened 2022-09-06
Remove 9.11 merge workflow

bdukes:
Per https://github.com/dnnsoftware/Dnn.Platform/pull/5268#issuecomment-1237446099


valadas:
Awesome, thanks!

Issue #5266, opened 2022-09-05
Upgrade to v9.11.0-rc2 may delete DLLs from DNN-Extensions like 2sxc with Imageflow


BeniFreitag:
## Description of bug
The Upgrade Wizard deletes DLLs from extensions, so the whole DNN application may not work after the upgrade completed.

## Steps to reproduce
1. Install DNN 9.6.0 with 2sxc 13 (or newer). It will place these files into the bin-folder: Imageflow.Net.dll, Imazen.Common.dll, Imazen.HybridCache.dll and Imageflow (entiry folder)
2. Unzip the DNN Upgrade Package
3. Run the Upgrade Wizard
4. When the Upgrade Wizard completed, the DNN-Application will not start anymore

## Current behavior
DNN application does not start anymore. Error because of missing DLLs.

## Expected behavior
DNN upgrade should not delete DLLs from DNN-Extensions.

## Error information
Error when the Upgrade Wizard completed:
`Could not load file or assembly 'Imazen.Common, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.`

## Additional context
2sxc ships with Imageflow (till version 12) and/or Imazen (version 13+). The installation places additional DLLs into the bin folder. They're own assemblies and not shared with other extensions.

## Affected version

* [x] 09.11.00 release candidate 2

## Affected browser

* [x] Chrome

david-poindexter:
[DNN 9.11.0 RC 3](https://github.com/dnnsoftware/Dnn.Platform/releases/tag/v9.11.0-rc3) is now out.  I am not sure if this is still an issue there or not, but you may want to try that to confirm or deny the existence of the issue.

valadas:
I don't know of any change in 9.11.0 that could cause this. I may be wrong but nothing pops to mind. Before trying another upgrade can you ensure that your install/modules folder has no other .zip files my guess here is that maybe there was some stray 2sxc version of an older or newer version in there that triggered those deletions maybe...

BeniFreitag:
The DLLs are deleted after the Upgrade Wizard has completed, after I click _Visit Website_. It seems there's something in the application startup. [I recorded the behavior here](https://www.screencast.com/t/eRXzFC0gxdz).

[...]

PR #5277, opened 2022-09-06
Fix issue with resources manifest from DNNSecurityHotFix20171

donker:
This fixes issue #5266

DNNSecurityHotFix20171 was badly packaged. Other similarly badly packaged modules might be impacted by uninstalling this module with "delete files" set. So to remedy this, the package is now uninstalled with delete files set to false and then the remaining dll is deleted if present.


mitchelsellers:
@donker It looks like the build broke

valadas (approved):
That look good to me

bdukes (approved):

Issue #5274, opened 2022-09-06
[v9.11.0-rc4] Folder icon for custom folder provider fails to load


lanthonyneville:
## Description of bug
Folder icon for custom folder provider fails to load (404) within the new Resource Manager module. The issue appears to be due to the Resource Manager module requesting the image with .svg extension, when it is defined in the folder provider code (and installed on disk) as .png.

## Steps to reproduce
1. An existing custom folder provider is installed which defines the folder icon path as ~/Providers/FolderProviders/TfsFolderProvider/icon-asset-manager-tfs-folder.png (see code image below)
2. After upgrade to v9.11.0-rc4, within the new Resource Manager module, the folder icon image is broken (see image below)

## Expected behavior
The folder icon should display as it did before the v9.11.0-rc4 upgrade. No change to folder provider code should be required.

## Screenshots
![image](https://user-images.githubusercontent.com/22200046/188649044-b5d60d6d-95d5-4d78-aebc-66a6850d7bb4.png)
![image](https://user-images.githubusercontent.com/22200046/188649999-a1433e1d-7e0c-4632-8d03-ca7348a28f69.png)
![image](https://user-images.githubusercontent.com/22200046/188650138-85b3be03-e778-49a3-af70-f5f85fc6b6da.png)

## Affected version
[ X ] 09.11.00 release candidate

valadas:
If memory serves good, it should look for an SVG and if not found use the png of the same name, but I will try that scenario and report back, maybe we missed something in that logic...

valadas:
Should be resolved in RC5, please comment if you can still reproduce the issue and we'll be happy to reopen
https://github.com/dnnsoftware/Dnn.Platform/releases/tag/v9.11.0-rc5

lanthonyneville:
I installed RC5 and yes, the icons are now loaded, thanks. Also I can see the Add Folder Type button, which is a good catch!

One thing - in both the left and right panels, the image is rendered with no resizing. In my case the icon I provide is a different size than the default so it looks too big in the left panel and too small in the right. Can you force resizing to 20x16px in the left and 50x40px in the right so it looks consistent? I'll open another issue for this if you need.

lanthonyneville:
> I installed RC5 and yes, the icons are now loaded, thanks. Also I can see the Add Folder Type button, which is a good catch!
> 
[...]

PR #5282, opened 2022-09-09
Fixed an issue where folder provider icons would not show

valadas:
Closes https://github.com/dnnsoftware/Dnn.Platform/issues/5274

The new logic is as follows:
1. If we don't have any folder mapping info, we show the standard folder svg
2. If we have in the resource manager an svg for the provider, we use that one (usually nicer and we have one for our own Azure provider connector)
3. If the provider has one, we use the one the provider dictates
If all else fails, we use the standard svg one.

In addition to this, I also added a button so users can create new folder mappings (totally forgot about that part when implementing the feature).


valadas:
This PR was already submitted but incorrectly targetted `develop` instead of `release/9.11.0` branch.

It is OK to have it in both but right know this was not on this branch and I kinda need it for my next PR.

I am self-merging it as it was already approved.

PR #5292, opened 2022-09-12
Fix for #5291: Non-Integer Thumbnail Sizes Break API Calls

### LINKED ISSUE #5291: Non-Integer Thumbnail Sizes Break API Calls
## Description of bug
Non-Integer Thumbnail Sizes Break API Calls

## Steps to reproduce

1. Try zooming out several times in your browser.
2. Reload a DNN Page with the Assets page of the Persona Bar open.
3. If the thumbnails fail to load, look in the network tab for the failing calls.

It is easy to reproduce the failed API calls with non-integer parameters: make an API call to http://{your dnn install}.dnndev.me/DnnImageHandler.ashx?mode=securefile&fileId={a file id}&MaxWidth=74.98&MaxHeight=42.01

## Current behavior
API calls return with HTTP 500

## Expected behavior
API parses the non-integer parameters and returns thumbnail based on rounded dimensions

## Screenshots
See here https://docs.google.com/document/d/1DgWvWwNloNGleA7zpt6sqb4fT_W1V9gSwzmZhRBWKBg/edit

## Error information
n/a

## Additional context
n/a

## Affected version

* [x] 10.00.00 alpha build
* [x] 09.11.00 release candidate

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ ] Microsoft Edge Chromium

v-karbovnichy:
Fixes #5291

## Summary

Implements a function ParseDimension to allow non-integer values for width and height, return a rounded value when decimals are used, and a DefaultDimension value when it is not a number.

Unit tests added.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5292) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5292) before we can accept your contribution.<br/><br/>:x: v-karbovnichy [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5292)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=5292) it.</sub>

mitchelsellers:
This is a behavior change for sure, I think we need to hold/merge this with a 10.0.0 release.  But option to discussion

valadas:
@mitchelsellers it looks to me like this is a bugfix, previously it would return a 500 and now it returns a sensible dimension. My only concern is performance, but this method should not be hammered much and it has client cache enabled by default. Unless I am missing something...

mitchelsellers:
@valadas I could see that, but we are going from "it must be an integer or it will error" to "it could be any type of number and we guess, or fall back"

v-karbovnichy:
@mitchelsellers I was initially against such code change (quoting internal ticket discussion):
>I think the backend API is totally fine to throw HTTP 500 when the caller asks to get a thumbnail with non-integer width and height, because the resulting image has discrete dimensions. It’s simply a violation of the API contract to ask for an image with non-integer size.

but then got a guidance from my top management which I had to follow:
>Why do I disagree with the rationale: the backend needs to be robust and always return a thumbnail. No thumbnail is worse than rounding up/down and breaking the layout, which would be broken anyway " for some customers when they heavily adjust the zoom and text-size on their browser."

valadas:
Yeah, I agree that I would prefer for the calling code to send an integer there.

bdukes:
@v-karbovnichy this branch now has conflicts, and the maintainers haven't been given access to make changes to the branch, so we can't merge this PR. Please merge these changes with `develop` and then we can merge the PR. Thanks!

v-karbovnichy:
@bdukes I had resolved the conflict, please proceed.

bdukes (approved):

PR #5295, opened 2022-09-13
Fix missing objectqualifier and db owner in script

donker:
Title says it all


bdukes:
@donker looks like we were both looking at this at the same time (see #5296). I think your change shouldn't have a `.` after `{databaseOwner}`, right?

sleupold:
I agree with @bdukes regarding the .

donker:
Superceded by another PR

PR #5297, opened 2022-09-13
Exclude web.*.config from distribution zips

donker:
As the title says


Build/Tasks/packaging.json
```diff
@@ -48,6 +49,7 @@
     "/favicon.ico",
     "/Robots.txt",
     "/web.config",
+    "/web.*.config",
```
bdukes, line 52: We shouldn't need this line. This is only what we need to exclude from the install package, the install excludes are applied automatically, https://github.com/dnnsoftware/Dnn.Platform/blob/e0b483e06c4c0a26ed3c6b27598c05436b0db3a5/Build/Tasks/CreateUpgrade.cs#L28-L30
```suggestion
```

bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #5314, opened 2022-09-22
Include SharpZipLib package

bdukes:
Resolves issue mentioned at https://github.com/dnnsoftware/Dnn.Platform/pull/5185#issuecomment-1254967053

## Summary
In #5185 SharpZipLib was removed as a dependency of DNN (except where required by deprecated APIs). A library package was introduced to manage upgrading the dependency, but that package was mistakenly left out of the distribution. This PR resolves that oversight.


david-poindexter (approved):
Good catch @bdukes 🎉

mitchelsellers (approved):

PR #5351, opened 2022-10-23
Fixed an issue where resource manager would not have appropriate root folder

### LINKED ISSUE #4825: Resource Manager module in Group mode allows user to navigate outside of their Group folder
## Description of bug
ResourceManager module in Group mode allows the user to navigate back up the folder tree. I would expect it to not allow this so the user can only see what's in their group folder.

## Steps to reproduce
1. Add the Resource Manager onto the Group Activity page of a DNN install.
2. Via the Resource Manager module settings, set the Mode to "Group" and save.
3. Visit the page via the Groups Directory and it will set the default folder of the Resource Manager to the appropriate Groups folder.
4. Via the Folder tree control in the header of the Resource Manager module, you can navigate back up the folder tree.

## Current behavior
Allows a group user to navigate outside of their group folder.

## Expected behavior
Lock the user into the Group folder

## Affected version

* [ ] 10.00.00 alpha build

valadas:
Resolves an issue where in group/user/folder mode the folder list would always start at the site root instead of the folder dictated by the module settings.

Closes #4825

User mode:
image

Site Assets:
image

Global Assets:
image

Group mode:
image

Specific folder mode:
image


david-poindexter (approved):
This is FANTASTIC @valadas - thank you!!!

mitchelsellers (approved):

mitchelsellers:
@valadas Looks like we have a build error

PR #5368, opened 2022-11-02
Modify Next Button on install wizard

uzmannazari:
## Summary
Modification of the next button in the get package information step, whose name could not be called from language resource files and could not be translated in multilingual portals.

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Extensions.Web/src/components/InstallExtensionModal/index.jsx
+++ b/Dnn.AdminExperience/ClientSide/Extensions.Web/src/components/InstallExtensionModal/index.jsx
@@ -149,7 +149,7 @@ class InstallExtensionModal extends Component {
                 onCancel={this.cancelInstall.bind(this)}
                 installationMode={true}
                 onSave={this.goToReleaseNotes.bind(this)}
-                primaryButtonText={Localization.get("Next")}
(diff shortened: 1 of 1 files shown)


mitchelsellers:
Can we get more detail on this particular change?  I don't see any addition to the resource files to then set this value, yet it was pulling from localization files already.

uzmannazari:
{Localization.get("Next.Button")} just added. 
The element is already available in the language resource file

valadas (approved):
Correct, fiund it here https://github.com/dnnsoftware/Dnn.Platform/blob/06e467e84fad58805e0fb04b10ec49d09e9a7144/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Extensions/App_LocalResources/Extensions.resx#L1068

LGTM

mitchelsellers (approved):
Thanks for the extra information that this was already in the language pack

uzmannazari:
> Correct, fiund it here
> 
> https://github.com/dnnsoftware/Dnn.Platform/blob/06e467e84fad58805e0fb04b10ec49d09e9a7144/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Extensions/App_LocalResources/Extensions.resx#L1068
> 
> LGTM

Thanks mr @valadas and @mitchelsellers

PR #5369, opened 2022-11-02
Modify search placeholder

uzmannazari:
## Summary
Modification of pages search place holder that can be read from language resource files.

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
@@ -44,7 +44,7 @@ class SearchPageInput extends Component {
                             value={this.state.searchTerm}
                             onChange={this.onSearchFieldChange.bind(this)}
                             onKeyPress={(e) => { e.key === "Enter" ? this.props.onSearch(this.state.searchTerm) : null; }}
-                            placeholder="{Localization.get("Search")}" />
+                            placeholder={Localization.get("Search")} />
                     </div>
                     {this.state.searchTerm ?
                         <div
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
@@ -44,7 +44,7 @@ class SearchPageInput extends Component {
                             value={this.state.searchTerm}
                             onChange={this.onSearchFieldChange.bind(this)}
                             onKeyPress={(e) => { e.key === "Enter" ? this.props.onSearch(this.state.searchTerm) : null; }}
-                            placeholder="Search" />
+                            placeholder="{Localization.get("Search")}" />
                     </div>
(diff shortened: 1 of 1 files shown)


mitchelsellers (changes requested):
Similar question to the other PR if you are moving this to be pulled from localization we need to update the localization files to include this element.

uzmannazari:
> Similar question to the other PR if you are moving this to be pulled from localization we need to update the localization files to include this element.

The element is already available in the language resource file

valadas (approved):
Yep appears to be here https://github.com/dnnsoftware/Dnn.Platform/blob/935540d3d05905113532586fe7285e74abbfe137/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Pages/App_LocalResources/Pages.resx#L177

Looks good to me

mitchelsellers (approved):

Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
```diff
@@ -44,7 +44,7 @@ class SearchPageInput extends Component {
                             value={this.state.searchTerm}
                             onChange={this.onSearchFieldChange.bind(this)}
                             onKeyPress={(e) => { e.key === "Enter" ? this.props.onSearch(this.state.searchTerm) : null; }}
-                            placeholder="Search" />
+                            placeholder="{Localization.get("Search")}" />
```
valadas, line 47: I don't think the double-quotes should be kept here, did you test this out ?

valadas (changes requested):

Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
uzmannazari: No, unfortunately, I could not test. I do not have the test conditions.

valadas:
You just have to run ./build.ps1 and get the installer from the artifacts folder and install a site with it. I am pretty sure those quotes need to be removed or it will just literally show that text.

Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
david-poindexter: > I don't think the double-quotes should be kept here, did you test this out ?

I agree 100% regarding the outer quotes.

david-poindexter:
The build is failing anyway.  I am pretty sure those outer quotes need to be removed.

Dnn.AdminExperience/ClientSide/Pages.Web/src/components/SearchPage/SearchPageInput.jsx
mitchelsellers, line 47:
```suggestion
                            placeholder={Localization.get("Search")} />
```

uzmannazari (commented):
I cant test personabar but i think this will work fine

mitchelsellers (approved):

valadas (approved):

Issue #5371, opened 2022-11-03
Request IP address not being handled correctly when using ARR (i.e. Azure App Service)


davidjrh:
## Description of bug
When hosting a DNN website on a platform using a load balancer that forwards the requests with the header "X-Forwarded-For" (i.e. the ARR used by Azure App Service), the remote IP Address in the header comes in the format "WWW.XXX.YYY.ZZZ:Port". The current logic validates this header to a v4 IP Address and returns an empty string because the port is not correctly handled. 

This causes lot of DNN features based on the Remote IP Address not working properly, like: 
- storing the LastIPAddress field on Users table 
- use of "Security > Login IP Filters" feature
- storing the IP Address on the event log properties on some type of requests
- and many more locations where the ```UserRequestIPAddressController.GetUserRequestIPAddress``` method is used (counting 23 references on the DNN Platform solution)

## Steps to reproduce
1. Install a DNN Platform website on Azure App service
2. Login into the website
3. Explore the Users table
4. Note that the LastIPAddress field is empty

## Current behavior
Any functionality based on the ```UserRequestIPAddressController.GetUserRequestIPAddress``` method doesn't work because the function returns an empty string.

## Expected behavior
The method ```UserRequestIPAddressController.GetUserRequestIPAddress``` should return the IP address part.

## Additional context
Reproduced on any version of DNN Platform hosted on Azure App Service.

## Affected version
* [X] 10.00.00 alpha build
* [X] 09.11.00 release candidate
* [X] 09.10.02 release candidate
* [X] 09.10.01 latest supported release

valadas:
Implemented in #5371

PR #5372, opened 2022-11-03
Fix issue #5371 to correctly handle Request IP Addresses

davidjrh:
Fixes issue #5371.

## Summary
Added code to correctly parse the request IP Address when original IP Address comes in the request header "X-Forwarded-For" (i.e. when using an ARR like in an Azure App Service) by removing the port.

### DIFF
--- a/DNN Platform/Library/Services/UserRequest/UserRequestIPAddressController.cs
+++ b/DNN Platform/Library/Services/UserRequest/UserRequestIPAddressController.cs
@@ -31,6 +31,10 @@ namespace DotNetNuke.Services.UserRequest
             {
                 userIPAddress = request.Headers[userRequestIPHeader];
                 userIPAddress = userIPAddress.Split(',')[0];
+                if (userIPAddress.Contains(':'))
+                {
+                    userIPAddress = userIPAddress.Split(':')[0];
+                }
             }
 
             if (string.IsNullOrEmpty(userIPAddress))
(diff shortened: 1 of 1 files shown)


valadas (commented):
Hmm, not sure if it's an issue or not but what about IPv6 addresses that use : as a separator ?

uzmannazari (approved):

bdukes:
That's a great point @valadas. I think we need to try calling `this.ValidateIP(userIPAddress, ipFamily)` and only if it fails do we try cutting off after the colon.

davidjrh:
Good point @valadas. I've moved the check to the `this.ValidateIP(userIPAddress, ipFamily)` methods as @bdukes suggested. Need to to it before`IPAddress.TryParse(ipString, out address)` because that method fails if contains an IPv4 with colons.

davidjrh:
Finnally brought back the port removal outside the ValidateIP method since it must return the IP without the port (returning the IP with the port is not for what the method is currently used)

valadas:
That looks better, it only leaves the possibility of an IPv6 with a port which would look like
`http://[1fff:0:a88:85a3::ac1f]:8001/index.html` which is a bit more of a pickle because the `[]` are optional and only needed if a port is specified.
So I guess if it's IPv6 we could split check if there is a `]`, if so we split on it and remove the first character `:` or something like that.

mitchelsellers (approved):

davidjrh:
Hi @valadas, I've added code to handle the same situation for IPv6 addresses, following your thoughts.

PR #5380, opened 2022-11-08
Fixed an issue where comments in manifest would make install fail

valadas:
If a package manifest would have a comment on the level of the first node, the installer would try to parse the comment instead of the real non-comment node.

0f60ce3a8ba7985e8fb835a9570cbd6855cc6b65 Fixes the actual issue and
50bb9e9328fc4af567e4e1125c41140e2d7c6712 just cleans-ups the file a bit and adds some documentation


bdukes (approved):

mitchelsellers (approved):

david-poindexter (approved):
Excellent catch!

PR #5385, opened 2022-11-11
Fix casing issue on SelectedItemTemplate property

donker:
In a previous PR the backing field has been changed and inadvertently the code in this property was set to the property name which creates an infinite loop.


mitchelsellers (approved):

valadas (approved):
Great catch

PR #5337, opened 2022-10-15
Apply analyzer (StyleCop) auto-fixes

bdukes:
## Summary
I ran Code Cleanup from Visual Studio 2022 with the _Fix analyzer warnings and errors set in EditorConfig_ option, and then resolved a couple of issues it created. I've closely reviewed and tweaked about half of the files.

It primarily adjusting naming conventions (removing underscores, making arguments and privates lowercase).

I also fixed an incorrect reference of SharpZipLib in Dnn.PersonaBar.Extensions (not sure why it wasn't causing trouble before).

PR #5390, opened 2022-11-11
Resolve properties made incorrectly recursive

bdukes:
PR #5337 implemented auto-fixes for style issues, but the renaming of backing fields sometimes caused a conflict which resulted in the property becoming recursive


valadas (approved):
Good catch, I love the more concise syntax too

donker (approved):
Awesome

PR #5396, opened 2022-11-13
Fix the library debug build to debug folder

donker:
Since the addition of library.build to the Library project, the old build logic with a copy to the debug folder is bypassed. This PR fixes that.


valadas (approved):
Awesome, I have been wondering what was happening there

bdukes (approved):

Issue #5401, opened 2022-11-22
Adding SEO URL's to pages causes extra 'automatic' rows to be added to grid


mwigley-trilogy:
## Description of bug

Whenever a URL is added using the 'pages => advanced => S.E.O.' dialog it addes duplicate copies of the existing 'Automatic' rows. Also, duplicate copies of 'Automatic' rows exist for each language that is enabled for the page regardless of any manually added URLs.

![image](https://user-images.githubusercontent.com/116310985/203433366-67e04bcb-bd3e-4242-8b9b-d5c37db6dd51.png)

## Steps to reproduce

Prerequisite:
Have the languages feature enabled and enable multiple languages for a non-home page:
![image](https://user-images.githubusercontent.com/116310985/203432024-e89d7129-c02b-41ac-b54c-144d4b1101b2.png)

1. Open the Pages -> Avanced -> SEO page for a non-home page
![image](https://user-images.githubusercontent.com/116310985/203430357-332d0d2c-7759-4b88-917a-13fdad3a8bdd.png)
2. Note the number of 'Automatic' URL's - there will be one per language.  This is the first issue - if multiple languages are enabled on the site you will see repeated indiscernable rows:
![image](https://user-images.githubusercontent.com/116310985/203432109-124f41ac-d7c6-40d5-83dc-a7a431c85b57.png)
3. Add a new URL
![image](https://user-images.githubusercontent.com/116310985/203430663-4c722141-1809-4627-ba8a-e957ab44ff84.png)

An extra version of the automatically generated URL appears:

![image](https://user-images.githubusercontent.com/116310985/203430773-20aa966b-cb5e-43c1-9629-b83c592dd5b4.png)

And new ones keep appearing as you add more SEO Urls:

![image](https://user-images.githubusercontent.com/116310985/203437173-25ffa439-a1ee-4c23-81aa-0c8c12fbb83a.png)

## Current behavior

1. Currently when there are multiple languages for a page, extra identical looking SEO URL rows are displayed for a page.

This appears to be because an 'automatic' page URL is generated for each language, and these are all displayed in the UI:
![image](https://user-images.githubusercontent.com/116310985/203432414-5fb1b103-dcd9-4654-8220-89b77153320d.png)

2. Currently when a new SEO URL is added existing rows get duplicated.

This appears to be because 'automatic' URL's are keyed by index, and this can cause issues in React:
[...]

PR #5402, opened 2022-11-22
Fix issues that can cause duplicate rows to show in page->advanced->SEO

mwigley-trilogy:
## Summary

Fix issues that can cause duplicate rows to show in page->advanced->SEO.

The 'extra duplicate rows appearing after adding new SEO URL' issue is fixed by keying by: `url_${url.id}_${url.path}_${url.statusCode.Key}`. This will be unique for non-automatic URLs because of the {url.id}, and is unique for automatic URLs because of the path and statusCode. It means that the human generated ones will be detected as new rows if the path is edited, but I don't think that's a big deal and this keeps the code simple.

The 'duplicated looking rows show in SEO dialog with multiple languages' issues is fixed by only including unique rows in the dialog. I could have changed it to show the language that's causing the duplicate row in the dialog, but I the extra rows aren't particularly relevant anyway so it makes sense to just strip them out.

Fixes https://github.com/dnnsoftware/Dnn.Platform/issues/5401

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/Table.jsx
+++ b/Dnn.AdminExperience/ClientSide/Pages.Web/src/components/Seo/PageUrls/Table.jsx
@@ -5,19 +5,25 @@ import Localization from "../../../localization";
 import UrlRow from "./UrlRow";
 import EditUrl from "./EditUrl";
 import styles from "./style.less";
+import uniqBy from "lodash/uniqBy";
 
 class Table extends Component {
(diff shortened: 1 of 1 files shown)


mwigley-trilogy:
@microsoft-github-policy-service agree [company="trilogy"]

mwigley-trilogy:
@microsoft-github-policy-service agree company="trilogy"

bdukes:
@mitchelsellers @valadas it looks like the CLA bot check needs to be adjusted so this can get merged

PR #5409, opened 2022-11-26
Added host level setting for browser and upload folders of html editor

### LINKED ISSUE #5408: CKEditor host level setting for browser and upload folders
## Description of problem
In CKEditor Provider Settings (through Personabar Site Settings) you can choose to change the settings for all portals:
image

However, if you do that, the dropdown lists for folders remain unchanged, containing the folders that exist for the current site:
image

Uploading to such a folder would obviously only work in the current site.

## Description of solution
I propose to add string settings for these folders, where a host can configure paths used for all sites. These will override the per-portal setting when used. We would implement just-in-time creation of these folders when they're needed. Setting will be visible for site admins, but can only be changed by host users.

I concidered the option to allow site admins to overide this setting, but that doesn't make sense for the use case where the host wants to force usage of certain folders.

## Screenshots
Added settings would look like this:
image
(also for Upload folders

skamphuis:
Closes #5408 

## Summary
Backgroup of the changes can be found in #5408 
In summary I added 4 settings for host level, and implmented their usage on opening the file browser, and on uploading files.

### DIFF
--- a/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/App_LocalResources/Options.aspx.resx
+++ b/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/App_LocalResources/Options.aspx.resx
@@ -518,4 +518,19 @@
   <data name="HostNoSettingLabel.Text" xml:space="preserve">
     <value>Contains No Host Settings</value>
   </data>
+  <data name="BrowserRootFolderForImg.Text" xml:space="preserve">
+    <value>Browser Root Folder for images</value>
+  </data>
+  <data name="HostBrowserRootFolder.Text" xml:space="preserve">
+    <value>Browser Root Folder Name for all portals (overrides setting below)</value>
+  </data>
+  <data name="HostBrowserRootFolderForImg.Text" xml:space="preserve">
+    <value>Browser Images Root Folder Name for all portals (overrides setting below)</value>
+  </data>
+  <data name="HostUploadFolderLabel.Text" xml:space="preserve">
+    <value>Default Upload Folder for all portals (overrides setting below)</value>
+  </data>
+  <data name="HostUploadFolderLabelForImg.Text" xml:space="preserve">
+    <value>Default Upload Folder for images for all portals (overrides setting below)</value>
+  </data>
 </root>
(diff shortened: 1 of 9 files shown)


david-poindexter (approved):
This looks good to me - thanks @skamphuis 🎉

mitchelsellers (approved):

valadas (commented):
The code looks ok to me.

Just a couple of questions.
I understand that the host level settings here win, so say I have 20 portals with mixed folder settings, if I change the host level setting, then that changes it for all portals. Now let's suppose that some admins of these portals don't want that host-level folder or the host was accidentally playing with the feature...

1. Is it right for the host setting to win over the portal level setting ? Or should it be a fallback instead of an override ?

2. Assuming this is the desired behaviour, can it be reversed easily, like unsettling the host level folders ?

valadas:
@skamphuis did you see my comments above ?

skamphuis:
> @skamphuis did you see my comments above ?

Did now, thanks.

skamphuis:
> The code looks ok to me.
> 
> Just a couple of questions. I understand that the host level settings here win, so say I have 20 portals with mixed folder settings, if I change the host level setting, then that changes it for all portals. Now let's suppose that some admins of these portals don't want that host-level folder or the host was accidentally playing with the feature...

The use case actually is a host that needs/wants to enforce these settings onto all portals, which I can imagine from a support point of view where the host is supporting many portals being administered by less-techy admins.

> 
>     1. Is it right for the host setting to win over the portal level setting ? Or should it be a fallback instead of an override ?

It was intended for the host setting to win.

> 
>     2. Assuming this is the desired behaviour, can it be reversed easily, like unsettling the host level folders ?

The host can obviously empty the settings again to reverse the behaviour. An admin has no power over that. We could add an extra setting, available for the host, to allow an admin to override. Is that what you meant?

valadas:
> We could add an extra setting, available for the host, to allow an admin to override. Is that what you meant?

Not necessarily, just wanted to know the intention here to make sure everyone is on the same page. This is fine for my use cases, I'll just hold on merging it to make sure the other reviewers understand this and are fine with it. Thanks for the update.

@dnnsoftware/approvers we may want to discuss this one on the next meeting just to make sure everyone is fine with this.

PR #5411, opened 2022-11-27
Added support for a Notes field in IP Filters

### LINKED ISSUE #4140: Column for note in IP filter
## Description of problem
If you add IP addresses to the IP filter you can not add a note why you added that IP. So, after a couple of months, there is a bunch of IP's and you have no idea if they are still needed. For instance an IP from someone's home address but they have moved. 

## Description of solution
When adding an IP address I'd like te add a note which will be visible in the IP filter list.

valadas:
Closes #4140

Hosts can now set a note with each IP Filter rule to remind them about the purpose of each rule.

image

### DIFF
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -3694,9 +3694,32 @@ namespace DotNetNuke.Data
             return this.ExecuteReader("GetIPFilters");
         }
 
+        /// <summary>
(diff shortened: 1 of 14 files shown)


david-poindexter (approved):
Awesome @valadas - thank you!

mitchelsellers (approved):

skamphuis:
Nice one. Did you notice, @Tychodewaard ?

PR #5412, opened 2022-11-28
Fix renamed public/protected properties and fields

bdukes:
## Summary
PR #5337 changed the casing of some `public` and `protected` field and property names. These are breaking changes, as pointed out by @skamphuis. This PR resolves these renames.


valadas (approved):
Looks good to me

mitchelsellers (approved):

PR #5414, opened 2022-11-29
Keep track of whether in global assets mode

bdukes:
## Summary
I haven't had a chance to test this, but I believe this should pass through to the API whether the module is in Global Assets mode or for the current portal.

### DIFF
--- a/DNN Platform/Modules/ResourceManager/Components/SettingsManager.cs
+++ b/DNN Platform/Modules/ResourceManager/Components/SettingsManager.cs
@@ -58,26 +58,30 @@ namespace Dnn.Modules.ResourceManager.Components
         /// </summary>
         public int Mode { get; set; }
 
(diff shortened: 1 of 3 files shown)


valadas (approved):
You are awesome!

valadas:
It looks right but we have another problem (don't let this hold merging this PR as it's probably not caused by this)...
But we are now missing all the scripts in the module zip:
image

I'll try to figure out what happened with that the next weekend...

mitchelsellers (approved):

valadas:
> I'll try to figure out what happened with that the next weekend...

Hmm, I tried a build from develop and fixed some stuff on another PR and the files were there, nothing in this PR appears to possible be able to break that. @bdukes could you rebase on develop to see if the files are there with a new build that includes that fix...

bdukes:
I've rebased @valadas

PR #5421, opened 2022-12-05
Possibly fixed intermittent build issues

valadas:
I was consistently experiencing locally the build issues we sometimes (well oftentimes) get on azure (can't resolved Dnn.Platform\Dnn.AdminExperience\Dnn.PersonaBar.Extensions\Dnn.PersonaBar.Extensions.csproj)...

So I was trying to pinpoint the possible cause of that issues and when I removed the Project GUIDS from this project file, the build started working. Many of the other .csproj files don't have any ProjectType defined at all and they work fine.

I have confirmed that the install files are still being produced.

One of the GUIDs defines this as an MVC project but it is clearly not, the other defines it as a C# project but the .csproj extension already defines that.

Is anyone is opposed to remove both, then I can bring back the one starting with FAE04....

I am absolutelly not sure this will fix the intermittent build issues, because of them beeing intermittent but if it works, we may have fixed it...


mitchelsellers (approved):
Build runs for me, worth a shot!

bdukes (approved):
worth a try

PR #5433, opened 2022-12-11
Changed build to run all npm builds before backend builds.

valadas:
This makes `yarn install` and `yarn build` before any project builds. It still uses lerna as before.

- I confirmed that a failed install makes the build fail.
- I confirmed that a failed npm build makes the build fail.
- I confirmed that it fixed the resource manager scripts missing from the final build.

The drawback of this approach is that one can no longer build only the persona bar in release mode without using cake. But for dev/debug, the current documented way with `yarn start --scope ...` is still valid.


david-poindexter (approved):
Thank you @valadas - this looks like the lessor of evils here.

DNN_Platform.sln
```diff
@@ -500,6 +500,9 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DotNetNuke.Providers.Caching.SimpleWebFarmCachingProvider", "DNN Platform\Providers\CachingProviders\DotNetNuke.Providers.Caching.SimpleWebFarmCachingProvider\DotNetNuke.Providers.Caching.SimpleWebFarmCachingProvider.csproj", "{2C25580C-A971-4F0B-9F70-436A35C2473E}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dnn.Modules.ResourceManager", "DNN Platform\Modules\ResourceManager\Dnn.Modules.ResourceManager.csproj", "{7D61A32C-0F21-453F-A981-BD8E5A3A5304}"
+	ProjectSection(ProjectDependencies) = postProject
+		{9CCA271F-CFAA-42A3-B577-7D5CBB38C646} = {9CCA271F-CFAA-42A3-B577-7D5CBB38C646}
+	EndProjectSection
```
bdukes, line 505: Do we know what this change is?

Build/BuildScripts/AEModule.build
```diff
@@ -1,14 +1,12 @@
 ﻿<Project ToolsVersion="4.0" DefaultTargets="Build" 
   xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Yarn.MSBuild.1.22.17\build\Yarn.MSBuild.props" Condition="Exists('..\..\packages\Yarn.MSBuild.1.22.17\build\Yarn.MSBuild.props')" />
-  <Import Project="..\..\packages\Yarn.MSBuild.1.22.17\build\Yarn.MSBuild.targets" Condition="Exists('..\..\packages\Yarn.MSBuild.1.22.17\build\Yarn.MSBuild.targets')" />
```
bdukes, line 4: Should we remove the `Yarn.MSBuild` package entirely?

valadas: Well, if we go with this solution then it is unused...

DNN_Platform.sln
valadas: I have no clue, it was done by Visual Studio...

donker (approved):
I quite like moving the npm build from being hidden in some build file to the main scripts

PR #5439, opened 2022-12-13
Remove now-unused Yarn.MsBuild package

bdukes:
## Summary
As a follow-up to #5433, this PR removes the Yarn.MsBuild package that is no longer used.


valadas (approved):
Oh, I missed that. And thanks for the simplification, had a hard time navigating the Cake documentation...

david-poindexter (approved):

PR #5441, opened 2022-12-13
Fix PersonaBar project referencing wrong version of Newtonsoft.Json

bdukes:
## Summary
All references to Newtonsoft.Json are for version 13.0.2 except Dnn.PersonaBar.Extensions which references 13.0.1. This PR resolves the issue


valadas (approved):
Good catch

david-poindexter (approved):

PR #5461, opened 2022-12-29
Fixes an issue where URLControl checks upload permissions wrong

### LINKED ISSUE #5423: UrlControl issue: cannot upload for any role (Only works for admin/Superuser)
## Description of bug
Upgraded to DNN 9.11
***Form & List module*** still using `UrlControl` on one of their field, and it seems there's an issue on field with type "URL".
It doesn't allow users who is not "Admin" or "Superuser" to Upload file.

This is not an issue with Form & List module, because the issue within `UrlControl`
It happen even if the User `Role` have "Browse, View & Write" permission to folder.
Whenever press "Upload New File" it throw error "You don't have permission to upload files to this site."
This is not happen for "Admin" or "Superuser"

After browsing through the source code, this line seems check if user has "ADD" permission on the folder
https://github.com/dnnsoftware/Dnn.Platform/blob/c7b04652d214bd40a8ee6c36f133f7054ae70017/DNN%20Platform/Library/UI/UserControls/URLControl.cs#L923
Don't know if this is the cause of the issue.
But somehow it always throw "You don't have permission to upload files to this site." on every folder, even if it was "My Folder"

I check table "Permission" and I don't see any PermissionKey "ADD" for SYSTEM_FOLDER, it only have "READ, WRITE & BROWSE"
image

So whenever it call SP `GetFoldersByPermissions` it always return NULL, because there's no "ADD" permission.

## Steps to reproduce
1. Using "Form & List" module, create a new form
2. At least one field type set to "URL"
3. Set module permission "Create Record/ Submit Form" for certain role 
4. Set the same role to have "Browse, View & Write" permission to folder.
5. "Add new Record", and change Link Type to "File", change folder and press "Upload new file"
6. It throw "You don't have permission to upload files to this site." immediately

image

skamphuis:
Fixes #5423

## Summary
The URLControl tries for "ADD" permission on the selected folder, but that doesn't exist (anymore?). It changed it to "WRITE" instead, which seems logical because that's actually the permission you're controlling by giving a user or role Edit permissions on a folder. This seems to fix the issues mentioned in #5423 and in [Links](https://github.com/DNNCommunity/DNN.Links/issues/65) and [Documents](https://github.com/DNNCommunity/DNN.Documents/issues/21).

However, I did find other references to the "ADD" permission in several other places:
- ResourceManager.PermissionsManager
- InternalServices.FileUploadController (but that checks for both ADD and WRITE)
- Modules.Export.ascx.cs

These might need changing too. Happy to do so, by the way. Just let me know.

### DIFF
--- a/DNN Platform/Library/UI/UserControls/URLControl.cs
+++ b/DNN Platform/Library/UI/UserControls/URLControl.cs
@@ -920,7 +920,7 @@ namespace DotNetNuke.UI.UserControls
         protected void cmdUpload_Click(object sender, EventArgs e)
         {
             string strSaveFolder = this.cboFolders.SelectedValue;
-            this.LoadFolders("ADD");
+            this.LoadFolders("WRITE");
             if (this.cboFolders.Items.FindByValue(strSaveFolder) != null)
             {
                 this.cboFolders.Items.FindByValue(strSaveFolder).Selected = true;
@@ -944,7 +944,7 @@ namespace DotNetNuke.UI.UserControls
                 else
                 {
                     // reset controls
-                    this.LoadFolders("BROWSE,ADD");
+                    this.LoadFolders("BROWSE,WRITE");
                     this.cboFolders.Items.FindByValue(strSaveFolder).Selected = true;
                     this.cboFiles.Visible = true;
                     this.cmdUpload.Visible = false;
@@ -1404,7 +1404,7 @@ namespace DotNetNuke.UI.UserControls
 
                         if (this.ViewState["FoldersLoaded"] == null || this.doReloadFolders)
                         {
-                            this.LoadFolders("BROWSE,ADD");
+                            this.LoadFolders("BROWSE,WRITE");
                             this.ViewState["FoldersLoaded"] = "Y";
                         }
 


sleupold:
Add permission is part of the advanced permission provider and should get mapped to write in the community provider.

skamphuis:
> Add permission is part of the advanced permission provider and should get mapped to write in the community provider.

Right. So the URLControl needs to check for ADD like it does without this PR, but the PermissionProvider needs to check for WRITE if ADD was requested by calling code?

sleupold:
> > Add permission is part of the advanced permission provider and should get mapped to write in the community provider.
> 
> Right. So the URLControl needs to check for ADD like it does without this PR, but the PermissionProvider needs to check for WRITE if ADD was requested by calling code?

yes

skamphuis:
> > > Add permission is part of the advanced permission provider and should get mapped to write in the community provider.
> > 
> > 
> > Right. So the URLControl needs to check for ADD like it does without this PR, but the PermissionProvider needs to check for WRITE if ADD was requested by calling code?
> 
> yes, that is correct

Hmm, doesn't look like that's true, because the LoadFolders method in the UrlControl uses the FolderManager which goes straight to the DB and doesn't use the PermissionProvider. Also, that looks like checks need to be somewhat more complex, like: if ADD permission exists in Permissions, then check for ADD permission, otherwise check for WRITE permission.

Any advice on how to fixed this further would be appreciated.

mitchelsellers (approved):

david-poindexter (approved):
Wow - hard to believe this has been this way for so long!  Thanks @skamphuis

PR #5465, opened 2023-01-04
Correctly save the attribute maxAllowedContentLength

### LINKED ISSUE #4839: After upgrade to 9.10.1 Error SecurityController.UpdateOtherSettings
## Description of bug
After upgrade to 9.10.1 Error SecurityController.UpdateOtherSettings

ExceptionType:System.NullReferenceException,StackTrace:   bij DotNetNuke.Common.Utilities.Config.SetMaxUploadSize(Int64 newSize)

## Steps to reproduce
1. Go to Security settings more settings where e.g. max file upload size is
2. Open Console in F12
3. Change something and Save
4. Error 500 in network tab on UpdateOtherSettings

## Expected behavior
No Error 500

## Error information
I would propose to add some checking in the code so if the attribute is missing then add it to the node
in the public static void SetMaxUploadSize(long newSize)

image

Basically in the web config it looks like

image

(please note that often the attribute maxAllowedContentLength is not there and just falls back to its default)
So the code wants to set an attribute that is not there and should be added in this case.

## Affected version

* [ ] 10.00.00 alpha build
* [X] 09.10.01 latest supported release

ryanjagdfeld:
Fixes #4839

## Summary
When saving the security settings, an error could occur if the web.config file has not had the attribute maxAllowedContentLength for node security//requestFiltering//requestLimits.  On some occasions users may have added the node but not the attribute.  This fix checks for the attribute, if it does not exist it adds it to the node before setting the value.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/Config.cs
+++ b/DNN Platform/Library/Common/Utilities/Config.cs
@@ -354,6 +354,11 @@ namespace DotNetNuke.Common.Utilities
                        configNav.SelectSingleNode("configuration//location//system.webServer//security//requestFiltering//requestLimits");
             if (httpNode != null)
             {
+                if (httpNode.Attributes["maxAllowedContentLength"] == null)
+                {
+                    httpNode.Attributes.Append(configNav.CreateAttribute("maxAllowedContentLength"));
+                }
+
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me

bdukes (approved):
Thanks!

PR #5470, opened 2023-01-08
Convert module icons to svg + Add Metrial Icons in SVG format.

### LINKED ISSUE #5469: Update Old Module icons.
## Description 
The old module icons look very old and have irritated me for years.  They shoudl be changed to something modern, that looks more professional.

## Description of solution
Get SVG icons from Material Icons into the ./images folder.
Update ModuleInstanceContext.cs, chnage .gif to .svg.  Keep icons names the same.

All old icons/images should not be removed for legacy.

## Screenshots
image

image

## Additional context
SVG icons should be 24px.  The smaller version looks odd compared the to other elements s in the admin page.

leedavi:
`Closes #5469`. 
https://github.com/dnnsoftware/Dnn.Platform/issues/5469

## Summary
Add SVG icons to the ./images folder.
Update code to get a solution to issue #5469.

In this push request I have included all 24px Material Icons.

## Copyright
Material Icons are open source and free to use.
https://developers.google.com/fonts/faq
https://www.apache.org/licenses/LICENSE-2.0 

## Risk
Very small --> None.
The worst case will only cause icons to not appear on the module menu.

## Test
The code change has been implemented and tested by myself on a local PC installation.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5470) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5470) before we can accept your contribution.<br/><br/>:x: leedavi [sign now](https://cla.dotnetfoundation.org/dnnsoftware/Dnn.Platform?pullRequest=5470)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dnnsoftware/Dnn.Platform?pullRequest=5470) it.</sub>

mitchelsellers:
I agree that the current icons are outdated, and in need of a refresh.  However, I have a few questions/considerations for this, one with the items contributed, but also with a longer-term strategic direction.

## Strategic Questions

We have an existing project [DNN Icons](https://github.com/DNNCommunity/dnnicons) that to my knowledge has not yet been used.  However, this moves to a much easier-to-maintain font-based solution rather than individual icons.  Should we prioritize a solution that goes this route, reducing the installation footprint for DNN and more modern, over yet another icon based solution?

We also have the 'icons/sigma' files that are included in DNN, should those be reviewed as well?

## Specific notes for this PR

I have a few concerns with this PR that would need to be addressed before I can approve.

* This introduces a new third-party licensed component.  This means that we need to have the proper license information added to `/DNN Platform/Website/Licenses` Including the where/how used and copy of the current license terms
* I don't like merging these icons into the root "images" folder as it mixes this library of images with all of the other images that have been carried along for the last 20 years with DNN platform.  I believe either a new sub-folder for `/material` or similar should be created to isolate these images from the rest of the assets
* This adds 1,500+ images to swap out maybe 10 icons
* I'm not sure what items in this contribution are actually material icons, and which are custom or otherwise.  For example `action_print.svg` is NOT the name of any Material icon, so it is hard to understand what is actually included here and who the possible copyright holder is

If we can get some clarity on this, it would be great.  

(Lastly, if you can accept the new CLA bot as well that would be awesome)

valadas:
I am not a big fan of bringing a copy of something like this into the platform.
It's very easy to bring such things in but then we can never get them out because anyone could possibly rely on them.
It is easy for 3rd-party themes or modules to get this library through npm by their developer instead of DNN providing them.

IMO, if we can get some basic (an a minimal quantity of them) svg icons that we do need inside our own platform code, that would be better.

david-poindexter:
I agree 100% with the sentiment that led to this contribution and I value the contribution.  I just believe we should use DNN Icons instead of bringing in a huge 3rd-party library.  We don't need all these icons in the platform and it is a recipe for disaster with potential consumers, thus creating another long term dependency.

leedavi:
OK, I thought the 1500 icons would freak you all out.

I'm not surprised with your reaction.  I was in 2 minds if I should have even tried.

The idea here was to get a solution for the module icons that makes DNN look second rate.  

The reason I went in this direction, rather than a font based solution, was for speed.  The third party problem with this should not be a issue.  The license for material icons allows this kind of integration and we should add a comment in the license, like Mich has pointed out.

As for the names of the material icons, they have been renamed, simply to make integration easier.

As for leaving some kind of legacy situation if we include the icons, I totally agree.  It might not be best to do this.

The DNN icon project I looked at with the font solution was not suitable for this improvement.  It would make the work into a large project which would probably never be completed.

If I may, can I make a suggestion that we remove the material icons and only leave the 7 or 8 SVG icons required for the improvement.  That way we can get a quick fix, no (very small) legacy issues for the future.  

However, those icons are still Material Icons, even with the name changed.  I could get our designer to create the SVG needed, but that would seem an act of work just to prove we can copy Material Icons and change a few pixels.  So I suggest we stay with the renamed Material Icons for now.

After we have this quick fix done, we can work out the strategic options.   At least DNN UI will not be looking like a CMS from the 1990s.

What are your thoughts on that idea?

leedavi:
As for the CLA, I thought I had.   I have not idea what to sign?

mitchelsellers:
> If I may, can I make a suggestion that we remove the material icons and only leave the 7 or 8 SVG icons required for the improvement. That way we can get a quick fix, no (very small) legacy issues for the future.

I think that this would be acceptable, update the icons impacted, then make a license comment that "__ Icons" came from Material to properly attribute them.

I do still think they should be in a sub-folder, and I'd prefer keeping their proper Material Names.

david-poindexter:
@leedavi if you'd like to discuss this one at DNN Connect, I'd be happy to help you get it to a point where it could achieve the great goal of this PR!

PR #5484, opened 2023-01-14
Fixed an issue that prevented managing folders in user mode

### LINKED ISSUE #5456: Resource Manager in user mode prevents sub directory file management
## Description of bug
The RM in user mode can allow users to view files in sub directories but they can only delete or add files in their root folder.

## Steps to reproduce

1. Go to a page with the RM module and it's settings in user mode with the Users directory as it's default path.
2. Log in as a user with access to the RM. Ensure the test user has a subdirectory with a file by adding it as an admin or uploading a zip with a directory and a test file.
3. Try to upload a file to the subdirectory or delete an existing file within a sub directory.
4. There is no error but these actions are not accessible outside the users root folder.

## Current behavior
Users cannot add or delete files from sub directories that are within their root folder (ie 011/11/11/testdir/).

## Expected behavior
Users should be able to add files and delete files from within any sub directory inside of their root folder.

## Error information
No errors occur.

## Additional context
The PermissionManager is using a IsUserFolder check that prevents the sub directories from passing the HasDeletePermission and HasAddFilesPermission bool function calls.

https://github.com/dnnsoftware/Dnn.Platform/blob/116c8bcff68b0a6ea007c0930edfd480c7b085c8/DNN%20Platform/Modules/ResourceManager/Components/PermissionsManager.cs#L62

## Affected version

* [ ] 10.00.00 alpha build
* [X ] 09.11.00 release candidate
* [ ] 09.10.01 latest supported release

## Affected browser

* [ X] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer 11
* [ X] Microsoft Edge Chromium

valadas:
Closes #5456

Basically there was a permission check on the user folder but this did not account for the current folder being a children of the user folder. The code now checks recursivelly to see if the user is inside of the user folder instead of exactly in that folder.


david-poindexter (approved):
Nice catch @valadas - thank you! 🎉

bdukes (approved):

PR #5488, opened 2023-01-17
Fixed an issue where sourcemaps would not load

valadas:
Because of the way we load the bundles (using jquery ajax) they are loaded with the context of the site root. Because of that browsers are looking for the .map file on the site root instead of the folder where it actually lives in.

This PR declares the full path to those map files so that browsers know where to find them.

Additionally, it fixes some errors and warnings reported by eslint in the webpack config files and removes some unused variables, etc.

Also, `yarn watch --scope xxx` does not support package names with an underscore, so I renamed all the npm packages so they use a dash instead of an underscore for their names.


bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #5498, opened 2023-01-19
Bumped Moq to v4.18.4

valadas:
Supercedes #5497 

This properly bumps Moq to v4.18.4.
Also fixes multiple unit tests due to this change or other reasons. One of the main reasons was that there was multiple tests that only tested a method was calling another overload of the same class. This was a very poor test to begin with and it looks like the new Moq did not like that anyway. I adjusted those tests so they actually test the effects of calling such method.

I believe there might still be one test failing but I am too tired right now to handle that one. I am sending this PR to test how CI likes it, but please don't merge it until all the tests pass (well the ones we run anyway)...


valadas:
Oh well, they all passed. Good for review then :)

bdukes (approved):
Awesome work, thanks for fixing those!

mitchelsellers (approved):

PR #5508, opened 2023-01-29
Fixed an issue that caused an infinite loop when saving pofiles

valadas:
An automated stylecop fix caused a casing issue on a variable (IsValid vs isValid) that was a private field and a public property. The property calling a method and the method using the field (but in this case it was using the property by mistake and causing the endless loop).

I also fixed all other stylecop issues on the file and tested the build to confirm it fixes the issue.


david-poindexter (approved):
Great catch - thank you @valadas 🎉

mitchelsellers (approved):
You know, this is why I like using the _ prefix which is still in the Microsoft documentation for private backers.

But great catch

PR #5544, opened 2023-02-22
Implemented a workaround to support node v18

valadas:
Azure pipelines has updated windows-latest to run on node v18.x.x on Febuary 17th.

Node 18 has bumped some requirements about open-ssl and Webpack 4 will not be updated to support this use case.

The correct solution is to upgrade to webpack 5 but this has the potential to be a massive task. In the meantime one can pass openssl-legacy-provider for node v18 to support the legacy ssl methods. This PR is a workaround to make our current build work and be able to release DNN v9.11.1 I'll create an issue to do this in a more proper manner after I can confirm this PR fixes our issue in CI.

### DIFF
--- a/Build/Tasks/BuildNpmPackages.cs
+++ b/Build/Tasks/BuildNpmPackages.cs
@@ -3,6 +3,8 @@
 // See the LICENSE file in the project root for more information
 namespace DotNetNuke.Build.Tasks
 {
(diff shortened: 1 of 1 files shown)


valadas:

david-poindexter:
Glad it worked!

PR #5545, opened 2023-02-22
Update azure-pipelines.yml for Azure Pipelines to use node v18

valadas:
Enforces using node >= v18.x.x


valadas:

PR #5546, opened 2023-02-22
Bumped lerna to v6.5.1

valadas:
This is an attempt to fix the current build issue with 9.11.1


valadas:

PR #5550, opened 2023-02-22
Updates issue template with v9.11.1 beeing current

valadas:
Updates issue template with v9.11.1 beeing current


bdukes (approved):

valadas:

Issue #5361, opened 2022-10-28
New Resource Manager Needs Setting(s) for Height and Some Way to also Control the Height via CSS


jeremy-farrance:
## Description of problem
 
There is no way to set or adjust the height of the new resource manager. On sites with lots of folders and files and a large monitor to work on, this reduces productivity significantly. My case is a somewhat extreme example, but presents the issue succinctly. :)
 
I included a screenshot below of what it looks like on a vertical QHD monitor with the browser full screen. Even though there is plenty of screen real estate to expand, the module appears to have it's height fixed at just 400px. 
 
After a little investigating to see if I could fix this (the usual way?) with CSS, I was unable to come up with a way to set or override the elements height without having to (manually, each time the page loads) inspect and drill into the DNN Vertical Splitview (DNN Element). Because this element with this setting is a web component, the shadow (dom, root, scope?) prevents page or module level CSS from overriding the height.

Once you are drilled in the Inspector to the Vertical Splitview, you can apply this simple CSS fix:
```
.container dnn-vertical-splitview {
    height: 75vw !important;
}
```

## Description of solution
For the Resource Manager, I am hoping that this can become a setting that we can control as well as have a better default.
 
For the Vertical-Splitview component, I am hoping this gets added as both a CSS setting and an attribute.
 
In both cases, it seems most flexible and useful to allow people to set the "height:" to any valid CSS (e.g. "400px", "80vh", "64rem", or even a calc, "calc(32rem + 60vh").

## Description of alternatives considered
If a quick fix is needed because there is no time for a well-thought out solution, then at least simply update the default so that height is something like "calc(20rem + 40vh)". This will give it a minimum height and provide some expansion when its on a taller page.

## Screenshots
![image](https://user-images.githubusercontent.com/653301/198590698-8dfa8435-d2ac-46b3-aa76-86cd35b977ba.png)

## Additional context
This only applies to the newer Resource Manager introduced in DNN v9.11.00.

Affects all modern browsers.

jeremy-farrance:
Me? More details? I prefer to call it OCD. :)

valadas:
Done as part of the duplicate issue.

PR #5557, opened 2023-02-24
Temp fix Resource Manager default height

jeremy-farrance:
Temporary CSS-only fix for next DNN patch release to give the Resource Manager a better default height.

## Summary

In brief, the default height is currently hard-coded to 400px. This fix instead bases the height on the browser's height using "100vh - 25rem" which basically means the resource manager's height will be (dynamically) about 400px less than the browser's height (to account for the theme's header and footer at least). We've also set a min-height of 30rem (appx 480px) so scroll bars appear, but everything functions as expected.

Temp fix, but does not close Issue https://github.com/dnnsoftware/Dnn.Platform/issues/5361


jeremy-farrance:
> @microsoft-github-policy-service agree company="Accuraty"

valadas (approved):
That looks fine to me

david-poindexter (approved):
Awesome - thank you for the contribution @jeremy-farrance 🎉

jeremy-farrance:

PR #5570, opened 2023-03-12
Bumped webpack to v5

valadas:
Bumped webpack to v5

This removes the workarounds put in place in 9.11.1 so the build works in CI where node was upgrade to v18. Our build should now work on node 16 and 18 without workaround in place... Hopefully.

I tested a full build and confirmed all the modules still load up fine.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -52,11 +52,6 @@ steps:
     key: 'yarn | "$(Agent.OS)" | yarn.lock'
     restoreKeys: 'yarn | "$(Agent.OS)"'
     path: $(YARN_CACHE_FOLDER)
-
-- task: NodeTool@0
-  inputs:
-    versionSource: 'spec'
-    versionSpec: '>=18.x'
   
 - task: UseDotNet@2
   displayName: 'Use .NET 6.x SDK'
--- a/Build/Build.csproj
+++ b/Build/Build.csproj
@@ -24,6 +24,7 @@
     <PackageReference Include="Cake.XdtTransform" Version="2.0.0" />
     <PackageReference Include="Cake.Yarn" Version="0.4.8" />
     <PackageReference Include="Dnn.CakeUtils" Version="2.0.2" />
+    <PackageReference Include="MSBuild.StructuredLogger" Version="2.1.787" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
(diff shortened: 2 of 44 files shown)


valadas:
I don't get it, the build error has nothing to do with these changes... @bdukes do you have any clue about the binlog version (assuming that is the problem here)...

valadas:
Adding the nuget package appears to fix it.

david-poindexter (approved):
@valadas this looks great!  One question.  I see the `webpack` script (which uses `webpack-dev-server ...`) has been removed from all the **package.json** files.  What is replacing this?  Are we recommending just using the webpack CLI now for this?  If so, do the Cake scripts need to be up updated accordingly?  (I honestly can't remember if those used this or not in the scoped build options.)

valadas:
It was because since we have `yarn start --scope something` now (which we did not have back then, using the `webpack` task is more complex and needs the developer to setup the view to load from :8080 manually which we don't have anywhere in our docs and many already have something running on 8080 and then get into troubles, etc. Happy to bring it back if anyone mentions they actually use that task...

david-poindexter:
> It was because since we have `yarn start --scope something` now (which we did not have back then, using the `webpack` task is more complex and needs the developer to setup the view to load from :8080 manually which we don't have anywhere in our docs and many already have something running on 8080 and then get into troubles, etc. Happy to bring it back if anyone mentions they actually use that task...

That makes perfect sense - thanks for the explanation.  And seriously...thank you for this PR!

azure-pipelines.yml
```diff
@@ -52,11 +52,6 @@ steps:
     key: 'yarn | "$(Agent.OS)" | yarn.lock'
     restoreKeys: 'yarn | "$(Agent.OS)"'
     path: $(YARN_CACHE_FOLDER)
-
-- task: NodeTool@0
-  inputs:
-    versionSource: 'spec'
-    versionSpec: '>=18.x'
```
bdukes, line 59: Do we not want to specify the node.js version anymore?

valadas: Well, the machine as some sort of default that was bumped to v18 in February, they use the latest LTS which I think is a good policy... I put that in place earlier for the workaround as it was apparently sometimes using v16 and sometimes v18 (I guess depending on the machine assigned or something).

I do believe we should support 16/18 as I doubt a lot of developers use v18 locally...

valadas: And since we no longer have the workaround, we build fine on v16 and v18 if this PR is merged.

bdukes (approved):

PR #5586, opened 2023-03-21
Scheduler Validate Servers - #5575

### LINKED ISSUE #5575: Adjust Scheduler to Validate Existence of Server Before Start
## Description of the problem
Currently within DNN Platform if you schedule a task to execute on a single server, and that server no longer exists the job will simply not run.  In the case of scaled solutions, this can result in tasks that no longer run.

## Description of solution
DNN Platform currently tracks a WebServers table that has a "last activity" date.  The proposed solution would be to do the following when starting a task

* If no specific servers are set, continue with the normal process, no impact
* If a single server is set
  *  Validate that the server is still listed in the `WebServer` table and that its last activity is within 10 minutes
  *  If it is not, find another server from the list that IS active an update the task
  *  Ensure that the task runs
* If more than one server is set
  * Follow a similar process to the above, but if a server is invalid remove it.  ONLY add another server if it would be left with no servers defined

We discussed possibly updating the health checks or otherwise, but it would put more load on the environment rather than targeting the task start process

## Additional context
This change is important for Azure Auto Scale, other cloud, and other multi-server situations

mtrutledge:
Fixes #5575 

## Summary
Adds a WebServerMonitor scheduled task that runs every 5 minutes. This scheduled task updates the Last Activity Date for a WebServer. If a server hasn't been seen in > 10 minutes then it is disabled. If a WebServer hasn't been seen in > 24 hours then delete the server from the table. Anytime we deactivate or delete a server, we updated the scheduled tasks related to the deactivated server to a new server.

### DIFF
--- a/DNN Platform/Library/Data/DataProvider.cs
+++ b/DNN Platform/Library/Data/DataProvider.cs
@@ -2206,6 +2206,11 @@ namespace DotNetNuke.Data
             this.ExecuteNonQuery("RemoveUser", userId, this.GetNull(portalId));
         }
 
+        public virtual void ReplaceServerOnSchedules(string oldServername, string newServerName)
+        {
+            this.ExecuteNonQuery("ReplaceServerOnSchedules", oldServername, newServerName);
+        }
+        
         public virtual void ResetTermsAgreement(int portalId)
         {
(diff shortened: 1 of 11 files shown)


valadas (approved):
Looks good to me

mitchelsellers (approved):

PR #5587, opened 2023-03-22
Fixed an issue resolving LiteDB

valadas:
In #5555 and #5556 LIteDB version was bumped by dependabot but the reference in the .csproj files were not updated. This PR fixes that so we have a working build. It also updates our yarn.lock after running a build following other dependencies updates.


david-poindexter (approved):
Nice - thank you @valadas

PR #5592, opened 2023-03-25
Enhance PersonaBarTheme.css to Resize the PersonaBar Logo

### LINKED ISSUE #5591: Add Ability to Resize the PersonaBar Logo
## Description of problem
A while back some :root {} variables were added (see Portals/_default/PersonaBarTheme.css) so that the PersonaBar could be branded with colors and a logo. But there is no way to adjust the size of the logo and it just defaults, hard-coded to `background-size: 32px auto;`.

## Description of solution
Add a new variable following the existing pattern: `--dnn-pb-menu-brand-background-size: 32px auto;`

Then consume that in main.css at appx. line 743 (DNN 9.11.01).

## Screenshots
Default at 32px width:
image

Example with `5rem auto`
image

## Affected browser
* [x] All

PS, I will do a pull request for this soon - JRF

jeremy-farrance:
Fixes #5591

## Summary

Adding the ability for themers or admins to adjust the size of the PersonaBar Logo while keeping the existing size (default).

### DIFF
--- a/Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/css/main.css
+++ b/Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/css/main.css
@@ -740,8 +740,8 @@ span.dnnFormError:after {
     width: 80px;
     height: 103px;
     background: var(--dnn-pb-menu-brand-background);
-    -ms-background-size: 32px auto;
-    background-size: 32px auto;
+    -ms-background-size: var(--dnn-pb-menu-brand-background-size);
+    background-size: var(--dnn-pb-menu-brand-background-size); /* 32px auto; */
     position: relative;
     border-bottom: 1px solid var(--dnn-color-pb-menu-divider);
     text-align: center;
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/css/main.css
```diff
@@ -740,8 +740,8 @@ span.dnnFormError:after {
     width: 80px;
     height: 103px;
     background: var(--dnn-pb-menu-brand-background);
-    -ms-background-size: 32px auto;
-    background-size: 32px auto;
+    -ms-background-size: var(--dnn-pb-menu-brand-background-size);
+    background-size: var(--dnn-pb-menu-brand-background-size); /* 32px auto; */
```
valadas, line 744: I would either remove the comment (it would lie if someone changes it and be just confusing) or make it a fallback (so that it defaults to that size if not specified):

var(--dnn-pb-menu-brand-background-size, 32px auto);

valadas (changes requested):

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/css/main.css
jeremy-farrance: I'll remove the comment. Not needed since the default (original) values `32px auto` is in theme.css and there is no case where a fallback is needed.

valadas (approved):

valadas:
Thanks, that works for me

PR #5600, opened 2023-04-04
Enable Dependency Injection for Prompt commands

bdukes:
## Summary
This PR updates creation of Prompt commands to use the dependency injection container.

This includes a breaking change in `ICommandRepository.GetCommand` to take an `IServiceProvider` (in addition to the command name).


david-poindexter (approved):
Awesomeness 👏🏼

PR #5606, opened 2023-04-04
Allow registration of custom IModuleControlFactory implementations

### LINKED ISSUE #5605: Allow registration of custom control factories
## Description of problem
It's not possible to register or override a custom [IModuleControlFactory](https://github.com/dnnsoftware/Dnn.Platform/blob/421210f9a20d1c272db77e7208f7a5e960011534/DNN%20Platform/Library/UI/Modules/IModuleControlFactory.cs#L11) in DNN. This is because the extensions are hard-coded in [ModuleControlPipeline](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs#L55-L62).

## Description of solution
1. Move the extension check out of the pipeline and into the factory.
2. Add a Priority property, which the pipeline will use to determinate which factory to use.
    By default, DNN control factories have a priority of 100. If you register a custom control factory (for example for `.html` to modify the content) you can give it a priority of 200 to overrule the DNN factory.

The new factory would look like this:

```cs
public interface IModuleControlFactory
{
    /// <summary>
    /// Gets the priority of the factory.
    /// The factory with the highest priority will be used.
    /// </summary>
    /// <remarks>
    /// DNN Control factories (WebForms, Razor, HTML, MVC) have a priority of 100
    /// and the default factory (Reflected) has a priority of -1.
    /// </remarks>
    int Priority { get; }

    /// <summary>Validates if the factory supports the given configuration and <see cref="controlSrc"/>.</summary>
    /// <param name="moduleConfiguration">Module configuration.</param>
    /// <param name="controlSrc">Control source.</param>
    /// <returns><c>true</c> if the factory supports the control; otherwise, <c>false</c>.</returns>
    bool SupportsControl(ModuleInfo moduleConfiguration, string controlSrc);

    // CreateControl and CreateSettingsControl are unchanged.
}
```

The pipeline will be changed that it'll resolve all `IModuleControlFactory` and get the first supported factory.

While it is currently possible to override the [IModuleControlPipeline](https://github.com/dnnsoftware/Dnn.Platform/blob/421210f9a20d1c272db77e7208f7a5e960011534/DNN%20Platform/Library/UI/Modules/IModuleControlPipeline.cs#L11) in the service collection:

```cs
public class Extension : IDnnStartup
{
    {
        services.AddSingleton<IModuleControlPipeline>()
    }
}

public class NewModuleControlPipeline : ModuleControlPipeline
{
    // ...
}
```

This brings two problems:

1. If multiple extensions are changing the `IModuleControlPipeline`, the last one will be used. All other pipelines will be ignored.
2. The default IModuleControlPipeline of DNN (`ModuleControlPipeline`) is in a class library (`DotNetNuke.ModulePipeline`) which is not in NuGet. So you'll have to add a DLL-reference. 😔

GerardSmit:
Fixes #5605

## Summary

This PR implements the proposed solution in #5605.

The target branch is 10.0.0, since the `IModuleControlFactory` has been changed. If this is incorrect I can change the branch and cherry-pick the commit. 🙂

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
@@ -9,6 +9,7 @@ namespace DotNetNuke.Web.Mvc
     using DotNetNuke.Common;
     using DotNetNuke.Common.Internal;
     using DotNetNuke.DependencyInjection;
+    using DotNetNuke.UI.Modules;
     using DotNetNuke.Web.Mvc.Extensions;
     using DotNetNuke.Web.Mvc.Routing;
     using Microsoft.Extensions.DependencyInjection;
@@ -20,7 +21,7 @@ namespace DotNetNuke.Web.Mvc
         public void ConfigureServices(IServiceCollection services)
         {
             services.TryAddSingleton<IControllerFactory, DefaultControllerFactory>();
-            services.AddSingleton<MvcModuleControlFactory>();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, MvcModuleControlFactory>());
             services.TryAddEnumerable(new ServiceDescriptor(typeof(IRoutingManager), typeof(MvcRoutingManager), ServiceLifetime.Singleton));
 
             services.AddMvcControllers();
--- a/DNN Platform/DotNetNuke.Web.Razor/Startup.cs
+++ b/DNN Platform/DotNetNuke.Web.Razor/Startup.cs
@@ -1,18 +1,22 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information
-
 namespace DotNetNuke.Web.Razor
 {
     using DotNetNuke.DependencyInjection;
+    using DotNetNuke.UI.Modules;
+
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
 
     public class Startup : IDnnStartup
     {
         /// <inheritdoc/>
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddSingleton<RazorModuleControlFactory>();
+#pragma warning disable CS0618 // Type or member is obsolete
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, RazorModuleControlFactory>());
+#pragma warning restore CS0618 // Type or member is obsolete
         }
     }
 }
--- a/DNN Platform/Library/Startup.cs
+++ b/DNN Platform/Library/Startup.cs
@@ -28,6 +28,7 @@ namespace DotNetNuke
     using DotNetNuke.UI.Modules.Html5;
 
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
 
     /// <inheritdoc />
     public class Startup : IDnnStartup
@@ -35,9 +36,10 @@ namespace DotNetNuke
         /// <inheritdoc />
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddSingleton<WebFormsModuleControlFactory>();
-            services.AddSingleton<Html5ModuleControlFactory>();
-            services.AddSingleton<ReflectedModuleControlFactory>();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, WebFormsModuleControlFactory>());
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, Html5ModuleControlFactory>());
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, ReflectedModuleControlFactory>());
+
             services.AddSingleton<IDnnContext, DotNetNukeContext>();
 
 #pragma warning disable CS0618
--- a/DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
+++ b/DNN Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs
@@ -6,6 +6,7 @@ namespace DotNetNuke.ModulePipeline
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
 #if NET472
     using System.Web.UI;
 #endif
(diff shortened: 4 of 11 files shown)


GerardSmit:
It seems like some whitespace changes have made it into the commit... I'm not sure how this happened.
You can check the diffs without whitespace changes by adding `?w=1` behind the URL, like this: https://github.com/dnnsoftware/Dnn.Platform/pull/5606/files?w=1
This makes it easier to read the changed code.

DNN Platform/Library/Startup.cs
```diff
@@ -30,8 +30,14 @@ public class Startup : IDnnStartup
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<WebFormsModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<WebFormsModuleControlFactory>());
```
mitchelsellers, line 33: Is this registration correct?  Since this registers a singleton with the interface does this result in proper behavior?  

As belowon on lines 36 and 39 the same type is being injected with a different implementation

GerardSmit: Yes, this is correct.

In the library `Microsoft.Extensions.DependenyInjection` it's possible to register the same type multiple times. For example:

```cs
interface IService {}
class ServiceA : IService {}
class ServiceB : IService {}

IServiceCollection services;
services.AddSingleton<IService, ServiceA>();
services.AddSingleton<IService, ServiceB>();
```

Whenever you resolve the service `IService`, you'll get the last registered service. So the following would resolve ServiceB:
```cs
IServiceProvider provider;
provider.GetRequiredService<IService>(); // Resolved ServiceB
```

However, it's also possible to get all the registered services of type `IService` with `IEnumerable<IService>`:
```cs
IServiceProvider provider;
provider.GetRequiredService<IEnumerable<IService>>(); // An enumerable with ServiceA and ServiceB
```

This means, if an extension registers their custom type:
```cs
class MyService : IService {}

IServiceCollection services;
services.AddSingleton<IService, MyService>();
```
It'll add the MyService to the IEnumerable.

This is what I'm doing in ModuleControlPipeline:
https://github.com/dnnsoftware/Dnn.Platform/blob/05852bfe270203f4fe4a0e291ce560e6e96e3f70/DNN%20Platform/DotNetNuke.ModulePipeline/ModuleControlPipeline.cs#L41-L46

Hence why I'm registering WebFormsModuleControlFactory, Html5ModuleControlFactory and ReflectedModuleControlFactory this way.

---

For the reason why I'm not doing the following:

```cs
services.AddSingleton<WebFormsModuleControlFactory>();
services.AddSingleton<IModuleControlFactory, WebFormsModuleControlFactory>();
```

When you resolve `WebFormsModuleControlFactory` it'll create a new instance and when you resolve `IEnumerable<IModuleControlFactory>`, it'll create another instance. The dependency injector doesn't try to check if there is already an instance of the registered type.

By doing the following:

```cs
services.AddSingleton<WebFormsModuleControlFactory>();
services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<WebFormsModuleControlFactory>());
```

It'll create an instance if you resolve `WebFormsModuleControlFactory` and when you try to resolve `IEnumerable<IModuleControlFactory>`, it'll get the instance of the `WebFormsModuleControlFactory` from the service provider. This way you won't get multiple instances of the factory in memory.

If you want to read more, this is better documented by Microsoft than explained by me: https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#service-registration-methods

bdukes: I've been using [`TryAddEnumerable`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddenumerable?view=dotnet-plat-ext-7.0&viewFallbackFrom=netframework-4.7.2) when I'm registering multiple implementations, just to make it clearer that it's intentional.

I wasn't aware about the issue with getting multiple instances of singletons, that's good to know.

GerardSmit: > Since we're no longer injecting these into `ModuleControlPipeline`, do we need the standalone `AddSingleton<WebFormsModuleControlFactory>()`?

We can remove them, yes.

The only issue is when external extensions are resolving `WebFormsModuleControlFactory` in their code, other than that we can remove them. But I highly doubt it that anyone would use the factories directly.

GerardSmit: > I wasn't aware about the issue with getting multiple instances of singletons, that's good to know.

Sadly using the `GetRequiredService`-workaround also brings a problem with disposing the class.

For example the following (Fiddle: https://dotnetfiddle.net/2UVbpb)
```cs
using System;

var services = new ServiceCollection();
services.AddSingleton<Service>();
services.AddSingleton<IService>(p => p.GetRequiredService<Service>());

var provider = services.BuildServiceProvider();
provider.GetRequiredService<IService>();
provider.Dispose();

interface IService {}

class Service : IService, IDisposable
{
	public Service() => Console.WriteLine("Construct");
	public void Dispose() => Console.WriteLine("Dispose");
}
```

Would dispose the class twice. There is an open issue about this problem: https://github.com/dotnet/runtime/issues/70004 but sadly it is not addressed yet.

DNN Platform/Library/Startup.cs
```diff
@@ -30,8 +30,14 @@ public class Startup : IDnnStartup
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<WebFormsModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<WebFormsModuleControlFactory>());
+
             services.AddSingleton<Html5ModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<Html5ModuleControlFactory>());
+
             services.AddSingleton<ReflectedModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<ReflectedModuleControlFactory>());
```
bdukes, line 39:
```suggestion
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, Html5ModuleControlFactory>());
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, ReflectedModuleControlFactory>());
```

bdukes (commented):
Let's combine these registrations

DNN Platform/DotNetNuke.Web.Razor/Startup.cs
```diff
@@ -2,17 +2,21 @@
...
-        {
-            services.AddSingleton<RazorModuleControlFactory>();
-        }
-    }
-}
+        public void ConfigureServices(IServiceCollection services)
+        {
+#pragma warning disable CS0618 // Type or member is obsolete
+            services.AddSingleton<RazorModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<RazorModuleControlFactory>());
```
bdukes, line 18:
```suggestion
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, RazorModuleControlFactory>());
```

DNN Platform/DotNetNuke.Web.Mvc/Startup.cs
```diff
@@ -2,27 +2,29 @@
...
-
-            DependencyResolver.SetResolver(new DnnMvcDependencyResolver(Globals.DependencyProvider));
-        }
-    }
-}
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.TryAddSingleton<IControllerFactory, DefaultControllerFactory>();
+            services.AddSingleton<MvcModuleControlFactory>();
+            services.AddSingleton<IModuleControlFactory>(p => p.GetRequiredService<MvcModuleControlFactory>());
```
bdukes, line 23:
```suggestion
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IModuleControlFactory, MvcModuleControlFactory>());
```

bdukes (approved):

mitchelsellers (approved):

PR #5616, opened 2023-04-11
Add IBusinessControllerProvider abstraction & Globals.GetOrCreateServiceScope

bdukes:
## Summary
This PR adds a couple of helpers used in implementing dependency injection for business controller class.

PR #5617, opened 2023-04-11
Add DI for EventMessageProcessorBase

bdukes:
## Summary
This PR relies on #5616. It enables dependency injection for `EventMessageProcessorBase` (i.e. the component which runs `UpgradeModule`)

### DIFF
--- a/DNN Platform/Library/Common/Initialize.cs
+++ b/DNN Platform/Library/Common/Initialize.cs
@@ -18,7 +18,6 @@ namespace DotNetNuke.Common
     using DotNetNuke.Entities.Host;
     using DotNetNuke.Entities.Urls;
     using DotNetNuke.Instrumentation;
-    using DotNetNuke.Services.Connections;
     using DotNetNuke.Services.EventQueue;
     using DotNetNuke.Services.Exceptions;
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/Common/Initialize.cs
```diff
@@ -428,41 +428,46 @@ private static string InitializeApp(HttpApplication app, ref bool initialized)
             {
                 // Check whether the current App Version is the same as the DB Version
                 redirect = CheckVersion(app);
-                if (string.IsNullOrEmpty(redirect) && !InstallBlocker.Instance.IsInstallInProgress())
+                if (!string.IsNullOrEmpty(redirect) || InstallBlocker.Instance.IsInstallInProgress())
```
valadas, line 431: Since this condition was reversed, would it make sense to bring all the core from the else block up here and kill the else block ?

bdukes: I finished simplifying this logic, let me know if you were looking for anything different.

valadas (commented):
Looks good to me except the failed test

valadas (approved):
Awesome!

mitchelsellers (approved):

PR #5627, opened 2023-04-14
Introduce FakeServiceProvider for tests

bdukes:
## Summary
This PR introduces `FakeServiceProvider` to the `DotNetNuke.Tests.Utilities` project. The fake service provider creates automatic mocks for all services requested that aren't registered, and provides a simple interface to register it as `Globals.DependencyProvider` and as the scope on `HttpContextBase`. Finally, when disposed, it resets that global state.


valadas (approved):
Looks good to me other than the build failing which appears to be unrelated, so re-fired the build to see if it was a fluke.

david-poindexter (approved):

PR #5640, opened 2023-05-08
Add params array constructor to SupportedModulesAttribute

### LINKED ISSUE #5638: Allow params array constructor for SupportedModulesAttribute
## Description of problem
Currently the only constructor for SupportedModulesAttribute takes a single string parameter that it attempts to split on commas in order to build the supportedModules string array like this:

```
public SupportedModulesAttribute(string supportedModules)
{
    this.supportedModules = supportedModules.Split(new[] { ',' });
}
```

This has the downside of being error prone. For example, our team made the mistake of adding a space after a comma (',') character which caused API requests to the second module to not be accepted. 

It also has other downsides like being hard to read especially once you have a handful of supported modules
`[SupportedModules("Test: Module1,Test: Module2,Test: Module3,Example: Module4")]`

## Description of solution
I suggest adding an **additional** constructor to SupportedModulesAttribute that takes a params string array as its only parameter that can be directly assigned to the supportedModulesArray like so:

```
public SupportedModulesAttribute(params string[] supportedModules)
{
    this.supportedModules = supportedModules;
}
```

An alternative would be to allow multiple `[SupportedModules]` attributes on a single class as currently it seems only the first of the attributes will function as intended.

If approved I will happily submit a PR and have the code ready to do so now.

CreggHancock:
Fixes #5638

### DIFF
--- a/DNN Platform/DotNetNuke.Web/Api/SupportedModulesAttribute.cs
+++ b/DNN Platform/DotNetNuke.Web/Api/SupportedModulesAttribute.cs
@@ -19,6 +19,11 @@ namespace DotNetNuke.Web.Api
             this.supportedModules = supportedModules.Split(new[] { ',' });
         }
 
+        public SupportedModulesAttribute(params string[] supportedModules)
+        {
+            this.supportedModules = supportedModules;
+        }
+
         /// <inheritdoc/>
         public override bool IsAuthorized(AuthFilterContext context)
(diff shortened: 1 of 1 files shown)


CreggHancock:

bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #5648, opened 2023-05-30
Enable lazy DI registration

### LINKED ISSUE #5646: Implement a Lazy<T> for DI
## Description of problem

DNN now has a lot of DI, but it's still missing the classic `Lazy<Service>`. 
This is important to request a service which may not be used, so we don't want to initialize it when not used. 

This was already discussed at DNN connect, and I think it's for @bdukes 

## Description of solution

This is a very classic setup:

1. Create the 3-lines Lazy-implementation
2. Register it
3. Usable

```c#
﻿using System;

namespace DotNetNuke.WhateverShouldBeTheNamespace
{
    /// <summary>
    /// Enables lazy requesting of objects - won't be available until needed.
    /// This is a classic plain-vanilla implementation of Lazy for ServiceProviders.
    /// </summary>
    public class LazyImplementation<TService> : Lazy<TService>
    {
        public LazyImplementation(IServiceProvider sp) : base(sp.Build<TService>)
        {
        }
    }
}
```

Register it like this:

```c#
        public static IServiceCollection AddLibDI(this IServiceCollection services)
        {
            // Lazy objects in General
            services.TryAddTransient(typeof(Lazy<>), typeof(LazyImplementation<>));
...
```

Use like this:

```c#
public class MyClass
{
  public MyClass(Lazy<ISomeService> serviceLazy) 
  {
    var actuallyUsed = serviceLazy.Value;
  }
```

This is super-standard, so there are not many alternatives (except for not using Lazy, or using a different DI system).

bdukes:
Fixes #5646

Thanks @iJungleboy!

### DIFF
--- a/DNN Platform/DotNetNuke.DependencyInjection/DotNetNuke.DependencyInjection.csproj
+++ b/DNN Platform/DotNetNuke.DependencyInjection/DotNetNuke.DependencyInjection.csproj
@@ -6,6 +6,7 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <RootDirectory>$(MSBuildProjectDirectory)\..\..</RootDirectory>
     <DocumentationFile>bin/$(Configuration)/$(TargetFramework)/DotNetNuke.DependencyInjection.xml</DocumentationFile>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


valadas (approved):
Awesome!

mitchelsellers (approved):

PR #5652, opened 2023-06-02
Add ComponentFactory fallback to service provider

bdukes:
This enables components registered in the component factory to migrate to the global DI container without breaking things


valadas (approved):
Awesome!

mitchelsellers (approved):

PR #5655, opened 2023-06-06
Resolve StackOverflowException

bdukes:
#5652 introduced an issue because [the Telerik Removal component registers `LocalizationProvider` with the DI container using a lambda pointing to `LocalizationProvider.Instance`](https://github.com/dnnsoftware/Dnn.Platform/blob/a716272d41a864b263cc4db413caa0ae33f02ce9/DNN%20Platform/DotNetNuke.Maintenance/Telerik/Removal/Startup.cs#L32-L37), which is [`ComponentBase.Instance`](https://github.com/dnnsoftware/Dnn.Platform/blob/a716272d41a864b263cc4db413caa0ae33f02ce9/DNN%20Platform/Library/ComponentModel/ComponentBase.cs#L18-L37). Since `LocalizationProvider` is not registered with `ComponentFactory`, this means that, due to #5652, it falls back to the DI container, which then asks the `ComponentFactory` _again_, creating an infinite loop, and ultimately a `StackOverflowException`.

This PR registers `LocalizationProvider` directly in the DI container, instead of via the `LocalizationProvider.Instance` lambda. 

Ultimately, the best resolution is to ensure that every component using `ComponentBase` is registered first in the DI container. I think we should also mark `ComponentBase` as obsolete.

I also considered adding a check to prevent the infinite loop or fallback in these scenarios, but couldn't come up with a safe way to do it which preserved the existing functionality.

### DIFF
--- a/DNN Platform/DotNetNuke.Maintenance/Telerik/Removal/Startup.cs
+++ b/DNN Platform/DotNetNuke.Maintenance/Telerik/Removal/Startup.cs
@@ -15,6 +15,7 @@ namespace DotNetNuke.Maintenance.Telerik.Removal
     using DotNetNuke.Services.Installer.Packages;
     using DotNetNuke.Services.Localization;
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
 
     /// <inheritdoc />
     public class Startup : IDnnStartup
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):

valadas (approved):
Awesome

PR #5666, opened 2023-06-11
Add `Google Analytics 4` Connector

### LINKED ISSUE #5665: Add Google Analytics 4 (GA4) Connector
## Description of problem
Currently, there is no UI to handle Google Analytics 4 (GA4) configuration.  It can be done using Google Tag Manager, but some people prefer to use the GA4 script directly.

## Description of solution
Add a new `Connector` for GA4.

Replace the existing `Connector` for GA.

## Screenshots
n/a

## Additional context
Based on conversations with @dnnsoftware/approvers it was decided to create a new GA4 connector for now.  Then once data collection ceases with Universal Analytics, we'll remove the existing Google Analytics connector.

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

david-poindexter:
## Summary
Resolves #5665 

image

image

image

image

image


valadas (approved):
Awesome!

mitchelsellers (approved):
Awesome

bdukes:
@dnnsoftware/approvers should we remove the old GA connector? There's no reason to include it in new installs anymore, right?

valadas:
Makes sense to me, upgrades would keep and people can uninstall it when not needed anymore. But... The module (UI) is only part of it, the actual engine is solely driven by the config file. Sooo, people would have to disconnect, then uninstall. One of the things I wanted to do (when I have some time) is to be able to call something like IUpgradeable but for uninstallation of a module.  Basically define a method that gets called before the rest of the normal uninstall happens. That could be a nice solution for this but I won't have enough time to do it in v9... Maybe v10...

david-poindexter:
> @dnnsoftware/approvers should we remove the old GA connector? There's no reason to include it in new installs anymore, right?

In principle, I agree.  In practice, there are some details about the base implementation that will require some thought as to how to unravel it all.  For instance, it depends on the default `SiteAnalytics.config` file (which contains the old GA script).

PR #5670, opened 2023-06-12
Add packaging for Microsoft.Extensions.DependencyInjection

bdukes:
## Summary
This PR introduces a package for the Microsoft.Extensions.DependencyInjection library (and dependencies) and upgrades the NuGet reference from 2.1.1 to 7.0.0.

At this point, this introduces a manual upgrade step, where the package is installed on a DNN 9.x site to upgrade the DI library before doing the full DNN upgrade.


david-poindexter (approved):
@bdukes this looks good.  Perhaps we should discuss the details of this tomorrow in our @dnnsoftware/approvers  meeting?

bdukes:
I added a commit to ensure the request scope is asynchronously disposed, per https://github.com/dnnsoftware/Dnn.Platform/pull/5014#issuecomment-1489360207. There are other scopes which we're currently creating which would ideally be disposed asynchronously, but that'll take some review to see how/if that can be done.

valadas (approved):

PR #5672, opened 2023-06-13
Update bug report issue form to require versions

david-poindexter:
## Summary
Update bug report issue form to require versions


valadas:

PR #5678, opened 2023-06-13
Correct `Yes` in `enhancement-request.yml`

david-poindexter:
## Summary
Correct `Yes` in `enhancement-request.yml`


valadas (approved):

valadas:

PR #5683, opened 2023-06-14
Cleaned up format of affected-versions dropdown

david-poindexter:
## Summary
Attempt to use multi-line markdown in `affected-versions` dropdown description.


valadas (approved):

valadas:

PR #5692, opened 2023-06-15
Fixed an issue where users module would not show user picture

valadas:
#5630 Fixed a backend issue for a feature that was not available on Platform. This PR adds that display into the Users module

Additionally `Updates versions as per release candidate creation` github action stopped working somehow (I'll take a look into this) so running a build here caused all those versions to update too.

And finally `--colors` was removed from all the `watch` tasks as that option is no longer available in Webpack 5.

image


bdukes (approved):

bdukes:
Is this a regression?

valadas:
> Is this a regression?

Yes and No,

I was under the assumption that #5630 was a bugfix for a feature that was existing in the frontend. But it ends up it is an addition and the screenshots in that PR did not come from a Platform install. So the release notes indicate this as a bugfix but users would have been surprised that this was not working on Platform. So I implemented the missing frontend part to complete that.

mitchelsellers (approved):

PR #5705, opened 2023-06-22
Updated bug-report.yml as per v9.12.0 release

valadas:
Updated bug-report.yml as per v9.12.0 release


valadas:

PR #5724, opened 2023-07-07
Add JS libraries for jQuery.fileupload, jQuery.hoverIntent, and DNN jQuery Plugins

bdukes:
## Summary
This PR removes the special cases for `CommonJs.jQueryFileUpload`, `CommonJs.HoverIntent`, and `CommonJs.DnnPlugins` by making them proper libraries (with dependencies).

This will help the removal of obsolete APIs in DNN 10.0.0 #5722 to have a migration path.


bdukes:
I also have a commit to add jQuery.hoverIntent, not sure if we want that in this PR or a separate one…

mitchelsellers (approved):

mitchelsellers:
@bdukes I think we can just make it one bit "Fix stuff that should have been libraries"

This is a better solution!

bdukes:
Works for me, I've updated this PR

mitchelsellers (approved):

DNN Platform/Library/Framework/JavaScriptLibraries/JavaScript.cs
```diff
@@ -80,15 +80,6 @@ public static void RequestRegistration(string jsname)
                 case CommonJs.jQuery:
                     RequestRegistration(CommonJs.jQueryMigrate);
                     break;
-                case CommonJs.DnnPlugins:
```
valadas, line 83: Everything looks good to me but I don't understand why the case statements are removed from here... Can you elaborate on that ?

bdukes: Because these are now being included as JS libraries, there's no more special behavior need, all of this logic is now part of the library itself as dependencies.

valadas: Sorry, I have not dug much in this area, but do we need to do anything (deprecate) in the CommonJs class or is that still usable in other parts of the code ?

bdukes: It's just a collection of strings for the JS libraries that come with DNN. I don't see a need to deprecate, unless we're planning to stop including a library.

david-poindexter (approved):

PR #5733, opened 2023-07-13
Fix error installing DNN after #5463

bdukes:
## Summary
#5463 created an incompatibility in the install and upgrade process because of how Microsoft.Web.Infrastructure was referenced. This PR removes that troublesome reference, making installs work again.

Thanks for [the testing and heads up on this](https://github.com/dnnsoftware/Dnn.Platform/pull/5722#issuecomment-1633425818) @valadas


valadas (approved):
Awesome, thank you!

david-poindexter (approved):

PR #5754, opened 2023-07-18
Implement DnnDeprecated source generator for types

bdukes:
## Summary
This PR introduces a [source generator](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview) into the DNN codebase. This allows the replacement of `Obsolete` attributes with a new attribute, `DnnDeprecated`. Usage of the new attribute allows the formatting of the deprecation message to be consistent. It could also, potentially, incorporate a simpler way to link to related documentation, when appropriate.

### Implementation Details
This source generator requires that the members which are decorated with the attribute to be marked `partial`, so that another file can be generated that includes the actual `Obsolete` attribute. This means that methods and nested types need their enclosing types to also be marked `partial`.

### Drawbacks
Because this source generator relies on [members being marked `partial`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/partial-method), it only applies to members which can be `partial`. This rules out constructors, finalizers, overloaded operators, property declarations, and event declarations.

It's possible to create a different system which generates these members with the `Obsolete` attribute, but it will be difficult to find a straightforward approach where the cost is less than the benefit.


bdukes:
This PR is ready for review.

All `Obsolete` attributes have been touched, either being turned into `DnnDeprecated` or being updated to use the template that `DnnDeprecated` generates.

I also removed a number of `Obsolete` attributes which were on members where the type was already deprecated.

Since `DnnDeprecated` requires a version in which the member was deprecated, I tried to find that version where it wasn't already marked. I used 7.0.0 as a placeholder if the history of the deprecation pre-dates the git repo.

DNN Platform/DotNetNuke.DependencyInjection/Extensions/TypeExtensions.cs
```diff
@@ -10,16 +10,17 @@ namespace DotNetNuke.DependencyInjection.Extensions
...
-    public static class TypeExtensions
+    public static partial class TypeExtensions
     {
         /// <summary>Safely Get all Types from the assembly. If there is an error while retrieving the types it will return an empty array of <see cref="Type"/>.</summary>
         /// <param name="assembly">The assembly to retrieve all types from.</param>
         /// <returns>An array of all <see cref="Type"/> in the given <see cref="Assembly"/>.</returns>
         /// <remarks>This is obsolete because logging is not added. Please use the SafeGetTypes with the ILog parameter.</remarks>
-        [Obsolete("Deprecated in DotNetNuke 9.9.0. Please use the SafeGetTypes overload with the ILog parameter. Scheduled removal in v11.0.0.")]
-        public static Type[] SafeGetTypes(this Assembly assembly)
+        [DnnDeprecated(9, 9, 0, "Please use the SafeGetTypes overload with the ILog parameter")]
```
david-poindexter, line 22: Is this missing the `RemovalVersion` on purpose?

bdukes: `RemovalVersion` defaults to `major + 2` and only needs to be supplied if that default is being overridden

david-poindexter: Okay, I get it now - thanks for the explanation @bdukes

mitchelsellers (approved):

PR #5762, opened 2023-07-24
CKE Adds option to allow browser by permissions

### LINKED ISSUE #5761: [Enhancement]: CKE option to allow file browser based on folder permissions

### Description of problem

In CKEditorProvider options, it can be set which roles are allowed to use the file browser popup. In many situations however, the browser popup is allowed for users that have access to the file browser root folder.
The browser already checks for permissions of the user to be able to use it, so it would make sense to also be able use the folder permissions to control the visibility of the button.

### Description of solution

I propose to add an option to allow the file browser button to be shown when the user has (at least) read access to the file browser's root folder.

This should add up to the current configuration based on roles: if the user has permission to the file browser button in either way: he'll get the button.

I considered adding a "fake" role for this purpose to the roles chjeckboxlist, but I figured it would be better as an extra option, so it can be configured as an "extra" way of allowing the user to access the browser.

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

skamphuis:
Closes #5761 

## Summary
This PR implements an extra option in the File Browser settings tab, and uses the setting to determine wether the file browser button is allowed for the user.
Supports Host, Site, Page and Module level setting.

### DIFF
--- a/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/App_LocalResources/Options.aspx.resx
+++ b/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/App_LocalResources/Options.aspx.resx
@@ -533,4 +533,10 @@
   <data name="HostUploadFolderLabelForImg.Text" xml:space="preserve">
     <value>Default Upload Folder for images for all portals (overrides setting below)</value>
   </data>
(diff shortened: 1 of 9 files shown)


bdukes (approved):
LGTM, thanks!

PR #5769, opened 2023-07-31
Updated to remove additional elements

mitchelsellers:
Fixes #5552 by introducing the final set of needed changes.

Three additional elements that should have been remove by prior upgrade scripts were updated to be removed `ImportDocumentLibraryCategories`, `ImportDocumentLibraryCategoryAssoc`, and `FitsExtendedPropertyPermission`.

Changes were also made to the `ExportImport_AddUpdateUsersBulk` stored procedure to correct a table reference.

There are two changes requested in #5552 that I do not believe we should fix at this time, given the nature of where those came from.

[dbo].[aspnet_AnyDataInTables] Invalid object name 'dbo.aspnet_WebEvent_Events' P
[dbo].[aspnet_Users_DeleteUser] Invalid object name 'dbo.aspnet_UsersInRoles' P

They were not included.

PR #5790, opened 2023-08-09
Fixed an sql issue in 9.12.1 clean install

### LINKED ISSUE #5789: [Bug]: Current develop branch can't perform clean install

### What happened?

Can install a clean DNN instance with the current 9.12.1 develop branch.

### Steps to reproduce?

1. Download the latest CI build from develop branch or build locally
2. Attempt using the install package to create a clean install

### Current Behavior

Install fails with: Cannot find data type dbo.ExportImportUserType.

### Expected Behavior

Install should work

### Relevant log output

```shell
Msg 2715, Level 16, State 3, Procedure ExportImport_AddUpdateUsersBulk, Line 3 [Batch Start Line 73]
Column, parameter, or variable #1: Cannot find data type dbo.ExportImportUserType.
```

### Anything else?

I think I just figured out the issue and I'll send a PR shortly.

### Affected Versions

9.12.1 (alpha)

Chrome

### Code of Conduct

valadas:
Part of was was done in #5769 would only work upon upgrades because it belongs in a module and sql objects don't exist before that module is installed.

I moved those parts in the ExportImport module upgrade script and it fixed a clean install scenario.

Closes #5789


donker (approved):
Good catch

mitchelsellers (approved):

PR #5806, opened 2023-09-08
Fixed an issue where filenames of downloads would get mangled

### LINKED ISSUE #5551: File download in Resource Manager
## Description of bug
We can download files with the Resource Manager that names contains spaces, accents or special characters like parenthesis but the names of the downloaded files are renamed with underscore at the beginning and at the end of files.

Files without spaces, accents or special characters are downloaded successfully with the same name.

For the files with accents, the downloaded names is really weird like this: 
_=_utf-8_B_UmFwcG9ydCBkdSBqdXJ5IHNvdXRlbmFuY2UgZGUgdGjDqHNlLnBkZg==_=_.pdf

But, file name will always start with **_=utf-8**

## Steps to reproduce
1. Go to 'Persona Bar'.
2. Click on 'Manage' and 'Site Assets'.
3. Scroll folders to find file with spaces, accents or special characters. If you can't find files like that, just rename an existing file with the 'Edit' option of the 'Resource Manager' by a right-click on the existing file.
4. After that, download the file with the 'Download' option on the 'Resource Manager' tools bar or by a right-click on the file.

## Current behavior
The downloaded file can be open on the desktop but we need to rename it if we would like that Windows recognize the type of file for open the good associated application of the downloaded file.

## Expected behavior
Conserve the name of the file.

## Screenshots
image

## Error information
No errors.

## Additional context
I tried with a lot of files.

If I have accents, spaces, or special characters, I have the issue. But if my files are no spaces, etc. the file name downloaded is right.

If I rename the file under the Resource Manager without spaces, etc. the file name downloaded is right. Conversely, if I rename the file under the Resource Manager with a space, etc. the file name downloaded has the issue.

I also tried with a new Uploaded file, and I have the same issue.

## Affected version
* [X] 09.11.01 latest supported release

## Affected browser
* [X] Chrome
* [X] Firefox: names are good with spaces and special characters but the issue exist with weird name for a file with accents

valadas:
Closes #5551

The string is in quotes when there are spaces in the filename and Chrome replaced those with underscores. Filenames with special characters are encoded in the response. This adds a function to convert all those special cases back into a plain string that matches the actual filename in the front end.

### DIFF
--- a/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/components/actions/dnn-action-download-item/dnn-action-download-item.tsx
+++ b/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/components/actions/dnn-action-download-item/dnn-action-download-item.tsx
@@ -18,7 +18,7 @@ export class DnnActionDownloadItem {
   }
 
   private handleClick(): void {
-    this.itemsClient.download(this.item.itemId, true);
+    this.itemsClient.download(this.item.itemId, false);
   }
 
   render() {
--- a/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/services/ItemsClient.ts
+++ b/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/services/ItemsClient.ts
@@ -169,7 +169,6 @@ export class ItemsClient{
                 signal: this.abortController.signal,
             })
             .then(response => {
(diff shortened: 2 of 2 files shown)


bdukes (approved):

david-poindexter (approved):
Awesome @valadas - thanks!  🎉

PR #5807, opened 2023-09-08
Update `web.config` max upload size defaults, correct backend calcs, and fix validation issue preventing increasing

### LINKED ISSUE #5778: [Bug]: Setting the `Max Upload Size (MB)` results in unexpected `web.config` values for `requestLengthDiskThreshold`

### What happened?

Currently, updating the `Max Upload Size (MB)` pictured below results in the `maxRequestLength` and `requestLengthDiskThreshold` values being the same.

image

The default for DNN is as follows:
```xml
    <httpRuntime targetFramework="4.7.2" shutdownTimeout="120" executionTimeout="1200" useFullyQualifiedRedirectUrl="true" maxRequestLength="29296" requestLengthDiskThreshold="81920" maxUrlLength="2048" requestPathInvalidCharacters="&lt;,&gt;,*,%,:,\,?" enableVersionHeader="false" requestValidationMode="2.0" fcnMode="Single" />
```

Updating the value for `Max Upload Size (MB)` to `12` in the UI results in the following
```xml
    <httpRuntime targetFramework="4.7.2" shutdownTimeout="120" executionTimeout="1200" useFullyQualifiedRedirectUrl="true" maxRequestLength="12288" requestLengthDiskThreshold="12288" maxUrlLength="2048" requestPathInvalidCharacters="&lt;,&gt;,*,%,:,\,?" enableVersionHeader="false" requestValidationMode="2.0" fcnMode="Single" />
```

Updating the value for `Max Upload Size (MB)` back to `28` in the UI results in the following
```xml
    <httpRuntime targetFramework="4.7.2" shutdownTimeout="120" executionTimeout="1200" useFullyQualifiedRedirectUrl="true" maxRequestLength="28672" requestLengthDiskThreshold="28672" maxUrlLength="2048" requestPathInvalidCharacters="&lt;,&gt;,*,%,:,\,?" enableVersionHeader="false" requestValidationMode="2.0" fcnMode="Single" />
```

For those that don't know, here is the math for calculating:  

```
1 Megabyte = 1024 Kilobyte
```

The result for a value of `28` (MB) in the UI, should translate to a value of `28672` in the `maxRequestLength` value, and it does when saved.  

However, this shows that the default value of `29296` does NOT in fact translate to `28` (MB).  It would translate to approximately `28.61` (MB).  So, if `28` (MB) is our desired default value, we need to update the default `web.config` value accordingly, and ensure the math for the PersonaBar UI is correct in reading/displaying it.

Microsoft default values are as follows: 
* `maxRequestLength` = `4096` (KB)
* `requestLengthDiskThreshold` = `81920` (B)

```
Note: Thank you Microsoft for not making the unit of measurement the same for both of these.  🤦🏼 
```

Based on the default values by Microsoft, it can be implied (as a general rule of thumb) to set the `requestLengthDiskThreshold` to 1.95% of the `maxRequestLength` value.  However, the value of `requestLengthDiskThreshold` should never exceed that of `maxRequestLength` (when converted for the unit of measurement).

So, our calculated values for `maxRequestLength` and `requestLengthDiskThreshold` should be proportionate to the default values set originally by Microsoft.

That said, I believe we should also add another field to the PersonaBar UI to allow manual changes to `requestLengthDiskThreshold` proportionate defaults (calculated when entering `maxRequestLength`), as there are situations where changing this value is warranted and there is currently no way of making these changes in the UI.  A value exceeding `maxRequestLength` should not be allowed.

### Steps to reproduce?

1. Go to `web.config` and note default values for `httpRuntime`
2. Go to `Settings` > `Security` > `More` > `More Security Settings`
3. Change value for `Max Upload Size (MB)`
4. Go to `web.config` and note new values for `maxRequestLength` and `requestDiskThresholdLength` in `httpRuntime`

### Current Behavior

Values for `maxRequestLength` and `requestDiskThresholdLength` in `httpRuntime` are the same after saving in UI.

### Expected Behavior

Values for `maxRequestLength` and `requestDiskThresholdLength` in `httpRuntime` should be based proportionately according to the defaults established by Microsoft.

### Relevant log output

```shell
n/a
```

### Anything else?

We have never, by default, added the following to the `web.config`.  Should we consider adding this?  It is required for working with large files in IIS 7+.  Please note the unit of measurement for `maxAllowedContentLength` is in Bytes (not KB).  🤦🏼 

```xml
<system.webServer>
    <security>
      <requestFiltering>
        <requestLimits maxAllowedContentLength="28672000" />
      </requestFiltering>
    </security>
</system.webServer>
```

### Affected Versions

9.12.0 (latest release), 10.0.0 (alpha)

_No response_

### Code of Conduct

david-poindexter:
## Summary
Resolves #5778 

The following was done in this PR:
* Updated default values for `maxRequestLength` in `development.config` and `release.config`.
* Added section for IIS7 support of `maxAllowedContentLength`
```xml
<system.webServer>
    <security>
      <requestFiltering>
        <requestLimits maxAllowedContentLength="28672000" />
      </requestFiltering>
    </security>
</system.webServer>
```
* Resolved backend calculation issues resulting in confusing values (by default and when saving)
* Resolved validation issue that was preventing an admin from increasing the Max Upload Size in the UI.  Now it validates against the max supported value of 4GB (minus one byte) for IIS.

Lastly, we decided it best to not add a new field in the UI to control `requestLengthDiskThreshold`.  Since this is not something DNN admins would typically know/understand, it is best to leave this to manual `web.config` changes by server admins wanting to optimize performance for specific scenarios/environments.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/Config.cs
+++ b/DNN Platform/Library/Common/Utilities/Config.cs
@@ -318,8 +318,7 @@ namespace DotNetNuke.Common.Utilities
 
             if (httpNode != null)
             {
-                var maxAllowedContentLength = XmlUtils.GetAttributeValueAsLong(httpNode.CreateNavigator(), "maxAllowedContentLength", 30000000);
-                return maxAllowedContentLength / 1024 / 1024;
+                return 4294967295 / 1024 / 1024; // 4GB (max allowedContentLength supported in IIS7)
             }
 
             return DefaultRequestFilter;
@@ -352,7 +351,6 @@ namespace DotNetNuke.Common.Utilities
             if (httpNode != null)
             {
                 httpNode.Attributes["maxRequestLength"].InnerText = (newSize / 1024).ToString("#");
-                httpNode.Attributes["requestLengthDiskThreshold"].InnerText = (newSize / 1024).ToString("#");
             }
 
             httpNode = configNav.SelectSingleNode("configuration//system.webServer//security//requestFiltering//requestLimits") ??
--- a/DNN Platform/Library/Common/Utilities/Config.cs
+++ b/DNN Platform/Library/Common/Utilities/Config.cs
@@ -318,7 +318,8 @@ namespace DotNetNuke.Common.Utilities
 
             if (httpNode != null)
             {
-                return 4294967295 / 1024 / 1024; // 4GB (max allowedContentLength supported in IIS7)
+                var maxAllowedContentLength = XmlUtils.GetAttributeValueAsLong(httpNode.CreateNavigator(), "maxAllowedContentLength", 30000000);
+                return maxAllowedContentLength / 1024 / 1024;
             }
 
             return DefaultRequestFilter;
@@ -351,6 +352,7 @@ namespace DotNetNuke.Common.Utilities
             if (httpNode != null)
             {
                 httpNode.Attributes["maxRequestLength"].InnerText = (newSize / 1024).ToString("#");
+                httpNode.Attributes["requestLengthDiskThreshold"].InnerText = (newSize / 1024).ToString("#");
             }
 
             httpNode = configNav.SelectSingleNode("configuration//system.webServer//security//requestFiltering//requestLimits") ??
--- a/DNN Platform/Tests/App.config
+++ b/DNN Platform/Tests/App.config
@@ -114,7 +114,7 @@
       <forms name=".DOTNETNUKE" protection="All" timeout="60" cookieless="UseCookies" />
     </authentication>
     <!-- allow large file uploads -->
-    <httpRuntime shutdownTimeout="120" executionTimeout="900" useFullyQualifiedRedirectUrl="true" maxRequestLength="12288" requestLengthDiskThreshold="12288" />
+    <httpRuntime shutdownTimeout="120" executionTimeout="900" useFullyQualifiedRedirectUrl="true" maxRequestLength="28672" requestLengthDiskThreshold="81920" />
     <httpCookies httpOnlyCookies="true" requireSSL="false" domain="" />
     <!--  GLOBALIZATION
     This section sets the globalization settings of the application. 
--- a/DNN Platform/Website/development.config
+++ b/DNN Platform/Website/development.config
@@ -117,6 +117,11 @@
         <add name="X-Frame-Options" value="SAMEORIGIN" />
       </customHeaders>
     </httpProtocol>
+    <security>
+      <requestFiltering>
+        <requestLimits maxAllowedContentLength="29360128" />
+      </requestFiltering>
+    </security>
   </system.webServer>
 
   <system.web>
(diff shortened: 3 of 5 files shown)


DNN Platform/Tests/App.config
```diff
@@ -114,7 +114,7 @@
       <forms name=".DOTNETNUKE" protection="All" timeout="60" cookieless="UseCookies" />
     </authentication>
     <!-- allow large file uploads -->
-    <httpRuntime shutdownTimeout="120" executionTimeout="900" useFullyQualifiedRedirectUrl="true" maxRequestLength="12288" requestLengthDiskThreshold="12288" />
+    <httpRuntime shutdownTimeout="120" executionTimeout="900" useFullyQualifiedRedirectUrl="true" maxRequestLength="28672" requestLengthDiskThreshold="81920" />
```
valadas, line 117:
```suggestion
    <httpRuntime shutdownTimeout="120" executionTimeout="900" useFullyQualifiedRedirectUrl="true" maxRequestLength="28672" requestLengthDiskThreshold="80" />
```

This value is in Kb and the Microsoft default is 80 https://learn.microsoft.com/en-us/dotnet/api/system.web.configuration.httpruntimesection.requestlengthdiskthreshold?view=netframework-4.8.1

valadas (changes requested):
Awesome, just one value is not the "Microsoft default" and is quite big here

DNN Platform/Tests/App.config
valadas: nevermind me, got some conflicting information

valadas (approved):

DNN Platform/Library/Common/Utilities/Config.cs
```diff
@@ -318,8 +318,7 @@ public static long GetRequestFilterSize(IApplicationStatusInfo appStatus)
 
             if (httpNode != null)
             {
-                var maxAllowedContentLength = XmlUtils.GetAttributeValueAsLong(httpNode.CreateNavigator(), "maxAllowedContentLength", 30000000);
-                return maxAllowedContentLength / 1024 / 1024;
+                return 4294967295 / 1024 / 1024; // 4GB (max allowedContentLength supported in IIS7)
```
bdukes, line 321: This value used to be dynamic, and is now hardcoded. Should this be adjusting the default on line 311, instead?

valadas: So before this change, it would send the currently configured maximum, preventing users from changing the value (in the UI) to anything larger (if on IIS7). With this change it allows using the maximum permitted in IIS7

david-poindexter: Yeah, what he said ☝🏼

valadas: In other words if you had manually configured for let's say 100MB, then in the UI you could only change the value to one between 28MB and 100MB

bdukes: It looks to me like this method was returning the correct value, but that [the usage by `SecurityController`](https://github.com/dnnsoftware/Dnn.Platform/blob/40b2bf5b63ec8c15e8f241cd385b469fec11ea11/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SecurityController.cs#L713) was incorrectly using is as the maximum supported size, which is not true. It seems to me like we should use this value in `SecurityController` and leave `Config.GetRequestFilterSize` as-is. Am I missing what we're trying to do here?

valadas: Oh, I understand, makes a lot of sense...

david-poindexter: @bdukes agreed - a commit was just pushed.

bdukes (approved):
👍🏻

PR #5813, opened 2023-09-13
CKE Provider fixes

### LINKED ISSUE #5811: [Bug]: CKE Provider host level settings not stored consistently

### What happened?

I found that host level settings are not working correctly in the CKEditor Provider.

In short: the options form has code to save host level settings, geving the settingname a prefix of DNNCKH# (note the H for HOST level). However, the property that sets the SettingMode to host level, is never set correctly. As a consequence, host level settings are being saved with a prefix of DNNCKP#-1# (note the P for portal level and the -1 for portalId).

When rendering the editor, the host level settings are tried by using the key DNNCKH# which obviously doesn't exist.

### Steps to reproduce?

The issue is hard to see, because of another issue where in the options page host level settings prevale over site level ones, but upon rendering, it's the other way around (I'll log another issue for that).

1. Open the html editor provider settings from Site Settings
2. Uncheck the checkbox "This site only" to enter host level settings
3. Change an obvious setting (e.g. "Editor Height")
4. Check the values in the database: `SELECT  * FROM  CKE_Settings WHERE SettingName LIKE '%#height'`
5. You will notice no setting starts with DNNCKH#

### Current Behavior

For some settings, the host level settings do work, because the DNNCKP#-1# key is used to read them. But in other places it doesn't.

### Expected Behavior

_No response_

### Relevant log output

_No response_

### Anything else?

I'm working on a PR already.

### Affected Versions

9.12.0 (latest release)

_No response_

### Code of Conduct

### LINKED ISSUE #5812: [Bug]: CKE Allow browser based on folder permissions doesn't work on host level

### What happened?

PR #5762 for issue #5761 introduced a setting for the CKEditor provider, to allow the filebrowser to be enabled or disabled, based on the user's permissions on the opened folder.

As it turns out, this setting does not work when set on host level. That is: you can set it on host level, and it will show as set on site level too, but when rendering the editor, a user with permissions still doesn't get a "Brose Server" button in the image popup.

### Steps to reproduce?

1. Go to CKE settings in Site Settings
2. Uncheck "This site only"
3. Check "Allow users with folder permissions" on the second tab
4. Make sure only administrators role is checked in the allowed roles setting
5. Save
6. Login with a user that has edit rights on a page, but is not an administrator
7. Edit and HTML/Text module
8. Notice ther's no browse server button in the image dialog

### Current Behavior

_No response_

### Expected Behavior

When this setting is set on host level, a non-admin user with edit rights, should be able to browse the server for images.

### Relevant log output

_No response_

### Anything else?

I'm wokring on a PR.

### Affected Versions

9.12.0 (latest release)

_No response_

### Code of Conduct

skamphuis:
This PR closes 2 issues:
Closes #5811 
Closes #5812 

## Summary
for #5811 :
Since I have not been able to find any site with a settingname starting with the host level prefix of DNNCKH#, I changed the host level prefix to be the one that's actually being used: DNNCKP#-1#
For obvious reasons I also changed the occurences of strings DNNCHP# to use SettingConstants.

for #5812 :
I had to move the part that reads the setting value outside of an if-block, and stop it from overriding the host level setting with the site level setting, when set on host level.

### DIFF
--- a/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
+++ b/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
@@ -4,6 +4,7 @@
 namespace DNNConnect.CKEditorProvider.Constants
 {
     using System;
+    using System.Dynamic;
 
     /// <summary>Provider Constants.</summary>
     public static class SettingConstants
@@ -128,5 +129,19 @@ namespace DNNConnect.CKEditorProvider.Constants
 
         /// <summary>The resize Upload height setting name.</summary>
         public const string RESIZEHEIGHTUPLOAD = "resizeheightupload";
+
+        /// <summary>Gets the prefix key for host level settings.</summary>
+        /// <returns>Host Key.</returns>
+        public static string HostKey => PortalKey(-1);
+
+        /// <summary>Gets the prefix key for portal level settings.</summary>
+        /// <param name="portalId">The portal id.</param>
+        /// <returns>Portal Key.</returns>
+        public static string PortalKey(int portalId) => PortalKey(portalId.ToString());
+
+        /// <summary>Gets the prefix key for portal level settings.</summary>
+        /// <param name="portalId">The portal id as a string.</param>
+        /// <returns>Portal Key.</returns>
+        public static string PortalKey(string portalId) => $"DNNCKP#{portalId}#";
     }
 }
--- a/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
+++ b/DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
@@ -4,7 +4,7 @@
 namespace DNNConnect.CKEditorProvider.Constants
 {
     using System;
-    using System.Dynamic;
+    using System.Globalization;
 
     /// <summary>Provider Constants.</summary>
     public static class SettingConstants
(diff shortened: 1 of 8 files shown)


DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
```diff
@@ -4,13 +4,14 @@
 namespace DNNConnect.CKEditorProvider.Constants
 {
     using System;
-
+    using System.Dynamic;
```
bdukes, line 7:
```suggestion
    using System.Globalization;
```

DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Constants/SettingConstants.cs
```diff
@@ -128,5 +129,19 @@ public static class SettingConstants
...
         public const string RESIZEHEIGHTUPLOAD = "resizeheightupload";
+
+        /// <summary>Gets the prefix key for host level settings.</summary>
+        /// <returns>Host Key.</returns>
+        public static string HostKey => PortalKey(-1);
+
+        /// <summary>Gets the prefix key for portal level settings.</summary>
+        /// <param name="portalId">The portal id.</param>
+        /// <returns>Portal Key.</returns>
+        public static string PortalKey(int portalId) => PortalKey(portalId.ToString());
```
bdukes, line 140:
```suggestion
        public static string PortalKey(int portalId) => PortalKey(portalId.ToString(CultureInfo.InvariantCulture));
```

bdukes (approved):
LGTM

mitchelsellers (approved):

PR #5816, opened 2023-09-19
Move DnnDeprecatedAttribute to Abstractions

### LINKED ISSUE #5815: [Bug]: v9.13.0-rc1 has System.Runtime.CompilerServices.Unsafe in bin folder of upgrade zip

### What happened?

Upgrading a site that had a module using System.Runtime.CompilerServices.Unsafe properly defined it manifest, I got a binding redirect error. The dll got replaced by DNN upgrade zip.

### Steps to reproduce?

1. Install a clean DNN v9.12.x
2. Install a module with System.Runtime.CompilerServices.Unsafe properly defined in its manifest
3. Confirm the binding redirect works in the web.config after installing the module
4. Upgrade to DNN v9.13.0
5. Notice the installer does not fire and a binding redirect error regarding that dependency

### Current Behavior

The DNN v9.13 upgrade.zip has System.Runtime.CompilerServices.Unsafe in its bin folder

### Expected Behavior

Either that file is there by mistake (I only saw that dependency on DNN side of things in tests)

or

We should have a package for it so binding redirects are updated.

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.0 (release candidate)

Chrome

### Code of Conduct

bdukes:
In order for projects to not require a full reference to the source generator project, the attribute cannot be defined in the project. The first attempt to resolve this was to generate the attribute via the source generator itself, but since the Library project exposes its internals to other projects (e.g. DotNetNuke.HttpModules), this resulted in ambiguous references. As a result, moving the type to a project that all other projects can reference seemed to be the best compromise. Since the attribute is conditional, it doesn't affect the public API of that project.

Fixes #5815

### DIFF
--- a/DNN Platform/DotNetNuke.Abstractions/DnnDeprecatedAttribute.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/DnnDeprecatedAttribute.cs
@@ -1,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Internal.SourceGenerators;
+
+using System;
+using System.Diagnostics;
+
+/// <summary>Marks a type or member as deprecated.</summary>
+[Conditional("DNN_SOURCE_GENERATOR")]
+public class DnnDeprecatedAttribute : Attribute
+{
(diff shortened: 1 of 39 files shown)


valadas (approved):
Awesome, thanks so much, I'll test out the artififact but from reading, it looks good to me...

mitchelsellers (approved):

david-poindexter (approved):
Awesome!

Issue #5819, opened 2023-09-22
[Enhancement]: various enhancements from a large installation with external cache


skamphuis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of problem

I have been working in with a rather large DNN installation in a quite specific environment. This single DNN instance has approximately 3000 portals with about 10 languages is running in a webfarm, using Redis as an external cache.

In this environment we ran into some issues we needed to address to improve performance.

### Description of solution

I'm creating 3 PR's related to this, by grouping the issues into 3 areas:
* Logging: we needed to use logging a lot to find issues related to the external cache. We found there are some log messages being added a lot, that don't really help for anything.
* Caching: using an external cache adds some overhead to using the cache. That way we found a few places where caching could be improved
* Fixes: we found a few places where exceptions could be prevented, and we ran into a few smaller issues (small to DNN, big impact for us, mostly).

### Do you be plan to contribute code for this enhancement?

- [X] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5820, opened 2023-09-22
Feature/5819 improvements logging

skamphuis:
This PR relats to #5819 and covers the logging-related changes:
- Skip warning for moniker not found, when moniker value is empty
- In DDRMenu.FilterNodes only do GetTab when tabId > 0
- DDRMenu: Don't GetTab when we already know TabId <= 0
- BasicUrlRewriter: no log after redirect

### DIFF
--- a/DNN Platform/Modules/DDRMenu/MenuBase.cs
+++ b/DNN Platform/Modules/DDRMenu/MenuBase.cs
@@ -222,7 +222,6 @@ namespace DotNetNuke.Web.DDRMenu
                                     return false;
                                 }
 
-
                                 var tab = TabController.Instance.GetTab(n.TabId, Null.NullInteger, false);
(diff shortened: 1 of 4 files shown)


DNN Platform/Modules/DDRMenu/MenuBase.cs
```diff
@@ -216,6 +216,13 @@ private void FilterNodes(string nodeString, bool exclude)
                         this.RootNode.Children.FindAll(
                             n =>
                             {
+                                // no need to check when the node is not a page
+                                if (n.TabId <= 0)
+                                {
+                                    return false;
+                                }
+
+
```
valadas, line 225:
```suggestion
```

valadas (approved):
Looks good to me other than an extra empty line

PR #5821, opened 2023-09-22
Feature/5819 improvements fixes

skamphuis:
This PR relats to #5819 and covers the fixes:
- Prevent exception for missing PortalSetting PageHeadText
- Fix UpdatePortalInfo for new language when en-US not present
- Fix mail Null Reference exception in SendMail found in upgraded portals
- When NodeManipulator adds items they have no tabid, do not filter on null.
- Scheduler Fix for error thrown when no servers active

### DIFF
--- a/DNN Platform/Library/Services/Scheduling/Scheduler.cs
+++ b/DNN Platform/Library/Services/Scheduling/Scheduler.cs
@@ -128,6 +128,12 @@ namespace DotNetNuke.Services.Scheduling
                 // Get the enabled servers with recent activity.
                 var enabledServers = ServerController.GetEnabledServersWithActivity();
 
+                if (!enabledServers.Any())
+                {
+                    // No enabled servers with activity, so just return true to indicate the scheduled item was not updated.
+                    return true;
+                }
+
                 // Validate that any server in the scheduled item is in the enabled server list with activity in the last 10 minutes.
                 bool serverHasRecentActivity = enabledServers.Any(i => servers.Contains(i.ServerName.ToLowerInvariant()));
 
--- a/DNN Platform/Library/Services/Mail/Mail.cs
+++ b/DNN Platform/Library/Services/Mail/Mail.cs
@@ -131,7 +131,7 @@ namespace DotNetNuke.Services.Mail
             // Send Notification to User
             var user = UserController.Instance.GetUserById(portalId, userId);
             int toUser = user.UserID;
-            string locale = user.Profile.PreferredLocale;
+            string locale = user.Profile.PreferredLocale ?? settings.DefaultLanguage;
             string subject;
             string body;
             ArrayList custom = null;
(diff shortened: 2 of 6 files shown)


DNN Platform/Library/Services/Scheduling/Scheduler.cs
```diff
@@ -128,6 +128,12 @@ public static bool ValidateServersAreActiveForScheduledItem(ScheduleItem schedul
                 // Get the enabled servers with recent activity.
                 var enabledServers = ServerController.GetEnabledServersWithActivity();
 
+                if (!enabledServers.Any())
```
mitchelsellers, line 131: This change breaks the desired behavior here to move the job to another server that has activity

skamphuis: Hmm, the intention is to fix the exception that occurs on `enabledServers.First()` when there are no enabledServers:


```
                // Get the enabled servers with recent activity.
                var enabledServers = ServerController.GetEnabledServersWithActivity();
```
if enabledServers is an empty `List<ServerInfo>` here.
```
                // Validate that any server in the scheduled item is in the enabled server list with activity in the last 10 minutes.
                bool serverHasRecentActivity = enabledServers.Any(i => servers.Contains(i.ServerName.ToLowerInvariant()));
```
`serverHasRecentActivity` will be `false` here.
```
                // If no servers in the scheduled item had recent activity, assign this scheduled item to the server with the most recent activity
                if (!serverHasRecentActivity)
                {
```
then this line will throw a `InvalidOperationException: Sequence contains no elements` because there is no `.First()`.
```
                    // Assign a new server to this task then run it
                    scheduleItem.Servers = $",{enabledServers.First().ServerName.ToLowerInvariant()},";

                    // Save the scheduled item back to the database.
                    SchedulingController.UpdateSchedule(scheduleItem);
                    return false;
                }

```

That's what we saw in the logs and what I tried to fix here.

mitchelsellers: Ah this makes sense but we don't want to just run everywhere

skamphuis: The way I see it, I'm not changing behavior in any way. I'm just preventing an exception by saying: if there are no available servers, then do nothing. 

That said, I'm happy to improve something if I know what to improve...

mitchelsellers: @mtrutledge Do you have any guidance on this one

mtrutledge: There should always be at least 1 enabled server. In what conditions are we seeing the WebServers table and the enabledServers count is 0?

skamphuis: Sorry for the delay,I had to dive into that. 

We experienced that after changing the frequency of the Montior Web Servers task. It looks like if that task isn't running (or not frequently enough?) it might occur that there are no enabled servers and the exception that I'm catching here, is thrown.

mtrutledge:
What did the frequency of the monitor get changed to? Was it more than 10
minutes? What were the initial reasons why the frequency of that got
changed? Just want to make sure we have the whole picture. I don't want a
side effect of this change to let the scheduler run when there are no
active servers in the list but a schedule has been told to run on a
specific server.

skamphuis:
@mtrutledge sorry, missed your questiojn again. I'll get back at this.

mitchelsellers:
@skamphuis I am gong to close this for now, you can re-open if you would like to adjust/submit for merging

PR #5822, opened 2023-09-22
Feature/5819 improvements caching

skamphuis:
This PR relats to #5819 and covers the caching-related changes:
- Fix serialization errors when using external cache
- HTML/Text: implement caching for GetWorkflows
- Only do a clearcache when unknown tab belongs to the portal whose cache is being flushed.
- TabController: only clear cache when tab is for the current portal

### DIFF
--- a/DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
+++ b/DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
@@ -13,16 +13,31 @@ namespace DotNetNuke.Modules.Html
     public class WorkflowStateController
     {
         private const string WORKFLOWCACHEKEY = "Workflow{0}";
+        private const string WORKFLOWS_CACHE_KEY = "Workflows{0}";
         private const int WORKFLOWCACHETIMEOUT = 20;
 
         private const CacheItemPriority WORKFLOWCACHEPRIORITY = CacheItemPriority.Normal;
 
-        /// <summary>GetWorkFlows retrieves a collection of workflows for the portal.</summary>
+        /// <summary>GetWorkFlows retrieves a collection of workflows for the portal from the cache.</summary>
         /// <param name="portalID">The ID of the Portal.</param>
         /// <returns>An <see cref="ArrayList"/> of <see cref="WorkflowStateInfo"/> instances.</returns>
         public ArrayList GetWorkflows(int portalID)
         {
-            return CBO.FillCollection(DataProvider.Instance().GetWorkflows(portalID), typeof(WorkflowStateInfo));
+            var cacheKey = string.Format(WORKFLOWS_CACHE_KEY, portalID);
+            return CBO.GetCachedObject<ArrayList>(new CacheItemArgs(cacheKey, WORKFLOWCACHETIMEOUT, WORKFLOWCACHEPRIORITY, portalID), this.GetWorkflowsCallBack);
+        }
+
+        /// -----------------------------------------------------------------------------
+        /// <summary>
+        ///   GetWorkflowsCallBack retrieves a collection of WorkflowStateInfo objects for the Portal from the database.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


mitchelsellers (approved):
This looks good to me!

DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
```diff
@@ -13,16 +13,31 @@ namespace DotNetNuke.Modules.Html
     public class WorkflowStateController
     {
         private const string WORKFLOWCACHEKEY = "Workflow{0}";
+        private const string WORKFLOWS_CACHE_KEY = "Workflows{0}";
```
valadas, line 16: Our current stylecop rules prevent us from using _ (underscores) for these variables which is making the build fail currently.

DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
```diff
@@ -13,16 +13,31 @@ namespace DotNetNuke.Modules.Html
...
+        }
+
+        /// -----------------------------------------------------------------------------
+        /// <summary>
+        ///   GetWorkflowsCallBack retrieves a collection of WorkflowStateInfo objects for the Portal from the database.
+        /// </summary>
+        /// <param name = "cacheItemArgs">Arguments passed by the GetWorkflowStates method.</param>
+        /// <returns>WorkflowStateInfo List.</returns>
+        /// -----------------------------------------------------------------------------
+        public object GetWorkflowsCallBack(CacheItemArgs cacheItemArgs)
```
valadas, line 37: Do we _have_ to return an object here ?

valadas (changes requested):

DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
skamphuis: Right, how did I not notice that...

DNN Platform/Modules/HTML/Components/WorkflowStateController.cs
```diff
@@ -13,16 +13,31 @@ namespace DotNetNuke.Modules.Html
...
         private const CacheItemPriority WORKFLOWCACHEPRIORITY = CacheItemPriority.Normal;
 
-        /// <summary>GetWorkFlows retrieves a collection of workflows for the portal.</summary>
+        /// <summary>GetWorkFlows retrieves a collection of workflows for the portal from the cache.</summary>
         /// <param name="portalID">The ID of the Portal.</param>
         /// <returns>An <see cref="ArrayList"/> of <see cref="WorkflowStateInfo"/> instances.</returns>
         public ArrayList GetWorkflows(int portalID)
         {
-            return CBO.FillCollection(DataProvider.Instance().GetWorkflows(portalID), typeof(WorkflowStateInfo));
+            var cacheKey = string.Format(WORKFLOWSCACHEKEY, portalID);
```
bdukes, line 26: Do we need to also clear this cache if a workflow gets updated/added/removed?

skamphuis: Probably, but I haven't found any code that does such a thing...

skamphuis: @bdukes  Would it make sense to add an optional clearCache parameter to the method so extensions can control the cache themselves? Or a new method "ClearWorkflowsCache" or something?

bdukes: You're right, there isn't any code to modify these HTML workflows (in contrast to the content workflows used by the platform). Clearing the cache would be the responsibility of whoever is updating that table. I'm okay with this as-is.

bdukes (approved):

mitchelsellers:
@valadas can we get another review from you to clear the requested changes review

valadas (approved):
Awesome

PR #5835, opened 2023-10-09
Prevent control characters in username

### LINKED ISSUE #5834: [Bug]: Login module allows control characters

### What happened?

In Discord a user reported that they received the following error:

> EXCEPTION: System.ArgumentException: '.', hexadecimal value 0x00, is an invalid character.
   at System.Xml.XmlEncodedRawTextWriter.WriteAttributeTextBlock(Char* pSrc, Char* pSrcEnd)
   at System.Xml.XmlEncodedRawTextWriter.WriteString(String text)
   at System.Xml.XmlWellFormedWriter.WriteString(String text)
   at System.Xml.XmlWriter.WriteAttributeString(String localName, String value)
   at DotNetNuke.Services.Log.EventLog.LogInfo.WriteXml(XmlWriter writer)
   at DotNetNuke.Services.Log.EventLog.LogInfo.Serialize()
   at DotNetNuke.Services.Log.EventLog.DBLoggingProvider.SendLogNotifications()
   at DotNetNuke.Services.Log.EventLog.SendLogNotifications.DoWork()

This happend because there was an attempt to log in with a username that contains an invalid character (`\0`). DNN tried to log this attempt, but the XML serializer doesn't allow null-characters causing an exception to be thrown. I don't think this is a security issue, since everything was handled fine, but maybe we should block control characters (like the null-character).

### Steps to reproduce?

I did not find a real way to reproduce this. You can however send invalid characters to the textbox with the `username` query.

1. Go to the login page (e.g. http://localhost/Login)
2. Add `?username=User%00name` to the query (e.g. http://localhost/Login?username=User%00name)

### Current Behavior

The null character will be set in the input field:

image

Causing the username field to show an null character:

image

### Expected Behavior

The null character won't be shown.

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.0 (latest release)

_No response_

### Code of Conduct

GerardSmit:
Fixes #5834

## Summary
Prevents control characters to be inserted in the username field.

Before (http://localhost/Login?username=User%00name):
image

After (http://localhost/Login?username=User%00name):
image

Noteworthy, this blocks _all_ control characters, including `\n` or `\r`. Instead of removing `\t` I've added a replacement of a space, but I don't think anyone uses a tab (since this switches the input field in the browser) or new lines (since this isn't a textarea) in their username.

### DIFF
--- a/DNN Platform/Library/Security/PortalSecurity.cs
+++ b/DNN Platform/Library/Security/PortalSecurity.cs
@@ -38,6 +38,7 @@ namespace DotNetNuke.Security
         private static readonly Regex StripTagsRegex = new Regex("<[^<>]*>", RegexOptions.IgnoreCase | RegexOptions.Singleline | RegexOptions.Compiled);
         private static readonly Regex BadStatementRegex = new Regex(BadStatementExpression, RegexOptions.IgnoreCase | RegexOptions.Compiled);
         private static readonly Regex ControlCharactersRegex = new Regex(@"\p{C}+", RegexOptions.Compiled);
+        private static readonly Regex ControlCharacterToWhitespaceRegex = new Regex(@"\r\n|\r|\n|\t", RegexOptions.Compiled);
 
         private static readonly Regex[] RxListStrings = new[]
         {
@@ -114,6 +115,9 @@ namespace DotNetNuke.Security
             /// <summary>
             /// Removes all unicode control characters (like \0, \t, \n, \r, etc.) from the string.
             /// </summary>
+            /// <remarks>
+            /// The control characters \r\n, \r, \n, and \t are replaced with a single space instead of being removed.
+            /// </remarks>
             NoControlCharacters = 64,
         }
 
@@ -1035,7 +1039,7 @@ namespace DotNetNuke.Security
         /// <remarks>This is a private function that is used internally by the <see cref="InputFilter"/> function.</remarks>
         private static string FormatRemoveControlCharacters(string str)
         {
-            return ControlCharactersRegex.Replace(str.Replace('\t', ' '), string.Empty);
+            return ControlCharactersRegex.Replace(ControlCharacterToWhitespaceRegex.Replace(str, " "), string.Empty);
         }
 
         /// <summary>This function determines if the Input string contains any markup.</summary>
--- a/DNN Platform/Library/Security/PortalSecurity.cs
+++ b/DNN Platform/Library/Security/PortalSecurity.cs
@@ -37,6 +37,7 @@ namespace DotNetNuke.Security
 
         private static readonly Regex StripTagsRegex = new Regex("<[^<>]*>", RegexOptions.IgnoreCase | RegexOptions.Singleline | RegexOptions.Compiled);
         private static readonly Regex BadStatementRegex = new Regex(BadStatementExpression, RegexOptions.IgnoreCase | RegexOptions.Compiled);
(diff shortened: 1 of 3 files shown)


DNN Platform/Library/Security/PortalSecurity.cs
```diff
@@ -1018,6 +1029,15 @@ private static string FormatRemoveSQL(string strSQL)
             return BadStatementRegex.Replace(strSQL, " ").Replace("'", "''");
         }
 
+        /// <summary> This function removes control characters from the string.</summary>
+        /// <param name="str">This is the string to be filtered.</param>
+        /// <returns>Filtered UserInput.</returns>
+        /// <remarks>This is a private function that is used internally by the <see cref="InputFilter"/> function.</remarks>
+        private static string FormatRemoveControlCharacters(string str)
+        {
+            return ControlCharactersRegex.Replace(str.Replace('\t', ' '), string.Empty);
```
bdukes, line 1038: Should removing control characters replace a tab with a space?

bdukes (approved):
Thanks!

DNN Platform/Library/Security/PortalSecurity.cs
GerardSmit: I wasn't sure about this one, since a tab is still some whitespace. I didn't know if you still want to see in the logging that a whitespace/tab was used.

If you want I can replace it with an empty string as well.

bdukes: I hadn't realized that [control characters](https://en.wikipedia.org/wiki/Unicode_control_characters) includes tabs and line feeds. Replacing with a space makes sense to me.

GerardSmit: Should I replace `\n` and `\r` (new lines) with spaces as well?

bdukes: 👍🏻 that would make sense to me

DNN Platform/Library/Security/PortalSecurity.cs
```diff
@@ -37,6 +37,8 @@ public class PortalSecurity
 
         private static readonly Regex StripTagsRegex = new Regex("<[^<>]*>", RegexOptions.IgnoreCase | RegexOptions.Singleline | RegexOptions.Compiled);
         private static readonly Regex BadStatementRegex = new Regex(BadStatementExpression, RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex ControlCharactersRegex = new Regex(@"\p{C}+", RegexOptions.Compiled);
+        private static readonly Regex ControlCharacterToWhitespaceRegex = new Regex(@"\r\n|\r|\n|\t", RegexOptions.Compiled);
```
bdukes, line 41: If we're matching `\r` and `\n`, does it make a difference to search for them together also?

bdukes: Oh, you're trying to avoid introducing multiple spaces, got it.

bdukes (approved):
👍🏻

DNN Platform/Library/Security/PortalSecurity.cs
GerardSmit: Windows uses `\r\n` for new lines and `\r\n` is seen as one new line, instead of two. If we were to make this case separate, they would show two spaces. If this isn't a problem, I can remove `\r\n`.

mitchelsellers (approved):

PR #5838, opened 2023-10-10
Add ISkinService for Abstractions Project

### LINKED ISSUE #5837: [Enhancement]: Add Dependency Injection Support for SkinController

### Description of problem

The current implementation of `SkinController` does not have abstraction layer that can be used with dependency injection.

### Description of solution

Create an interface for this class called `ISkinService`.

Don't add an interface and keep on using `SkinController`.

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

GerardSmit:
Fixes #5837 

## Summary
This PR adds `ISkinService` which is an abstraction layer for `SkinController`.

## Things to validate
1. I've renamed properties with `ID` to `Id`.
    Only `DotNetNuke.Abstractions.Users.IUserInfo` uses `ID` (`AffiliateID`, `PortalID`, `UserID`), the rest uses `Id`. 
2. I've moved 2 enums to the abstractions; I've added `[TypeForwardTo]` so extensions don't have to recompile.
3. I didn't mark the static methods as `[Obsolete]` and I haven't modified the calls to `PortalController`.
4. I've added the service as a scoped service; just in case the service will use scoped services in the future.


DNN Platform/Library/UI/Skins/SkinController.cs
```diff
@@ -305,9 +348,12 @@ public static void SetSkin(string skinRoot, int portalId, SkinType skinType, str
                     }
 
                     break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(skinRoot), skinRoot, "Expected either 'Skins' or 'Containers'");
```
GerardSmit, line 352: I've added an exception to this method, since the method could fail silently if you made a typo. Is this OK?

DNN Platform/Library/UI/Skins/SkinScope.cs
```diff
@@ -1,13 +1,10 @@
...
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+using System.Runtime.CompilerServices;
+
+using DotNetNuke.UI.Skins;
+
+// The enum 'SkinScope' has been moved to DotNetNuke.Abstractions
+[assembly: TypeForwardedTo(typeof(SkinScope))]
```
valadas, line 10: How does this know to forwared to DotNetNuke.Abstractions ? Is this right or we need to fully qualify that type?

DNN Platform/Library/UI/Skins/SkinType.cs
```diff
@@ -1,12 +1,10 @@
...
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+using System.Runtime.CompilerServices;
+
+using DotNetNuke.UI.Skins;
+
+// The enum 'SkinType' has been moved to DotNetNuke.Abstractions
+[assembly: TypeForwardedTo(typeof(SkinType))]
```
valadas, line 10: Same concern here about forwaring to DotNetNuke.Abstractions

valadas (commented):
I have 2 small concernes but I am not sure if relevant, just something to double-check...

DNN Platform/Library/UI/Skins/SkinScope.cs
GerardSmit: Because DotNetNuke.Library has a reference to DotNetNuke.Abstractions. The `typeof` is the type in the assembly DotNetNuke.Abstractions, so the compiler knows which assembly the type has moved to.

It's also not possible to add a fully qualifier in this attribute, [since it only accepts `System.Type`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.typeforwardedtoattribute.-ctor?view=netframework-4.7.2#system-runtime-compilerservices-typeforwardedtoattribute-ctor(system-type)).

More info about TypeForwardTo: https://learn.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding

valadas: What would happen if SkinScope type existsted in multiple other referenced assemblies ? How would it know which one to forward to ?

GerardSmit: I think it depends on the namespace.

If the class/enum exists in multiple assemblies with the same namespace, you would get a ambiguous reference. In this case you can't even compile the library.

Since the enum has been moved to Abstractions, with the same namespace, it'll be fine.

bdukes: I haven't used type forwarding, but AFAIK it should work fine in this scenario. I guess the only concern here is that we don't get to change the namespace to `DotNetNuke.Abstractions.Skins`, it has to remain `DotNetNuke.UI.Skins`. Is it worth introducing the conversion code between the two separate `enum` types so that we can have cleaner organization in the long-term?

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
```diff
@@ -27,6 +28,7 @@ public class SkinPackageInfo : BaseEntityInfo, IHydratable
         private string skinType;
         private Dictionary<int, string> skins = new Dictionary<int, string>();
 
+        /// <inheritdoc cref="ISkinPackageInfo.PackageId"/>
         public int PackageID
```
bdukes, line 32: Do we want to mark these properties as deprecated?

DNN Platform/DotNetNuke.Abstractions/Skins/SkinType.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// The skin is a portal skin.
+    /// </summary>
+    Portal = 0,
+
+    /// <summary>
+    /// The skin is an admin skin.
+    /// </summary>
+    Admin = 1,
```
bdukes, line 20: Do we want to rename these to `Site` and `Edit` (matching `Dnn.PersonaBar.Themes.Components.ApplyThemeScope`)? It took me a few minutes to figure out what these meant, since we don't use these terms/concepts in the same way anymore.

bdukes (commented):
Thanks! We appreciate your work, this is a nice and clean PR. Hopefully we can quickly resolve our couple of questions and get it included.

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
GerardSmit: I'm against this if `int PackageId` is not explicit.

This happend to PortalId as well:

https://github.com/dnnsoftware/Dnn.Platform/blob/3af88c5078e102d2316aac90a72c0abb634882f3/DNN%20Platform/Library/Entities/Portals/PortalInfo.cs#L212-L221

`ID` was deprecated but this makes it really awkward for current code. If you try to get the portal ID, you get a deprecated warning:

```cs
portalInfo.PortalID; // Deprecate warning.
```

But if you try to get `PortalId`, you'll get an error:
```cs
portalInfo.PortalId; // Error: PortalId does not exists.
```

This is because `PortalId` is an implicit implementation, so you have to cast the instance to an interface before you can get it (because `ThisAsInterface` is private):
```cs
((IPortalInfo)portalInfo).PortalID;
```

---

I suggest the following:

Add `[Obsolete]` and `[EditorBrowsable]` to the property, and make `PackageId` public:
```cs
public class SkinPackageInfo
{
    [Obsolete("Deprecated in DotNetNuke 10.0.0. Use DotNetNuke.Abstractions.Portals.IPortalInfo.PortalId instead. Scheduled removal in v12.0.0.")]
    [EditorBrowsable(EditorBrowsableState.Never)]
    public int PackageID
    {
        get => this.PackageId;
        set => this.PackageId = value;
    }
    
    public int PackageId
    {
        get => this.packageID;
        set => this.packageID = value;
    }
}
```

This way:
1. You can still access `skinPackageInfo.PackageID` with a deprecated warning.
2. You can change it to `skinPackageInfo.PackageId` without an error.
3. You won't see `PackageID` and `PackageId` in Visual Studio, since the property `PortalID` is hidden.
    _Note, in [Rider](https://www.jetbrains.com/rider/) the property is still visible [if the setting in "Settings → Editor → General → Code Completion → Filter members by [EditorBrowsable] attribute" isn't enabled](https://youtrack.jetbrains.com/issue/RIDER-17669/Respect-EditorBrowsableEditorBrowsableState.Never#focus=Comments-27-6455778.0-0)._

DNN Platform/Library/UI/Skins/SkinController.cs
```diff
@@ -51,17 +55,32 @@ public static string RootContainer
...
+        string ISkinService.RootContainer => RootContainer;
+
+        /// <inheritdoc cref="ISkinService.AddSkin" />
         public static int AddSkin(int skinPackageID, string skinSrc)
         {
             return DataProvider.Instance().AddSkin(skinPackageID, skinSrc);
         }
 
+        /// <inheritdoc cref="ISkinService.AddSkinPackage" />
         public static int AddSkinPackage(SkinPackageInfo skinPackage)
```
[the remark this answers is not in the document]
GerardSmit: If we deprecate this method, you'll need to cast every call to `AddSkinPackage`, but then you'll get another deprecations if we're deprecating all static methods in https://github.com/dnnsoftware/Dnn.Platform/pull/5838#discussion_r1358388559.

```cs
SkinPackageInfo info;

AddSkinPackage(info); // Warning about deprecation
AddSkinPackage((ISkinPackageInfo)info); // Warning if all static methods are deprecated
```

We can also make `static void AddSkinPackage(ISkinPackageInfo skinPackage)` private and only allow access through `ISkinService.AddSkinPackage(ISkinPackageInfo skinPackage)`.

DNN Platform/Library/UI/Skins/SkinScope.cs
GerardSmit: Should we rename the enum in the abstractions assembly?
Otherwise, you'll have two `SkinType`'s; one deprecated and one that you should use. You'll have to make sure to get the right namespace or fill in the full enum name (`DotNetNuke.Abstractions.Skins.SkinType`)

DNN Platform/DotNetNuke.Abstractions/Skins/SkinType.cs
GerardSmit: Should I rename the following methods in ISkinService as well?

| Current                     | New                       |
|-----------------------------|---------------------------|
| `GetDefaultAdminContainer`  | `GetDefaultEditContainer` |
| `GetDefaultAdminSkin`       | `GetDefaultEditSkin`      |
| `GetDefaultPortalContainer` | `GetDefaultSiteContainer` |
| `GetDefaultPortalSkin`      | `GetDefaultSiteSkin`      |

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
bdukes: The problem with this is that usage of the property becomes ambiguous in VB.NET, since it's case insensitive (this was what we originally did with `PortalInfo.PortalId`, until users reported errors https://github.com/dnnsoftware/Dnn.Platform/pull/4397).

I agree that it's awkward to deprecate the member without having the replacing member visible. I don't have strong feelings about deprecating the property or not. Any thoughts @dnnsoftware/approvers?

DNN Platform/Library/UI/Skins/SkinController.cs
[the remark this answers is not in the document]
bdukes: If we decide to deprecate all of the static methods, then we might as well make the `ISkinPackageInfo` overloads private.

DNN Platform/Library/UI/Skins/SkinScope.cs
bdukes: The other enums that we've moved have not been renamed, we just require using the right namespace. It is certainly messier in the short term to do it that way, I'm not certain if it's worth it in the long term or not.

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
GerardSmit: I didn't know VB.NET was case insensitive...

So either VB.NET does a cast (`DirectCast(info, ISkinPackageInfo).PackageId`) when we make it public or C# does a cast (`((IPortalInfo)ISkinPackageInfo).PackageId`) when we make it an implicit implementation 😅.

Both are annoying, but VB.NET will not compile and is being used a lot in skins; which will break a lot of skins.
C# is just a warning, still annoying if you want to fix all the deprecations but it won't break anything.

In my opinion the best thing to do is to not deprecate the property.

DNN Platform/Library/UI/Skins/SkinController.cs
GerardSmit: I've created a new enum called `SkinPackageType` that can be either Skin or Container.

DNN Platform/DotNetNuke.Abstractions/Skins/ISkinPackageInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>Gets or sets the ID of the portal.</summary>
+    /// <remarks>If the portal ID is <c>-1</c>, then the skin package is a global skin package.</remarks>
+    int PortalId { get; set; }
+
+    /// <summary>Gets or sets the name of the skin.</summary>
+    string SkinName { get; set; }
+
+    /// <summary>Gets the skins in the skin package.</summary>
+    IList<ISkinInfo> Skins { get; }
```
bdukes, line 26: If we're going to use a collection interface, I think the read-only version is more expressive of the intent of the method (i.e. changing the collection would/should not have any effect).
```suggestion
    IReadOnlyList<ISkinInfo> Skins { get; }
```

DNN Platform/Library/Common/Utilities/SkinUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+        };
+    }
+
+    /// <summary>Gets the skin package type from a database name.</summary>
+    /// <param name="databaseName">The database name of the skin package type.</param>
+    /// <returns>The skin package type.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="databaseName"/> is not a valid skin package type database name.</exception>
+    public static SkinPackageType FromDatabaseName(string databaseName)
+    {
+        return databaseName switch
```
bdukes, line 36: Is there any benefit to doing a case-insensitive check here?

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
```diff
@@ -79,9 +87,10 @@ public string SkinName
             }
         }
 
-        [XmlIgnore]
+        /// <inheritdoc cref="ISkinPackageInfo.Skins"/>
+        [XmlIgnore]
         [JsonIgnore]
-        public Dictionary<int, string> Skins
+        public List<SkinInfo> Skins
```
bdukes, line 93: Is this intended to be a breaking change?

GerardSmit:
I've changed the abstraction a little bit, in my eyes it makes more sense this way.

## SkinPackageType
I've created a new enum named "SkinPackageType", this can be either Skin or Container.
You can use this enum to get or update everything in the interface. For example:

Instead of having a property "RootSkin" or "RootContainer", you call: `GetFolderName(SkinPackageType.Skin)`.
Instead of having 4 methods (e.g. "GetDefaultPortalSkin"), you call: `GetDefaultSkinSrc(SkinPackageType.Skin, SkinType.Edit)`
Instead of changing the skin with the folder name, you call: `SetSkin(SkinPackageType.Skin, ...)`

## Skin in SkinPackageInfo
I also did see that there is a SkinInfo, but it's never used. I changed SkinPackageInfo.Skins to a list of SkinInfo. If we ever add more details in the table `dbo.Skins`, we can add these to this class. Before the Skins was a Dictionary with ID as key and SkinSrc as value. **This means it's a breaking change**, but it's already targeting v10. Is this OK?

Because of the abstractions, I cannot implement `IList<ISkinInfo>` because in the implementation this is a `List<SkinInfo>`. I've created a new wrapper class called `AbstractionList<TInterface, TImplementation>` which casts all the interfaces to the implementation.

In the abstractions I've added `IObjectList<T>`, which is used in SkinPackageInfo.Skins. This interface has an extra method called `T AddNew()` which creates a new instance of the implementation, adds it to the list and returns the instance. Because you don't have `SkinInfo` in the abstractions; you can never create it. Now you can do the following:

```cs
ISkinPackageInfo skinPackageInfo; // Instance of the Skin Package

ISkinInfo newSkin = skinPackageInfo.Skins.AddNew(); // New skin
```

## Naming
I've also changed the naming a little bit, for example `GetSkins` returns an enumerable with the skins in a folder; not from the database. I've changed the name to `GetSkinsInFolder`.

---

This means the abstraction is quite different than the implementation now, but it's way clearer what every method does, just by the name and arguments. If there is any feedback I like to hear it.

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
GerardSmit: You were too fast; I was still typing my comment 😉

But yes, it's a breaking change. I didn't see any usage of Skins in DNN; but for extensions this is a breaking change.
`SkinInfo` exists in the DNN library but is never used. With this change we can future proof the Skins table.

I can make my own Dictionary and update the skins objects based on what is changed, and renamed this property to `SkinObjects`.

DNN Platform/DotNetNuke.Abstractions/Skins/ISkinPackageInfo.cs
GerardSmit: The problem however is that you can't add a new skin to the package, but you can edit an existing skin:

```cs
ISkinPackageInfo skinPackageInfo; // Instance of the package

ISkin newSkin = skinService.CreateSkin();
skinPackageInfo.Skins.Add(newSkin); // Error!

ISkin existingSkin = skinPackageInfo.Skins[0];
existingSkin.SkinSrc = "New.ascx"; // Changed... So are we read-only or not 🤔

skinService.UpdateSkinPackage(skinPackageInfo)
```

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
GerardSmit: Note, if these changes can't be approved because we rather have backward combability than future proofing, I can revert this change without a problem.

DNN Platform/DotNetNuke.Abstractions/Skins/ISkinService.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>Gets the global default skin src.</summary>
+    /// <param name="packageType">The type of the skin package.</param>
+    /// <param name="skinType">The type of the skin.</param>
+    /// <returns>The global default edit skin.</returns>
+    string GetDefaultSkinSrc(SkinPackageType packageType, SkinType skinType);
+
+    /// <summary>Gets a skin package by its id.</summary>
+    /// <param name="packageId">The skin package id.</param>
+    /// <returns>The skin package.</returns>
+    ISkinPackageInfo GetSkinByPackageId(int packageId);
```
valadas, line 28: Maybe this could be `GetSkinPackageById` to match what it returns and avoid confusion?

valadas: Awesome! Thanks!

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
valadas: Mayby we can have SkinList and deprecate Skin

bdukes:
@dnnsoftware/approvers I've rebased on `develop`, should be good to merge.

bdukes (approved):
Thanks so much!

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
bdukes: Unfortunately, we decided that our best option was to continue down the route we've already started, and live with pain of casting to the interface until we can remove the conflicting properties in DNN v11. We're not especially happy with it as a resolution either, but think it's the best long-term option for the platform. We really appreciate your input and contributions.

bdukes:
Just thinking about this PR, should we call the new abstraction `IThemeService`?

valadas:
> Just thinking about this PR, should we call the new abstraction `IThemeService`?

I don't have a strong preference. I would love it but then a lot of other code uses Skin terminology so not sure if it would add or remove confusion :)

bdukes:
I think we can merge this PR as-is. I opened [a new PR](https://github.com/GerardSmit/Dnn.Platform/pull/1) to discuss the potential naming change.

DNN Platform/Library/UI/Skins/SkinPackageInfo.cs
```diff
@@ -79,8 +92,10 @@ public string SkinName
             }
         }
 
-        [XmlIgnore]
+        /// <summary>Gets or sets a dictionary mapping from <see cref="SkinInfo.SkinId"/> to <see cref="SkinInfo.SkinSrc"/>.</summary>
+        [XmlIgnore]
         [JsonIgnore]
+        [Obsolete($"Deprecated in DotNetNuke 9.13.1. Use {nameof(ISkinPackageInfo)}.{nameof(ISkinPackageInfo.Skins)} instead. Scheduled for removal in v11.0.0.")]
```
GerardSmit, line 98: Note, the SkinController only updates the `SkinsList` (`skinPackage` is `ISkinPackageInfo` here):

https://github.com/dnnsoftware/Dnn.Platform/blob/2d859732a381bfcdea259e3a5274b44c25a3566b/DNN%20Platform/Library/UI/Skins/SkinController.cs#L641-L656

Even though the property is deprecated, it's read-only now.

GerardSmit: We currently have two properties:

`Skins` = Dictionary<int, string>
`SkinsList` = List<ISkinInfo>

If you change the SkinsList and call UpdateSkinPackage the skins will be updated.
However, if you change the Skins and call UpdateSkinPackage the skins will not be updated, which happens in the current version.

> Or that we're missing an update to keep the two collection in sync?

Correct.

david-poindexter:
I propose we move this PR to a 9.14.0 milestone.  @dnnsoftware/approvers ?

david-poindexter:
Per Technology meeting today, we agreed on a 9.14.0 milestone for this PR.  Thanks!

14oneplusn205g-code (approved):

mitchelsellers:
At this time this PR has bcome stale with the current changes, if someone would like to update this and make it possible with the current version we can reconsider merging, but closing this for now for clarity.

PR #5841, opened 2023-10-11
Add IPermissionService for Abstractions Project

### LINKED ISSUE #5840: [Enhancement]: Add Dependency Injection Support for PermissionController

### Description of problem

The current implementation of `PermissionController` does not have abstraction layer that can be used with dependency injection.

### Description of solution

Create an interface for this class called `IPermissionService`.

Don't add an interface and keep on using `PermissionController`.

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

GerardSmit:
Fixes #5840 

## Summary
This PR adds `IPermissionService` which is an abstraction layer for `PermissionController`.

Instead of returning an ArrayList in the `GetPermissions*`-methods, I've made the return type `IEnumerable<IPermissionInfo>`; this reduces allocations (ArrayList instance and two arrays). For all these methods I've created an Enumerable method, that's being used by the current implementation and interface implementation; this is so we don't have duplicated code.

I've only created an interface for this class. All calls to the PermissionController haven't been refactored.

## TODO
- [x] Merge `IPermissionInfoBase` and `IPermissionInfo` into a single interface `IPermissionInfo`


valadas (approved):
Nothing sticks out to me. Just no in love with IPermissionInfoBase name which makes it look like a base class. What about we use IPermission and IPermissionInfo instead of IPermissionInfo and IPermissionInfoBase ? Just a thought...

GerardSmit:
The only downside to IPermission is that it's an already existing interface in .NET Framework: [System.Security.IPermission](https://learn.microsoft.com/en-us/dotnet/api/system.security.ipermission?view=netframework-4.8).

We can also merge the two interfaces to a single interface `IPermissionInfo`, since the current implementations of *PermissionInfo all extends PermissionInfoBase and PermissionInfo:

image

valadas:
Sounds good. Hey what tool you use for that graph ?

GerardSmit:
I used ReSharper to create that diagram.

What I did:
1. Left click the interface name and click "Derrived Symbols":
    image
2. Click on the first icon "Show on Diagram":
    image
3. Then I changed the view to "Business Logic":
    image

DNN Platform/Library/Security/Permissions/PermissionController.cs
```diff
@@ -67,27 +68,37 @@ public static string BuildPermissions(IList permissions, string permissionKey)
...
-            return new ArrayList(GetPermissions().Where(p => p.PermissionCode == "SYSTEM_DESKTOPMODULE").ToArray());
+            return new ArrayList(GetPermissionsByPortalDesktopModuleEnumerable().ToArray());
         }
 
+        /// <inheritdoc cref="IPermissionService.GetPermissionsByTab" />
         public static ArrayList GetPermissionsByTab()
         {
-            return new ArrayList(GetPermissions().Where(p => p.PermissionCode == "SYSTEM_TAB").ToArray());
+            return new ArrayList(GetPermissionsByTabEnumerable().ToArray());
         }
```
bdukes, line 87: Let's make sure and deprecate these methods, since they're especially inefficient now (with copying from one collection to another)

DNN Platform/Library/Security/Permissions/PermissionController.cs
```diff
@@ -215,6 +270,39 @@ private static IEnumerable<PermissionInfo> GetPermissions()
...
+        private static IEnumerable<PermissionInfo> GetPermissionsByModuleDefIdEnumerable(int moduleDefId)
+        {
+            return GetPermissions().Where(p => p.ModuleDefID == moduleDefId);
+        }
+
+        private static IEnumerable<PermissionInfo> GetPermissionsByModuleEnumerable(int moduleId, int tabId)
+        {
+            var module = ModuleController.Instance.GetModule(moduleId, tabId, false);
+
+            return GetPermissions().Where(p => p.ModuleDefID == module.ModuleDefID || p.PermissionCode == "SYSTEM_MODULE_DEFINITION");
```
bdukes, line 303: Let's not capture the `ModuleInfo` instance in the closure.

```suggestion
            var module = ModuleController.Instance.GetModule(moduleId, tabId, false);
            var moduleDefId = module.ModuleDefID;

            return GetPermissions().Where(p => p.ModuleDefID == moduleDefId || p.PermissionCode == "SYSTEM_MODULE_DEFINITION");

```

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
```diff
@@ -1,50 +1,60 @@
...
-    public class PermissionInfo : BaseEntityInfo
+    public class PermissionInfo : BaseEntityInfo, IPermissionInfo
     {
-        /// <summary>Gets or sets the Mdoule Definition ID.</summary>
-        /// <returns>An Integer.</returns>
-        [XmlIgnore]
+        /// <inheritdoc cref="IPermissionInfo.ModuleDefId" />
+        [XmlIgnore]
         [JsonIgnore]
         public int ModuleDefID { get; set; }
```
bdukes, line 25: Do we want this to be deprecated?

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
```diff
@@ -1,50 +1,60 @@
...
-        /// <summary>Gets or sets the Permission Name.</summary>
-        /// <returns>A String.</returns>
-        [XmlIgnore]
+        /// <inheritdoc />
+        [XmlIgnore]
         [JsonIgnore]
         public string PermissionName { get; set; }
 
+        /// <inheritdoc />
+        int IPermissionInfo.ModuleDefId
```
bdukes, line 45: Do we need to add `XmlIgnore` and `JsonIgnore`, like we have on the "real" property?

bdukes (commented):
Thanks!

bdukes:
I agree, let's merge `IPermissionInfo` and `IPermissionBase`

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
GerardSmit: Not sure how the JSON/XML-serializer handles implicit implemented interfaces, but better safe than sorry. I'll add them.

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
GerardSmit: See my comment in https://github.com/dnnsoftware/Dnn.Platform/pull/5838#discussion_r1358508810, the same thing applies here as well.

GerardSmit:
Sadly, I couldn't merge `IPermissionInfo` and `IPermissionInfoBase`, since I misunderstood PermissionInfo.

PermissionInfo is from `dbo.Permissions`, which are the definitions.
PermissionInfoBase is the base for `dbo.TabPermission`, `dbo.ModulePermission` etc. they are the actual permission to the object they belong to (e.g. Tab or Module) with their definition.

I've renamed `IPermissionInfo` to `IPermissionDefinition` to prevent confusion, and `IPermissionInfoBase` to `IPermissionInfo`. This only applies to the abstractions, the actual implementation still have the same name (class `PortalInfo` and `PortalInfoBase`).

Is this OK?

bdukes:
Sounds great to me, thanks!

DNN Platform/Library/Security/Permissions/PermissionController.cs
GerardSmit: I've added the deprecations with `[DnnDeprecated]`. Is the way I've added the deprecations OK, or should I change the message?

Example:
```cs
[DnnDeprecated(10, 0, 0, $"Use {nameof(IPermissionService)}.{nameof(IPermissionService.GetDefinitionsByFolder)} instead.")]
```

DNN Platform/Library/Common/Globals.cs
```diff
@@ -3444,6 +3444,13 @@ public static bool IsHostTab(int tabId)
             return PortalSecurity.Instance.InputFilter(strSQL, PortalSecurity.FilterFlag.NoSQL);
         }
 
+        /// <summary>Gets an <see cref="IServiceProvider"/> for the current request, or the global provider if not available.</summary>
+        /// <returns>An <see cref="IServiceProvider"/> instance.</returns>
+        public static IServiceProvider GetCurrentServiceProvider()
```
GerardSmit, line 3449: I made this helper method public, since I needed it in the following:

https://github.com/dnnsoftware/Dnn.Platform/blob/d302d0755d44f04217fbfdf8368ac62936aa4ebc/DNN%20Platform/Modules/ResourceManager/Services/Dto/FolderPermissions.cs#L72-L76

Is this OK?

The only problem with calling `GetCurrentServiceProvider` is that it falls back to the global service provider if the HttpContext is not available. Since IPermissionService is a scoped service, it'll create a scoped instance in the root provider that won't be disposed. This is currently not an issue (since we don't implement IDisposable), but maybe it'll be in the future. 🤷‍♂️

GerardSmit: Alternativly I could add an `[assembly: InternalsVisibleTo("Dnn.Modules.ResourceManager")]`... But this sounds hacky 😅

bdukes: I would update `Permissions.LoadPermissionDefinitions` to take an `IPermissionService` instance.

GerardSmit: I refactored LoadPermissionDefinitions; but for this I to add `IPermissionService` as a parameter in a constructor, which I don't know if I like.

This only affects `Dnn.Modules.ResourceManager`. Can we make breaking changes in here or is this still the core of DNN?

bdukes: We can make breaking changes in Resource Manager

valadas: I agree, we don't publish a Nuget for this one and it is a module that could in some scenarios not even be installed...

GerardSmit:
I've checked `/Admin/File-Management` if the permissions are being loaded and saved. Everything works as expected 🎉

DNN Platform/Library/Security/Permissions/PermissionController.cs
bdukes: Looks good to me. @dnnsoftware/approvers are we targeting this for 10, or is this okay for a 9.x release?

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
bdukes: @dnnsoftware/approvers opinions?

mitchelsellers (approved):

GerardSmit:
I'm still thinking about the interface name. If we're gonna make abstractions for `PermissionProvider`, should this be done in the `IPermissionService` as well?

Another solution is to rename this interface to `IPermissionDefinitionService`, and when we make abstractions for `PermissionProvider` the interface name `IPermissionService`.

bdukes:
> I'm still thinking about the interface name. If we're gonna make abstractions for `PermissionProvider`, should this be done in the `IPermissionService` as well?
> 
> Another solution is to rename this interface to `IPermissionDefinitionService`, and when we make abstractions for `PermissionProvider` we can call the interface name `IPermissionService`.

I think `IPermissionDefinitionService` is a better name, since every member deals with `IPermissionDefinitionInfo`.

GerardSmit:
There are still 2 comments open: https://github.com/dnnsoftware/Dnn.Platform/pull/5841#discussion_r1362596744, https://github.com/dnnsoftware/Dnn.Platform/pull/5841#discussion_r1362597013; other than that everything is done.

bdukes:
Thanks @GerardSmit, this is looking great. The @dnnsoftware/approvers will meet next Tuesday, so we may not get a definitive answer on those open questions until then. We appreciate your contribution and your patience.

DNN Platform/Library/Security/Permissions/PermissionController.cs
valadas: As long as it is non-breaking (I did not see any breaking changes here). I would prefer we do this in v9 so we can remove the deprecated stuff in v11. It is fine to do in 10 too but then we need to get to 12 to do the actual removal.

DNN Platform/Library/Security/Permissions/PermissionInfo.cs
mitchelsellers: The discussion around this is to mark as [Obsolete] as needed

bdukes:
@dnnsoftware/approvers I've rebased on `develop`, updated the deprecations to 9.13.0, added deprecations for the properties which we're changing casing on, and also extracted `IFolderPermissionInfo` as part of adjusting usages of the deprecated properties. This should be good to merge.

bdukes (approved):

valadas (approved):
Awesome, great work!

PR #5846, opened 2023-10-14
Fixed module cache

### LINKED ISSUE #5844: [Bug]: Replace Tokens in HTML module settings gets Unchecked after upgrade

### What happened?

During a test upgrade from DNN 9.10.02 to 9.13.00 any module that had Replace Tokens enabled (check marked) in the HTML Module Settings was changed to disabled (or no longer check marked) after the upgrade. This occurred in every html module across every portal in the DNN installation that had the Replace Tokens enabled under HTML Module Settings. Which caused the token to display on the website since it was no longer parsed.

### Steps to reproduce?

Have a html module using a token before upgrade with Replace Tokens enabled (check marked) in the HTML Module Settings. Upgrade to 9.13.00 and the Replace Tokens box is unchecked.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.0 (latest release)

_No response_

### Code of Conduct

GerardSmit:
Fixes #5844

## Summary
Currently the module cache is being retrieved from the portal cache because we're only using the portal ID as key. Now we're using both the portal and module ID to generate a key.

If there is no module, then we use -1 as module ID; which is the null-value.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
+++ b/DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
@@ -164,7 +164,13 @@ namespace DotNetNuke.Entities.Modules.Settings
                     }
                 }
             });
-            DataCache.SetCache(this.CacheKey(portalId, moduleContext?.TabModuleID ?? -1), settings);
+
+            DataCache.ClearCache(this.CacheKeyPortalPrefix(portalId));
+            DataCache.SetCache(this.CacheKey(portalId, -1), settings);
+            if (moduleContext is not null)
+            {
+                DataCache.SetCache(this.CacheKey(portalId, moduleContext.TabModuleID), settings);
+            }
         }
 
         private T Load(CacheItemArgs args)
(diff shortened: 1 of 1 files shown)


valadas (approved):
Makes sense to me, thanks

DNN Platform/Library/Entities/Modules/Settings/SettingsRepository.cs
```diff
@@ -164,7 +164,13 @@ private void SaveSettings(int portalId, ModuleInfo moduleContext, T settings)
                     }
                 }
             });
-            DataCache.SetCache(this.CacheKey(portalId, moduleContext?.TabModuleID ?? -1), settings);
+
+            DataCache.ClearCache(this.CacheKeyPortalPrefix(portalId));
+            DataCache.SetCache(this.CacheKey(portalId, -1), settings);
```
GerardSmit, line 169: Maybe I'm missing something, but shouldn't this only be happening if the `moduleContext` is null?

If the module context is null, then we're talking about global portal settings, otherwise module-specific settings. Now the module-specific settings can be set in the portal settings.

bdukes: A settings class can contain both module settings and portal settings, so when a settings class is updated, it needs to invalidate all of the other settings classes for that portal.

This is caching module-specific settings with the non-module-specific cache key, under the assumption that any code which is only passing in a portal ID will know to not use the values of the module-specific properties. Perhaps this is not a safe-enough (or efficient-enough) optimization and should be removed.

bdukes: I've pushed a commit to remove that optimization

GerardSmit:
Looks good 👍, thanks for the addition.

mitchelsellers (approved):

PR #5851, opened 2023-10-27
Implemented core support for portal styles

valadas:
Implementation of portal styles for the core level support.
This is complete but the overall support, in next PRs I'll submit a module to manage those settings and we can then consume those css values in dnn-elements, resource-manger, new theme, etc.


DNN Platform/DotNetNuke.Abstractions/Portals/IPortalStyles.cs
```diff
@@ -0,0 +1,227 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the main shade of the primary color.
+        /// </summary>
+        IStyleColor ColorPrimary { get; set; }
+
+        /// <summary>
+        /// Gets or sets the light shade of the primary color.
+        /// </summary>
+        public IStyleColor ColorPrimaryLight { get; set; }
```
bdukes, line 20: These shouldn't have `public` since it's an interface, right?

DNN Platform/DotNetNuke.Abstractions/Portals/IStyleColor.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Abstractions.Portals
+{
+    /// <summary>
+    /// Represents a CSS color and its components.
+    /// </summary>
+    public interface IStyleColor
```
bdukes, line 10: Is it worth thinking about (1) alpha, (2) HSL, (3) LCH?

DNN Platform/Library/Entities/Portals/StyleColor.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    using DotNetNuke.Abstractions.Portals;
+
+    /// <summary>
+    /// Represents a CSS color and its components.
+    /// </summary>
+    public class StyleColor : IStyleColor
```
bdukes, line 15: Maybe this should be a `struct`, since it represents a single value?

DNN Platform/Website/Default.aspx.cs
```diff
@@ -645,6 +653,110 @@ private void InitializePage()
             }
         }
 
+        private string GenerateCssCustomProperties()
+        {
+            string cacheKey = $"Dnn_Css_Custom_Properties_{this.PortalSettings.PortalId}";
```
bdukes, line 658: Does this cache key get set? Should this be using `CBO.GetCachedObject`?

DNN Platform/Website/Default.aspx.cs
```diff
@@ -645,6 +653,110 @@ private void InitializePage()
...
+            {
+                return cache;
+            }
+
+            var styles = this.PortalSettings.Styles;
+            var sb = new StringBuilder();
+            sb
+                .AppendLine(@"<style type=""text/css"">")
+                .AppendLine(@":root {")
+                .AppendLine($"--dnn-color-primary: #{styles.ColorPrimary.MinifiedHex};")
```
bdukes, line 671: If we're using a `StringBuilder`, we probably want to use it to add the values, rather than string interpolation

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalStyles.cs
valadas: Good catch, fixed in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/396c252c0f7b05533356b27f454094b7abe30d01

DNN Platform/DotNetNuke.Abstractions/Portals/IStyleColor.cs
valadas: I don't believe so, dnn-elements expects opaque colors and internally may use transparancy on top of those for like disabled stuff, etc. we provide rgb so that consumers can do math on top of a base color. For instance there was css functions to offset a hue or javascript functions to find if either white or black has a better contrast over a certain rgb value. Also we do have color utilities class in dnn-elements that people can use to get those values out of any color: https://github.com/DNNCommunity/dnn-elements/blob/develop/packages/stencil-library/src/utilities/colorInfo.ts so I think we just need to provide the color and let developers use frontend tools if they need converting.

DNN Platform/Library/Entities/Portals/StyleColor.cs
valadas: Good good, done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/22256682bde0280ee10268af525e3d820d34990c

DNN Platform/Website/Default.aspx.cs
valadas: You mean like this?

```
  .AppendFormat("--dnn-color-secondary-dark: #{0};\n", styles.ColorSecondaryDark.MinifiedHex)
```

I kinda find it less readable. Is this a performance concern or did I misunderstand the idea ?

DNN Platform/Website/Default.aspx.cs
valadas: I kinda ran out of time last week on this, it will get used soon... I actually probably have to move that in some other place as the module to manage it will need to clear that cache.

DNN Platform/Website/Default.aspx.cs
bdukes: Yes, I'm thinking about performance. You should be able to do it all as a single multi-line interpolated string, and get rid of the `StringBuilder`.

DNN Platform/Website/Default.aspx.cs
valadas: Ok, we will need that key in the module too and I don't want to module and DNN to both have it (don't repeat yourself).... I think this is a nice place to reuse it ?

DNN Platform/Website/Default.aspx.cs
valadas: Yeah, good idea... I`ll revisit this

DNN Platform/Library/Entities/Portals/StyleColor.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        private bool IsValidCssColor(string hexValue)
+        {
+            if (string.IsNullOrWhiteSpace(hexValue))
+            {
+                throw new ArgumentNullException("You need to provide a CSS color value in the constructor");
+            }
+
+            var regex = new Regex(@"([\da-f]{3}){1,2}", RegexOptions.IgnoreCase);
```
GerardSmit, line 127: Should this Regex instance be made a static field (maybe with RegexOptions.Compiled) instead of recreating it every time that IsValidCssColor is called?

DNN Platform/Library/Entities/Portals/StyleColor.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    using DotNetNuke.Abstractions.Portals;
+
+    /// <summary>
+    /// Represents a CSS color and its components.
+    /// </summary>
+    public struct StyleColor : IStyleColor
```
GerardSmit, line 15: I see that `StyleColor` is a struct: in PortalStyles all the properties use `IStyleColor`. When a struct is being converted to an interface, it's being boxed:  it's still allocating to the heap.

DNN Platform/Library/Entities/Portals/StyleColor.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            return true;
+        }
+
+        private byte GetComponent(Component comp)
+        {
+            switch (comp)
+            {
+                case Component.Red:
+                    return byte.Parse(this.hex.Substring(0, 2), System.Globalization.NumberStyles.HexNumber);
```
GerardSmit, line 141: Every time `Red`, `Green` or `Blue` is being accessed, a substring is being created and being parsed.
Since it's always being used, maybe it's possible to cache this value in a field?

GerardSmit (commented):
Just some small things I saw while looking at this PR. These aren't big issues since the PortalStyles and the CSS are being cached: it's only when the cache doesn't exist these allocations happen.

DNN Platform/Website/Default.aspx.cs
valadas: @bdukes any thoughts ?

DNN Platform/Website/Default.aspx.cs
valadas: Done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/1b8b566f9f2c7aeb4d5d43fa657a8ed0919a1df6

DNN Platform/Library/Entities/Portals/StyleColor.cs
valadas: Good idea, done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/0a675fae3646d49e8100f225414fa35e227c401e

DNN Platform/Library/Entities/Portals/StyleColor.cs
valadas: Should we just move StyleColor to the abstractions project and kill IStyleColor?
Or do we want to only keep interfaces in that project ?

@bdukes any thoughts on this ?

DNN Platform/Library/Entities/Portals/StyleColor.cs
valadas: Good idea, done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/c352f05565e50ecbb1320082aeaf0dbce882a2d2

DNN Platform/Website/Default.aspx.cs
bdukes: I would just add it to the collection of keys in `DataCache`
https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Common/Utilities/DataCache.cs#L43-L74

DNN Platform/Library/Entities/Portals/StyleColor.cs
bdukes: I'm okay with it being in the abstractions project

DNN Platform/Website/Default.aspx.cs
valadas: Oh, good point, done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/4f3966f44bfb3c5140a290415d5430cfdb449c84

DNN Platform/Library/Entities/Portals/StyleColor.cs
valadas: Cool cool, done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/f74f0e71e2df8969b614e9af4f246851ebe25877

david-poindexter (approved):
This is looking good @valadas 🎉

valadas:
Ok, I think this is good to merge now, I started work on the actual management module which I would like to submit a different PR for to make it easier to review than a huge PR and also makes this available in DNN for the theme development.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
```diff
@@ -370,5 +370,10 @@ public interface IPortalSettings
 
         /// <summary>Gets a value indicating whether to display the dropdowns to quickly add a moduel to the page in the edit bar.</summary>
         bool ShowQuickModuleAddMenu { get; }
+
+        /// <summary>
+        /// Gets the dnn styles for this portal, these get injected on the page top as css custom properties.
+        /// </summary>
+        IPortalStyles Styles { get; }
```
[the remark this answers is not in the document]
valadas: IMO we should try to get it as right as we can and attempt to prevent changes, but when it is inevitable, I tend to like a version. With modern tooling, the name gets autocompleted anyway and I have seen that practice in some places including DirectX `IDirect3DDevice9` for v9; Windows COM 64 bit APIs, Android `Camera2` APIs when they revamped the whole thing.

I think this approach would allow us to keep the old interface (deprecated) for 2 major versions.

Now we should probably not do this very often and still try to think the initial interface through, we can also use extension methods if we are just adding some utility to an existing interface, etc.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalStyles.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        /// <summary>
+        /// Gets or sets the controls padding in pixels.
+        /// </summary>
+        public int ControlsPadding { get; set; }
+
+        /// <summary>
+        /// Gets or sets the base font size in pixels.
+        /// </summary>
+        int BaseFontSize { get; set; }
```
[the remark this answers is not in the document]
valadas: I don't believe I ever saw a non-integer value for those, but good point, it is technically a supported use case. Javascript Number type is a 64 bit double, so I guess I can change these for doubles...

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
```diff
@@ -0,0 +1,171 @@
...
+        public string MinifiedHex
+        {
+            get
+            {
+                if (
+                    this.hex[0] == this.hex[1] &&
+                    this.hex[2] == this.hex[3] &&
+                    this.hex[4] == this.hex[5])
+                {
+                    return this.hex[0].ToString() + this.hex[2].ToString() + this.hex[4].ToString();
```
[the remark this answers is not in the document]
valadas: Hmm, not sure what I was thinking there, your suggestion makes sense, I'll revisit end of week

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            string value;
+            var r = hexValue[0];
+            var g = hexValue[1];
+            var b = hexValue[2];
+            value = string.Concat(r, r, g, g, b, b);
+            return value.ToUpperInvariant();
+        }
+
+        private bool IsValidCssColor(string hexValue)
```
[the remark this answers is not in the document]
valadas: Assert?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalStyles.cs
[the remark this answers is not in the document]
valadas: Thanks, done in :

https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/f2f128f99a3d64c928ac443d238a835110fb6c3d

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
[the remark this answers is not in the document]
valadas: Done in Commit a2538cbf

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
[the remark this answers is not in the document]
valadas: Done in fb3682db

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                throw new ArgumentNullException("You need to provide a CSS color value in the constructor");
+            }
+
+            if (!HexColorRegex.IsMatch(hexValue))
+            {
+                throw new ArgumentOutOfRangeException($"The value {hexValue} that was provided is not valid, it needs to be 3 or 6 character long hexadecimal string without the # sing");
+            }
+
+            return true;
```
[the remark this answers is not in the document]
valadas: Done in fb3682db

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                return this.hex;
+            }
+
+            set
+            {
+                if (this.IsValidCssColor(value))
+                {
+                    this.hex = ExpandColor(value);
+                }
```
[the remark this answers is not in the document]
valadas: Done in fb3682db

DNN Platform/DotNetNuke.Abstractions/Portals/StyleColor.cs
```diff
@@ -0,0 +1,171 @@
...
+        public StyleColor(string hexValue)
+        {
+            if (this.IsValidCssColor(hexValue))
+            {
+                this.HexValue = ExpandColor(hexValue);
+            }
+            else
+            {
+                this.HexValue = ExpandColor("FFFFFF");
+            }
```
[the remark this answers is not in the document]
valadas: Done in fb3682db

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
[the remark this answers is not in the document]
bdukes: I think this is the only remaining issue I have with this PR. Do we call this type `IPortalSettings2` or `IPortalSettings10` (and deprecate `IPortalSettings` for DNN 12)?

Thoughts @dnnsoftware/approvers?

DNN Platform/Website/Default.aspx.cs
```diff
@@ -645,6 +653,117 @@ private void InitializePage()
...
+            ";
+            Common.Utilities.DataCache.SetCache(
+                cacheKey,
+                cssVars,
+                null,
+                System.Web.Caching.Cache.NoAbsoluteExpiration,
+                TimeSpan.FromMinutes(Common.Utilities.DataCache.PortalStylesCacheTimeOut),
+                Common.Utilities.DataCache.PortalStylesCachePriority,
+                null);
+            return cssVars;
```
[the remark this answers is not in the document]
valadas: This is awesome, learned something new there!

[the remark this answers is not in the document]
valadas: Done in https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/301d87966ab6b0316d4bb600a6c21c1ee05d0e14

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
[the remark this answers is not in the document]
valadas: I kinda like the idea of using the major version, so IPortalSettings10 IMO. I guess it would make it easier to know in which DNN version the new interface started existing.

Now should we duplicate the existing interface or inherit from it ?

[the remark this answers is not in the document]
bdukes: I want to say inherit, but does that mean that it would inherit the `Obsolete` attribute?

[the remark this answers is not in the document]
valadas: Yeah, I am not sure, let me try that out...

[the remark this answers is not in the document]
valadas: With this:
image

It appears to be fine:
image

I'll go with that approach...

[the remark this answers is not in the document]
valadas: That has a lot of ramifications though :)
image

Which I can handle, but just want to make sure that creating new overloads and deprecating the old methods is the proper way...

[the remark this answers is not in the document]
valadas: Another approach could be to not make this part of IPortalSettings at all... Open to suggestions.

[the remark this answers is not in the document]
valadas: One idea I can think would be to create an extension method so consumers woudl do:

portalSettings.GetStyles()

does that sound like a good idea to you ?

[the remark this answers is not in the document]
valadas: Could also create IPortalSettings10 but not deprecate IPortalSettings, those who want the new features just use the new interface...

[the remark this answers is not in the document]
valadas: @bdukes Ok, I removed the change on IPortalSettings and created an extension method. Please see https://github.com/dnnsoftware/Dnn.Platform/pull/5851/commits/51dce1e05684ea210c6eb7b553d319875967d535 and let me know what you thing, we can revert that commit if we come up with a better idea. But this way we don't change the interface at all.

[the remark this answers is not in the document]
david-poindexter: I like this approach @valadas (extension method).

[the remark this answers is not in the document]
bdukes: Looks good to me, thanks!

bdukes (approved):

valadas:
Awesome, I'll merge this to make https://github.com/dnnsoftware/Dnn.Platform/pull/5884 easier to review :)

Issue #5858, opened 2023-11-05
[Bug]: Unable to build: Error: Yarn Runner: Could not locate executable.


KirillOsenkov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I cloned Dnn.Platform and ran build.ps1. It failed:

![image](https://github.com/dnnsoftware/Dnn.Platform/assets/679326/6422de6c-9c9f-47e6-990a-ca63d22861f4)

### Steps to reproduce?

1. git clone https://github.com/dnnsoftware/Dnn.Platform on Windows
2. build.ps1

### Current Behavior

Build fails

### Expected Behavior

Build succeeds

### Relevant log output

```shell
========================================
BuildNpmPackages
========================================
An error occurred when executing task 'BuildNpmPackages'.

----------------------------------------
TearDown
----------------------------------------
git $reset --hard
HEAD is now at f62e90699f 'backup'
git $reset HEAD^
Unstaged changes after reset:
D       global.json
Error: Yarn Runner: Could not locate executable.
```

### Affected Versions

10.0.0 (alpha)

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

valadas:
Hmm, I am pretty sure the build should get yarn for you, but just to rule it out, do you have yarn installed globally? You can check by typing `yarn -v` in a terminal...

KirillOsenkov:
Nope, no yarn installed. If it's a prerequisite, it should be at least documented, or issue an error if it's not installed. But of course ideally the build should install a local copy if possible and use that.

I wasn't able to figure out how to install yarn or what it is.

valadas:
We have our build documentation here: https://github.com/dnnsoftware/Dnn.Platform/blob/develop/.github/BUILD.md

It may need expanding to define prerequisites that we may have wrongly assumed everyone has...
- I am guessing you have a recent version of Visual Studio installed, if not, you can install the latest community (free) edition.
- In addition to that you will need to have node.js installed, see https://nodejs.org/en and probably get the LTS version installed. This will give you `node` and `npm` and some other command line tools.
- After that, you should be able to run `npm install --global yarn` which will get you yarn installed globally as a tool

After that try ./build.ps1 again and let us know... If that fixes it, then you could send a PR to update those docs :)

KirillOsenkov:
That page does mention Yarn, but just links to the front page:
[...]

PR #5859, opened 2023-11-06
Use corepack to run Yarn

bdukes:
## Summary
As demonstrated in #5858, the Cake build does not work when Yarn has not been installed. [The recommended way to install Yarn](https://yarnpkg.com/getting-started/install) is via [Corepack](https://nodejs.org/dist/latest/docs/api/corepack.html), which is included with Node.js since version 16.9.0.

Corepack will read [the `packageManager` field from `package.json`](https://github.com/dnnsoftware/Dnn.Platform/blob/dd0e72d006711064fb38133f0b5f541291d74c5b/package.json#L44) to determine the correct package manager and its version, and get a copy dynamically, instead of using the JS file checked into the repository.

This PR updates the build task which runs Yarn commands to do so via Corepack, to avoid an extra install step. It also adds a check for Corepack and fails the build if it is not available.


mitchelsellers (approved):
This is awesome!

valadas (approved):
AwesomeGrandmaGIF

KirillOsenkov:
Hey @bdukes, this is amazing, thanks so much!

PR #5880, opened 2023-12-11
Merges latest change from `develop` into `release/10.0.0`

valadas:
Merges latest change from `develop` into `release/10.0.0`


valadas:
Oh boy, I'll revisit this later...

PR #5884, opened 2023-12-15
Implementation the new module to manage styles

valadas:
This is still a work in progress and when https://github.com/dnnsoftware/Dnn.Platform/pull/5851 will be merged, this PR will become way easier to review


valadas:
Got some progress this weekend, still needs some polishing and some logic but it is a start:

image

valadas:
Polished a bit the design / layout...

image

valadas:
I had time today to implement 2 missing properties, and implement saving/reset/restore as well as admin permissions. This means a host can decide on a portal by portal level to allow or not their admins to change the portal styles. It defaults to false, so one has to opt-in to allow that.

image

valadas:
@dnnsoftware/approvers I also took this PR out of draft, I think from the module perspective, this is pretty stable now and the rest can be done in other PRs.

david-poindexter (approved):
Looks great!

DNN Platform/Website/Default.aspx.cs
```diff
@@ -672,89 +687,98 @@ private string GenerateCssCustomProperties()
                     return $$"""
                              <style type="text/css">
```
mitchelsellers, line 688: Should we consdier adding a nonce or hash to this inclusion to help with CSP management for those looking to set security policies?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Styles/App_LocalResources/Styles.resx
```diff
@@ -0,0 +1,273 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="nav_Styles.Text" xml:space="preserve">
+    <value>Styles</value>
+  </data>
+  <data name="ModuleDescription" xml:space="preserve">
+    <value>DNN Provides css-varialbes (custom properties) to designers and developers to allow them to style their solutions to the site branding preferences. These style preferences may or may not be implemented by a theme or extension.</value>
```
mitchelsellers, line 124:
```suggestion
    <value>DNN provides css-variables (custom properties) to designers and developers to allow them to style their solutions to the site branding preferences. These style preferences may or may not be supported by a theme or extension.</value>
```

There is a typo in variables, but I'm wondering if there is a better way to emphasize that these colors will not impact anything if they are not adjusted.

Is there any impact to DNN regardless of theme usage?  If so, we should call that out here as well

mitchelsellers (changes requested):
Just one small note with a typo and a few questions.

My last question is there any reason we should consider, possibly allowing this to be not injected into the page if not needed/used by a theme?

DNN Platform/Website/Default.aspx.cs
valadas: Sure I'll look into that next time I can

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Styles/App_LocalResources/Styles.resx
valadas: Well yes, the goal is to harmonize the UI and provide a single source of truth for such preferences. So it will affect all of DNN UI components, persona bar, etc. the new default CSS will use them. It would make no sense to have an option to not inject them. It will not affect 3rd party solutions that don't use them or the older UI/UX patterns. The defaults should be close the the existing Dnn branding so unless they are changed, it won't look too far from what it looks now. The exception to that is we currently have different styles in the persona bar than it the rest of the editing UI. With the new default.css, the "others" will be more aligned with the current person's bar styling (flatter).

valadas: But if a theme does unload default.css and does its own thing, that wins... So that would be the "opt-out" exit route as in the past.

valadas:
Alternativelly we could spit it out to a css file, opinions? We could have variables.css next to default.css or such (for CSP support). as a bonus it would allow theme developers to also refuse it (although that makes less sense).

mitchelsellers:
I personally think I would prefer the file route. Anything to avoid processing and online items that would make CSP harder

valadas:
Definition of the css-variables is now removed from the theme, so changing the values using the module now affect the rendered look.
Ex:
image

valadas:
Converting this back to draft to work on making this a file instead of injection in default.asxc.cs...

valadas:
And my last commit removes the resource manager hardcoding of styles.

image

image

david-poindexter:
Great work @valadas - it is so nice to see the plan starting to come together!

valadas:
> I personally think I would prefer the file route. Anything to avoid processing and online items that would make CSP harder

Ok, what do you think of this https://github.com/dnnsoftware/Dnn.Platform/pull/5884/commits/514ba5703b0e742239de00d48b6fc208f3d3e6a4

- We still have a cache to know if it has changed since the file was last saved.
- The file gets generated either if it does not exists or if the cache has changed.
- It saves it in portals system folder so that admins don't go play with it in the resource manager.

mitchelsellers (commented):
I think this is a much better approach. 

My only pending reservation here is that we will be going back to disk to check for existence of a file on a somewhat regular basis for a random HtTP request when it should be assumed to be there. 

This is a different behavior than that of what we expect for other similar files (portals.css). 

The overhead should be minimal but it does require disk lookups etc that we are not using in other places and this is an every web request behavior. 

I'm leaving this as a comment, as I'm not going to say yes/no but wanted to make that observation. I am out of the country until 1/2 so do not want to delay this any and will respect the direction that others confirm. 

I'm just always very cautious with anything that adds to default.aspx especially in high performance situations

valadas:
@mitchelsellers unless I did something wrong and did not notice, it should only check for file existence if it is not yet in cache. So worst case scenario each 20 minutes. But yes it is tied to a request because then we can handle only the portal.

Alternative solutions I was thinking:
1: We could do this on the application startup but then we need to loop through all the portals and then we may need additional checks for being just post-install or such odd scenarios where the portal might not yet be ready to use. (happy to try/investigate this options if it would be preferred).
2: We do generate the file only upon portal creation and then not check if it exists afterwards, but then we need a potentially complex upgrade scenario for existing portals and well, it just feels brittle to me.
3: I could swear we had some sort of sliding-expiration mechanism for DataCache and I could not see it for some reason. I can look harder on that as then we could only check if the file exists if the site receives no visits at all for 20 minutes.

I need to leave but let me know if you have any other ideas and I'll be happy to try.

david-poindexter (approved):
The changes to `Aperture` look great @valadas - thanks for working through all the challenges there with me - much appreciated.

As for the handling of the stylesheet loading, this approach seems good to me as is, but we could of course revisit in a future release if a better way to handle it is determined.  My two cents is to accept it as is for 10.0.0.

mitchelsellers (approved):
I'm with @david-poindexter let's approve as is. 

I'll do some performance testing later on.

david-poindexter:
Awesome @mitchelsellers - thanks!

PR #5890, opened 2023-12-18
Stop execution of script if there is an error

### LINKED ISSUE #5889: [Enhancement]: Stop SQL script execution on error

### Description of problem

Currently SQL scripts with batching (scripts separated by 'GO' statements) get run to completion even if there is an error in one of the batches. The error is reported, but often later pieces of scripts rely on previous ones being executed first. Therefore continuing after an error can leave the database in a bad state.

### Description of solution

I propose adding break statement in the catch block of the loop. This will prevent the rest of the script from running if there an error.

Considered leaving the implementation as is, and engineering my scripts to either not use batching or to be safe if a batch in the middle fails. This has proved to be very cumbersome and not always possible.

### Anything else?

This method in question: https://github.com/dnnsoftware/Dnn.Platform/blob/c3f4725b6445d9ed6ea495b9cded76606a3d2418/DNN%20Platform/Library/Data/SqlDataProvider.cs#L213

- [X] Yes

- [ ] Yes

### Code of Conduct

DanielBolef:
Fixes #5889 

Stop execution of script if there is an error in one of the batches.

### DIFF
--- a/DNN Platform/Library/Data/SqlDataProvider.cs
+++ b/DNN Platform/Library/Data/SqlDataProvider.cs
@@ -234,6 +234,7 @@ namespace DotNetNuke.Data
                     {
                         Logger.Error(objException);
                         exceptions += objException + Environment.NewLine + Environment.NewLine + sql + Environment.NewLine + Environment.NewLine;
+                        break;
                     }
                 }
             }


valadas:
I have mixed feelings about this, can this be some sort of opt-in behaviour? I could see this being a big breaking change specially for older extensions where blocks could have been used to try to do something but failing was ok to proceed. I know it is not "right" but seen that here and there.

PR #5892, opened 2023-12-22
Fixed an issue that kept rm items selected after changing folders

### LINKED ISSUE #5758: [Bug]: Accidental deletion of folder in ResourceManager when drilling down in the items view

### What happened?

When on the new resource manager to select a folder in the items view the user can doubleclick to open the folder.
If this doubleclick happens after a single click on the folder, the folder will not only be opened, but also selected.
This because every click executes a select / unselect of the clicked item.

Users drilling down to a file might therefore accidentally also select a folder.
If the user wants to delete the file also the folder is deleted.

### Steps to reproduce?

1. Go to {website}Admin/File-Management
2. SingleClick on a folder in the items view
3. Doubleclick on same folder to drill down
4. Repeat until a file is available
5. Select the file and click on delete

### Current Behavior

Drilling down in the items list leads to accidental selection of folders. (not good)
In this case the delete confirmation popup shows the folder and the file to be deleted (correct), confirming the deletion deletes the file and the selected folder. (also correct)

### Expected Behavior

Double clicking a folder should not select/deselect the folder. The behaviour should be like it was in Version 9.11.00: only drilldown happens..

### Relevant log output

_No response_

### Anything else?

This is not a real "bug" as the files as well as the selected folders are clearly visible on the delete confirmation dialog.
But, as the users tend not to read the beaviour leads to the deletion of whole folders.
As those files are hard deleted, they have to be restored from backups.

This behaviour was different on Version 9.11.00 where the double click does not work on the items pane.

### Affected Versions

9.12.0 (latest release)

Chrome, Microsoft Edge

### Code of Conduct

valadas:
This could cause confusion and users to accidently perform batch operations (like deletions) by accident on items they were not expecting to be selected.

Closes #5758

### DIFF
--- a/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/components/dnn-rm-left-pane/dnn-rm-left-pane.tsx
+++ b/DNN Platform/Modules/ResourceManager/ResourceManager.Web/src/components/dnn-rm-left-pane/dnn-rm-left-pane.tsx
@@ -17,6 +17,7 @@ export class DnnRmLeftPane {
   }
 
   private handleFolderClicked(e: CustomEvent<FolderTreeItem>): void {
+    state.selectedItems = [];
     this.itemsClient.getFolderContent(
       Number.parseInt(e.detail.data.key),
       0,
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,6 +46,19 @@ pr:
 
 steps:
 
+- task: PowerShell@2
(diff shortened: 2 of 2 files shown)


mitchelsellers (approved):

david-poindexter (approved):
Awesomeness! 🎉

valadas:
I fixed the build issue in Azure but that commit applied only on the master branch, so cherrypicked it into this PR @mitchelsellers @david-poindexter it should fix the build issue.

PR #5895, opened 2023-12-25
DNN 10 Theme: Update styling of login 

david-poindexter:
## Summary
This PR resolves a few mixin issues as well as styles the login.

image

image


mitchelsellers (approved):

valadas (approved):
Looks good to me, we may want to test the login screen with other providers enabled before release. I know there are a couple of provided ones with DNN that don`t work but they do display though...

david-poindexter:
@valadas as a follow-up, I tested the four other auth providers' layout/styling and everything looks good to me.

image

image

PR #5917, opened 2024-01-13
Aperture theme - refine various styles and utility classes

david-poindexter:
## Summary
This PR:
* Updates margins on base styles.
* Adds a utility class for `aperture-divider`.
* Resolves an issue with the `flex` utility class naming convention to bring it more in line with what people coming from Bootstrap would expect.
* Add support to `margin` utility classes for `auto`.
* Update release notes to remove version to avoid an ongoing maintenance nightmare.


valadas (approved):
Looks good

bdukes (approved):

PR #5918, opened 2024-01-13
Aperture theme - add smooth scrolling on click for hashed nav items

david-poindexter:
## Summary
This PR updates the new DNN 10 Aperture theme to have smooth scrolling for all hashed nav menu items.  The content that will be in the new `Default` template will have divs with ids for targeting those sections on the home page.

ApertureThemeSmoothScrolling

PR #5921, opened 2024-01-15
Aperture theme - mobile menu refinement

david-poindexter:
## Summary
**NOTE: Please review/approve/merge #5918 before this one, as I had to stack commits to test these changes against the previously submitted PR.**

This PR adds multiple UI/UX refinements/fixes to the mobile menu.

ApertureThemeMobileMenu

### DIFF
--- a/DNN Platform/Skins/Aperture/src/scripts/menu.ts
+++ b/DNN Platform/Skins/Aperture/src/scripts/menu.ts
@@ -12,6 +12,8 @@ export const navFunctions = (() => {
 				const target: HTMLAnchorElement | null = document.querySelector(anchorWithHash.hash);
 				// check if target exists
 				if (target !== null) {
+					// toggle mobile menu and overlay
+					hamburgerMenu.checked = !hamburgerMenu.checked;
 					// get target position
 					const targetTop = target.getBoundingClientRect().top;
 					// get window position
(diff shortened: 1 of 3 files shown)


valadas (approved):
Awesome!

DNN Platform/Skins/Aperture/src/scripts/menu.ts
```diff
@@ -1,31 +1,64 @@
...
+		});
+	});
+
 	// add event listeners to expands
-	const navExpand = document.querySelectorAll('.nav-expand');
-  
-	navExpand.forEach(item => {
-	  item.querySelector('.nav-expand-link')?.addEventListener('click', () => item.classList.add('active'));
-	  item.querySelector('.nav-back-link')?.addEventListener('click', () => item.classList.remove('active'));
+	const navExpand = [].slice.call(document.querySelectorAll('.nav-expand'));
```
bdukes, line 32: Would `Array.from` be clearer here?
```suggestion
	const navExpand = Array.from(document.querySelectorAll('.nav-expand'));
```

david-poindexter: @bdukes if you think it is better to do it this way, I am fine with it.  However, there is a small fix that will be required due to the default type of `Array.from` of `Element` instead of `HTMLElement`.  I just pushed a commit to cast it to an `HTMLElement`.

bdukes (approved):

PR #5930, opened 2024-01-20
Fix #5894: remove the child node with null reference check.

### LINKED ISSUE #5894: cannot export modules

### What happened?

 cannot export html module to xml archive

### Steps to reproduce?

soBAD
1. in page setting go edit mode
2. in right corner of thml module press settings
3. click in export content
4. in new windows select folder where you will save the file
5. press in export bottom

### Current Behavior

An error occurred during the export.

### Expected Behavior

the file is saved in folder specify

### Relevant log output

```shell
An error occurred during the export.
```

### Anything else?

no

### Affected Versions

9.13.0 (latest release)

Chrome

### Code of Conduct

zyhfish:


valadas (commented):
I would like to discuss this in the next TAG meeting. This would fix the issue at hand but I am not sure this is what we should do or find another way for that obsolete serialization higher up.

david-poindexter:
Agreed.  Thanks so much for submitting this PR @zyhfish 🎉 Our next meeting is on Jan 23.

zyhfish:
Hi @valadas , another possible solution is handle the XmlSerializer.UnknownElement or UnknownNode event handler to add the obsoleted properties manually, but IMO this is not needed as we need to remove them finally in the result.

valadas:
HI @zyhfish my concern is not just about this specific bug but overall serialization by others on these objects. Being public and providing a base class any 3rd party could potentially have their own permissions inherited by one of these classes to do something custom and #5841 caused a breaking change in that respect. 

I took a stab at implementing IXmlSerializable and quickly discovered that it was a bad idea because then any subclass would call the highest inherited class implementation unless we override those methods in the whole chain of inheritance. So it would not serialize any lower details.

I ran out of time to dig further, so I'll have to revisit this but curious about what people think on #5943 which has a different approach to simply having both casings present and making the wrong case property [CLSCompliant(false)]

valadas:
That being said I did not yet try UnkownElement or UnkownNode, not sure on that specific one but could be something as long as we don't have the same issue as in IXmlSerializable implementation

bdukes (approved):

mitchelsellers (approved):

PR #5933, opened 2024-01-22
Improve API tokens

donker:
During a Demo of the new DNN 10 API Tokens feature, a few things struck me that needed improvement:

- Fallback for when a resourcefile cannot be found so that API keys will show up with some text
- Do not show or allow users to create things when it's switched off

This PR addresses these issues


valadas (approved):
Nothing is sticking out to me.

PR #5943, opened 2024-01-29
Fixed a serialization issue with PermissionInfo

valadas:
This is a work in progress

This way appears to work for serializing as before the changes in #5841

But now I am bumping into an installation issue where Activator.CreateInstance inside CBO.CreateObject inside DTO.FillCollection (don't quote me on the exacts here) but it returns a null instead of an instance which causes this error during installation:

```
1/29/2024 07:12:44 [ERROR] DotNetNuke.Services.Upgrade.Upgrade Error: Value cannot be null.
Parameter name: source   at System.Linq.Enumerable.Where[TSource](IEnumerable`1 source, Func`2 predicate)
   at DotNetNuke.Security.Permissions.PermissionController.GetPermissionByCodeAndKeyEnumerable(String permissionCode, String permissionKey) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Security\Permissions\PermissionController.cs:line 306
   at DotNetNuke.Security.Permissions.PermissionController.GetPermissionByCodeAndKey(String permissionCode, String permissionKey) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Security\Permissions\PermissionController.cs:line 137
   at DotNetNuke.Entities.Modules.ModuleController.InitialModulePermission(ModuleInfo module, Int32 tabId, Int32 permissionType) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Entities\Modules\ModuleController.cs:line 1138
   at DotNetNuke.Services.Upgrade.Upgrade.AddModuleToPage(TabInfo page, Int32 moduleDefId, String moduleTitle, String moduleIconFile, Boolean inheritPermissions, Boolean displayTitle, String paneName) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Services\Upgrade\Upgrade.cs:line 290
   at DotNetNuke.Services.Upgrade.Upgrade.AddModuleToPage(TabInfo page, Int32 moduleDefId, String moduleTitle, String moduleIconFile, Boolean inheritPermissions) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Services\Upgrade\Upgrade.cs:line 242
   at DotNetNuke.Services.Upgrade.Upgrade.AddModuleToPage(TabInfo page, Int32 moduleDefId, String moduleTitle, String moduleIconFile) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Services\Upgrade\Upgrade.cs:line 2248
   at DotNetNuke.Services.Upgrade.Upgrade.UpgradeToVersion920() in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Services\Upgrade\Upgrade.cs:line 2750
   at DotNetNuke.Services.Upgrade.Upgrade.UpgradeApplication(String providerPath, Version version, Boolean writeFeedback) in C:\dnnvaladas\Dnn.Platform\DNN Platform\Library\Services\Upgrade\Upgrade.cs:line 1394
```

I am creating this WIP PR to probe your minds about some sort of pattern for the furue ID renamings for Id (which will happen in may other places" I'd like to see what people think about simply having both casisings during the deprecation phase and bypassing the CLSCompliance on only the wrong one.

I also needed some code to be here to refer to it in comments of other issues :)

### DIFF
--- a/DNN Platform/Library/Security/Permissions/PermissionInfo.cs
+++ b/DNN Platform/Library/Security/Permissions/PermissionInfo.cs
@@ -19,27 +19,32 @@ namespace DotNetNuke.Security.Permissions
     [Serializable]
     public class PermissionInfo : BaseEntityInfo, IPermissionDefinitionInfo
     {
-        /// <inheritdoc cref="IPermissionDefinitionInfo.ModuleDefId" />
+        private int moduleDefId;
+        private int permissionId;
+
+        /// <inheritdoc cref="ModuleDefId" />
         [XmlIgnore]
         [JsonIgnore]
-        [Obsolete($"Deprecated in DotNetNuke 9.13.1. Use {nameof(IPermissionDefinitionInfo)}.{nameof(IPermissionDefinitionInfo.ModuleDefId)} instead. Scheduled for removal in v11.0.0.")]
+        [Obsolete($"Deprecated in DotNetNuke 9.13.1. Use {nameof(ModuleDefId)} instead. Scheduled for removal in v11.0.0.")]
+        [CLSCompliant(false)]
         public int ModuleDefID
         {
-            get => ((IPermissionDefinitionInfo)this).ModuleDefId;
-            set => ((IPermissionDefinitionInfo)this).ModuleDefId = value;
+            get => this.moduleDefId;
+            set => this.moduleDefId = value;
         }
 
         /// <inheritdoc />
         [XmlElement("permissioncode")]
         public string PermissionCode { get; set; }
 
-        /// <inheritdoc cref="IPermissionDefinitionInfo.PermissionID" />
-        [XmlElement("permissionid")]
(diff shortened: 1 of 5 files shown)


bdukes:
Allowing members which differ only by casing will break users of VB.Net (they should be able to cast to the interface to resolve the issue, but it would be breaking until they make that change).

valadas:
@bdukes do you know if making the old one [CLSCompliant(false)] solves that ? My guess was that it would use the one without that attribute for VB but I did not try it, just assuming...

bdukes:
If you add a new property in an assembly and don't recompile the VB.Net assembly, it will ignore the new property (without regard for `CLSCompliant` attributes). Once compiled, it is no longer case insensitive.

However, any compile step will now cause build errors, specifically [BC31429](https://learn.microsoft.com/en-us/dotnet/visual-basic/misc/bc31429). This will be annoying for developers when updating which version they compile against. However, the major issue will be for code that is compiled _at runtime_ (e.g. anything in `.ascx` files, `App_Code`, or codebehind files). These will stop working all of a sudden and take the site down. This seems less likely with permissions than with `PortalId`, but still a realistic scenario.

valadas:
I am closing this one, working on an analyzer...

PR #5961, opened 2024-02-07
Fixed a node typing issue in resource manager

valadas:
Fixed a node typing issue in resource manager

This add `@types/node` to the `dev-dependencies` so we hopefully have better typings and also adds `skipLibCheck` to the typescript configuration so that basically only "our" code is checked.

@bdukes this may be the fix for the current build issue with https://github.com/dnnsoftware/Dnn.Platform/pull/5960 if you want to rebase it after this fix.


david-poindexter (approved):
Looks good assuming the build is successful.

bdukes (approved):

Issue #5920, opened 2024-01-14
[Feature]: Integrate `Bulk Install` (previously known as `PolyDeploy`) and improve UI/UX


david-poindexter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of problem

Currently, there is no way to install multiple modules at a time within DNN Platform.  It would be nice to have a solution for this.

Over the past several years, a popular solution has been to utilize `PolyDeploy`, an open-source solution originally developed by Cantarus.  It has been a staple solution for many since its launch and several community members have contributed to it to improve various aspects.  

For the sake of better maintainability, Cantarus ultimately contributed the project to the DNN Community.  See [GitHub repo here](https://github.com/DNNCommunity/PolyDeploy).

### Description of solution

The first step was to place the solution into the Dnn.Platform GitHub repo.  This work has already been started on [this feature branch](https://github.com/dnnsoftware/Dnn.Platform/tree/feature/bulk-install).  Once the module is in a workable state, a decision will be made as to whether or not it is ready for inclusion within DNN 10.

The front end is currently using AngularJS (v1).  There is no desire to bring in this dependency.  So, the plan is to rewrite the front end using [DNN Elements](https://github.com/dnncommunity/dnn-elements) (pure web components).  This is the same as was used for the new Resource Manager.

### Description of alternatives considered

n/a

### Anything else?

Following are some rudimentary wireframes to help guide the desired UI/UX improvements:

## Install
![Install_1_v1](https://github.com/dnnsoftware/Dnn.Platform/assets/4568451/e52541ba-2908-493a-8098-590cd265debd)

![Install_2_v1](https://github.com/dnnsoftware/Dnn.Platform/assets/4568451/1e8ec53e-938a-41cd-8bd1-e9de6f32ecfb)

## Event Log
![Event_Log_v1](https://github.com/dnnsoftware/Dnn.Platform/assets/4568451/918e8d01-7f92-4ad3-b276-4c7da7c29af9)

## API Users
![API_Users_v1](https://github.com/dnnsoftware/Dnn.Platform/assets/4568451/59f2a9e9-6e51-4672-9aa5-69a72d9e7059)

## IP Safelist
![IP_Safelist_v1](https://github.com/dnnsoftware/Dnn.Platform/assets/4568451/ccd9a18d-df3e-47dd-8bb4-c6200b336c62)

[...]

PR #5966, opened 2024-02-10
Move module out of `PolyDeploy` folder and rename a few base project files

david-poindexter:
## Summary
This is the first of many PRs in support of #5920


valadas (commented):
Looks like this is mostly a move/rename of files but the build is failing and I could not quickly find out why from the logs here. @david-poindexter do you see more details when doing a full build locally or it just works fine?

david-poindexter:
@valadas I am pretty sure the project didn't build correctly before.  We just didn't see it because it had not been added to the `DNN_Platform.sln`.  I was honestly expecting the build to fail until we get a lot of this project converted to the new namespace, etc.  How would you like for me to handle this?  I can do one of the following:

* Remove the project from the solution for now.
* Try to make a much larger PR and get the project to compile.

If you have another idea, I am open to that as well.

david-poindexter:
Okay, I just verified.  The project was unable to build successfully before my changes, so would you like for me to remove it from the solution, or we ignore the build issues for now?

valadas:
Ok sounds good. Unless anyone disagrees so we can work faster on this, only on this (feature) branch we will work with one reviewer and backpedal only if needed. We will then follow the branch rules when we merge the whole bulk-install feature in a protected branch (release/10.0.0 for instance).

Until we get farther along it is kinda normal to get build failures here.

valadas (approved):

PR #5971, opened 2024-02-15
Completely remove the ability to edit another user from control.

donker:
## Summary

The EditUser control used to be used to edit users in a (now defunct) module in DNN pre PersonaBar. Currently it is only used to edit one's details when you click on Edit on your profile page. It no longer needs the ability to edit another user and this should be disabled.


bdukes (approved):

david-poindexter (approved):
I did not have time to test this yet, but it looks good from a browser review perspective.  I will not merge it to allow for testing if others want or feel the need to.

mitchelsellers (approved):

valadas (approved):
Gave it a test too and everythhign appears to be working to me.

PR #5989, opened 2024-03-01
Bulk Install: Resolved various build and watch issues

david-poindexter:
## Summary
This PR is part of ongoing work for #5920.  

It aims to:
* Resolve all MSBuild issues for the new Bulk Install module.
* Resolve all packaging issues within the MSBuild system for DNN.
* Ensure the front-end project is picked up for the `yarn` workspace, and thus `lerna`.
* Ensure the front-end placeholder component is rendered in the module's `view.html` file.

image


valadas (approved):
Nice

valadas:
Oh, except this:

image

I believe it may be caused by not have xml documentation turned on in release mode maybe... Ping me up if that is not clear

david-poindexter:
> Oh, except this:
> 
> image
> 
> I believe it may be caused by not have xml documentation turned on in release mode maybe... Ping me up if that is not clear

I just pushed a commit to see if this fixes it.  Great catch!

valadas:
It fixed, it, merging...

PR #6012, opened 2024-04-24
Use Cake Frosting optimized addins

pascalberger:
## Summary
Updates build to make use of addins optimized to be used with Cake Frosting. Compared to the previously used addins, these addins contain NuGet dependencies instead of shipping them as part of the package. 

In case of Cake.Git this fixes some issues with native LibGit2Sharp DLLs. Beside that it gives you control about the version of dependencies which should be used in your build, which can be useful for example for Cake.Issues.MsBuild and MsBuild.StructuredLogger if a newer binlog version should be supported and Cake.Issues.MsBuild is not available yet.

Also updates addins to latest version. For Cake.Git to a version which support Cake Frosting 4.x, and for Cake Issues addins to the latest versions, which also brings support for binlog format version 20.

### DIFF
--- a/Build/Build.csproj
+++ b/Build/Build.csproj
@@ -17,9 +17,9 @@
     <PackageReference Include="Cake.BuildSystems.Module" Version="7.0.0" />
     <PackageReference Include="Cake.FileHelpers" Version="7.0.0" />
     <PackageReference Include="Cake.Frosting" Version="4.0.0" />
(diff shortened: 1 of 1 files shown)


pascalberger:
Please let me know if you prefer to additionally have an issue for this kind of change.

bdukes (approved):

valadas (approved):
Awesome, thanks

PR #6014, opened 2024-04-26
Clean up additional references to `Xcillion`

### LINKED ISSUE #5998: [Bug]: In DNN V10 alpha Activity-Feed ViewProfile module throws Could Not Load Container ... /Xcillion/NoTitle.ascx

### What happened?

After a fresh install of v10, and you are already logged in as a superuser, just immediately click on "SuperUser Account" (1) in the upper right of the eyebrow menu, just to the right of the Logout link.

This is the URL
https://dnn100000rc1-deleteme.dnndev.me/Activity-Feed/userId/1

When the Profile page displays, it looks like (2) the ViewProfile module is using a Xcillion container in its settings still. 

image

### Steps to reproduce?

This is covered above.

### Current Behavior

This is covered above.

### Expected Behavior

After a clean install, clicking on the user's profile should not result in an error or a module expecting a container from Xcillion.

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

10.0.0 (alpha)

Chrome

### Code of Conduct

david-poindexter:
## Summary
Resolves #5998 

There were several rogue references to `Xcillion` that needed to be cleaned up.  Thanks for pointing out some of these @jeremy-farrance 🎉


valadas (approved):
Awesome

PR #6017, opened 2024-04-30
ENH: Return Id when inserting item into PetaPoco repository

### LINKED ISSUE #5983: [Enhancement]: Return Id (primary key) of INSERTed entity in PetaPoco repository

### Description of problem

When calling DAL2 I`nsert()` method, a frequent strategy for determining the new Id / primary key of the inserted object is to requery the database directly after the insert to get the highest id. 

### Description of solution

Not only is this wasteful, in that it requires an additional database roundtrip, is not 100% reliable since another `INSERT` could have happened in between, but it's unnecessary. PetaPoco patches in the `INSERT`ed Id into the object, so you really only need to call Insert(item), and then item.Id has the new primary key. What would be even better is to return the Id as an additional Insert method, e.g. `int Id = repo.Insert<int>(item);`

```
repo.Insert(item);
int Id = item.Id;
```

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

johnhenley:
Close #5983

## Summary

ENH: Return Id when inserting item into PetaPoco repository

### DIFF
--- a/DNN Platform/Library/Data/IRepository.cs
+++ b/DNN Platform/Library/Data/IRepository.cs
@@ -96,6 +96,12 @@ namespace DotNetNuke.Data
         /// <param name="item">The item to be inserted.</param>
         void Insert(T item);
 
+        /// <summary>Inserts an Item into the repository and returns value of Id field.</summary>
+        /// <typeparam name="TProperty">The type of the Id field.</typeparam>
+        /// <param name="item">The item to be inserted.</param>
+        /// <returns>Primary key of item inserted.</returns>
+        TProperty Insert<TProperty>(T item);
+
         /// <summary>Updates an Item in the repository.</summary>
         /// <param name="item">The item to be updated.</param>
         void Update(T item);
--- a/DNN Platform/Library/Data/RepositoryBase.cs
+++ b/DNN Platform/Library/Data/RepositoryBase.cs
@@ -112,6 +112,14 @@ namespace DotNetNuke.Data
         }
 
         /// <inheritdoc/>
+        public TProperty Insert<TProperty>(T item)
+        {
+            this.InsertInternal(item);
+            this.ClearCache(item);
+            return this.GetPrimaryKey<TProperty>(item);
+        }
+
+        /// <inheritdoc/>
         public void Update(T item)
         {
             this.UpdateInternal(item);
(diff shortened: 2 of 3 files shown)


valadas (commented):
Does'nt the exisint Insert method update the item itself with the new id in it ?

johnhenley:
> Does'nt the exisint Insert method update the item itself with the new id in it ?

Yes it does. But it's not obvious and I see a lot of DAL2 examples where an insert is immediately followed by a (full) get then orderbydescending by create date then firstordefault-- all just to find the new ID.

valadas:
> > Does'nt the exisint Insert method update the item itself with the new id in it ?
> 
> Yes it does. But it's not obvious and I see a lot of DAL2 examples where an insert is immediately followed by a (full) get then orderbydescending by create date then firstordefault-- all just to find the new ID.

No I mean does it not update the ID of the passed item to save ? Like one does not need an extra fetch to begin with ?

johnhenley:
> > > Does'nt the exisint Insert method update the item itself with the new id in it ?
> 
> > 
> 
> > Yes it does. But it's not obvious and I see a lot of DAL2 examples where an insert is immediately followed by a (full) get then orderbydescending by create date then firstordefault-- all just to find the new ID.
> 
> 
> 
> No I mean does it not update the ID of the passed item to save ? Like one does not need an extra fetch to begin with ?

That is correct. All I added was a new method that returns the ID rather than void. Just to make it obvious because a lot of people write code like "var id = insert ...

valadas:
Well one can't do var id = Insert.... was it returns void.

What I mean is that one should be able to do:
```cs
var dog = new Dog{Name="Clara", Age=2};
this.repository.Insert(dog);
var newId = dog.Id;
```

valadas:
Well one can't do var id = Insert.... was it returns void.

What I mean is that one should be able to do:
```cs
var dog = new Dog{Name="Clara", Age=2};
this.repository.Insert(dog);
var newId = dog.Id;
```

johnhenley:
> Well one can't do var id = Insert.... was it returns void.
> 
> 
> 
> What I mean is that one should be able to do:
> 
> ```cs
> 
> var dog = new Dog{Name="Clara", Age=2};
> 
> this.repository.Insert(dog);
> 
> var newId = dog.Id;
> 
> ```
> 
> 

Exactly why I thought the new method would be helpful, since it's a typical code pattern.  
var newId = this.repository.Insert(dog);

But no problem if you don't think it's worthwhile.

valadas:
Well, I am not opposed on the idea, but this way means clearing cache to get an ID that is already available and potentially an additional database roundtrip... Plus the original way would also return the results of any other computed and/or autogenerated fields in the entity... Also in this scenario what would be the type of a compound key ?

PR #6018, opened 2024-05-01
Fix #5994: Sorting (A-Z) members of roles in PersonaBar

### LINKED ISSUE #5994: [Bug]: Sorting (A-Z) members of roles

### What happened?

We manage a very large DNN site with different role groups and roles. The members of a role are now shown interchangeably. It would be nice if these were shown alphbetic (A-Z).

### Steps to reproduce?

1. Log in as Administrator
2. Choose in the admin personabar for Roles
3. Select a role with a lot of members

### Current Behavior

Notes that the members of the group are not shown to be alphabetic.

### Expected Behavior

Notes that the members of the group must be shown alphabetic (A-Z).

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.2 (latest release)

Firefox, Chrome, Safari, Microsoft Edge

### Code of Conduct

Mostafa-Moafi:
Fix #5994: Sorting (A-Z) members of roles in PersonaBar

### DIFF
--- a/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/RolesController.cs
+++ b/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/RolesController.cs
@@ -246,7 +246,7 @@ namespace Dnn.PersonaBar.Roles.Services
                         Localization.GetString("InvalidRequest", Components.Constants.LocalResourcesFile));
                 }
 
-                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName);
+                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName).ToList();
                 if (!string.IsNullOrEmpty(keyword))
                 {
                     users =
@@ -257,7 +257,7 @@ namespace Dnn.PersonaBar.Roles.Services
                 var totalRecords = users.Count;
                 var startIndex = pageIndex * pageSize;
                 var portal = PortalController.Instance.GetPortal(this.PortalId);
-                var pagedData = users.Skip(startIndex).Take(pageSize).Select(u => new UserRoleDto()
+                var pagedData = users.OrderBy(x => x.FullName).Skip(startIndex).Take(pageSize).Select(u => new UserRoleDto()
                 {
                     UserId = u.UserID,
                     RoleId = u.RoleID,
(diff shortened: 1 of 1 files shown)


Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/RolesController.cs
```diff
@@ -246,11 +246,11 @@ public HttpResponseMessage GetRoleUsers(string keyword, int roleId, int pageInde
                         Localization.GetString("InvalidRequest", Components.Constants.LocalResourcesFile));
                 }
 
-                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName);
+                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName).OrderBy(x => x.FullName).ToList();
```
bdukes, line 249: Can we move the `OrderBy` call to the beginning of line 260? This would allows us to only specify it once, and avoid creating another copy of the list.

bdukes (commented):
Thanks!

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/RolesController.cs
Mostafa-Moafi: yes of course, I fix it.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/RolesController.cs
```diff
@@ -246,7 +246,7 @@ public HttpResponseMessage GetRoleUsers(string keyword, int roleId, int pageInde
                         Localization.GetString("InvalidRequest", Components.Constants.LocalResourcesFile));
                 }
 
-                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName);
+                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName).ToList();
```
bdukes, line 249:
```suggestion
                var users = RoleController.Instance.GetUserRoles(this.PortalId, Null.NullString, role.RoleName);
```

bdukes (approved):

valadas (approved):
Looks good

PR #6031, opened 2024-05-09
Make MvcServicePath regex matching ignore case.

### LINKED ISSUE #6030: [Bug]: Regex checks for MVC paths in some HttpModules are not case insensitive and fail to match.

### What happened?

The Regex used to do the checks in [MvcHttpModule.cs](https://github.com/dnnsoftware/Dnn.Platform/blame/develop/DNN%20Platform/DotNetNuke.Web.Mvc/MvcHttpModule.cs#L50) and [MobileRedirectModule.cs](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/HttpModules/MobileRedirect/MobileRedirectModule.cs#L47) is `new Regex(@"DesktopModules/MVC/", RegexOptions.Compiled);`, which only matches paths with that exact case, while urls can have arbitrary casing and also the check in MvcHttpModule.cs is done on the `.ToLowerInvariant()` of the path.

Looks like it was changed/broken a long time ago by [this commit](https://github.com/dnnsoftware/Dnn.Platform/commit/6510c4e6ced88652322997e157184d40fcae99ac).

I am not sure what was the original intent with the checks, if changes since then have made them unnecessary, or whether fixing them now could break existing behaviors. I just found them while looking through the code and thought that would be good to share.

### Steps to reproduce?

Not really any steps to reproduce, but accessing an MVC view through the browser (domain.com/en-us/dEsKtopMoDules/mvC/<view path>) should trigger a match for the checks, which does not happen.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.3 (latest release), 9.13.4 (alpha), 10.0.0 (alpha)

_No response_

### Code of Conduct

dimarobert:
`Fixes #6030

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/MvcHttpModule.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/MvcHttpModule.cs
@@ -23,7 +23,7 @@ namespace DotNetNuke.Web.Mvc
 
     public class MvcHttpModule : IHttpModule
     {
-        public static readonly Regex MvcServicePath = new Regex(@"DesktopModules/MVC/", RegexOptions.Compiled);
+        public static readonly Regex MvcServicePath = new Regex(@"DesktopModules/MVC/", RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
         static MvcHttpModule()
         {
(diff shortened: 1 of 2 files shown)


bdukes (approved):
Thanks!

mitchelsellers (approved):

PR #6034, opened 2024-05-13
Add MaximumPageSize const property for limit page size

### LINKED ISSUE #6033: [Enhancement]: Add Limits for search result

### Description of problem

There is a search service between the content and the blog, which contains a parameter called pageSize. The attacker can change the output value of the program by increasing this parameter. For example, by increasing this number, the output of the program can be several megabytes.
This problem causes the query of the program with the database to run longer and in case of successive attacks, it will cause the server to waste resources and eventually fail.

this service is `/API/internalservices/searchService/search` in DotNetNuke.Web project

### Description of solution

we can add a number to maximum page sizes in `SearchServiceController`

_No response_

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

Mostafa-Moafi:
Add MaximumPageSize const property for limit page size
Fixes #6033


DNN Platform/DotNetNuke.Web/InternalServices/SearchServiceController.cs
```diff
@@ -37,6 +37,7 @@ public class SearchServiceController : DnnApiController
         private const string ModuleTitleCacheKey = "SearchModuleTabTitle_{0}";
         private const CacheItemPriority ModuleTitleCachePriority = CacheItemPriority.Normal;
         private const int ModuleTitleCacheTimeOut = 20;
+        private const int MaximumPageSize = 100;
```
bdukes, line 40: 100 seems a little small to me for an absolute limit. If we made this configurable via a host setting, I'm fine with 100 as a default.

Mostafa-Moafi: I agree with you but when the user searches in DNN, shows a maximum of 100 results per page.

image

valadas:
Just to be sure this maximum is only per page right, one could still obtain more results but on other pages correct?

Mostafa-Moafi:
@valadas yes, of course. I just set the maximum page size per page.

valadas:
Ok, I am not too concerned with this but like Brian said, it would be nicer if that value comes from the web.config or a host setting and could be changed without having to recompile. I could see potential implementations where that API gets called from 3rd party modules that may need more results and not support paging (would only get page 1).

Mostafa-Moafi:
@bdukes  @valadas OK, I add a new input to the search section in the persona bar. do you agree?

valadas:
Sounds like a decent place to put that yes.

Mostafa-Moafi:
@valadas Thanks.
I added a new field in the search section.

bdukes (approved):

valadas (approved):
Looks good to me

PR #6036, opened 2024-05-14
prevent writing invalid module title

### LINKED ISSUE #6035: [Enhancement]: Prevent writing invalid module title (such as HTML, CSS, JS)

### Description of problem

I think that we shouldn't set HTML or JS in the module title and should prevent writing invalid code.

### Description of solution

It's very simple to solve it. we should use `WebUtility.HtmlEncode()` when update module title.

_No response_

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

Mostafa-Moafi:
Fixes #6035

### DIFF
--- a/DNN Platform/Website/admin/Modules/Modulesettings.ascx.cs
+++ b/DNN Platform/Website/admin/Modules/Modulesettings.ascx.cs
@@ -9,6 +9,7 @@ namespace DotNetNuke.Modules.Admin.Modules
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
+    using System.Net;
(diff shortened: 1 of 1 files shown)


valadas:
@Mostafa-Moafi can you change this as discussed in the issue. We need to make this an opt-in behavior with a web.config setting or some such.

valadas (changes requested):
As discussed in the issue if we go this route, it would need to be a setting in the web.config or some such. This would break many existing implementations as it is

mitchelsellers:
We are closing this for now, if this is updated to support opt-in it can be re-opened for consideration

Issue #6028, opened 2024-05-08
[Bug]: Unable to Install DNN 9.13.3 on Azure SQL


mitchelsellers:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Attempting to Install DNN 9.13.3 into a clean Azure SQL database is failing starting on, or about, May 1st, 2024.  

This has worked previously, without any issues, however, now it is not possible to install as there are abundant errors (See below)

Suspect issues with SQL Server compatibility with 2024 product releases on Azure.  

### Steps to reproduce?

1. Setup a new database in SQL Azure
2. Attempt to install DNN using the blank Azure database
3. BOOM!

### Current Behavior

Multiple SQL Failures at about 20% completion and no success

### Expected Behavior

DNN should install and work

### Relevant log output

```shell
The following error logs happen, no install is successful.

`
2024-05-08 04:04:17.697+00:00 [REDACTED][D:7][T:13][ERROR] DotNetNuke.Data.SqlDataProvider - System.Data.SqlClient.SqlException (0x80131904): Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
'CREATE/ALTER PROCEDURE' must be the first statement in a query batch.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near 'GO'.
Incorrect syntax near the keyword 'ALTER'.
Must declare the scalar variable "@RemoveIncompatibleSchema".
Must declare the scalar variable "@Feature".
Must declare the scalar variable "@IsCurrentVersion".
Must declare the scalar variable "@Feature".
Must declare the scalar variable "@Feature".
Incorrect syntax near 'GO'.
Incorrect syntax near the keyword 'EXEC'.
Incorrect syntax near 'GO'.
Incorrect syntax near the keyword 'ALTER'.
Must declare the scalar variable "@Feature".
A RETURN statement with a return value cannot be used in this context.
A RETURN statement with a return value cannot be used in this context.
Incorrect syntax near 'GO'.
Incorrect syntax near the keyword 'ALTER'.
Must declare the scalar variable "@ApplicationId".
Must declare the scalar variable "@ApplicationId".
Must declare the scalar variable "@ApplicationId".
[...]

PR #6038, opened 2024-05-16
Fixes to sql script and new rollup scripts for version 10

donker:
This PR adds a few missing objectQualifiers from the v10 SQL script and replaces the old 8.0.4  rollup scripts with a new one for v10.

# Methodology for creating the rollup scripts

1. A new/empty DB was created
2. All SqlDataProvider SQL scripts (i.e. not the membership provider scripts) that the current installation process runs were run using a unique schema (not dbo.) and objectqualifier (not empty)
3. In SSMS a schema was generated by selecting all tables, views, procedures and functions (i.e. none of the security objects)
4. Similarly a data script was created.
5. The data script was parsed and changed to conform to the way DNN runs scripts (remove GO statements within IDENTITY_INSERT blocks).
6. With search and replace all databaseOwners and objectQualifiers were put back in place.

Note with item 2: DNN runs "Schema", then "Data" and then all scripts higher than 8.0.4 (starting thus with 08.00.04.01). This process was mimicked using a console app to isolate the exact process.

Note this will probably resolve issue #6028


mitchelsellers:
@donker This looks awesome, I'll test the build results in Azure in the next few days, very curious to see how much this helps install

valadas (approved):
I tested a clean install on SQL Server 2017 and everything worked fine. I tried an upgrade and it failed with `ICSharpCode.SharpZipLib` wich is probably unrelated. So I think this is good since it would only apply to clean installs anyway

mitchelsellers (approved):

PR #6041, opened 2024-05-17
Set SameSite and Secure property for authentication cookie

### LINKED ISSUE #6039: [Enhancement]: Set SameSite and Secure property for authentication cookie

### Description of problem

1. Defining SAMESITE to ensure that cookies are not sent to a third-party site or address. The value of this tag must be defined as "Strict".
2. Definition of the SECURE tag to ensure that the cookie is transmitted through a secure communication path and that it cannot eavesdrop (the SECURE tag should be activated after activating the HTTPS protocol).

### Description of solution

For the secure enabled, I think we should check SSL enabled in DNN and update the cookie when logging user.
for set SamaSite to Stric, we should just update the authentication cookie

_No response_

### Anything else?

_No response_

- [X] Yes

- [ ] Yes

### Code of Conduct

Mostafa-Moafi:
Fixes #6039

### DIFF
--- a/DNN Platform/Library/Security/PortalSecurity.cs
+++ b/DNN Platform/Library/Security/PortalSecurity.cs
@@ -614,7 +614,6 @@ namespace DotNetNuke.Security
         /// <param name="createPersistentCookie">if set to <c>true</c> [create persistent cookie].</param>
         public void SignIn(UserInfo user, bool createPersistentCookie)
         {
-            var settings = PortalController.Instance.GetCurrentPortalSettings();
             if (PortalController.IsMemberOfPortalGroup(user.PortalID) || createPersistentCookie)
             {
                 // Create a custom auth cookie
@@ -670,7 +669,7 @@ namespace DotNetNuke.Security
                 if (!string.IsNullOrEmpty(authCookie?.Value))
                 {
                     authCookie.SameSite = SameSiteMode.Strict;
-                    if (settings.SSLEnabled)
+                    if (UrlUtils.IsSecureConnectionOrSslOffload(HttpContext.Current.Request))
                     {
                         authCookie.Secure = true;
                     }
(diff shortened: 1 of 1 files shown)


DNN Platform/Library/Security/PortalSecurity.cs
```diff
@@ -668,6 +669,14 @@ public void SignIn(UserInfo user, bool createPersistentCookie)
                 var authCookie = HttpContext.Current.Response.Cookies[FormsAuthentication.FormsCookieName];
                 if (!string.IsNullOrEmpty(authCookie?.Value))
                 {
+                    authCookie.SameSite = SameSiteMode.Strict;
+                    if (settings.SSLEnabled)
```
mitchelsellers, line 673: Should we be only setting this for SSLEnabled & SSLEnforced?

Mostafa-Moafi: Secure cookies are sent along only with requests that travel over a secure connection; usually this means a HTTPS connection.

bdukes: This should use `UrlUtils.IsSecureConnectionOrSslOffload` to check if the current request is secure

mitchelsellers:
There are situations here where existing settings are able to control this via the various cookie settings in the web.config.

Long-term, after discussion, ideally, we should be working to set these values based on the SSL Configurations to drive the setting of web.config files such as "Secure" for HttpCookies and the Auth Cookies settings via the web.config.  

A similar configuration of SameSite mode via a setting so that it is user-controllable would be preferred due to the diversity of implementations and system requirements for usage of the DNN Platform.

mitchelsellers:
This issue was discussed at the most recent Git Approvers meeting and the end result is that we do not support this process, given the possibility of conflicting situations with other configurations based on web.config settings and how this will interact with SSL Offloading and otherwise.

We need to have a solution that will not introduce a conflict of that nature.

PR #6061, opened 2024-05-31
Adds lock on calls to ComputeHash to handle concurrency

### LINKED ISSUE #6059: [Bug]: JWT login fails with Unknown error "-1073741816" on ComputeHash

### What happened?

Hi, recently I started having some errors using JWT login API. From time to time the system starts throwing errors on all login and extend tokens requests:
When calling /DesktopModules/JwtAuth/API/mobile/login

Message: Unknown error "-1073741816".

StackTrace:

   at System.Security.Cryptography.BCryptHashAlgorithm.HashCore(Byte[] array, Int32 ibStart, Int32 cbSize)
   at System.Security.Cryptography.HashAlgorithm.TransformBlock(Byte[] inputBuffer, Int32 inputOffset, Int32 inputCount, Byte[] outputBuffer, Int32 outputOffset)
   at System.Security.Cryptography.SHA384Managed.HashCore(Byte[] rgb, Int32 ibStart, Int32 cbSize)
   at System.Security.Cryptography.HashAlgorithm.ComputeHash(Byte[] buffer)
   at Dnn.AuthServices.Jwt.Components.Common.Controllers.JwtController.LoginUser(HttpRequestMessage request, LoginData loginData)
   at Dnn.AuthServices.Jwt.Services.MobileController.Login(LoginData loginData) 

I think it might have something related to the problem found on this thread:
https://answers.microsoft.com/en-us/windows/forum/all/sha256-computehash-started-throwing/4e18a046-8bd6-4369-968c-93f34a1088bb

Looking at the source code it seems that concurrency is not being handled when calling the ComputeHash function here:
https://github.com/dnnsoftware/Dnn.Platform/blob/1e04695c9937d2a7e282ab00e909479501a1411b/DNN%20Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs#L154

As the Hasher is static is shared across threads thus can this be the source of the problem?

If anyone can have a look at this I would gladly appreciate.

### Steps to reproduce?

I haven't been able to force the error when the system is working normally but after it errors one time any call to the login api will fail with the above error.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Affected Versions

9.13.3 (latest release)

_No response_

### Code of Conduct

miguelvaz78:
Fixes #6059 

## Summary
Added a new static object that will be used to lock execution across threads.
Both ComputeHash calls were wrapped inside a lock statement so no concurrent executions can be done from different threads. Because of this, the result of the hash function had to be assigned to a new variable that is then used outside of the lock code block.

### DIFF
--- a/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
+++ b/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
@@ -43,6 +43,8 @@ namespace Dnn.AuthServices.Jwt.Components.Common.Controllers
         private static readonly HashAlgorithm Hasher = SHA384.Create();
         private static readonly Encoding TextEncoder = Encoding.UTF8;
 
+        private static object hasherLock = new object();
+
         /// <inheritdoc/>
         public string SchemeType => "JWT";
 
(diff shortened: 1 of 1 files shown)


miguelvaz78:

________________________________

bdukes (approved):

valadas (approved):
Looks good to me, thanks!

miguelvaz78:
Thanks guys!

valadas:
Thank YOU

TagGIF

PR #6098, opened 2024-07-13
Removed optimize-css-assets-webpack-plugin dependency

valadas:
This was not used, this PR removes that unused dependency and replaces #5598


bdukes (approved):

david-poindexter (approved):
Thanks @valadas 🎉

Issue #6119, opened 2024-08-29
[Bug]: Persona Bar does not use Resource files for "All Users"


Timo-Breumelhof:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I edited SharedResources.resx to fix #6091. This works for the Module permissions grid, but not the Persona Bar.

![image](https://github.com/user-attachments/assets/e3d9954e-8f0c-424d-ad94-3e17c927fb25)

![image](https://github.com/user-attachments/assets/ae4a1ee0-8c5e-4e72-884d-79fdf5c96477)

### Steps to reproduce?

1. Edit SharedResources.resx and change "All Users to "Everyone"
2. Open the Permissions of Page in the Persona Bar
3. It still says "All Users" in the permissions grid.

### Current Behavior

Persona Bar has the virtual role "All Users" text hard coded

### Expected Behavior

The text should be loaded from the resource file SharedResources.resx

### Anything else?

Related to #6091 and both should be fixed at the same time IMO.
All users > Everyone change : https://dnncommunity.org/forums/future-development/dnn-ideas/quotall-usersquot-is-not-the-best-term-for-security/

### Affected Versions

9.13.3 (latest release)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bdukes:
The current [permission grid calls `Localization.LocalizeRole`](https://github.com/dnnsoftware/Dnn.Platform/blob/4dc00f90e2fddf7a996bfeec16d4dee0e8778587/DNN%20Platform/Library/Security/Permissions/Controls/PermissionsGrid.cs#L819) on the role name. If the Persona Bar's [`RoleController`](https://github.com/dnnsoftware/Dnn.Platform/blob/4dc00f90e2fddf7a996bfeec16d4dee0e8778587/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Roles/RolesController.cs) API methods are updated to call that before returning to the frontend, that should get us pretty close.

Timo-Breumelhof:
@donker I think this is fixed and can be closed right?

PR #6120, opened 2024-08-29
Change "All Users" to "Everyone"

### LINKED ISSUE #6091: [Enhancement]: Replace "All users" with "Everyone"

### Description of problem

Please read: https://dnncommunity.org/forums/future-development/dnn-ideas/quotall-usersquot-is-not-the-best-term-for-security/

### Description of solution

Use "Everyone" instead of "All Users" as text for that "virtual role"

All visitors, but most seem to prefer "Everyone", see Forum Thread.

### Anything else?

_No response_

- [ ] Yes

- [ ] Yes

### Code of Conduct

Timo-Breumelhof:
fixes #6091

## Summary

This must be the simplest PR in our history? ;-)
I edited SharedResources.resx to replace "All Users" with "Everyone".

But there's one issue remaining, the Persona Bar does not use this resource key, in there "All Users" is hard coded.
So IMO we should only implement this one if #6119 is fixed too..

### DIFF
--- a/DNN Platform/Website/App_GlobalResources/SharedResources.resx
+++ b/DNN Platform/Website/App_GlobalResources/SharedResources.resx
@@ -796,7 +796,7 @@
     <value>EndJob</value>
   </data>
   <data name="AllUsers.Text" xml:space="preserve">
-    <value>All Users</value>
+    <value>Everyone</value>
   </data>
   <data name="Superuser.Text" xml:space="preserve">
     <value>SuperUser</value>


bdukes:
The role name _should_ be localized in all scenarios (see #6119 that @Timo-Breumelhof also opened). Changing the localized text should be a completely safe change. So, once #6119 is resolved, this should be a pretty clean change.

valadas:
Good point, however I would still prefer we handle all of "All Users" localization everywhere in one PR. It is referenced in a couple of localization paragraphs and I'd prefer it is named the same everywhere to avoid making it even more confusing if it is "All Users" in some places and "Everyone" in other places.

Timo-Breumelhof:
Closed as superseded by #6133

PR #6128, opened 2024-09-02
Updated Initialize.cs to show modern dotnet4+ versions

### LINKED ISSUE #6127: [Bug]: DotNet Framework Version is not Accurate

### What happened?

After confirming .NET Framework 4.8.1 is installed, DNN still shows just 4.8:

image

### Steps to reproduce?

Ensure you have .NET Framework updated to v4.8.1, then login to DNN as superuser. Hover on the PersonaBar logo to see the info panel (screen clip above).

### Current Behavior

Currently displays v4.8

### Expected Behavior

Should show v4.8.1

### Relevant log output

```shell
n/a
```

### Anything else?

Enhanced code a little so that if there is a v4.8.2 someday, the update will be simple(r).
 
I plan to contribute the code for this, pull request coming soon.

### Affected Versions

9.13.3 (latest release), 9.13.4 (alpha)

Chrome, Microsoft Edge

### Code of Conduct

jeremy-farrance:
Fixes #6127 

## Summary
Quick update so that the .NET Framework version shown (e.g. in the PersonaBar) is accurate up to v4.8.1.

Also updated the code style (no more `else`) making it easy to added new versions in the future.

### DIFF
--- a/DNN Platform/Library/Common/Initialize.cs
+++ b/DNN Platform/Library/Common/Initialize.cs
@@ -383,14 +383,20 @@ namespace DotNetNuke.Common
             // Otherwise utilize release DWORD from registry to determine version
             // Reference List: https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies
             var release = GetDotNet4ReleaseNumberFromRegistry();
+            if (release >= 533320)
+            {
+                version = "4.8.1 or later";
+            }
             if (release >= 528040)
             {
                 version = "4.8";
             }
-            else
+            if (release >= 461808)
             {
                 version = "4.7.2";
             }
+            // This code should never execute
+            version = "No 4.5 or later version detected (or unexpected result)";
 
             return new Version(version);
         }


valadas (changes requested):
Except the build failed...

jeremy-farrance:
> Except the build failed...

Sorry about that. Works for me now (local build).

valadas:
There is still 2:

image

jeremy-farrance:
> There is still 2:

Are you messing with me? I just fixed the opposite above, "SA1513 - Closing brace should be followed by blank line"
 
I tried this in both VS Code and VS 2022. The only way not to get one of the two errors is ...
 
Oh, so both editors are just inserting whitespace up to the indented cursor. I am supposed to know to spot that and fix it? 
 
Okay, sorry, lesson learned. Follow the rules even if you can't see the difference. I think this will fix it, at least I get neither error in VS 2022 when I force the blank lines.

DNN Platform/Library/Common/Initialize.cs
```diff
@@ -383,11 +383,17 @@ private static Version GetNETFrameworkVersion()
             // Otherwise utilize release DWORD from registry to determine version
             // Reference List: https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies
             var release = GetDotNet4ReleaseNumberFromRegistry();
+            if (release >= 533320)
+            {
+                version = "4.8.1";
+            }
+
             if (release >= 528040)
```
bdukes, line 391: Does this actually work? Doesn't this need to be in descending order? If `release` is 533320, that should mean 4.8.1, but then it will _also_ be greater than 528040 and 461808, so won't `version` end up being set to `"4.7.2"`? Maybe I'm getting this mixed up in my head, but it doesn't read to me like it should work correctly without an `else` before each `if`.

jeremy-farrance: Good catch, it doesn't. It was lazy-converted from using `return "4.8";`. I'll switch this pull to draft until I can get this done properly. Thanks!

jeremy-farrance (commented):
Okay, code is updated.

DNN Platform/Library/Common/Initialize.cs
jeremy-farrance: Logic is correct now. Humbly, thanks!

mitchelsellers (approved):

PR #6138, opened 2024-09-27
Fix SharpZipLib issues

donker:
This PR removes a few unnecessary SharpZipLib references and ensures the library is not included in the bin folder of the main Upgrade zip. The latter ensures DNN will still upgrade as the references in the web.config during upgrade are still pointing to the old version.

This follows up issues mentioned in #2735


mitchelsellers (approved):

valadas (approved):
Awesome, just tested an upgrade and it worked good. Thanks!

Issue #6153, opened 2024-10-05
[Enhancement]: PersonaBar, Manage Roles - Add Role ID to Visible Information


jeremy-farrance:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of problem

Needing Role ID while managing a DNN site is common for administrators or developers. Currently, the only ways to get this detail is to use the SQL Console or go directly digging in the database.

### Description of solution

As with previous enhancements that added User ID to the User Management page (see Issue #4412) and Page ID to Pages, we simply need to surface the Role ID on the Manage Roles page.

Visual for my proposed solution.
![image](https://github.com/user-attachments/assets/29807b7a-8b03-46dd-8f41-631b6ef80150)

### Description of alternatives considered

Might be better to move it up to the Role row so its visible in the immediate list w/o clicking edit?

### Anything else?

I plan to submit code and a pull request for this.

### Do you be plan to contribute code for this enhancement?

- [X] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

david-poindexter:
Closed per #6154

PR #6154, opened 2024-10-05
Update PersonaBar Manage/Roles to show Role ID

jeremy-farrance:
Fixes Issue #6153

## [Enhancement]: PersonaBar, Manage Roles - Add Role ID to Visible Information

This screenshot shows the result of these changes.
image


valadas (commented):
Looks good but:
1. Could it maybe not be bolder than the reset, it stands out a bit much IMO.
2. The added props already existed but were not typed I guess, so it fixes warnings ?

jeremy-farrance:
> Looks good but:
> 
> 1. Could it maybe not be bolder than the reset, it stands out a bit much IMO.
> 2. The added props already existed but were not typed I guess, so it fixes warnings ?

Agree on 1, updated/removed bolder from the label. For 2, yes, left them in because it was nice to see it compile with no warnings! :)

valadas (approved):
Awesome, thanks for the quick turnaround

david-poindexter (approved):
Looks good to me - thank you @jeremy-farrance 🎉

PR #6197, opened 2024-11-06
Merges `develop` latest changes back into `release/10.0.0`

valadas:
Merges `develop` latest changes back into `release/10.0.0`

Also ran a local build after the rebase and had to fix some packaging issue with the new Aperture Theme.


valadas:

PR #6199, opened 2024-11-06
ENH: Use message subject when sending instant subscription messages

### LINKED ISSUE #6176: [Enhancement]: Use message Subject as subject of DNN Messaging Notification Emails

### Description of problem

When a module uses DNN messaging, the email being sent out always have a generic subject.
See: 

image

Although DNN messaging is a great way to notify users and also send out email notifications, the fact that the subject is always the same is not ideal. 
As an example the forum on DNN Community.
I get a lot of emails sent out by the community site.
During a busy period I prefer to only open the messages that are Forum replies to a thread I contributed to.
Currently you have to open an email to see what it's about, which is not very efficient.

### Description of solution

IMO if a message has a subject, we should also use that as the Subject for the email notification.

I have been thinking of reasons why you would not want this behavior but I cant think of any TBH..

### Anything else?

We would really like this change for the Community Forums Module :-)

- [ ] Yes

- [ ] Yes

### Code of Conduct

johnhenley:
## Summary

When sending instant/individual email notifications, uses message subject rather than generic "New message from {PortalName}".
Closes #6176


DNN Platform/Library/Services/Social/Messaging/Scheduler/CoreMessagingScheduler.cs
```diff
@@ -708,8 +707,8 @@ private void SendMessage(MessageRecipient messageRecipient)
             {
                 var senderName = GetSenderName(author.DisplayName, portalSettings.PortalName);
                 var senderAddress = GetSenderAddress(senderName, portalSettings.Email);
-                var emailBodyItemContent = GetEmailItemContent(portalSettings, messageRecipient, emailBodyItemTemplate);
-                var subject = string.Format(emailSubjectTemplate, portalSettings.PortalName);
+                var emailBodyItemContent = GetEmailItemContent(portalSettings, messageRecipient, emailBodyItemTemplate);
+                var subject = InternalMessagingController.Instance.GetMessage(message.MessageID).Subject;
```
mitchelsellers, line 711: Should this have a check for a possible missing subject?

donker (approved):
LGTM

valadas (approved):
LGTM

PR #6205, opened 2024-11-14
Update with correct version info

david-poindexter:
This PR updates the bug report template with correct DNN Platform version information.


bdukes (approved):

valadas:
Thanks @david-poindexter 

PR #6242, opened 2024-12-06
Add .NET SDK to dependabot

bdukes:
See https://devblogs.microsoft.com/dotnet/using-dependabot-to-manage-dotnet-sdk-updates/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,15 @@ updates:
       - "javascript"
       - "Type: Maintenance"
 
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):
Great idea!

DO we want to have any sort of Ignore of items, or just keep up always and decide when the pR comes in?

bdukes:
I don't see a reason to avoid using a newer SDK, so long as it doesn't break the build.

donker (approved):
Good idea

PR #6243, opened 2024-12-06
Updated Quick Settings Performance

### LINKED ISSUE #6241: Protect performance when a quick settings control exists
### Discussed in https://github.com/dnnsoftware/Dnn.Platform/discussions/6239

<div type='discussions-op-text'>

<sup>Originally posted by **thienvc** December  6, 2024</sup>
After logging in, the following code slowed down the web performance
https://github.com/dnnsoftware/Dnn.Platform/blob/2e7a5eb203941b590a5dba9e820bd286e5b2a9e8/DNN%20Platform/Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs#L116
We should save the "QS_FirstLoad" setting somewhere other than in "ModuleSettings"
Because this action leads to calling ClearCache(tabid), making the cache useless</div>

### Suggested Fix

Ensure that settings values are only updated if/when they need to be updated.

mitchelsellers:
- Fixes #6241

Given the reported issue in #6239, each time a module setting is updated the cache is cleared.  The current code, was always updating the first load to be false, even if the first load value was already set to false.

This change utilizes the retreived setting value to ensure that we are ONLY updateing the module to flip the quick settings flag to false if it is currently set to true.  This has a substantial impact to performance for modules using Quick Settings.

### DIFF
--- a/DNN Platform/Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs
+++ b/DNN Platform/Website/admin/Menus/ModuleActions/ModuleActions.ascx.cs
@@ -113,7 +113,12 @@ namespace DotNetNuke.Admin.Containers
                     this.quickSettings.Controls.Add(control);
 
                     this.DisplayQuickSettings = this.ModuleContext.Configuration.ModuleSettings.GetValueOrDefault("QS_FirstLoad", true);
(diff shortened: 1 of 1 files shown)


bdukes (approved):

thienvc:
That's great, thank you

donker (approved):
LGTM

PR #6267, opened 2024-12-12
Sometimes we need to get additional keyFound variable to know if ResourceKey does not exist or exists but empty value

thienvc:
## Summary

Sometimes we need to get additional keyFound variable to know if ResourceKey does not exist or exists but empty value

```
bool keyFound;
string strLocalization = Localization.GetString(ResourceKey, LocalResourceFile,  PortalSettings, null, false, out keyFound);
if(string.IsNullOrEmpty(strLocalization)){
  if(!keyFound)
  {
      strLocalization = "Key not found, click here to create....";
  }
  else {
      strLocalization = "Key found but empty value";
  }
}
return strLocalization;
```

### DIFF
--- a/DNN Platform/Library/Services/Localization/ILocalizationProvider.cs
+++ b/DNN Platform/Library/Services/Localization/ILocalizationProvider.cs
@@ -22,6 +22,8 @@ namespace DotNetNuke.Services.Localization
 
         string GetString(string key, string resourceFileRoot, string language, PortalSettings portalSettings, bool disableShowMissingKeys);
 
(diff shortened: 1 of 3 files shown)


valadas:
We discussed this little bit today and we would prefer to have an extension method to the interface that would return only a boolean and one can call that only if they obtain an empty string from the existing code. It would make this non-breaking.

mitchelsellers:
Closing this item for now, the above suggeted improvement if created could be updated for reconsideration

PR #6272, opened 2024-12-28
Bulk Install: Remove CSS var typo now that dnn-elements has been updated

david-poindexter:
## Summary
This PR takes care of a TODO to correct a CSS variable name once the correction was made in `dnn-elements` (which it now has been).

This is part of the ongoing effort related to #5920.


valadas (approved):
Awesome

PR #6293, opened 2025-01-04
Rebased bulk-install branch with latest changes from release/10.0.0

valadas:
Rebased bulk-install branch with latest changes from release/10.0.0


david-poindexter:
This is a release management task and per our policy, we are self-approving it.

PR #6305, opened 2025-01-10
BulkInstall: Changed to use new form pattern from dnn-elements

valadas:
This makes validation a breeze. Also moved the form into a modal

image
image


david-poindexter:
This is part of the ongoing effort for #5920

DNN Platform/Modules/BulkInstall/BulkInstall.Web/src/components/tabs/api-users/api-users.tsx
```diff
@@ -84,6 +67,31 @@ export class ApiUsers {
...
+            onSubmit={(event) => {
+              event.preventDefault();
+              this.createUser(this.newUser);
+            }}
+          >
+            <h4>New API User</h4>
+            <dnn-input
+              type="text"
+              label="Name"
+              helpText="Enter API User name"
```
david-poindexter, line 85:
```suggestion
              helpText="Enter API User Name"
```

david-poindexter (changes requested):
Looks great.  Should we remove the `clearfix` SCSS now that it is no longer used in the DOM?

valadas:
> Looks great. Should we remove the `clearfix` SCSS now that it is no longer used in the DOM?

It is still used after the table.

david-poindexter (approved):

PR #6307, opened 2025-01-11
Implemented localization on bulk-install

valadas:
Implemented localization on bulk-install

Part of #5920


david-poindexter (approved):
Very nicely done 🎉

PR #6314, opened 2025-01-16
Feature/6313 unique profile property per portal

### LINKED ISSUE #6313: [Bug]: A portal cannot have a List ProfileProperty identical to another portal

### What happened?

In an installation with multiple portals, different portals can't have a profile property with the same name and values. If you enter a list value, you will get a popup saying it has been added, but it's not:

Image

This is caused by Unique constraints on the Lsts table, not containing the PortalId column.

### Steps to reproduce?

1. Create a profile prop of type list (FavoriteColor)
2. give it a few values (Red: 1, Green: 2, Blue: 3)
3. create a new portal in the same installation
4. Create a profile prop of type list (FavoriteColor) in the new portal
5. Try to give it a value you already used in the other portal (Red: 1)

### Current Behavior

If you enter a list value, you will get a popup saying it has been added, but it's not:

Image

### Expected Behavior

The value should be added.

### Relevant log output

```shell
none
```

### Anything else?

I'm working on a fix.

### Affected Versions

9.13.7 (alpha)

Firefox, Chrome, Safari, Microsoft Edge

### Code of Conduct

skamphuis:
Fixes #6313

## Summary
To allow multiple portals in the same installation tto have identical lists (for example for profile properties) I have:
* Added PortalId to the Unique constraint `Lists_ListName_Value_Text_ParentID`
* Added PortalId to the Unique index `Lists_ParentID`
* Added portalId to the check in SP `AddListEntry`


donker (approved):
I'm good with this fix

bdukes (approved):

PR #6316, opened 2025-01-19
Setup architecture to rebuild default.css using scss

valadas:
This adds support for building client-side parts (styles/scripts) using modern technologies such as scss and typescript.

This PR is a work in progress and in the current state only has what is needed to build and spit the files out at the proper places and switches the new empty v10 default.css

I'll keep working on this pending PR until I feel it is presentable...


DNN Platform/Website/Resources/Shared/stylesheets/dnndefault/10.0.0/default.css
bdukes, line 1: We want this and the `.map` in the `.gitignore`, since it's generated from the `.scss` file now, right?

mitchelsellers (approved):

DNN Platform/Website/Resources/Shared/stylesheets/dnndefault/10.0.0/default.css
bdukes: I would prefer to keep the source code for old versions.

valadas: In the last commits, I realized that some files were not included in the final package or would be old build. I noticed that the frontend build happens after CopyWebsite and well with this in mind, I changed the output path to go to that copy instead which is not under source control. This way we have the result out of git already.

valadas:
Ok, so I had to do a full first pass before doing adjustments, in the last PRs, I fixed to build issues, change a few things and refactored the install.css too. I think this is good for review as a first step. I'd like to relocate some stuff to make more sense and some sizing is not perfect but that could be polished in subsequent PRs.

mitchelsellers (approved):

PR #6325, opened 2025-01-29
Resolved issue with default template image src paths

david-poindexter:
## Summary
This PR resolves an image path issue in the `Default` site template that was causing images to not load in certain non-standard environments.  The change implements the use of a token path prefix instead of the leading `/`.

This...

```
src=&amp;quot;/portals/0/images/DnnSeal.png&amp;quot;
```

...becomes

```
src=&amp;quot;{{PortalRoot}}images/DnnSeal.png&amp;quot;
```


mitchelsellers (approved):

david-poindexter:
I tested the install package using the `Default` template and all images rendered correctly using the new token-based paths.  I think this is good to merge now.

david-poindexter:
I tested the install package using the `Default` template and all images rendered correctly using the new token-based paths.  I think this is good to merge now.

PR #6334, opened 2025-02-01
Fixed an issue with distribution path of Aperture

valadas:
The destination path for Aperture could not exist on a clean clone (or in this case in CI). This PR ensures the directory exists before the copy.


david-poindexter (approved):
Great catch!

mitchelsellers (approved):

PR #6335, opened 2025-02-01
Refactored Aperture build to fix CI issue

valadas:
In CI or on a brand new clone, we do not have a WebsitePath in the settings, so in this case we spit out the files to a temp directory so we can grab them from there to orchestrate our packaging.


david-poindexter (approved):
Thank you @valadas 🎉 🌮

donker (approved):
LGTM

PR #6337, opened 2025-02-01
Changed some default styles

valadas:
default.css was too opiniated on the color of heading, we should just use the ambiant foreground color at that level and leave it to theme developers to override if they want and how they want.

We did not have the DNN dark brown in the color system, this is now our foreground color which makes foreground text the DNN brown color and looks almost black on smaller text and keeps a soft not 100% contrast that feels nice to me.

image

image


donker (approved):
OK, I guess

david-poindexter (approved):

PR #6346, opened 2025-02-02
Fixed a font weight issue

valadas:
default.css and Aperture were having a bit of a fight for a font weight that made the theme not look as designed initially. default.css should not try to be too opiniated, so I removed the font weight from it to let the theme win that fight.
image


mitchelsellers (approved):

david-poindexter (approved):
Awesome - thank you @valadas 🎉

PR #6354, opened 2025-02-03
Include JWT Auth by default

bdukes:
## Summary
I believe the @dnnsoftware/approvers had decided at some point that the JWT auth package should be included by default starting in DNN 10.


valadas (approved):

david-poindexter (approved):
Thanks @bdukes 🎉

PR #6355, opened 2025-02-03
Remove TelerikRemoval module after removing Telerik

### LINKED ISSUE #6351: [Enhancement]: Remove telerik removal extention and PB entry in fresh DNN 10 installs

### Description of problem

Image

### Description of solution

No reason for this to exist on a fresh DNN 10 install.

_No response_

### Anything else?

_No response_

- [ ] Yes

- [ ] Yes

### Code of Conduct

bdukes:
Fixes #6351

### DIFF
--- a/Build/PackagingPatterns.cs
+++ b/Build/PackagingPatterns.cs
@@ -15,6 +15,9 @@ namespace DotNetNuke.Build
         /// <summary>Gets or sets a set of files to exclude from the upgrade package.</summary>
         public string[] UpgradeExclude { get; set; }
 
+        /// <summary>Gets or sets a set of files to include from the upgrade package.</summary>
+        public string[] UpgradeInclude { get; set; }
+
         /// <summary>Gets or sets a set of files to include from the symbols package.</summary>
         public string[] SymbolsInclude { get; set; }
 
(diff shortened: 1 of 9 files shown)


donker (approved):
Good one

bdukes:
I had to adjust how I was including the host page in the manifest to get it to automatically uninstall (there's code looking for it in two places and I put it in the wrong one, which was still partially read, but unused).

bdukes:
@valadas I just pushed a change which should cause it to uninstall itself on installs. Though maybe we should remove it altogether from clean installs… Is there an easy way to do that?

valadas:
Awesome, yes I thing we could probably just not include it in the install package. I have a few things to handle here but I may be able to take a look later tonight

valadas:
Ok, added a commit to skip that module only for install but keep it upon upgrades. I have tested a clean install which looks fine without the module or the page.

Then I also tested:
1. Installed DNN 9.10.2
2. Installed FAQ Module (which depends on Telerik)
3. Installed DNN v9.13.7 and did not remove Telerik
4. Install DNN v10.0.0

Telerik was remove, the page and Telerik removal module are gone, the site works but FAQ is broken.

This looks perfect to me and we can remove all that Telerik removal code in v11.

Thanks @bdukes for handling this, I'll merge after the build completes.

PR #6350, opened 2025-02-03
Bump dependencies

bdukes:
## Summary
Try #6330 again.

I removed the bump of Microsoft.Bcl.AsynInterfaces

PR #6356, opened 2025-02-04
Replace legacy System.IdentityModel.Tokens.Jwt

bdukes:
This PR builds on top of #6350, which upgrades the System.IdentityModel.Tokens.Jwt NuGet package. That package is now [marked as "legacy"](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e5272528a608b903e321223dd05f132d1d31a5dd/src/System.IdentityModel.Tokens.Jwt/System.IdentityModel.Tokens.Jwt.csproj#L7), so this follow-on PR replaces its usage with the recommended Microsoft.IdentityModel.JsonWebTokens package (which was already referenced). This was mostly straightforward, other than the header validation check that needed to be moved around a bit.


valadas (approved):
I tested a jwt login and it worked fine, nice one.

david-poindexter (approved):
Awesome! 🎉

PR #6359, opened 2025-02-07
Use valid syntax for CSS gradients

bdukes:
## Summary
Many of the CSS gradients defined in the platform used the outdated syntax which specified the direction the gradient comes _from_, whereas the final approved syntax uses the keyword `to` along with the direction that the gradient _goes to_. This PR adjusts those usages.

### DIFF
--- a/DNN Platform/Modules/CoreMessaging/module.css
+++ b/DNN Platform/Modules/CoreMessaging/module.css
@@ -129,7 +129,7 @@ button, input[type="button"], input[type="reset"], input[type="submit"], .dnnPri
 	background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,rgba(31,152,241,1)), color-stop(100%,rgba(2,111,196,1))); /* Chrome,Safari4+ */
 	background: -webkit-linear-gradient(top, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* Chrome10+,Safari5.1+ */
 	background: -o-linear-gradient(top, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* Opera 11.10+ */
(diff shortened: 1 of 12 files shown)


mitchelsellers (approved):

DNN Platform/Modules/CoreMessaging/module.css
```diff
@@ -314,7 +314,7 @@ button, input[type="button"], input[type="reset"], input[type="submit"], .dnnPri
 		background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,rgba(31,152,241,1)), color-stop(100%,rgba(2,111,196,1))); /* Chrome,Safari4+ */
 		background: -webkit-linear-gradient(top, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* Chrome10+,Safari5.1+ */
 		background: -o-linear-gradient(top, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* Opera 11.10+ */
-		background: linear-gradient(top, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* W3C */
+		background: linear-gradient(t bottom, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* W3C */
```
valadas, line 317:
```suggestion
		background: linear-gradient(to bottom, rgba(31,152,241,1) 0%,rgba(2,111,196,1) 100%); /* W3C */

```

valadas (changes requested):
Found one little typo.
In the scss files only (not the css ones) would could also remove the prefixed versions of those gradients as we do have autoprefixer setup to add only the ones needed according to our browserlists settings.

valadas (approved):

david-poindexter (commented):
Awesome - thanks @bdukes 🎉

PR #6361, opened 2025-02-08
Clean up DesktopModules folder in dev

donker:
This removes odd remnants left in the DesktopModules folder:

1. What is in the Admin/Dnn.PersonaBar folder gets overwritten by the actual installation of that module
2. Same for SiteExportImport
3. The portal template xsd has no place in the DesktopModules folder but should be in a core folder like Components. The only reason it's here is because there used to be a Portals module in this location.


mitchelsellers (approved):

valadas (approved):
Looks good, thanks

PR #6376, opened 2025-02-12
Fix reflection calls in Client Resources

bdukes:
## Summary
DotNetNuke.Web.Client calls DNN methods via reflection. However, the methods it calls were removed in DNN 10, so those calls are failing. This means that configuring client resources by portal does not work in DNN 10 anymore b/c it can't find which portal the request is for.

I've fixed the worst offending method, but I think there are more.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client/ClientResourceSettings.cs
+++ b/DNN Platform/DotNetNuke.Web.Client/ClientResourceSettings.cs
@@ -9,8 +9,11 @@ namespace DotNetNuke.Web.Client
     using System.Reflection;
     using System.Web;
 
+    using DotNetNuke.Abstractions.Portals;
     using DotNetNuke.Instrumentation;
     using DotNetNuke.Internal.SourceGenerators;
+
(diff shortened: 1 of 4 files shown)


mitchelsellers (approved):

bdukes:
This PR is based on #6350 now, because I needed to add a dependency to the project. It now uses DI to create abstractions where possible instead of using as much reflection.

valadas:
I am still getting this on this PR. Also am I correct that this PR includes both #6350 and 6356 @bdukes ? If so can we close the other 2 or there is a reason for them ?
image

bdukes:
I would prefer that #6350 get merged in first, so the other two can be reviewed on their own. This PR includes #6350, but does not include #6356.

It looks like we can't exclude Microsoft.Bcl.AsyncInterfaces from the install package because it's required by DI, so I'll have to downgrade it so it doesn't break things on upgrade.

valadas (approved):
Nice, tested working here.

PR #6381, opened 2025-02-16
Refactored upgrade code after rollup of sql scripts

### LINKED ISSUE #6364: [Bug]: User profile on user management shows empty

### What happened?

User profile on user management shows empty

### Steps to reproduce?

Login as Host.
Open the user management on the PB.
Click the profile icon on any user.

### Current Behavior

Image

### Expected Behavior

Profile editor opens.

### Relevant log output

```shell

```

### Anything else?

No errors on console or network.
No changes were made to the profile settings. 
Fresh DNN10 install. Tested on the SU and a normal user created by SU.

### Affected Versions

10.0.0 (alpha)

Firefox

### Code of Conduct

valadas:
In #6038 a rollup script was put in place to revamp our clean installation logic, get rid of some upgrade logic and potentially speed up installs, etc.

In RC1 issue #6364 was reported. I spent quite a lot of time trying to figure it out. Had to do a `git bisect` and it identified the commit of the rollup script as being the first commit to cause that issue. After a lot of head scratching I realized that the code in upgrade.cs was the one creating the virtual host page that should hold the module that allows editing user profiles from the Persona Bar. This solved the issue for HOST users but not for admins. For admins, this lives in the portal templates, it was cleaned up from modules we don't want, but this one needed to be brought back as the Persona Bar needs it for admins.

As I was in upgrade.cs I though since we will document in our upgrade steps to upgrade to latest v9 before going to v10, we can get rid of all the other upgrade logic for all versions <10 in there.

Closes #6364

### DIFF
--- a/DNN Platform/Library/Services/Upgrade/Upgrade.cs
+++ b/DNN Platform/Library/Services/Upgrade/Upgrade.cs
@@ -8,7 +8,6 @@ namespace DotNetNuke.Services.Upgrade
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
-    using System.Reflection;
     using System.Security.Cryptography;
     using System.Text;
     using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


mitchelsellers (approved):
I think this is ok, we need to track a "YOU MUST" upgrade following the upgrade logic, as this makes a 9.6.0+ stop prior to 10 required.

@dnnsoftware/approvers any concerns with this?

david-poindexter (approved):
Amazing work @valadas - I am impressed with your persistence.  

@mitchelsellers I think we just need to recommend (via DNN Docs and in the release notes) an upgrade path that takes users to the latest 9.x.x release prior to jumping to 10.0.0.  It would also be the cleanest approach to document, and should serve people well.

PR #6394, opened 2025-02-20
Updates versions to 9.13.8 and other autogenerated files

valadas:
We have a 9.13.8 script already which makes it hard to test 9.13.8 fixes without creating a huge git diff.

This is a release management task and as per our policy I will self-merge it after the build


DNN Platform/Website/DotNetNuke.Website.csproj
```diff
@@ -1275,6 +1275,7 @@
     <Content Include="Providers\DataProviders\SqlDataProvider\09.12.01.SqlDataProvider" />
     <Content Include="Install\Config\09.13.04.config" />
     <Content Include="Providers\DataProviders\SqlDataProvider\09.13.04.SqlDataProvider" />
+    <Content Include="Providers\DataProviders\SqlDataProvider\09.13.08.SqlDataProvider" />
```
bdukes, line 1278: I don't see this file in the PR, should this be here?

valadas: It was pre-existing but never got added to the project file https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Website/Providers/DataProviders/SqlDataProvider/09.13.08.SqlDataProvider

PR #6408, opened 2025-02-24
#6399 Theme Header and Footer class

### LINKED ISSUE #6399: [Bug]: DNN 10 > Aperture Theme Footer styling is not specific enough

### What happened?

I opened a page with the Community Forums on it and the footer of a topic picks up styling from the Theme.

### Steps to reproduce?

1. Place a module which has in internal footer element on a page
2. Use the Aperture Theme.

### Current Behavior

The internal module's header and footer pick up styling meant for the Theme

Image

### Expected Behavior

The internal module's footer and header should not pickup Theme styling

I spotted 2 issues.
1. The fact that there is styling for the header element makes internal headers sticky too
 https://github.com/dnnsoftware/Dnn.Platform/blob/release/10.0.0/DNN%20Platform/Skins/Aperture/src/scss/sections/_header.scss

2. Internal footer pick up the background color of the Theme footer.
https://github.com/dnnsoftware/Dnn.Platform/blob/release/10.0.0/DNN%20Platform/Skins/Aperture/src/scss/sections/_footer.scss

### Anything else?

I suggest not to style on any general elements like header, footer etc. from the theme as that might lead to clashes with element in modules.

```
<header class="theme-header">

<footer class="theme-footer">
```

I what I tend to use to avoid this.

### Affected Versions

10.0.0 (alpha)

Firefox

### Code of Conduct

Timo-Breumelhof:
## Summary

Add a CSS class to the Theme Header and Footer (+ associated SCSS) so their styling does not influence headers and footer in the page / module content.

fixes #6399

### DIFF
--- a/DNN Platform/Skins/Aperture/partials/_footer.ascx
+++ b/DNN Platform/Skins/Aperture/partials/_footer.ascx
@@ -1,4 +1,4 @@
-<footer>
+<footer class="aperture-footer">
   <div class="aperture-container">
         <div class="footer-left">
           <div class="footer-menu">
--- a/DNN Platform/Skins/Aperture/partials/_header.ascx
+++ b/DNN Platform/Skins/Aperture/partials/_header.ascx
@@ -1,4 +1,4 @@
-<header>
+<header class="aperture-header">
   <div class="eyebrow-bar">
     <div class="aperture-container">
       <dnn:Login runat="server" id="dnnLogin" />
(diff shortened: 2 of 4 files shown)


bdukes (approved):

david-poindexter (approved):
Great catch @Timo-Breumelhof - thanks for the contribution! 🎉

PR #6409, opened 2025-02-24
Min height of Page Content #6402

### LINKED ISSUE #6402: [Enhancement]: DNN 10 Aperture > empty pages look a bit weird?

### Description of problem
For us this is totally clear, but for someone new to DNN, IMO this looks weird:

Image

I know it's the footer but still..

### Description of solution

I see two options:

1. A Sticky Footer

Image

2. Add a minimum height for main when the PB is visible
.personabar-visible .theme-main{
min-height: 30vh;
}

Image

_No response_

### Anything else?

_No response_

- [x] Yes

- [ ] Yes

### Code of Conduct

Timo-Breumelhof:
## Summary

This PR makes sure the content has a minimum height of 20vh even when there are no module son the page.
This is only the case when the PersonaBar is visible.

image

I did wonder if we should add a _dnn-edit.scss file for this?
What do you think @valadas / @david-poindexter ?

fixes #6402

### DIFF
--- a/DNN Platform/Skins/Aperture/default.ascx
+++ b/DNN Platform/Skins/Aperture/default.ascx
@@ -6,7 +6,7 @@
   <!--#include file="partials/_header.ascx" -->
   
   <!-- Main Content -->
-  <main>
+  <main class="aperture-main">
     <div id="BannerPane" runat="server"></div>
     <div id="ContentPane" class="aperture-content-pane" runat="server"></div> 
     <div id="FluidPane" runat="server"></div>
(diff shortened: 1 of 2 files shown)


valadas (changes requested):
If we make something like:
html, body, #Form { height: 100% }
#Form { display: flex; flex-direction: column;}
.aperture-theme>main {flex-grow: 1;}

Then we make sure the footer is always pushed all the way to the bottom filling the window

Timo-Breumelhof (commented):
@valadas ok, see new commit

valadas (approved):
@Timo-Breumelhof  Does 100vh vs 100% make any difference regarding the edit bar ? If they both work fine, I am good with this

Timo-Breumelhof:
> @Timo-Breumelhof Does 100vh vs 100% make any difference regarding the edit bar ? If they both work fine, I am good with this

It's the same with less CSS.
See:
image

mitchelsellers (approved):

PR #6414, opened 2025-02-25
Merge `develop` into `release/10.0.0`

bdukes:


valadas:
Thanks @bdukes

PR #6416, opened 2025-02-26
Fixed an issue where auth providers config would use wrong layout

### LINKED ISSUE #6365: [Bug]: Site Settings in Authentication Provider Modules Shows Site Header and Footer

### What happened?

I assume this will happen with any module that has a settings.aspx page loaded in the persona bar. I noticed it with the Active Directory authentication provider as an example but confirmed it happens with the default authentication provider as well. When editing the module and then clicking the Site Settings tab, the settings.aspx page is loaded along with the header and footer of the actual site. It appears the page is loaded in an iframe.

The header and footer of the site should not be loaded into this view.

### Steps to reproduce?

1. Go to extensions
2. Switch to Authentication Systems
3. Click the edit pencil for the default authentication provider
4. Click the Site Settings tab
5. You will see the header and footer of the site loaded in the site settings window.

### Current Behavior

Image

### Expected Behavior

The header and footer of the site should not be loaded in this view.

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.0.0 (alpha)

Chrome

### Code of Conduct

valadas:
Popup skin should always be used if requested no matter if it lives in an actual popup or in an iframe in the PersonaBar

Closes #6365

### DIFF
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -33,6 +33,7 @@ namespace DotNetNuke.Framework
     using DotNetNuke.UI;
     using DotNetNuke.UI.Internals;
     using DotNetNuke.UI.Modules;
+    using DotNetNuke.UI.Skins;
     using DotNetNuke.UI.Skins.Controls;
     using DotNetNuke.UI.Utilities;
     using DotNetNuke.Web.Client;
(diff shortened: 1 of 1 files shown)


mitchelsellers (approved):

david-poindexter (approved):
Awesome @valadas 🎉

PR #6417, opened 2025-02-26
Fix styling issue with API Tokens Settings page on the PersonaBar

donker:
Due to an error the styling of this personabar page was absent. This PR fixes that.


mitchelsellers (approved):

valadas (approved):
Good catch

PR #6256, opened 2024-12-06
Add GitHub Action to keep caniuse-lite updated

bdukes:
## Summary
This PR adds a GitHub Action using [browserslist-update-action](https://github.com/c2corg/browserslist-update-action), which opens a PR to bump the database used by [browserslist](https://browsersl.ist/) on a regular basis.

This PR also:
 - updates `caniuse-lite` (the package with the database used by browserslist)
 - bumps Yarn from 4.5.0 to 4.5.3
 - runs `yarn dedupe` to reduce duplicate versions of dependencies

PR #6418, opened 2025-02-26
Enable corepack for browserslist action

bdukes:
## Summary
The Browserslist update GitHub action has been failing since it was introduced in #6256, with an error about needing to enable `corepack`. This PR adds that step to the workflow.


valadas (approved):

valadas:

PR #6419, opened 2025-02-26
Updated versions and bug report template for next release

valadas:
Updated versions and bug report template for next release


valadas:

PR #6428, opened 2025-03-04
Mark unused CoreCacheType enum as obsolete

bdukes:
## Summary
`CoreCacheType` is never used within DNN. Ideally, we could mark the type as deprecated in 9.x so it can be removed in 11.0.0.


valadas (approved):
Great

valadas:

PR #6436, opened 2025-03-07
Fix typo in styles text

bdukes:
Replace "varialbes" with "variables"

Also remove `\n` from confirmation text that was rendering literally. Since this is just HTML content, adding a line break does not render anything (without adding a `white-space` style).


valadas (approved):
Good catch

mitchelsellers (approved):

PR #6437, opened 2025-03-07
Use correct terminology in Tag Manager connector

bdukes:
## Summary
It could be confusing whether the container or account ID should be entered into the connector.


valadas (approved):
Nice

david-poindexter (approved):

PR #6445, opened 2025-03-12
Fix UrlUtils tests

bdukes:
## Summary
The build for #6442 is failing because of an unrelated issue with tests.


valadas (approved):

valadas:
Since this is only touching tests, I'll force-merge it as a build/release task as it makes all builds fail in subsequent PRs.

PR #6446, opened 2025-03-12
Fix two tab deletion issues

### LINKED ISSUE #6438: [Bug]: TabController.DeleteTab with deleteDescendants fails to delete the passed tab

### What happened?

When calling `DeleteTab(tabId, portalId, true)` you would expect the descendants to get deleted as well as the tab associated with the passed tabId. However, the call fails to do so on the initial call to the method. On subsequent calls, when no children remain after being deleted from the first call, the passed tab does get deleted. 

### Steps to reproduce?

1. Call `DeleteTab` on an instance of `TabController` while passing the overload `deleteDescendants` as true.
2. Check to see if the tab passed into the `DeleteTab` call has been deleted. It will not be.
3. Call `DeleteTab` on an instance of `TabController` again and verify that this time the tab has actually been deleted. 

### Current Behavior

At first glance, the cause seems to be related to caching and the order of operations. When calling the method with the overload, it first gets all descendant tabs and caches them. 

```
        public void DeleteTab(int tabId, int portalId, bool deleteDescendants)
        {
            List<TabInfo> descendantList = this.GetTabsByPortal(portalId).DescendentsOf(tabId);
            if (deleteDescendants && descendantList.Count > 0)
            {
                // Iterate through descendants from bottom - which will remove children first
                for (int i = descendantList.Count - 1; i >= 0; i += -1)
                {
                    this.HardDeleteTabInternal(descendantList[i].TabID, portalId);
                }
            }

            this.DeleteTab(tabId, portalId);
        }
```

At the end of the method it calls the `DeleteTab` method without the overload, which first checks for descendants and only deletes when no descendants exist. I don't think the cache has been cleared at this point, so descendants are found and the tab isn't deleted. Then the cache gets cleared, allowing for subsequent calls to work.

```
        public void DeleteTab(int tabId, int portalId)
        {
            // parent tabs can not be deleted
            if (this.GetTabsByPortal(portalId).WithParentId(tabId).Count == 0)
            {
                this.HardDeleteTabInternal(tabId, portalId);
            }

            this.ClearCache(portalId);
        }
```

### Expected Behavior

I would expect the initial call to clear the cache prior to calling to delete the main passed tab, or for the methods to not utilize a cache at all when determining descendants. Either way, I would expect a single call with the overload as `true` to delete the passed tab as well as descendants. 

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

9.13.7 (latest release), 10.0.0 (RC), 9.13.8 (RC)

_No response_

### Code of Conduct

bdukes:
## Summary
This PR fixes two bugs with deleting tabs. The first is described in #6438. The second comes when calling `DeleteTab` on a tab that has already been deleted (perhaps related to a workaround for the first issue 😉).

Fixes #6438

### DIFF
--- a/DNN Platform/Library/Entities/Tabs/TabController.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabController.cs
@@ -2903,7 +2903,10 @@ namespace DotNetNuke.Entities.Tabs
                 contentController.DeleteContentItem(tab);
             }
 
-            EventManager.Instance.OnTabDeleted(new TabEventArgs { Tab = tab });
+            if (tab != null)
+            {
+                EventManager.Instance.OnTabDeleted(new TabEventArgs { Tab = tab });
+            }
         }
 
         private bool SoftDeleteChildTabs(int intTabid, PortalSettings portalSettings)
--- a/DNN Platform/Library/Entities/Tabs/TabController.cs
+++ b/DNN Platform/Library/Entities/Tabs/TabController.cs
@@ -1221,7 +1221,8 @@ namespace DotNetNuke.Entities.Tabs
                 }
             }
 
-            this.DeleteTab(tabId, portalId);
+            this.HardDeleteTabInternal(tabId, portalId);
+            this.ClearCache(portalId);
         }
 
         /// <summary>Delete a Setting of a tab instance.</summary>


valadas (commented):
Is HardDeleteTabInternal ok here in the context ? Does that skip the recycle bin ?

valadas (approved):

donker (approved):
LGTM

Issue #5633, opened 2023-04-28
Module permissions lost when using "Place existing Module"


Timo-Breumelhof:
## Description of bug
When  you place an existing module that has the checkbox "Inherit View permissions from Page" turned off, you end up with a module without any permissions

## Steps to reproduce

List the precise steps to reproduce the bug:
As Host;
1. Create 2 pages
2. Place the HTML module on page 1
3. Open the module settings > Permissions
4. uncheck [ ]Inherit View permissions from Page
5. Set view and edit rights for Registered Users
6. Open the second page
7. Choose > Add existing module (make copy)

## Current behavior
When opening the module permissions for the copied module you will see that the right for "Registered Users" were not copied. Also the module is not visible to administrators.
This is due to the fact that there are no records for the module in the ModulePermissions Table.
`
SELECT * FROM vw_ModulePermissions where ModuleId = [moduleid] `

give back no records

If you set the Permissions for the module after you copied it, they do get saved to the ModulePermissions Table and the module is visible to Admins again.

## Expected behavior
Permissions of the original module should be copied to the new module

## Affected version

* [ ] 10.00.00 alpha build
* [x] 09.11.02 release candidate
* [ ] 09.11.01 latest supported release

Timo-Breumelhof:
First search, looks like this is missing in clone module:

![Image](https://github.com/user-attachments/assets/acb6233d-39a7-469f-abda-dd4a4cc08ee3)

https://github.com/dnnsoftware/Dnn.Platform/blob/cfed83c291d5e5072b2fa70924a8b7c35b1cdf9e/DNN%20Platform/Library/Entities/Modules/ModuleInfo.cs#L957

PR #6458, opened 2025-03-20
Fixes issue with copy module when inherit permissions is unchecked

schotman:
When a user with edit rights (aka admin, NOT host) copies an existing module that has "inherit permissions" unchecked, no permissions are saved for the new module. 

For the admin user the module "vanishes" from the page (seems to be not added). However the host user still sees the module.

line 957 

fix issue #5633


valadas (commented):
Was the scenario of cloning the module on a different portal tested, I am curious what would happen regarding permissions in that scenario.

schotman:
Yes it needs a change

PR #6459, opened 2025-03-20
Fixes issue with copy module when inherit permissions is unchecked

### LINKED ISSUE #5633: Module permissions lost when using "Place existing Module"

schotman:
Fixes #5633 

## Summary

When a user with edit rights (aka admin, NOT host) copies an existing module that has "inherit permissions" unchecked, no permissions are saved for the new module.

For the admin user the module "vanishes" from the page (seems to be not added). However the host user still sees the module.

reverted line 957 in ModuleInfo.cs and added code starting on line 688 in ControlBarController.cs

fix issue #5633


DNN Platform/DotNetNuke.Web/InternalServices/ControlBarController.cs
```diff
@@ -2,914 +2,919 @@
...
+                        PortalSettings.Current.AdministratorRoleId,
+                        Null.NullInteger,
+                        true);
+
+                    // Set PortalID correctly
+                    newModule.OwnerPortalID = newModule.PortalID;
+                    newModule.PortalID = PortalSettings.Current.PortalId;
+                    ModulePermissionController.SaveModulePermissions(newModule);
+                }
+                else
```
valadas, line 688: The `if` of this `else` has logic to set view permissions if `remote` is true, just thinking here if maybe we should just remove that condition would be the bug fix. Because now if the copy is on the same site it will copy all permissions (which users might not be expecting to happen). I think if at least the admin sees the module it would be good enough to fix the reported issue IMO.

meffj00:
This PR was not included in the latest release. 
Why not? I think it solves our practical problems! 
In which release and when will this PR be included?
Thanks!

valadas:
Honestly just lack of time, and as it relates to permissions too, we need to test and think about possible wrong scenarios like module sharing on multiportal with unrelated portal owners, etc. We just need to better review/test and discuss it and a lot of the reviewers have been pretty busy lately.

PR #6464, opened 2025-03-24
Avoid SQL error for existing ContentWorkflowActions rows

bdukes:
For https://github.com/dnnsoftware/Dnn.Platform/issues/6429#issuecomment-2748855512

Note that there's still an open question as to whether this fix improves the experience or not.


mitchelsellers (approved):
I think this is a good fix

valadas (approved):

PR #6470, opened 2025-04-01
Fixed an issue where normal users could not exit edit mode

### LINKED ISSUE #6363: [Bug]: Normal users with edit permissions can enter edit mode but cannot leave

### What happened?

The only way to (normally) leave edit mode is through the close button on the bottom bar that shows on edit mode. No such bar on edit permission.

### Steps to reproduce?

Give a normal user the edit or content permissions to a page.
Login as this normal user.
Enter edit mode clicking the pencil icon on the PB.
The bottom bar never shows, so the only way to exit edit mode is to move away from the page.

### Current Behavior

Bottom bar does not show, so no way to exit edit mode.

Image

### Expected Behavior

SuperUser:
Image

### Relevant log output

```shell

```

### Anything else?

The edit mode makes sense for edit permission.
Does it also make sense for the new "Content" permission?
Many content editors (specially for 2sxc) should not have access to the PB or edit mode.

### Affected Versions

10.0.0 (alpha)

Firefox

### Code of Conduct

valadas:
Closes #6363


david-poindexter (approved):
Looks good - thanks @valadas

bdukes (approved):
Validated fix, thanks!

PR #6476, opened 2025-04-03
Return empty string if no row is found

### LINKED ISSUE #6475: [Bug]: Data is Null in ModuleController.UpdateModuleSettingInternal

### What happened?

When trying to update a module setting with ModuleController, if the existing setting value is equal to "FileID=<Some ID that doesn't exist in the Files table>", ModuleController.UpdateModuleSettingInternal throws an exception due to a NULL value when fetching the setting value before setting it.

### Steps to reproduce?

1. Create a ModuleSetting with value "FileID=<Some ID that doesn't exist in the Files table>".
2. Try to update that setting using the ModuleController.

### Current Behavior

Exception thrown.

### Expected Behavior

Set the Module Setting.

### Relevant log output

```shell
Message:Data is Null. This method or property cannot be called on Null values.
StackTrace:
   at System.Data.SqlClient.SqlBuffer.get_String()
   at DotNetNuke.Entities.Modules.ModuleController.UpdateModuleSettingInternal(Int32 moduleId, String settingName, String settingValue, Boolean updateVersion)
InnerMessage:
InnerStackTrace:
Source:System.Data
```

### Anything else?

Appears to be due to a bug in the `FilePath` SQL Function.

### Affected Versions

9.13.8 (latest release)

_No response_

### Code of Conduct

DanielBolef:
Fixes #6475 

## Summary
Fix FilePath function so it won't return NULL.

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.13.09.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.13.09.SqlDataProvider
@@ -7,3 +7,27 @@
 /*****       for {databaseOwner} and {objectQualifier}  *****/
 /*****                                                  *****/
 /************************************************************/
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'{databaseOwner}[{objectQualifier}FilePath]') AND type = N'FN')
+    DROP FUNCTION {databaseOwner}[{objectQualifier}FilePath]
+GO
+
+CREATE FUNCTION {databaseOwner}[{objectQualifier}FilePath]
+(
+    @StrMayContainFileId nvarchar(255)
+)
+	RETURNS 			 nvarchar(500)
+AS
+	BEGIN
+		DECLARE @Path AS nvarchar(500);
+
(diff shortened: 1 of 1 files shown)


valadas (approved):
Nothing sticks out to me.

mitchelsellers (commented):
Given this change in behavior, I want to review the older script to see the behavior difference, as it seems that in this case  you are going to now provide an inconsistent behavior of users thinking things are ok, but the value isn't set to anything

bdukes:
FWIW the only meaningful change between this and [the previous version](https://github.com/dnnsoftware/Dnn.Platform/blob/6fabd76dc9ca67f5bfa01de9e23076521f60e118/DNN%20Platform/Website/Providers/DataProviders/SqlDataProvider/07.02.00.SqlDataProvider#L1021) is `RETURN @Path -- never Null!` became `RETURN ISNULL(@Path, '')`, so it's theoretically acting as it was supposed to all along. The current behavior makes it impossible to update a setting that is set incorrectly (at least, via that particular API), so I think this makes sense.

But I haven't reviewed in what other places this is called, so maybe somewhere we are relying on the function to return `NULL` sometimes.

PR #6477, opened 2025-04-03
Add cascade delete to PortalPermissions table

### LINKED ISSUE #6434: [Bug]: Cannot delete Portal when Portal Permissions exist

### What happened?

I tried to delete a portal, but got a SQL error.

### Steps to reproduce?

1. Create a portal (Don't think you can delete a portal if your instance only has one)
2. Create a role
3. Set up at least one portal permission for that role
4. (Try to) delete that portal via the sites menu

### Current Behavior

Foreign key error on the RoleID column of the PortalPermission table, probably when trying to cascade delete roles for the portal in question.

### Expected Behavior

Portal is deleted without error.

### Relevant log output

```shell
Error executing SQL: ;Exec dbo.DeletePortalInfo @0
The DELETE statement conflicted with the REFERENCE constraint "FK_PortalPermission_RoleId". The conflict occurred in database "<my database>", table "dbo.PortalPermission", column 'RoleID'.
```

### Anything else?

The FKs on PortalPermission are missing cascade delete on PortalId, PermissionId, RoleID, and UserID. I suspect this would happen if trying to delete the Portal, Permission, Role, or User associated with Portal Permissions.

Portal Permissions were added in 9.11.0 and not touched since, so this almost certainly affects all versions since then.

### Affected Versions

9.13.8 (RC), 9.13.7 (latest release), 10.0.0 (RC)

_No response_

### Code of Conduct

DanielBolef:
Fixes #6434 

## Summary
Add cascade delete to PortalPermissions table

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.13.09.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/09.13.09.SqlDataProvider
@@ -7,3 +7,24 @@
 /*****       for {databaseOwner} and {objectQualifier}  *****/
 /*****                                                  *****/
 /************************************************************/
+
+/*    Add cascade delete to PortalPermissions table         */
+/************************************************************/
+
+ALTER TABLE {databaseOwner}[{objectQualifier}PortalPermission] DROP CONSTRAINT [FK_{objectQualifier}PortalPermission_PermissionId]
+ALTER TABLE {databaseOwner}[{objectQualifier}PortalPermission] DROP CONSTRAINT [FK_{objectQualifier}PortalPermission_RoleId]
+ALTER TABLE {databaseOwner}[{objectQualifier}PortalPermission] DROP CONSTRAINT [FK_{objectQualifier}PortalPermission_UserId]
+GO
+
+ALTER TABLE {databaseOwner}[{objectQualifier}PortalPermission] WITH CHECK ADD CONSTRAINT [FK_{objectQualifier}PortalPermission_PermissionId] FOREIGN KEY([PermissionId])
+REFERENCES {databaseOwner}[{objectQualifier}Permission] ([PermissionId]) ON DELETE CASCADE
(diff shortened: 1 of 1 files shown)


valadas:
I was not able to reproduce the initial issue though, I created a child portal and a role with a user and was able to delete the portal just fine using the UI. Is there more to know about how to reproduce this issue ?

bdukes:
This is about _portal permissions_, which were added in 9.11 via https://github.com/dnnsoftware/Dnn.Platform/issues/5103. There was not a UI added to manage these yet, so you need to add them via SQL. @DanielBolef do you happen to have an example SQL query to add a permission?

DanielBolef:
Yeah, this should do it:
```
DECLARE @TopLevelPagePermissionId INT

SELECT @TopLevelPagePermissionId = PermissionId FROM Permission
WHERE PermissionKey = 'ADDTOPLEVELPAGE'

INSERT INTO PortalPermission (PortalId, PermissionId, AllowAccess, RoleID, UserID)
VALUES (<PortalID to be deleted>, @TopLevelPagePermissionId, 1, <RoleID on Portal to be deleted>, NULL)
```
That works for the real case I was encountering, but the same issue exists for User based Portal Permissions.
```
DECLARE @TopLevelPagePermissionId INT

SELECT @TopLevelPagePermissionId = PermissionId FROM Permission
WHERE PermissionKey = 'ADDTOPLEVELPAGE'

INSERT INTO PortalPermission (PortalId, PermissionId, AllowAccess, RoleID, UserID)
VALUES (<PortalID to be deleted>, @TopLevelPagePermissionId, 1, NULL, <UserID on Portal to be deleted>)
```
In my testing while creating the above SQL, I found that the bug still exists when Users have Portal Permissions on the Portal to be deleted. This is due to the fact that Users are not deleted (I think, only UserPortals are deleted?) when a Portal is deleted, so no cascade path exists to clean up the Portal Permission. I tried adding a cascade delete for the PortalId as well, but I get a `may cause cycles or multiple cascade paths` conflict between the cascade path for Roles and Portals. Not sure how to resolve this.

valadas (approved):

bdukes (approved):

donker (approved):
LGTM

PR #6505, opened 2025-04-14
Fixed an issue with wrong localization detection

### LINKED ISSUE #6503: [Bug]: Page Language Attribute

### What happened?

Page Language Attribute (`<html lang="...">`) Not Set in Latest DNN Versions

In the latest version of DNN Platform, the `<html lang="...">` attribute is missing from the rendered page, even when the current thread's `UICulture` is correctly set.

### Steps to reproduce?

1. Install the latest DNN Platform.
2. Configure a portal with default language `fa-IR` (or any other culture).
3. Access a normal page (not installwizard.aspx).
4. View the page source and inspect the `<html>` tag.

### Current Behavior

The `<html>` tag should contain the correct `lang` attribute based on the current culture, e.g.:
```html
<html lang="fa-IR">

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.0.0 (latest v10 release)

_No response_

### Code of Conduct

valadas:
I did a `git bisect` which pointed me to the commits in #6038 to rollup the sql scripts for v10. This made no sense to me since it's not really related to web.config and @uzmannazari pointed out in #6503 that a similar issue was caused by the LocalizationModule missing from the web.config. Also I could not reproduce the issue on upgrades. Then I thought since we have that v10 rollup, it uses the 10.00.00.config for the web.config xml merge scripts and up. Investigating the commit history of that file and comparing with 9.8.0 one I found that in #4251, a bugfix was done to ensure that the LocalizationModule is right after the UrlRewriteModule.

This means that upon upgrades it would be in the right place but with clean installs, that would never run and it was also missing from the base web.config, thus making it not be there at all.

I was debating between putting this fix in the 10.00.00.config or just in the web.config I thought the web.config is fine as upon upgrades, the order would already have been fixed (as long as we document to upgrade to latest v9 before upgrading to v10). For clean installs it would just already be at the right place to start with.

Closes #6503

### DIFF
--- a/DNN Platform/Website/development.config
+++ b/DNN Platform/Website/development.config
@@ -59,9 +59,9 @@
         <add key="UsePortNumber" value="true" /> -->
     <!-- Services Framework Tracing is primarily useful for developing and debugging -->
     <add key="EnableServicesFrameworkTracing" value="false" />
-	  <add key="UpdateServiceUrl" value="https://dnnplatform.io" />
+    <add key="UpdateServiceUrl" value="https://dnnplatform.io" />
     <add key="PreserveLoginUrl" value="true" />
-	  <add key="loginUrl" value="~/Login.aspx" />
+    <add key="loginUrl" value="~/Login.aspx" />
     <add key="ValidationSettings:UnobtrusiveValidationMode" value="None" />
     <add key="MobileViewSiteCookieName" value="dnn_IsMobile" />
     <add key="DisableMobileViewSiteCookieName" value="dnn_NoMobile" />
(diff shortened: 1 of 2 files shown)


david-poindexter (approved):
Excellent @valadas - thank you so much! 🎉

uzmannazari:
@valadas 
 investigation and after logging the entire installation process, we discovered that even other critical nodes from the 10.00.00.config file are not being merged into the final web.config during a clean install.

For instance, the ServiceRequestScopeModule, which is clearly defined in the config merge file, never gets added to the output web.config. This indicates a broader problem—it seems the installer is unable to properly read or apply the XML config merge files during installation.

We confirmed this by adding detailed logging to the XmlMerge.cs logic and saw that neither the expected InsertNode actions were triggered, nor were the relevant configuration files even being processed as expected.

This goes beyond a simple placement or syntax issue; it looks like the config merge engine is either silently skipping or failing to locate and process valid .config files, which can lead to a broken runtime environment after a clean install.

If needed, I can provide the logs and the exact config files we used for testing.

valadas:
@uzmannazari absolutely, the more information the better.

uzmannazari:
Thanks! I’ve prepared the full logs and the related config files we used during testing.

They clearly show that the 10.00.00.config was reached during installation, but certain <node> actions like insertbefore or insertafter were never executed—even though other parts of the same config file were successfully applied.

I'll upload everything shortly for your review. Let me know if you'd prefer a GitHub gist, direct attachment here, or a different way to share them.

[XmlMerge.zip](https://github.com/user-attachments/files/19728430/XmlMerge.zip)


Also, if there's anything specific you'd like us to trace further in the XmlMerge process, I'm happy to run

mitchelsellers (approved):
LGTM

mitchelsellers:
I've merged this one @valadas @uzmannazari maybe we create a separate issue to document other misses?

PR #6506, opened 2025-04-14
Fixed an issue where ServiceRequestScopeModule was missing from web.config

valadas:
This was only done in 10.00.00.config but xml-merge scripts only run upon upgrades of that version, hence this would be missing from clean installs and needs to exist in the base web.config for a clean install.

I have also verified the rest of 10.00.00.config and everything else appears to be pre-exising in the base web.config files.

Completes an extra detail from https://github.com/dnnsoftware/Dnn.Platform/issues/6503


david-poindexter (approved):
Nice catch!

mitchelsellers (approved):

PR #6517, opened 2025-04-17
Fix exception checking folder permissions

### LINKED ISSUE #6516: [Bug]: Error viewing folder that exists only in database

### What happened?

In the Site Assets, when the `EnableFileAutoSync` host setting is set to `Y`, clicking on a folder which exists in the database but does not exist on disk results in a 500 error.

### Steps to reproduce?

1. Ensure `EnableFileAutoSync` host setting is set to `Y`
2. Delete folder on disk for a folder that exists in the database
3. Select that folder from the tree

### Current Behavior

500 error (without any displayed error message)

### Expected Behavior

API call does not error

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

9.13.8 (latest v9 release)

_No response_

### Code of Conduct

bdukes:
Fixes #6516


donker (approved):
LGTM

mitchelsellers (approved):

PR #6523, opened 2025-04-28
Remove obsolete types

bdukes:
## Summary
There were a couple of methods marked to be removed in DNN 10 that did not get removed. Do we want to get those in DNN 10.0.1?


valadas (approved):
I am not opposed to those removals

mitchelsellers (approved):
I too am ok with this!  Good catch @bdukes

PR #6525, opened 2025-04-29
Add missing assembly

bdukes:
## Summary
The DNN 10.0.1 installer fails with a message that the System.Runtime.CompilerServices.Unsafe assembly could not be found.

screenshot of YSOD

PR #6526, opened 2025-04-29
Consolidate System.Runtime.CompilerServices.Unsafe.dll and ensure it's in the bin folder during install

donker:
This PR is an alternative to #6525 Both aim to address an issue where the installation of DNN will fail because System.Runtime.CompilerServices.Unsafe.dll is not in the bin folder when it first starts up. This assembly is required by the ServiceRequestScopeModule which is now part of the default installation.

This PR also consolidates the versions of this assembly in the project as we had 3 different versions until now.


donker:
Closed in favor of #6525

Issue #6547, opened 2025-05-15
[Bug]: SQL: DROP IF not supported on all SQL versions


leedavi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

There is an error upgrading on older versions of SQL DROP IF is only valid from 2016 and above.

https://dnncommunity.org/forums/using-dnn/upgrading-dnn/upgrade-issue-going-from-9-13-9-to-10-0-0/

### Steps to reproduce?

Upgrade in a version less than 2016.  Is in the 10.00.00.SqlDataProvider.

DROP VIEW IF EXISTS {databaseOwner}[{objectQualifier}vw_ApiTokens]

This shoudl be chnaged to to:

IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{objectQualifier}vw_ApiTokens]') AND OBJECTPROPERTY(id, N'IsView') = 1)
BEGIN
    DROP VIEW {databaseOwner}[{objectQualifier}vw_ApiTokens]
END
GO

### Relevant log output

```shell

```

### Affected Versions

10.0.0 (latest v10 release)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

valadas:
As stated in the v10 release notes, the minimal SQL Server version is now 2017

PR #6549, opened 2025-05-15
SQL: DROP IF not supported on all SQL versions #6547

leedavi:
Fixes: #6547
## Summary
Make SQL command backward compatible.

### DIFF
--- a/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/10.00.00.SqlDataProvider
+++ b/DNN Platform/Website/Providers/DataProviders/SqlDataProvider/10.00.00.SqlDataProvider
@@ -416,7 +416,10 @@ ALTER TABLE {databaseOwner}{objectQualifier}ApiTokenKeys ADD CONSTRAINT [FK_{obj
 END
 GO
 
-DROP VIEW IF EXISTS {databaseOwner}[{objectQualifier}vw_ApiTokens]
+IF EXISTS (SELECT * FROM dbo.sysobjects WHERE id = object_id(N'{databaseOwner}[{objectQualifier}vw_ApiTokens]') AND OBJECTPROPERTY(id, N'IsView') = 1)
+BEGIN
+    DROP VIEW {databaseOwner}[{objectQualifier}vw_ApiTokens]
+END
 GO
 
 CREATE VIEW {databaseOwner}[{objectQualifier}vw_ApiTokens]


valadas:
We have bumped the minimal SQL dependency to 2017, so if this works on 2017 I'd rather not rollback to an older syntax.

mitchelsellers:
The current version of DNN supports SQL Server 2017 and later (https://docs.dnncommunity.org/content/getting-started/setup/requirements/index.html)

This syntax has been supported since SQL 2016 and later and provides a cleaner syntax, there is no reason to go backwards in this change given the supported versions.

leedavi:
> The current version of DNN supports SQL Server 2017 and later (https://docs.dnncommunity.org/content/getting-started/setup/requirements/index.html)
> 
> This syntax has been supported since SQL 2016 and later and provides a cleaner syntax, there is no reason to go backwards in this change given the supported versions.

I think using backward compatible script on software that has been installed for over 20 year is normal, unless the script format is deprecated and will disappear?    I see no reason why you should refuse this pull request, it does not break things and saves people thinking there is a problem.  The statement that only SQL Server 2017 does not change.  If anything should happen , it should be a warning that SQL server needs to be upgrade, not an installation error.

If DNN should not use old script format all the existing scripts should be changed.  There is currently only 1 DROP IF statement, it makes sense to change that rather than all others.

mitchelsellers:
My $0.04 here.

It is not practical, or SAFE, for us to update prior scripts to a new format, as that would change the prior behaviors and would not be consistent with the evolution, why would we want a change in DNN 10.x to change a SQL script from say DNN 7.x or 8.x.  Those scripts are really truly just history.

Long term, the technology is continuing to evolve, the new syntax/format which is fully supported offers many benefits.

* It is more succint and easy to understand what is being done
* It is less prone to risk, as you have a SINGLE location for the name rather than two.  Yes slight risk but smaller
* It is more efficient and less statements to parse/process since it is a single atomic action

This behavior, if just like we have with other areas of the platform as we move to newer technology, there will be times and places that not everything will be supported with the new technology.  (Take Dependency Injection support as an example.)  But we are not rolling back contributions that are using the newer technology, which is that this would be doing.

Lastly, this is a PR against a release that has already been made, we CANNOT change this script for the above listed reason as it is already out in the wild and was already used by individuals.

valadas:
For clean installs, there is already a check for sql version, but for upgrades we can't do much as even if we could detect that, the files are already replaced and the user has to restore anyway. So for upgrades, one needs to just read the release notes and/or minimal requirements https://docs.dnncommunity.org/content/getting-started/setup/requirements/index.html.

From a PR validation standpoint, we also set an expectation of what we test and support. Setting our minimum at SQL 2017, everyone is on the same page and large SQL PRs like https://github.com/dnnsoftware/Dnn.Platform/pull/6038 are validated/tested against it and not older versions. In other words, we support what we test and test what we support.

As a final note SQL 2016 has been out of mainstream support for 4 years and will be out of extended support next year. It is a safe assumption that many of the contributors and approvers in this project will have uninstalled that version or do so soon, and that CI runners might also start removing it from their environment's defaults, breaking our CI builds.

leedavi:
OK.  I'm not sure I agree on changing the latest SQL script for a SQL format that can throw an  error and give people work.
But I understand and will accept your answer.
Thank you for listening.

PR #6575, opened 2025-06-05
Resolve build failure when SSMS is installed

bdukes:
## Summary
After installing SQL Server Management Studio (SSMS) 21, I noticed that I started getting errors during the build of the platform. SSMS 21 started using the Visual Studio Installer to manage installations and updates, so it appears that NuGet is finding and using its version of MSBuild instead of Visual Studio's.

This PR uses `vswhere` to find the latest Visual Studio version and pass that to `NuGet.exe`

See also https://github.com/nuget/home/issues/14349


valadas (approved):
Good catch!

valadas:
Looks like CI did not like it though...

bdukes:
Weird. I adjusted it a smidge and it looks to be working.

Issue #6581, opened 2025-06-06
[Enhancement]: Avoid per-folder DB calls in FolderManager.SearchFiles (honour recursive flag)


timi-ty:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of problem

Searching **Global Assets** (or any folder tree) on large portals causes thousands of database round-trips and saturates CPU.

**Repro (customer scale)**  
1. Portal with ≈ 67 000 files in ≈ 2 700 folders.  
2. Open **Global Assets** → search for a single letter (e.g. **“d”**).  
3. Runtime ≈ 60 s, server at 100 % CPU. Logs show ~2 700 calls to `dbo.GetFiles` with `@Recursive = 0`.

**Why it happens**  
`FolderManager.SearchFiles` always calls `GetFiles(folderId, false, false)` and then iterates every sub-folder in C#, producing one DB hit *per folder*. Each individual query is cheap; the **volume** is the bottleneck.

### Description of solution

Refactor **`FolderManager.SearchFiles`** to pass the **recursive** flag straight through to the stored procedure, then filter the single result set in memory while respecting permissions.

    private IEnumerable<IFileInfo> SearchFiles(IFolderInfo folder, Regex regex, bool recursive)
    {
        var files = CBO.Instance.FillCollection<FileInfo>(
            DataProvider.Instance().GetFiles(folder.FolderID, false, recursive));

        if (!recursive)
            return files.Where(f => regex.IsMatch(f.FileName)).Cast<IFileInfo>();

        var allowed = GetFolders(folder.PortalID)
            .Where(f => f.FolderPath.StartsWith(folder.FolderPath) &&
                        FolderPermissionController.Instance.CanViewFolder(f))
            .Select(f => f.FolderPath)
            .ToHashSet();

        return files
            .Where(f => regex.IsMatch(f.FileName) && allowed.Contains(f.Folder))
            .Cast<IFileInfo>();
    }

*No schema or stored-procedure changes required.*

### Description of alternatives considered

| Alternative | Outcome |
|-------------|---------|
| Additional indexes (`IX_Folders_PortalPath`, `IX_Files_PortalID_FolderID_INC`) | ≈ 7 % fewer reads but still ~60 s runtime. |
| Modify the GetFiles SP | Larger impact radius. |

### Observation

It’s likely that the `SearchFiles` function was written this way to enforce **per-folder permission checks**.  
[...]

PR #6582, opened 2025-06-06
refactor: Allow SQL to do recursive search for files in subfolders

timi-ty:
## Summary

Eliminates excessive database load during file searches by making a single recursive call instead of one per sub-folder.

* **Closes:** #6581 
* **Area:** Core – `FolderManager.SearchFiles`  
* **Type:** Performance enhancement (non-breaking)

---

## Problem statement

On portals with deep folder hierarchies every search triggers **N × `dbo.GetFiles`** calls (N = folder count). Though each query costs < 35 ms CPU, thousands executed serially drive SQL Server to 100 % CPU and users wait ~1 minute.

---

## Implementation details

* Propagate the `recursive` flag to `GetFiles`.  
* Consolidate result filtering & permission checks in memory.  
* No DB schema changes, no new indexes, no public API changes.  
* Existing unit tests remain green.

---

## Performance results

| Metric (67 k files / 2 700 folders) | **Before** | **After** |
|------------------------------------|-----------:|----------:|
| Total runtime (search “d”)         | ~60 s | **~2 s** |
| DB calls to `GetFiles`             | ~2 700 | **5** |
| SQL CPU utilisation                | 100 % | **< 15 %** |

---

## Test plan

1. Seed sample portal.  
2. Search “d” in **Global Assets** and record timings/logs.  
3. Run **DotNetNuke.Tests.Core.Providers.Folder** – all pass.  

---

### Evidence

**Before Fix**

https://github.com/user-attachments/assets/f131ad22-f798-4246-b116-448a1bed9ede

**After Fix**

https://github.com/user-attachments/assets/c5ce50cf-c97d-43a4-a831-c351efaf5f5e

### DIFF
--- a/DNN Platform/Library/Services/FileSystem/FolderManager.cs
+++ b/DNN Platform/Library/Services/FileSystem/FolderManager.cs
@@ -2027,7 +2027,7 @@ namespace DotNetNuke.Services.FileSystem
 
             // Simple filter with folder permission lookup
             return fileCollection
-                .Where(f => regex.IsMatch(f.FileName) && allowedFolderPaths.Contains(f.Folder))
+                .Where(f => allowedFolderPaths.Contains(f.Folder) && regex.IsMatch(f.FileName))
                 .Cast<IFileInfo>();
         }
 


mitchelsellers:
@timi-ty Thank you so much for such a detailed submission, we will be sure to get this reviewed as quickly as possible

DNN Platform/Library/Services/FileSystem/FolderManager.cs
```diff
@@ -2009,23 +2009,26 @@ private IFolderInfo GetParentFolder(int portalId, string folderPath)
...
+            // Pre-compute allowed folders once
+            var allowedFolderPaths = this.GetFolders(folder.PortalID)
+                .Where(f => f.FolderPath.StartsWith(folder.FolderPath) &&
+                           FolderPermissionController.Instance.CanViewFolder(f))
+                .Select(f => f.FolderPath)
+                .ToHashSet();
+
+            // Simple filter with folder permission lookup
+            return fileCollection
+                .Where(f => regex.IsMatch(f.FileName) && allowedFolderPaths.Contains(f.Folder))
```
bdukes: My intuition is that the folder check via the hash set is going to be `false` often and that it will be quicker than the regex check.
```suggestion
                .Where(f => allowedFolderPaths.Contains(f.Folder) && regex.IsMatch(f.FileName))
```

DNN Platform/Library/Services/FileSystem/FolderManager.cs
```diff
@@ -2009,23 +2009,26 @@ private IFolderInfo GetParentFolder(int portalId, string folderPath)
...
+            if (!recursive)
             {
-                foreach (var subFolder in this.GetFolders(folder))
-                {
-                    if (FolderPermissionController.Instance.CanViewFolder(subFolder))
-                    {
-                        files.AddRange(this.SearchFiles(subFolder, regex, true));
-                    }
-                }
+                return fileCollection.Where(f => regex.IsMatch(f.FileName)).Cast<IFileInfo>();
```
bdukes, line 2018: This method used to return a `List` and now it returns an `IEnumerable`. In the usages within DNN itself, this appears to be a performance boost, since the two callers do sorting and then materialize the results into a list (thus, this change avoids materializing the list an extra time).

It's possible that 3rd party usages of this method will be less performant if there were reusing the returned collection, but I think the benefit it higher than the drawback.

bdukes (approved):
This is a great contribution, thanks for the robust write up and investigation.

DNN Platform/Library/Services/FileSystem/FolderManager.cs
timi-ty: Yes, yes. Committing this.

valadas:
Ok, I just tested with an Azure folder provider with a lot of subfoldes and file and performance was great. Merci. Thanks @timi-ty

PR #6538, opened 2025-05-02
Introduce IHostSettings to replace Host class

bdukes:
## Summary
In order to resolve usages of the deprecated `HostController.Instance` in `DotNetNuke.Entities.Host.Host`, this PR introduces `DotNetNuke.Abstractions.Application.IHostSettings` and its default implementation `DotNetNuke.Entities.Host.HostSettings`. This included creating a copy of the `PerformanceSettings` and `SchedulerMode` enums inside `DotNetNuke.Abstractions`, as well as extracting `IFileExtensionAllowList` from the `FileExtensionWhitelist` class.

`IHostSettings` differs from the old `Host` class in a few ways.
1. All of the SMTP properties take a `portalId` parameter, rather than implicitly using the current portal context.
2. Some of the return types are changed from `int` to `TimeSpan` to make date math easier and less error prone.
3. Casing has been normalized (e.g. `Guid` instead of `GUID`, `HostPortalId` instead of `HostPortalID`)

PR #6602, opened 2025-06-20
Remove implicit dependencies in PortalController

bdukes:
## Summary
This PR removes usages of `Instance` in `PortalController`


bdukes:
Build is green now (apparently there were a couple of projects that didn't include all of the warnings when building in the Debug configuration, so I didn't see the errors locally)

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -265,60 +303,120 @@ public static void DeletePortalFolder(string serverPath, string portalFolder)
...
         /// <summary>GetPortalsByName gets all the portals whose name matches a provided filter expression.</summary>
         /// <param name="nameToMatch">The email address to use to find a match.</param>
         /// <param name="pageIndex">The page of records to return.</param>
         /// <param name="pageSize">The size of the page.</param>
-        /// <param name="totalRecords">The total no of records that satisfy the criteria.</param>
-        /// <returns>An ArrayList of PortalInfo objects.</returns>
+        /// <param name="totalRecords">The total number of records that satisfy the criteria.</param>
+        /// <returns>An <see cref="ArrayList"/> of <see cref="PortalInfo"/> objects.</returns>
         public static ArrayList GetPortalsByName(string nameToMatch, int pageIndex, int pageSize, ref int totalRecords)
+            => GetPortalsByName(Globals.GetCurrentServiceProvider().GetRequiredService<DataProvider>(), nameToMatch, pageIndex, pageSize, ref totalRecords);
```
valadas, line 333: Should we deprecated this one in favor of the one that takes DataProvider ?

DNN Platform/Library/Entities/Portals/PortalController.cs
```diff
@@ -396,9 +538,28 @@ public static void DeletePortalSettings(int portalID)
         /// <param name="portalID">The portal ID.</param>
         /// <param name="cultureCode">The culture code or <see cref="Null.NullString"/>.</param>
         public static void DeletePortalSettings(int portalID, string cultureCode)
-        {
-            DataProvider.Instance().DeletePortalSettings(portalID, cultureCode);
-            EventLogController.Instance.AddLog("PortalID", portalID.ToString() + ((cultureCode == Null.NullString) ? string.Empty : " (" + cultureCode + ")"), GetCurrentPortalSettingsInternal(), UserController.Instance.GetCurrentUserInfo().UserID, EventLogController.EventLogType.PORTAL_SETTING_DELETED);
+            => DeletePortalSettings(
```
valadas, line 554: Should we deprecate the old overload ?

DNN Platform/Library/Entities/Portals/PortalController.cs
valadas, line 1: Is there a reason to do so much method injection instead of just constructor injection in this class ?

bdukes: Doing method injection because these are static methods (though I did notice a couple of times that I had done method injection and then saw that it wasn't a static method, so if you notice places where it could use fields instead of arguments, call those out and I'll get them switched back).

DNN Platform/Library/Entities/Portals/PortalController.cs
bdukes: Yes, I've pushed this change. Thanks!

DNN Platform/Library/Entities/Portals/PortalController.cs
bdukes: Yes, I've pushed this change. Thanks!

DNN Platform/Library/Entities/Portals/PortalController.cs
valadas: Oh, makes sense, I'll continue my review tonight and call out what I can spot. Thanks for the info.

mitchelsellers (approved):

DNN Platform/Library/Entities/Host/MailSettings.cs
```diff
@@ -13,7 +13,7 @@ namespace DotNetNuke.Entities.Host;
 /// <summary>The default <see cref="IMailSettings"/> implementation.</summary>
 /// <param name="hostSettingsService">The host settings service.</param>
 /// <param name="hostSettings">The host settings.</param>
-public class MailSettings(IHostSettingsService hostSettingsService, IHostSettings hostSettings) : IMailSettings
+public class MailSettings(IHostSettingsService hostSettingsService, IHostSettings hostSettings, IPortalController portalController) : IMailSettings
```
valadas, line 16: Do we need to add an overload with the original parameters since this class is public ?

valadas (changes requested):
Just one little knit

DNN Platform/Library/Entities/Host/MailSettings.cs
bdukes: This class was added in https://github.com/dnnsoftware/Dnn.Platform/pull/6538, which also targeted 10.0.2, so there shouldn't be a need for an additional overload.

valadas: Perfect, thanks.

valadas (approved):
Looks good to me

PR #6616, opened 2025-07-08
Add RTL (Right-To-Left) language support for Persian localization

mnouraei:
This PR adds basic RTL language support to ensure proper layout and localization for Persian (fa-IR) users:

Sets Persian culture (fa-IR) to enable correct localization (e.g. for 404 pages).

Adds the rtl CSS class to the <body> tag when the current culture is right-to-left to handle RTL layouts properly.

These changes improve usability and layout rendering for RTL languages in DNN.

Notes:
Tested with Persian culture to ensure layout direction and localization are applied correctly.

### DIFF
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -209,17 +209,19 @@ namespace DotNetNuke.Framework
         {
             base.OnInit(e);
 
-            // Set Persian culture to support localization when culture is fa-IR (e.g. for 404 pages)
-            if (this.PortalSettings.CultureCode == "fa-IR")
-            {
-                var newCulture = Services.Localization.Persian.PersianController.GetPersianCultureInfo();
-                System.Threading.Thread.CurrentThread.CurrentUICulture = newCulture;
-            }
-
             // Add 'rtl' class to body for right-to-left language support
             if (CultureInfo.CurrentUICulture.TextInfo.IsRightToLeft)
             {
-                this.Body.Attributes.Add("class", "rtl ");
+                string existingClass = this.Body.Attributes["class"];
+
+                if (string.IsNullOrEmpty(existingClass))
+                {
+                    this.Body.Attributes.Add("class", "rtl ");
+                }
+                else if (!existingClass.Contains("rtl"))
+                {
+                    this.Body.Attributes["class"] = existingClass + " rtl ";
+                }
             }
 
             // set global page settings
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -209,6 +209,19 @@ namespace DotNetNuke.Framework
         {
             base.OnInit(e);
 
+            // Set Persian culture to support localization when culture is fa-IR (e.g. for 404 pages)
+            if (this.PortalSettings.CultureCode == "fa-IR")
+            {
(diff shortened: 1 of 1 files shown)


DNN Platform/Website/Default.aspx.cs
```diff
@@ -209,6 +209,19 @@ protected override void OnInit(EventArgs e)
...
+                var newCulture = Services.Localization.Persian.PersianController.GetPersianCultureInfo();
+                System.Threading.Thread.CurrentThread.CurrentUICulture = newCulture;
+            }
+
+            // Add 'rtl' class to body for right-to-left language support
+            if (CultureInfo.CurrentUICulture.TextInfo.IsRightToLeft)
+            {
+                this.Body.Attributes.Add("class", "rtl ");
+            }
+
```
valadas: I would prefer a solution where it is not as specific to a single culture.
Like `CultureInfo.CurrentUICulture.TextInfo.IsRightToLeft`

`this.Body.Attributes.Add` will replace the attribute, we need to get its current value, append the class and then .Add the attribute again.

valadas (changes requested):

DNN Platform/Website/Default.aspx.cs
mnouraei: Thank you for the feedback. 👍

You're right – we can update the code as below to ensure that the class attribute is not overwritten and rtl is only added if it's not already included:

```
if (CultureInfo.CurrentUICulture.TextInfo.IsRightToLeft)
{
    string existingClass = this.Body.Attributes["class"];

    if (string.IsNullOrEmpty(existingClass))
    {
        this.Body.Attributes.Add("class", "rtl ");
    }
    else if (!existingClass.Contains("rtl"))
    {
        this.Body.Attributes["class"] = existingClass + " rtl ";
    }
}
```
Please let me know if you prefer a utility method for class handling to keep it cleaner.

If you agree with this change, I can update the PR accordingly.

mnouraei: When the page hits a 404 error, the system's default culture falls back to en-US.
The solution I found was to check for "fa-IR" and explicitly set the Persian culture like this:

```
if (this.PortalSettings.CultureCode == "fa-IR")
{
    var newCulture = Services.Localization.Persian.PersianController.GetPersianCultureInfo();
    System.Threading.Thread.CurrentThread.CurrentUICulture = newCulture;
}
```
Do you have a better approach to ensure the correct culture is set for 404 pages?

valadas: I think (but have not tried it) this line of the web.config defines the fallback language for when DNN can't know the language (which may be the case for 404 pages).

`<globalization culture="en-US" uiCulture="en" requestEncoding="UTF-8" responseEncoding="UTF-8" fileEncoding="UTF-8" />`

If that is not the case then I guess we would need to figure out for the 404 page. If we know the portal info in that context we could use the portal primary language (which could be a different PR maybe.

Your first change looks good to me if you can adjust the PR, I am fine with the described changes there.

mnouraei: Alright, I will create a new PR for the rtl class

mnouraei:
A new Pull Request has been created at:
[6617](https://github.com/dnnsoftware/Dnn.Platform/pull/6617)

Therefore, I am closing this PR.

PR #6617, opened 2025-07-08
Add 'rtl' class to body for right-to-left language support

mnouraei:
This PR adds basic RTL language support by updating the <body> tag classes when the current UI culture is right-to-left:

Checks if the current UI culture is RTL.

If the ```<body>``` tag has no existing class, it adds ```"rtl"```.

If it already has classes but not ```"rtl"```, it appends ```"rtl"``` to ensure proper RTL layout rendering.

These changes improve layout direction handling for RTL languages such as Persian (fa-IR).

Notes:
Tested with Persian culture to confirm that the RTL class is applied correctly for proper layout rendering.

### DIFF
--- a/DNN Platform/Website/Default.aspx.cs
+++ b/DNN Platform/Website/Default.aspx.cs
@@ -209,6 +209,21 @@ namespace DotNetNuke.Framework
         {
             base.OnInit(e);
 
+            // Add 'rtl' class to body for right-to-left language support
+            if (CultureInfo.CurrentUICulture.TextInfo.IsRightToLeft)
+            {
+                string existingClass = this.Body.Attributes["class"];
+
+                if (string.IsNullOrEmpty(existingClass))
+                {
+                    this.Body.Attributes.Add("class", "rtl ");
+                }
+                else if (!existingClass.Contains("rtl"))
+                {
+                    this.Body.Attributes["class"] = existingClass + " rtl ";
+                }
+            }
+
             // set global page settings
             this.InitializePage();
 


valadas:
Also (I am no expert on rtl languages) but it looks like the standard would be to add the dir attribute to the html element... Maybe we could stick to that?

mnouraei:
> Also (I am no expert on rtl languages) but it looks like the standard would be to add the dir attribute to the html element... Maybe we could stick to that?

I have been localizing this repository for Persian users for the past 6 years in a fork based on this project. During these years, I have gained extensive experience with DNN, and I intend to gradually contribute these experiences to the core repository with the approval of the DNN team.

The approach I used in this PR is very reliable and has been tested with over 5,000 downloads.

You can find my fork here: [Persian-DnnSoftware/Dnn.Platform](https://github.com/Persian-DnnSoftware/Dnn.Platform)

Mostafa-Moafi:
@valadas Hi Valadas, I would also be interested if DNN supported RTL, but I think we need a full PR that handles this feature.
What do you think?

mnouraei:
Thanks, @Mostafa-Moafi . As a first step, I just wanted to enable this feature in DNN so that we can apply CSS changes and other adjustments in the next phases. Making too many changes at once might not get initial approval.

bdukes (approved):

bdukes:
I see you added `@microsoft-github-policy-service agree` to the issue description, I believe the CLA bot needs that as a comment, rather than in the description.

azarbara:
Such a helpful improvement—thanks so much!

valadas (approved):
Awesome!

Behnam-Emamian:
Thanks everyone! Adding right-to-left support is a great enhancement that helps accommodate all RTL languages such as Persian, Arabic, Hebrew, and more.

mnouraei:
```
```

PR #6620, opened 2025-07-10
Clean up legacy code

donker:
It seems we are not doing multitargeting and the NETFRAMEWORK directive can sometimes lead to code being erroneously marked as absent when it is not.

This also begs the question whether we really need this library in the project and if we should move the code back in, IMHO.


bdukes:
I agree, I don't see the utility of this being in its own project.

mitchelsellers (approved):

PR #6625, opened 2025-07-17
Use MSBuild instead of `nuget.exe` to restore packages

bdukes:
## Summary
This is a follow-up to #6575. In the related NuGet issue, it was indicated that using `nuget.exe restore` could have other issues, and `msbuild --target restore --property:RestorePackagesConfig=true` was a preferred approach. This PR implements that change, and removes the old `NuGet.targets` approach.

See https://github.com/NuGet/Home/issues/14349#issuecomment-3010732878


valadas (approved):
Yay, much simpler

PR #6637, opened 2025-07-29
Bumped versions to 10.1.0

valadas:
Since we have a 10.1.0 sql script now, we need builds to produce that version in order to do any testing. This commits the version so that developers don't have a gazilion changes when doing a fix locally.


bdukes (approved):

valadas:

PR #6638, opened 2025-07-29
Update password hash algorithm to SHA256

### LINKED ISSUE #6614: [Enhancement]: Change password hash from SHA1 to SHA256

### Description of problem

DNN uses the SqlMembershipProvider with hashed passwords. Passwords are encrypted one-way using the SHA1 hashing algorithm. Passwords are hashed including an additional PasswordSalt (16 characters), which is random for every user.

> NIST formally **deprecated** use of SHA-1 in **2011** and disallowed its use for digital signatures in 2013, and declared that it should be phased out by 2030

Source: [Wikipedia](https://en.wikipedia.org/wiki/SHA-1)

> “Modules that still use SHA-1 after 2030 will not be permitted for purchase by the federal government,”

Source: [NIST Retires SHA-1 Cryptographic Algorithm](https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm) (December 15, 2022):

### Description of solution

Change Hash Algorithm from SHA1 to SHA256.

This requires a few changes, especially in:
1. [web.config](https://learn.microsoft.com/en-us/dotnet/api/system.web.security.membershippasswordformat?view=netframework-4.8.1#examples): `<membership ... hashAlgorithmType="SHA256">`
2. DNN [MembershipPasswordController](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Entities/Users/Membership/MembershipPasswordController.cs)
3. Backward compatibility for Passwords and PasswordHistory, which currently have everything in SHA1

_No response_

### Anything else?

- [Password encryption used by DNN](https://dnnsupport.dnnsoftware.com/article/53528-password-encryption-used-by-dnn) (DNN Corp. Support)
- [Hashed Passwords](https://www.dnnsoftware.com/wiki/hashed-passwords) (DNN Community Wiki)

- [ ] Yes

- [ ] Yes

### Code of Conduct

bdukes:
This PR updates the default hash algorithm to SHA256. It does _not_ add any logic for migrating users from SHA1 hashes, it only affects new installations.

I have tested that password history works as expected.

Fixes #6614

### DIFF
--- a/DNN Platform/Website/development.config
+++ b/DNN Platform/Website/development.config
@@ -207,7 +207,7 @@
           domain="[domain]"                                   Enables output of the "domain" cookie attribute set to the specified value
         -->
     <anonymousIdentification enabled="true" cookieName=".ASPXANONYMOUS" cookieTimeout="100000" cookiePath="/" cookieRequireSSL="false" cookieSlidingExpiration="true" cookieProtection="None" domain=""/>
-    <membership defaultProvider="AspNetSqlMembershipProvider" userIsOnlineTimeWindow="15">
+    <membership defaultProvider="AspNetSqlMembershipProvider" userIsOnlineTimeWindow="15" hashAlgorithmType="SHA256">
       <providers>
         <clear/>
         <!-- Configuration for AspNetSqlMembershipProvider:
--- a/DNN Platform/Website/release.config
+++ b/DNN Platform/Website/release.config
@@ -208,7 +208,7 @@
           domain="[domain]"                                   Enables output of the "domain" cookie attribute set to the specified value
         -->
     <anonymousIdentification enabled="true" cookieName=".ASPXANONYMOUS" cookieTimeout="100000" cookiePath="/" cookieRequireSSL="false" cookieSlidingExpiration="true" cookieProtection="None" domain=""/>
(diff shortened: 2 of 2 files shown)


r90727:
Wouldn't this invalidate all existing passwords? Am I missing something? Or this is only for new installs?

valadas:
What we discussed is that would be for new installs only. But we would document how to migrate to it for upgrades. We need to test the process but if I remember correctly you just have run an sql script to force password reset upon next login.

r90727:
@valadas ok, deleted my comment as I didn't notice the "new installs only" remark. And yes, password reset is the only way foward for existing users.

valadas:
@r90727 you got me scratching my head a bit there, I was replying to a ghost lol...

mitchelsellers:
@valadas I thought we had targeted included this one in the RC, did I mis-remember?

valadas (approved):

valadas:
Ok, what do you guys think of this documentation here https://github.com/DNNCommunity/DNNDocs/pull/766
I approved this PR and I think since it only affects new installs, we should do it in the next minor, agreed?

bdukes:
Yes, this should be in 10.2.0

PR #6652, opened 2025-08-07
Enhance Login Module Detection Flexibility (Multiple Keyword Support)

uzmannazari:
Add support for multiple keywords in login module detection

## Summary
The original ValidateModuleInTab method only checked for a specific module name (such as "Account Login"). This was not flexible enough for scenarios where the login module might have a different name (like "User Login", "Authentication", "Account Auth", etc.), leading to incorrect login page validation.

## Solution

- Refactored ValidateModuleInTab to accept an array of keywords and return true if any of these keywords is found (case-insensitive) in any module's name on the given tab.
- Added an overload for backward compatibility to accept a single module name as before.
- Updated ValidateLoginTabID to use an array of relevant keywords (login, auth, account) for more robust login module detection.
- Added XML documentation for all updated methods.

## Benefits

- More flexible and robust detection of login modules, supporting custom and varied naming.
- No longer strictly tied to a specific module name.
- Backward compatibility preserved.

### DIFF
--- a/DNN Platform/Library/Common/Globals.cs
+++ b/DNN Platform/Library/Common/Globals.cs
@@ -3156,7 +3156,37 @@ namespace DotNetNuke.Common
         /// <returns><see langword="true"/> if the tab contains "Account Login" module, otherwise, <see langword="false"/>.</returns>
         public static bool ValidateLoginTabID(int tabId)
         {
-            return ValidateModuleInTab(tabId, "Account Login");
+            return ValidateModuleInTab(tabId, new[] { "login", "auth", "account" });
+        }
+
+        /// <summary>
+        /// Check whether the tab contains any module whose name contains any of the specified keywords.
+        /// </summary>
+        /// <param name="tabId">The tab id.</param>
+        /// <param name="moduleKeywords">
+        /// The array of keywords to check in the module names (case-insensitive, e.g., "login", "auth").
+        /// </param>
+        /// <returns>
+        /// <c>true</c> if the tab contains at least one module whose name includes any keyword; otherwise, <c>false</c>.
+        /// </returns>
+        public static bool ValidateModuleInTab(int tabId, string[] moduleKeywords)
+        {
+            bool hasModule = Null.NullBoolean;
+            foreach (ModuleInfo objModule in ModuleController.Instance.GetTabModules(tabId).Values)
+            {
+                var name = objModule.ModuleDefinition.FriendlyName?.ToLower() ?? string.Empty;
+                if (moduleKeywords.Any(keyword => name.Contains(keyword.ToLower())))
+                {
+                    TabInfo tab = TabController.Instance.GetTab(tabId, objModule.PortalID, false);
+                    if (TabPermissionController.CanViewPage(tab))
+                    {
+                        hasModule = true;
+                        break;
+                    }
+                }
+            }
+
+            return hasModule;
         }
 
         /// <summary>Check whether the tab contains specific module.</summary>


anajteam (approved):
This change is very useful for us who have programmed a separate login module. I agree.

shahubofficial (approved):
I approve and support this enhancement.

DNN Platform/Library/Common/Globals.cs
```diff
@@ -3156,7 +3156,37 @@ public static string GetOnLineHelp(IHostSettings hostSettings, string helpUrl)
         /// <returns><see langword="true"/> if the tab contains "Account Login" module, otherwise, <see langword="false"/>.</returns>
         public static bool ValidateLoginTabID(int tabId)
         {
-            return ValidateModuleInTab(tabId, "Account Login");
```
mitchelsellers, line 3159: This change would be a very large, and breaking change to the platform and has a high likelyhood of breaking the DNN installation resulting in an inability for a user to login to the system.

The "Account Login" module is the core DNN module that handles ALL authentication providers.

Custom providers that utlize the standard process ALL expose themselves through this module, and this protection helps to prevent someone from creating a login page that is not usable because it doesn't have a login module.

If you are using a non-provider based authentication method, you have to do a two step method to make that page the login page, but it is there for protection.  

I'm not sure that I am ok with this change, and will defer to the other @dnnsoftware/approvers for additional comments.

uzmannazari: Thank you for your feedback and the important explanation about the role of the “Account Login” module.

Just to clarify: the original method (ValidateModuleInTab with a single module name as a parameter) remains completely unchanged and fully functional. We have only added an overload that allows checking for multiple keywords at once. The core logic and the strict check for the “Account Login” module are both preserved.

This change is fully backward compatible and purely additive—it does not affect any default DNN authentication behavior or the protections you mentioned. All standard authentication mechanisms and security remain unchanged.
The new flexibility is only for cases where a site uses a custom or localized login module name, but still relies on the standard provider-based mechanism and needs to be detected for login page validation.

If there are still any concerns or suggestions from the DNN community, I am happy to discuss further or provide more details.

Thank you again for your careful review!

bdukes: The concern is in this line, where `ValidateLoginTabID` is changed to call the new API instead of the current one.

uzmannazari: You’re right—ValidateLoginTabID now checks for multiple keywords instead of only "Account Login". The intention is to add flexibility for custom and localized login modules, while still supporting the original core module.

If there’s a concern about maintaining strict default behavior for standard installations, we can easily adjust the implementation to first check for the exact "Account Login" module as before, and only fall back to the keyword-based approach if that isn’t found.

Happy to discuss or update the logic to better address platform security and compatibility requirements.

uzmannazari: @bdukes Just a quick note — I think we may be over-complicating things a bit.
This method is ultimately just a helper to detect which page is set as the login page, but the login page itself is explicitly configured by the site admin in the site settings.
If the admin chooses to designate a different page as the login page, even if it doesn't contain the standard "Account Login" module, that's their deliberate configuration. This flexibility is sometimes necessary for custom scenarios.

So, while platform security is important, this method mainly exists to help DNN recognize the intended login page — the final decision is always up to the site administrator.

mitchelsellers: @uzmannazari I think you are missing the overall issue here.

This check is in place as there is only 1 known supported module that guarantees that you can login and that is the Account Login module.

If we change to allow an arbitrary check, based on names it is highly likely that a user, without knowing what they are doing could change the login page to a page without any method to login.

If they save this, and log out, they are done.  They have no way to get back into the site without going to the database, making a change, and then restarting the site. and that is only possible if they have the tools and skills to do this in the DB and know what values need to be adjusted.

I think the risk here far outweights the benefits.

uzmannazari: Thank you for your input and for highlighting the security implications.

What we are aiming for here is to offer a cleaner and more flexible solution for scenarios where a site admin wants to add a custom or third-party login module on a page and set that page as the login page in DNN.
Right now, as a workaround, the admin has to add the standard DNN Account Login module to the page (even if it’s hidden or not used) just to pass the check. This is not ideal, and can even lead to situations where the module is present on the page but hidden from all users, defeating the purpose of the current check.

Our intention is to support these valid advanced scenarios more efficiently and transparently—not to weaken the default security or increase risk for regular users.
Of course, I agree that protecting users from accidental lockouts is crucial. If there is a way to balance both needs (such as an advanced setting, or a warning when changing the login page), I would be happy to explore and implement that approach.

Thanks again for your thoughtful review and discussion.

mtrutledge:
Custom modules should have a login page where the CS class inherits from
'AuthenticationLoginBase'. When the module is installed the type in the
manifest.dnn file must be 'Auth_System'. This allows the page that your
custom module is added to, to show as an optional login page in the Site
settings. You dont need to add the built in account login module to the
page for this to work.

uzmannazari:
@mtrutledge 
Thank you for clarifying that point about custom authentication modules and the manifest settings.

However, my change is specifically related to how the platform validates and allows a page to be selected as the login page.
When an admin chooses a login page in site settings, the validation process is directly tied to the ValidateLoginTabID method—which is exactly the method being updated here.

I also tested the scenario you mentioned. In practice, the list of pages available for selection as the login page in the Site Settings UI is determined by whether ValidateLoginTabID returns true.
You can see this directly in the PersonaBar code, where it sets the validation criteria with:
`let TabParameters_Login = Object.assign(Object.assign({}, TabParameters), { disabledNotSelectable: false, validateTab: "Account Login" });
`
So even if a custom module is implemented correctly, it will only appear in the login page selection list if it passes this validation check.

My update is intended to make this process more flexible for advanced use cases, but it does not affect how DNN manages custom authentication modules according to the standards you described.

uzmannazari:
@mtrutledge I have also tested the approach you described with custom authentication modules and the manifest settings. However, this does not affect whether a page appears in the list of selectable login pages in the Site Settings. The selection logic still strictly relies on the module name being "Account Login," so only pages containing a module with this exact name are shown in the list, regardless of the underlying authentication provider implementation.

mtrutledge:
You are not enabling your Authentication System correctly. Once you install
your modules with the Type Authentication System you must go into the
Settings->Extensions. Select Autherntication System and enable it (Click
the edit pencil, then go to the Site Settings tab). Then you disable the
Default Aiuthentication System. That is what enables your page to show on
the site settings and how custom logins are controlled. You still add the
base module to the page, but you enable your authentication system. This is
how I do all of my custom registrations and logins and has always worked
perfectly.

uzmannazari:
@mtrutledge Thank you for sharing your experience and the correct process for setting up a custom Authentication System provider in DNN.

Just to clarify, in this case, we are not creating a separate authentication provider or custom authentication system. We are still relying on the core DNN authentication process—our scenario involves implementing the login UI and flow in a separate custom module, but it uses the same standard DNN authentication logic and provider.

This means the module isn't registered as a new Authentication System in Extensions, but rather as a regular module that simply presents the login interface differently, while still using DNN's built-in authentication mechanism.

Because of this, the login page selection in Site Settings is still strictly tied to having the "Account Login" module on the page, which is the limitation we are trying to address for more flexibility.

Thank you again for the discussion and for sharing your insights.

mtrutledge:
But you need to implement it in the structure I outlined. Once you are in
your custom authentication system you can do whatever you want, make it
look however you want, and call out to whatever you want. You dont have to
use the built-in DNN methods in your custom authentication system. But for
it to show properly it needs to be done with an Authentication System
extension.  Although, in your case you may also want to implement a custom
IMembershipProvider. That would tell DNN to call out to your custom
provider when a login or registration occurs.

david-poindexter:
I highly recommend using the built-in structure within DNN for doing custom authentication providers.  I would not be in favor of this change.  That said, we could certainly do a better job in documenting the way it should be done.

PR #6706, opened 2025-09-16
Deny anonymous users access to CKE endpoints

bdukes:
## Summary
As a defense-in-depth strategy, we realized that only in very exceptional circumstances does there need to be anonymous access to upload files via the CKEditor. This PR adds a `web.config` level denial of access to the CKEditor upload endpoints for anonymous users.

Thanks @r90727 for calling this out and the initial implementation.


david-poindexter (approved):

donker (approved):
Awesome

PR #6707, opened 2025-09-16
Fix cacheability settings issue by fixing inconsistencies

### LINKED ISSUE #6703: [Bug]: Output Caching UI completely wrong in DNN 10

### What happened?

We're working on Output Caching - specifically for Cloudflare.
A feature which used to work in DNN 9 is now broken in DNN 10. 

1. The DB has value 2 stored
2. But the backend delivers `4`
3. and the UI says it doesn't know what to show.

Image

Image

Trying to save the correct value probably works:

Image

But after that the UI UI never gets the memo - it again receives a `4` from the backend and believes 'not configured'

Image

### Steps to reproduce?

1. Go to performance settings
2. try to set the value - to `Public`
3. reload - it should end up in the DB, but the UI doesn't know it. 

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.1.0 (latest v10 release)

_No response_

### Code of Conduct

donker:
The framework uses three places to deserialize and deserialize the cacheability settings. Unfortunately the enumerations differ between these places. This, plus the fact that the UI was receiving these options in the wrong format, has meant these settings would not be used correctly.

This should fix #6703

### DIFF
--- a/DNN Platform/DotNetNuke.Abstractions/Application/CacheControlHeader.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Application/CacheControlHeader.cs
@@ -22,11 +22,11 @@ public enum CacheControlHeader
     Server = 3,
 
     /// <summary>Applies the settings of both <see cref="Server" /> and <see cref="NoCache" /> to indicate that the content is cached at the server but all others are explicitly denied the ability to cache the response.</summary>
-    ServerAndNoCache = 3,
+    ServerAndNoCache = 4,
 
     /// <summary>Sets <c>Cache-Control: public</c> to specify that the response is cacheable by clients and shared (proxy) caches.</summary>
-    Public = 4,
+    Public = 5,
 
     /// <summary>Indicates that the response is cached at the server and at the client but nowhere else. Proxy servers are not allowed to cache the response.</summary>
-    ServerAndPrivate = 5,
+    ServerAndPrivate = 6,
 }
--- a/DNN Platform/Library/Entities/Host/HostSettings.cs
+++ b/DNN Platform/Library/Entities/Host/HostSettings.cs
@@ -404,12 +404,12 @@ public class HostSettings(IHostSettingsService hostSettingsService) : IHostSetti
     private static CacheControlHeader ToCacheControlHeader(string headerId)
         => headerId switch
         {
-            "0" => CacheControlHeader.NoCache,
-            "1" => CacheControlHeader.Private,
-            "2" => CacheControlHeader.Public,
+            "1" => CacheControlHeader.NoCache,
+            "2" => CacheControlHeader.Private,
             "3" => CacheControlHeader.Server,
             "4" => CacheControlHeader.ServerAndNoCache,
-            "5" => CacheControlHeader.ServerAndPrivate,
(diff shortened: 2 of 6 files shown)


mitchelsellers (approved):
You rock @donker  Thank you!

DNN Platform/DotNetNuke.Abstractions/Application/CacheControlHeader.cs
```diff
@@ -22,11 +22,11 @@ public enum CacheControlHeader
     Server = 3,
 
     /// <summary>Applies the settings of both <see cref="Server" /> and <see cref="NoCache" /> to indicate that the content is cached at the server but all others are explicitly denied the ability to cache the response.</summary>
-    ServerAndNoCache = 3,
+    ServerAndNoCache = 4,
 
     /// <summary>Sets <c>Cache-Control: public</c> to specify that the response is cacheable by clients and shared (proxy) caches.</summary>
```
bdukes, line 27: If we're not going to use the values from [System.Web.HttpCacheability](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpcacheability), we will need to update the mapping in `Default.aspx`:
https://github.com/dnnsoftware/Dnn.Platform/blob/437a2488f706d99d835f040b3e60a03a6351a024/DNN%20Platform/Website/Default.aspx.cs#L442-L459

donker: Good call. I will tweak the code. I took the enum we had in the framework as leading. We should probably use this. Also: why didn't we use this to begin with? Weird.

donker: OK, maybe I am misunderstanding the HttpCacheability Enum. Why two labels for the same value (3)? This was the reason that the value wouldn't properly deserialize out of the DB.

bdukes: 🤷🏻‍♂️ I can't find reasoning, but the docs state that `Server` is the same as `ServerAndNoCache`

donker:
> I think this introduces a breaking change and probably needs to go in a minor release. I understand it was broken prior though, so...thoughts?

I used the values from the UI dropdown and in SQL changed them to what they would be in the future. This is because users, when setting these values, would see the dropdown and that was actually stored correctly (BTW the value is stored as the number, not the text).

There is an alternative approach which is to ditch storing the cacheability as number and go to text. Then we get rid of the problem of the two 3's in the enum.

bdukes:
The breaking change would be if someone had code referencing the enum in DNN 10.0.0, it wouldn't work as expected when run on a newer version of DNN. The behavior for end users should be fine.

mitchelsellers:
I think it is important to get a proper fix out, but agree it is a slight breaking change, mostly low risk as it would require someone compiled against DNN 10.0.0 or later to have the issue and be using that enum

donker:
OK, I'll see if I can come up with something tomorrow morning. At least we know what the problem is.

PR #6711, opened 2025-09-17
Fix cacheability settings

### LINKED ISSUE #6703: [Bug]: Output Caching UI completely wrong in DNN 10

donker:
The mechanism to get and update the cacheability settings had a bug which prevented them from properly getting used.

This PR should fix #6703

### DIFF
--- a/DNN Platform/DotNetNuke.Abstractions/Application/CacheControlHeader.cs
+++ b/DNN Platform/DotNetNuke.Abstractions/Application/CacheControlHeader.cs
@@ -18,10 +18,7 @@ public enum CacheControlHeader
     /// <summary>Default value. Sets <c>Cache-Control: private</c> to specify that the response is cacheable only on the client and not by shared (proxy server) caches.</summary>
     Private = 2,
 
-    /// <summary>Specifies that the response is cached only at the origin server. Similar to the <see cref="NoCache" /> option. Clients receive a <c>Cache-Control: no-cache</c> directive but the document is cached on the origin server. Equivalent to <see cref="ServerAndNoCache" />.</summary>
-    Server = 3,
-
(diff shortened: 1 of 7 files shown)


bdukes (approved):
works for me

valadas:
I tested:

- Clean v9.13.9 (ServerAndNoCache was default) it migrated fine and kept ServerAndNoCache for the value.
- Clean v10.1.0 install (default to 4 which meant ServerAndNoCache but UI was broken), upgraded to this build, migrated fine to ServerAndNoCache and UI is fixed.
- Clean install of this build, works fine but defaults to 4 in the database, so when visiting the persona bar, nothing appears to be selected. I think we need to hunt down where that out-of-the-box default gets set and have a current value on a new install for this version or future ones... I'll see if I can find that...

valadas:
This commit changes the default to match the new way to store them https://github.com/dnnsoftware/Dnn.Platform/pull/6711/commits/29eae341822d52d368054d4436040ce5becc713e

valadas (approved):
Ok, I tested the default now works with last commit. Merging

valadas:
Oh, @bdukes was faster than me ;)

bdukes:
> Oh, @bdukes was faster than me ;)

The magic of auto-merge 🪄

Issue #6643, opened 2025-07-31
[Enhancement]: Pivot away from ClientDependencyFramework


donker:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of problem

The Client Dependency Framework that DNN uses is a fork of [the original project](https://github.com/Shazwazza/ClientDependency). This project is marked as deprecated and has not had any maintenance for 4 years.

### Description of solution

I see one of the following options to get us out of this:

1. Maintain CDF ourselves in case issues arise (current situation)
2. Look for an alternative (there is a more modern product created by the original maker called Smidge)
3. Roll our own

I'd like to propose and expand on 3. The CDF takes care of:

1. Keeping a list of JS and CSS files to include.
2. Managing where these resources are injected (top or bottom of page)
3. Ordering these resources
4. Adding a version nr to aid in caching and upgrades
5. Minification and compression (optional)
6. Bundling (optional)

By default we only use the first 4. 5 and 6 are optional. 6 has become obsolete with the adoption of HTTP/2. And I'd like to make the case that 5 should really be a concern of the developer and not the framework. In today's world, most developers no longer write plain JS code that is sent to the browser as is. Instead they use frameworks and tools to transpile their work before it goes live.

Without the need to minify or bundle the question arises why we don't do our own CDF with the aforementioned features. It will greatly simplify matters.

### Description of alternatives considered

See text above

### Do you plan to contribute code for this enhancement?

- [x] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

valadas:
I totally agree that 5 and 6 is the job of the developer + with http2 and protocol level compression is often of very minimal impact.

jeremy-farrance:
Attribute and capability wise, both [LINK](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link) and [SCRIPT](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script) have gotten feature rich over time and I suspect they will continue to evolve further in the years to come. 
 
If something new is developed, keep it simple. One suggestion I have from personal experience is to allow the developer to completely describe the resulting HTML while still retaining 1-4 and the ability eliminate dupes from multiple module instances per page.

[...]

PR #6729, opened 2025-09-29
Replace CDF

donker:
This PR will move DNN away from the ClientDependencyCore project and replace this with our own (simplified) logic for managing client dependencies.

### Motivation
We do this for two main reasons:

1. The CDF has gone stale and all immediate potential replacements have also gone stale
2. The CDF will not allow us to include an MVC pipeline and hinders further development of DNN

### Overview of Changes
The main changes are in the System.Web.Client library that managed most interaction with the CDF. The central class ClientDependencyManager now reroutes calls to the new logic and has been marked as deprecated. The new logic is concentrated in System.Web.Client.ResourceManagement and consists of an injectable IClientResourceController. This controller is used as follows to add a resource:

``` cs
    this.clientResourcesController
        .CreateScript()
        .FromSrc("~/Resources/Search/Search.js")
        .SetPriority(FileOrder.Js.DefaultPriority)
        .SetProvider(ClientResourceProviders.DnnFormBottomProvider)
        .Register();
```

It is important to call ```Register()``` at the end to add the resource. Various extension methods allow you to specify specific attributes to be added for the resource such as async and defer. There are also shorthand extensions in the DotNetNukelIbrary project to register a script with defaults: ```this.clientResourcesController.RegisterScript("~/Resources/Search/Search.js");```

### Breaking Changes
Note that this PR removes the ability to minify and bundle from the framework. Our motivation to cut support for this is:

1. It is a non-trivial task to reimplement these and carries a significant risk
2. Modern front-end development uses minifying as a build step when rolling out to production, begging the question why the framework should do this
3. HTTP2 has reduced the need for bundling as a way to speed page load. Instead, server side bundling would now just add needless processing and management.

This PR addresses #6643


DNN Platform/DotNetNuke.Abstractions/ClientResources/ClientResourceProviders.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Contains constants for the supported Client Resource Providers within the framework's registration system.
+    /// These values determine where a resource gets injected in the page.
+    /// </summary>
+    public class ClientResourceProviders
+    {
+        /// <summary>
+        /// In the page header.
+        /// </summary>
+        public const string Default = "";
```
mitchelsellers: Why a blank value here?

DNN Platform/DotNetNuke.Abstractions/ClientResources/FileOrder.cs
```diff
@@ -0,0 +1,103 @@
...
+            DefaultCss = 5,
+
+            /// <summary>The admin.css file has a priority of 6.</summary>
+            AdminCss = 6,
+
+            /// <summary>The feature.css file has a priority of 7.</summary>
+            FeatureCss = 7,
+
+            /// <summary>The ie.css file has a priority of 8.</summary>
+            IeCss = 8,
```
mitchelsellers, line 81: Can we at least mark this one as Obsolete?

DNN Platform/DotNetNuke.Abstractions/ClientResources/IClientResourcesController.cs
```diff
@@ -0,0 +1,99 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Abstractions.ClientResources;
+
+/// <summary>Provides an interface for managing client resources.</summary>
+public interface IClientResourcesController
```
mitchelsellers: In matching naming in other places would it be more proper for this to be `IClientResourceController` rather than the plural Resources?  Not hard set on it, just a question

DNN Platform/DotNetNuke.Abstractions/ClientResources/IFontResource.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Abstractions.ClientResources;
+
+/// <summary>Marker interface for font resources.</summary>
+public interface IFontResource : IResource
+{
```
mitchelsellers, line 8: Should we be supporting a "type" attribute here as well to inject for example such as `type="font/woff2"`.

I have a secondary comment about possibly having preload here rather than at the `IResource` level

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/FontResource.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <inheritdoc />
+        public new void Register()
+        {
+            this.clientResourcesController.AddFont(this);
+        }
+
+        /// <inheritdoc />
+        public new string Render(int crmVersion, bool useCdn, string applicationPath)
+        {
+            var htmlString = "<link";
```
mitchelsellers: Given this is a hot-path item within the execution and there are more than 6 concatentations at minimum we should look at doing this with a `StringBuilder` to avoid excessive memory usage.

DNN Platform/DotNetNuke.Abstractions/ClientResources/IResource.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the client resource is blocking.
+    /// </summary>
+    bool Blocking { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the client resource should be preloaded.
+    /// </summary>
+    bool Preload { get; set; }
```
mitchelsellers: Given this is only used on Links, shoudl this be set here?

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        /// <summary>
+        /// Renders the attributes.
+        /// </summary>
+        /// <returns>The attributes.</returns>
+        protected string RenderAttributes()
+        {
+            var htmlString = string.Empty;
+            foreach (var attribute in this.Attributes)
```
mitchelsellers, line 216: Not sure of usage here, could be another consideration for StringBuilder

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ScriptResource.cs
```diff
@@ -0,0 +1,80 @@
...
+        public string Type { get; set; } = string.Empty;
+
+        /// <inheritdoc />
+        public new void Register()
+        {
+            this.clientResourcesController.AddScript(this);
+        }
+
+        /// <inheritdoc />
+        public new string Render(int crmVersion, bool useCdn, string applicationPath)
```
mitchelsellers: Similar note here regarding StringBuilder

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            return string.Join(string.Empty, sortedList);
+        }
+
+        private List<T> AddResource<T>(List<T> resources, T resource)
+            where T : IResource
+        {
+            resource.ResolvedPath = this.ResolvePath(resource.FilePath, resource.PathNameAlias);
+            resource.Key = resource.ResolvedPath.ToLowerInvariant();
+            resources.RemoveAll(l => l.Key == resource.Key); // remove any existing link with the same key (i.e. exactly the same resolved path)
```
mitchelsellers: Just confirming the expectation here, if I somehow have 2 things that reference the same file, but that file has different configurations, only the last one in wins correct?  I don't know we can do much about it, but may want to document this behavior along the way.

DNN Platform/DotNetNuke.Web.Client/Cdf/BasicFile.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    using System.Collections.Generic;
+
+    internal class BasicFile
+    {
+        public BasicFile(ClientDependencyType type)
+        {
+            this.DependencyType = type;
+            this.HtmlAttributes = new Dictionary<string, string>();
+            this.Priority = 100;
```
mitchelsellers, line 15: Can we reference or at least comment back to the relative names for these values for future debugging assistance?

DNN Platform/DotNetNuke.Web.Client/Cdf/ClientDependencyType.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Represents a JavaScript client file.
+        /// </summary>
+        Javascript,
+
+        /// <summary>
+        /// Represents a CSS client file.
+        /// </summary>
+        Css,
```
mitchelsellers, line 20: Should this include Font as well?

DNN Platform/DotNetNuke.Web.Client/Cdf/CssMediaType.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Specifies the CSS media types that can be used.
+    /// </summary>
+    public enum CssMediaType
+    {
+        /// <summary>
+        /// Applies to all media types.
+        /// </summary>
+        All,
```
mitchelsellers, line 20: Is this something intended to be allowed within the `StylesheetResource` declaration?  If so I think it would be a great add

DNN Platform/DotNetNuke.Web.Client/Controls/ClientResourceExclude.cs
```diff
@@ -2,47 +2,53 @@
...
-
-                if (ctlToRemove != null)
-                {
-                    loader.Controls.Remove(ctlToRemove);
-                }
-            }
-        }
-    }
-}
+        protected override void OnInit(EventArgs e)
```
mitchelsellers, line 39: By moving this process from OnPreRender which is executed after postbacks & event information has been processed to OnInit which is before all of that and very early, I believe we will be introducing a breaking change as inside of a PageLoad, etc you wouldn't be able to unload a resource, which you can do today.

DNN Platform/DotNetNuke.Abstractions/ClientResources/IResource.cs
donker: Good catch. I'll move that one.

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
donker: Yes, you're right. These are the limitations of our (current and future) system. It's near impossible to determine which one will win but also impossible to let everyone have their way. There is a "force version" option which allows someone to force a version upon the rest. But if another dev does the exact same thing you get unpredictable results. I don't see a way to solve this.

mitchelsellers: Yeah, more than anyhthing it "shouldn't" happen, but I'm sure it can if people reference shared files etc., I'm going to pose a more global question in a minute

DNN Platform/DotNetNuke.Web.Client/Cdf/ClientDependencyType.cs
donker: We should discuss this. The stuff in Web.Client only remained to ensure no existing code would break, but I think this should all be removed at some point.

DNN Platform/DotNetNuke.Web.Client/Controls/ClientResourceExclude.cs
donker: The new approach makes this a moot point. Before, the resource would be added as a control to the page and removed. This is why these controls (CssExclude/Include etc) all do their magic on PreRender. But in this case whoever comes first finds whatever in the controls and does its thing. In the new situation the controller just keeps a list of stuff to add and to remove and does the necessary sorting and filtering on render.

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/FontResource.cs
donker: Done!

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
donker: Done!

DNN Platform/DotNetNuke.Abstractions/ClientResources/IClientResourcesController.cs
donker: Done!

DNN Platform/DotNetNuke.Abstractions/ClientResources/IFontResource.cs
donker: I've added two solution paths. One is that the user specifies the mimetype in FromSrc, the other is that there is a level of auto detection if the user doesn't do this. Obviously in the latter case we'll need to update code when new font/mimetypes appear.

DNN Platform/Library/Services/ClientDependency/DependencyController.cs
```diff
@@ -0,0 +1,34 @@
...
+    using System.Web.UI;
+
+    using DotNetNuke.Abstractions.ClientResources;
+    using Microsoft.Extensions.DependencyInjection;
+
+    internal class DependencyController
+    {
+        internal static IClientResourceController GetClientResourcesController(Page page)
+        {
+            var serviceProvider = GetCurrentServiceProvider(page.Request.RequestContext.HttpContext);
```
sachatrauwaen: Maybe Globals.GetCurrentServiceProvider() can do the job in a more uniform way inside dnn

bdukes: This is a pattern we're actively trying to discourage, it should only be used for backwards compatibility reasons. That's why [`Globals.GetCurrentServiceProvider` is `internal`](https://github.com/dnnsoftware/Dnn.Platform/blob/6267a7ba78028ccc8645f5b94a8bc165ace10b30/DNN%20Platform/Library/Common/Globals.cs#L3541-L3546)

sachatrauwaen: Yes sure but i think this method is also for backwards compatibility reasons. And done a littlebit different way. Or maybe i am wrong.

bdukes: My intention is that this method is equivalent to `Globals.GetCurrentServiceProvider`, but I have removed some parts that should be unnecessary here (e.g. the methods calling this have a `Page`, so we don't need to support a scenario without an `HttpContext`).

sachatrauwaen:
Great work !

This can maybe simplified 
```
this.clientResourcesController
        .CreateScript()
        .FromSrc("~/Resources/Search/Search.js")
        .SetPriority((int)FileOrder.Js.DefaultPriority)
        .SetProvider(ClientResourceProviders.DnnFormBottomProvider)
        .Register();
```

as

```
this.clientResourcesController
        .AddScript("~/Resources/Search/Search.js")
        .SetPriority((int)FileOrder.Js.DefaultPriority)
        .SetProvider(ClientResourceProviders.DnnFormBottomProvider);
```

This make required parameters visible to developers.
Then you can remove dependecy of the resources to the controller. Then the resources can be created outside of the context of the controller.
Then AddScript("~/Resources/Search/Search.js") can be also an extension method... witch simplify the Controller Interface.

It is not very clear if you want the resources only been created from the controller because the controller have also methods AddScript(IScriptResource script).

valadas:
It would be a nice touch if the .SetPriority() accepted an int or the enum value directly for discoverability.

donker:
> It would be a nice touch if the .SetPriority() accepted an int or the enum value directly for discoverability.

But wouldn't that tie this in a dependency on that enum? Meaning: if we want to change something it will break people's code? The priority enum is wild and is very likely to change in the future.

donker:
> It is not very clear if you want the resources only been created from the controller because the controller have also methods AddScript(IScriptResource script).

The reason it needs register in the end is because the controller needs this script to be added. The AddScript is used in case you have a full IScriptResource object.

bdukes:
We could add an extension method in the same place as the enum. If we're going to deprecate that enum, we would also deprecated that extension method.

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,250 @@
...
+        public void Register()
+        {
+            throw new System.NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public string Render(int crmVersion, bool useCdn, string applicationPath)
+        {
+            throw new System.NotImplementedException();
+        }
```
bdukes: Mark these as abstract instead of throwing?

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        /// <summary>
+        /// Renders the attributes.
+        /// </summary>
+        /// <param name="htmlString">The HTML string builder to append to.</param>
+        protected void RenderAttributes(StringBuilder htmlString)
+        {
+            foreach (var attribute in this.Attributes)
+            {
+                htmlString.Append($" {attribute.Key}=\"{attribute.Value}\"");
```
bdukes: We need to encode these attribute values
```suggestion
                htmlString.Append($" {attribute.Key}=\"{WebUtility.HtmlEncode(attribute.Value)}\"");
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        /// <summary>
+        /// Renders the mimetype attribute.
+        /// </summary>
+        /// <param name="htmlString">The HTML string builder to append to.</param>
+        protected void RenderType(StringBuilder htmlString)
+        {
+            if (!string.IsNullOrEmpty(this.Type))
+            {
+                htmlString.Append($" type=\"{this.Type}\"");
```
bdukes:
```suggestion
                htmlString.Append($" type=\"{WebUtility.HtmlEncode(this.Type)}\"");
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        /// <summary>
+        /// Renders the integrity attribute.
+        /// </summary>
+        /// <param name="htmlString">The HTML string builder to append to.</param>
+        protected void RenderIntegrity(StringBuilder htmlString)
+        {
+            if (!string.IsNullOrEmpty(this.Integrity))
+            {
+                htmlString.Append($" integrity=\"{this.Integrity}\"");
```
bdukes:
```suggestion
                htmlString.Append($" integrity=\"{WebUtility.HtmlEncode(this.Integrity)}\"");
```

sachatrauwaen:
> > It is not very clear if you want the resources only been created from the controller because the controller have also methods AddScript(IScriptResource script).
> 
> The reason it needs register in the end is because the controller needs this script to be added. The AddScript is used in case you have a full IScriptResource object.

You can register at the begin.

> The AddScript is used in case you have a full IScriptResource object.

The use of this is complicated because you need to pass the controller to the constructor of the resource. And the the register is not needed.
 
What is the indend ?

var r = new ScriptResource(controller);
r.Register();

or
 controller.AddScript(r);

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
bdukes:
```suggestion
        public abstract void Register();

        /// <inheritdoc />
        public abstract string Render(int crmVersion, bool useCdn, string applicationPath);
```

DNN Platform/DotNetNuke.Web.Client/Cdf/BasicFile.cs
donker: This is temporary old code to avoid breaking changes

DNN Platform/DotNetNuke.Web.Client/Cdf/ClientDependencyType.cs
donker: This is temporary old code to avoid breaking changes

DNN Platform/DotNetNuke.Web.Client/Cdf/CssMediaType.cs
donker: This is temporary old code to avoid breaking changes

johnhenley:
@donker did some testing with latest forums release (9.1.1) and there were no issues.

valadas:
@donker nice, I just went through it all and nothing sticks out to me. However I would like to test this out with some WebAPI (SPA) modules I have here, do we need to do something different than documented here? https://docs.dnncommunity.org/content/tutorials/modules/about-modules/spa-module-development/index.html#accessing-dnn-features

Or will that just work ase before?

donker:
> @donker nice, I just went through it all and nothing sticks out to me. However I would like to test this out with some WebAPI (SPA) modules I have here, do we need to do something different than documented here? https://docs.dnncommunity.org/content/tutorials/modules/about-modules/spa-module-development/index.html#accessing-dnn-features
> 
> Or will that just work ase before?

It should all work without any change in code. The underlying old CDF classes have all been redirected to the new CDF.

valadas:
I have started to test this with my solutions and then got sidetracked with finding an unrelated bug (which I just sent a PR for, the jQueryUI deprecated thingy).

Anyway, it worked fine with the way I use this in SPA modules.

I am just thinking however that if we no longer do composite files and minification, then we should probably kill the UI for it or users will be confused about options existing that don't work...
image

bdukes (approved):

DNN Platform/DotNetNuke.Abstractions/Application/IHostSettings.cs
```diff
@@ -40,12 +41,15 @@ public interface IHostSettings
...
     /// <summary>Gets a value indicating whether Composite Files are enabled at the host level.</summary>
+    [DnnDeprecated(10, 2, 0, "Bundling and minification have been deprecated in DNN 10.2. These references will be removed in DNN 12.")]
     public bool CrmEnableCompositeFiles { get; }
 
     /// <summary>Gets a value indicating whether CSS Minification is enabled at the host level.</summary>
+    [DnnDeprecated(10, 2, 0, "Bundling and minification have been deprecated in DNN 10.2. These references will be removed in DNN 12.")]
     public bool CrmMinifyCss { get; }
 
     /// <summary>Gets a value indicating whether JS Minification is enabled at the host level.</summary>
+    [DnnDeprecated(10, 2, 0, "Bundling and minification have been deprecated in DNN 10.2. These references will be removed in DNN 12.")]
```
bdukes: The `DnnDeprecated` attribute will create a message which includes the version it was deprecated in and the version it will be removed in, so I don't think we need that in the message provided.

```suggestion
    [DnnDeprecated(10, 2, 0, "Bundling is no longer supported, there is no replacement within DNN for this functionality.")]
    public bool CrmEnableCompositeFiles { get; }

    /// <summary>Gets a value indicating whether CSS Minification is enabled at the host level.</summary>
    [DnnDeprecated(10, 2, 0, "Minification is no longer supported, there is no replacement within DNN for this functionality.")]
    public bool CrmMinifyCss { get; }

    /// <summary>Gets a value indicating whether JS Minification is enabled at the host level.</summary>
    [DnnDeprecated(10, 2, 0, "Minification is no longer supported, there is no replacement within DNN for this functionality.")]
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+            this.RegisterPathNameAlias("SharedScripts", "~/Resources/Shared/Scripts/");
+        }
+
+        private List<IFontResource> Fonts { get; set; } = new List<IFontResource>();
+
+        private List<IScriptResource> Scripts { get; set; } = new List<IScriptResource>();
+
+        private List<IStylesheetResource> Stylesheets { get; set; } = new List<IStylesheetResource>();
+
+        private Dictionary<string, string> PathNameAliases { get; set; } = new Dictionary<string, string>();
```
bdukes: Instead of converting keys to lowercase, let's make the lookup be case-insensitive

```suggestion
        private Dictionary<string, string> PathNameAliases { get; set; } = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        private List<string> FontsToExclude { get; set; } = new List<string>();
```
bdukes: We don't need to preserve order here, right?
```suggestion
        private HashSet<string> FontsToExclude { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        private List<string> ScriptsToExclude { get; set; } = new List<string>();
+
+        private List<string> StylesheetsToExclude { get; set; } = new List<string>();
```
bdukes:
```suggestion
        private HashSet<string> ScriptsToExclude { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);

        private HashSet<string> StylesheetsToExclude { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+        public void RemoveFontByName(string fontName)
+        {
+            this.FontsToExclude.Add(fontName.ToLowerInvariant());
+        }
+
+        /// <inheritdoc />
+        public void RemoveFontByPath(string fontPath, string pathNameAlias)
+        {
+            var fullPath = this.ResolvePath(fontPath, pathNameAlias).ToLowerInvariant();
+            this.FontsToExclude.Add(fullPath.ToLowerInvariant());
```
bdukes: We don't need to call `ToLowerInvariant` twice (or at all if we change it to a HashSet with a case-insensitive comparer).
```suggestion
            var fullPath = this.ResolvePath(fontPath, pathNameAlias);
            this.FontsToExclude.Add(fullPath.ToLowerInvariant());
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+        private List<T> AddResource<T>(List<T> resources, T resource)
+            where T : IResource
+        {
+            resource.ResolvedPath = this.ResolvePath(resource.FilePath, resource.PathNameAlias);
+            resource.Key = resource.ResolvedPath.ToLowerInvariant();
+            resources.RemoveAll(l => l.Key == resource.Key); // remove any existing link with the same key (i.e. exactly the same resolved path)
+            if (!string.IsNullOrEmpty(resource.Name))
+            {
+                // if a resource with the same name and force version is already present we ignore this one
+                if (resources.Exists(l => l.Name.ToLowerInvariant() == resource.Name.ToLowerInvariant() && l.ForceVersion))
```
bdukes:
```suggestion
                if (resources.Exists(r => string.Equals(r.Name., resource.Name, StringComparison.OrdinalIgnoreCase) && r.ForceVersion))
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+                if (!string.IsNullOrEmpty(resource.Version))
+                {
+                    resources.RemoveAll(l => l.Name.ToLowerInvariant() == resource.Name.ToLowerInvariant() && string.Compare(l.Version, resource.Version, System.StringComparison.InvariantCultureIgnoreCase) < 0);
+
+                    // If we have an existing link with the same name and a higher version, we do not add this link
+                    if (resources.Exists(l => l.Name.ToLowerInvariant() == resource.Name.ToLowerInvariant() && string.Compare(l.Version, resource.Version, System.StringComparison.InvariantCultureIgnoreCase) >= 0))
+                    {
+                        return resources;
+                    }
+                }
```
bdukes, line 184:
```suggestion
                // If we are forcing the version, we need to remove any existing link with the same name
                if (resource.ForceVersion)
                {
                    resources.RemoveAll(r => string.Equals(r.Name, resource.Name, StringComparison.OrdinalIgnoreCase));
                }

                // If we have a version, we need to remove any existing link with the same name and a lower version
                if (!string.IsNullOrEmpty(resource.Version))
                {
                    resources.RemoveAll(r => string.Equals(r.Name, resource.Name, StringComparison.OrdinalIgnoreCase) && string.Compare(r.Version, resource.Version, StringComparison.InvariantCultureIgnoreCase) < 0);

                    // If we have an existing link with the same name and a higher version, we do not add this link
                    if (resources.Exists(r => string.Equals(r.Name, resource.Name, StringComparison.OrdinalIgnoreCase) && string.Compare(r.Version, resource.Version, StringComparison.InvariantCultureIgnoreCase) >= 0))
                    {
                        return resources;
                    }
                }
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+                // If we are forcing the version, we need to remove any existing link with the same name
+                if (resource.ForceVersion)
+                {
+                    resources.RemoveAll(l => l.Name.ToLowerInvariant() == resource.Name.ToLowerInvariant());
+                }
+
+                // If we have a version, we need to remove any existing link with the same name and a lower version
+                if (!string.IsNullOrEmpty(resource.Version))
+                {
+                    resources.RemoveAll(l => l.Name.ToLowerInvariant() == resource.Name.ToLowerInvariant() && string.Compare(l.Version, resource.Version, System.StringComparison.InvariantCultureIgnoreCase) < 0);
```
bdukes: Is doing a string comparison of versions going to be good enough? Will 10.0.0 be higher than 9.0.0?

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        private string ResolvePath(string filePath, string pathNameAlias)
+        {
+            if (string.IsNullOrEmpty(filePath))
+            {
+                return filePath;
+            }
+
+            if (filePath.ToLowerInvariant().StartsWith("http"))
```
bdukes:
```suggestion
            if (filePath.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/GenericResourceExtensions.cs
```diff
@@ -0,0 +1,221 @@
...
+        public static T AddAttribute<T>(this T input, string attributeName, string attributeValue)
+            where T : IResource
+        {
+            if (input == null)
+            {
+                throw new ArgumentNullException(nameof(input));
+            }
+
+            if (!string.IsNullOrEmpty(attributeName) && !string.IsNullOrEmpty(attributeValue))
+            {
```
bdukes: Are there no scenarios where someone would want to set an attribute to an empty value? Can you clear out an attribute some way other than adding an empty/`null` value?

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// <inheritdoc />
+        public string Provider { get; set; } = ClientResourceProviders.DnnPageHeaderProvider;
+
+        /// <inheritdoc />
+        public string Name
+        {
+            get
+            {
+                return string.IsNullOrEmpty(this.name) ? this.FilePath : this.name;
+            }
```
bdukes:
```suggestion
            get => string.IsNullOrEmpty(this.name) ? this.FilePath : this.name;
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// <param name="applicationPath">The application path to use for resolving relative paths.</param>
+        /// <returns>The versioned path.</returns>
+        internal string GetVersionedPath(int crmVersion, bool useCdn, string applicationPath)
+        {
+            var path = this.ResolvedPath;
+            if (useCdn && !string.IsNullOrEmpty(this.CdnUrl))
+            {
+                return this.CdnUrl;
+            }
+            else if (path.ToLowerInvariant().StartsWith("http"))
```
bdukes:
```suggestion
            else if (path.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ScriptResource.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            this.RenderType(htmlString);
+            this.RenderBlocking(htmlString);
+            this.RenderCrossOriginAttribute(htmlString);
+            this.RenderFetchPriority(htmlString);
+            this.RenderIntegrity(htmlString);
+            this.RenderReferrerPolicy(htmlString);
+            this.RenderAttributes(htmlString);
+            htmlString.Append(" type=\"text/javascript\"></script>");
```
bdukes: Should this have both `RenderType` and `type=\"text/javascript\"`?

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ScriptResource.cs
```diff
@@ -0,0 +1,75 @@
...
+        public override void Register()
+        {
+            this.clientResourceController.AddScript(this);
+        }
+
+        /// <inheritdoc />
+        public override string Render(int crmVersion, bool useCdn, string applicationPath)
+        {
+            var htmlString = new StringBuilder("<script");
+            htmlString.Append($" src=\"{this.GetVersionedPath(crmVersion, useCdn, applicationPath)}\"");
```
bdukes:
```suggestion
            htmlString.Append($" src=\"{WebUtility.HtmlEncode(this.GetVersionedPath(crmVersion, useCdn, applicationPath))}\"");
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/FontResource.cs
```diff
@@ -0,0 +1,58 @@
...
+        public override void Register()
+        {
+            this.clientResourceController.AddFont(this);
+        }
+
+        /// <inheritdoc />
+        public override string Render(int crmVersion, bool useCdn, string applicationPath)
+        {
+            var htmlString = new StringBuilder("<link");
+            htmlString.Append($" href=\"{this.GetVersionedPath(crmVersion, useCdn, applicationPath)}\"");
```
bdukes:
```suggestion
            htmlString.Append($" href=\"{WebUtility.HtmlEncode(this.GetVersionedPath(crmVersion, useCdn, applicationPath))}\"");
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/StylesheetResource.cs
```diff
@@ -0,0 +1,68 @@
...
+        public override void Register()
+        {
+            this.clientResourceController.AddStylesheet(this);
+        }
+
+        /// <inheritdoc />
+        public override string Render(int crmVersion, bool useCdn, string applicationPath)
+        {
+            var htmlString = new StringBuilder("<link");
+            htmlString.Append($" href=\"{this.GetVersionedPath(crmVersion, useCdn, applicationPath)}\"");
```
bdukes:
```suggestion
            htmlString.Append($" href=\"{WebUtility.HtmlEncode(this.GetVersionedPath(crmVersion, useCdn, applicationPath))}\"");
```

DNN Platform/DotNetNuke.Web.Client/ClientResourceManager.cs
```diff
@@ -14,13 +15,15 @@ namespace DotNetNuke.Web.Client.ClientResourceManagement
...
-    using ClientDependency.Core.Config;
+    using DotNetNuke.Abstractions.ClientResources;
     using DotNetNuke.Instrumentation;
     using DotNetNuke.Internal.SourceGenerators;
+    using DotNetNuke.Web.Client.ResourceManager;
+
+    using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>Provides the ability to request that client resources (JavaScript and CSS) be loaded on the client browser.</summary>
+    [Obsolete("Deprecated in DotNetNuke 10.2.0. Please use IClientResourceController instead. Scheduled removal in v12.0.0.")]
```
bdukes:
```suggestion
    [DnnDeprecated, 10, 2, 0, "Please use IClientResourceController instead.")]
```

DNN Platform/DotNetNuke.Web.Client/Providers/DnnCompositeFileProcessingProvider.cs
```diff
@@ -1,95 +1,97 @@
...
+    using System;
+    using System.IO;
+
+    using ClientDependency.Core;
+    using ClientDependency.Core.CompositeFiles;
+    using ClientDependency.Core.CompositeFiles.Providers;
+    using DotNetNuke.Internal.SourceGenerators;
+
+    /// <summary>A provider for combining, minifying, compressing and saving composite scripts/css files.</summary>
+    [DnnDeprecated(10, 2, 0, "Bundling and minification have been deprecated in DNN 10.2. These references will be removed in DNN 12.")]
```
bdukes:
```suggestion
    [DnnDeprecated(10, 2, 0, "Bundling and minification is no longer supported, there is no replacement within DNN for this functionality.")]
```

DNN Platform/Website/admin/Skins/DnnCssInclude.ascx.cs
```diff
@@ -36,9 +35,7 @@ protected override void OnLoad(EventArgs e)
         {
             base.OnLoad(e);
             this.ctlInclude.AddTag = this.AddTag;
-            this.ctlInclude.CssMedia = this.CssMedia;
```
bdukes, line 39: Can we support the `media` attribute on a `link` element?

DNN Platform/Library/Services/ClientDependency/Extensions.cs
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            if (string.IsNullOrEmpty(stylesheetPath))
+            {
+                throw new System.ArgumentException("Value cannot be null or empty.", nameof(stylesheetPath));
+            }
+
+            if (checkIfExists)
+            {
+                var physicalPath = HttpContext.Current.Server.MapPath(stylesheetPath);
```
bdukes: prefer [`HostingEnvironment.MapPath`](https://learn.microsoft.com/en-us/dotnet/api/system.web.hosting.hostingenvironment.mappath) over `Server.MapPath`, since it doesn't depend on `HttpContext`

DNN Platform/Library/Services/ClientDependency/DependencyController.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    using System;
+    using System.Collections;
+    using System.Web;
+    using System.Web.UI;
+
+    using DotNetNuke.Abstractions.ClientResources;
+    using Microsoft.Extensions.DependencyInjection;
+
+    internal class DependencyController
```
bdukes: It doesn't look like this type is used

DNN Platform/Library/Framework/JavaScriptLibraries/JavaScript.cs
```diff
@@ -653,15 +674,32 @@ private static int GetFileOrder(JavaScriptLibrary jsl)
...
+        private static IServiceProvider GetCurrentServiceProvider(HttpContextBase context)
+        {
+            return GetScope(context.Items).ServiceProvider;
+
+            // Copy of DotNetNuke.Common.Extensions.HttpContextDependencyInjectionExtensions.GetScope
+            static IServiceScope GetScope(IDictionary httpContextItems)
+            {
+                return httpContextItems[typeof(IServiceScope)] as IServiceScope;
             }
         }
```
bdukes:
```suggestion
        private static IClientResourceController GetClientResourcesController(HttpContextBase context)
        {
            var serviceProvider = Globals.GetCurrentServiceProvider();
            return serviceProvider.GetRequiredService<IClientResourceController>();
        }
```

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
donker: No, order is unimportant

DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
donker: Good point. Maybe that needs to be changed.

DNN Platform/DotNetNuke.Web.Client.ResourceManager/GenericResourceExtensions.cs
donker: Good point. I'll remove that check.

DNN Platform/Website/admin/Skins/DnnCssInclude.ascx.cs
donker: OK, I've brought it back

bdukes:
@donker I've adjusted a few other places, including the version comparison. This is ready from my perspective.

mitchelsellers:
@valadas or @david-poindexter Can I get a review from one of you on this one, since Brian made the last edits

DNN Platform/Library/UI/Containers/Container.cs
```diff
@@ -19,22 +20,30 @@ namespace DotNetNuke.UI.Containers
...
         private HtmlContainerControl contentPane;
         private ModuleInfo moduleConfiguration;
         private ModuleHost moduleHost;
 
+        /// <summary>Initializes a new instance of the <see cref="Container"/> class.</summary>
+        /// <param name="clientResourceController">The client resources controller.</param>
+        public Container(IClientResourceController clientResourceController)
+        {
+            this.clientResourceController = clientResourceController ?? Globals.GetCurrentServiceProvider().GetRequiredService<IClientResourceController>();
+        }
```
valadas, line 52: Should we add a parameterless constructor (obsolete) just so this is not a breaking change and if anyone creates an instance they know to use the new constructor?

donker: I looked into this and currently it's not supported to mark a constructor as deprecated using our own deprecation message. So we need to use the Obsolete attribute.

PR #6749, opened 2025-10-03
First code submission for MVC Pipeline

donker:
### Core Purpose
Introduces a new ASP.NET MVC-based rendering pipeline as an alternative to the traditional WebForms pipeline, allowing DNN to modernize its architecture while maintaining backward compatibility.
### Key Features
Dual Pipeline Architecture:
- Traditional WebForms pipeline continues through /default.aspx
- New MVC pipeline routes through /DesktopModules/MvcPipeline/Default/Page/{tabId}/{locale}

[Video showcase the general usage](https://www.youtube.com/watch?v=L7ZHTP8e_7w)

### Module Support:

[More info about MVC Module Control Implementation](https://github.com/dnnsoftware/Dnn.Platform/pull/6700)
[Razor+ Module Development Guide](https://github.com/dnnsoftware/Dnn.Platform/blob/feature/mvc-pipeline-old/DNN%20Platform/DotNetNuke.Web.MvcPipeline/razor-module-development.md)

- HTML module converted to MVC
- Sample modules demonstrating [Razor+](https://github.com/dnnsoftware/Dnn.Platform/tree/aa87d3fcd79268cd096f5674cc95d2bedccf1445/DNN%20Platform/Modules/Samples/Dnn.ContactList.Razor),  MVC and SPA patterns

[How to compile Samples](https://github.com/dnnsoftware/Dnn.Platform/tree/d39d9e4b2eec73c2d529ebb6c24b55880f6bda1d/DNN%20Platform/Modules/Samples)

### Configuration:
- Pipeline selection at portal and page level
- Settings UI integrated into Site and Page Settings
- URL rewriter modifications to route based on pipeline settings

[More info about MVC pipeline (routing / settings)](https://github.com/dnnsoftware/Dnn.Platform/pull/6721)

#### DotNetNuke.Web.MvcPipeline.csproj - New core MVC pipeline library
- Skins : Html Helpers (replacement for skin objects)
- Containers : Html Helpers (replacement for skin objects for containers)
- ModelFactories : construction of the Page Model (Model of the DefaultController)
- ModuleControl : Module Control infrastructure
- Utils : MvcViewEngine (Replacement for RazorEngine based on MVC 5 and can only be used inside HttpContext)
- UI/Utilities : MvcClientAPI (replacement for ClientAPI for mvc)

#### DDRMenu 
- Adapted to suport MVC Pipeline

#### DotNetNuke.Library
- Added Site and Page pipeline settings
- Added Module Control settings
- URL rewriter changes to route based on pipeline settings
- Update SPA (Html5) to use new CDF

#### DotNetNuke.Abstractions
- Added PagePipeline property to IPortalSettings

#### DotNetNuke.Web.Mvc
- Integration with new MVC pipeline
- Make actual DNN MVC modules run on the MVC Pipeline

#### HTML
- Add support for MVC Pipeline rendering
- Add MVC views and controllers
- This is implemented as true MVC (with form tags) using AJAX

#### Dnn.PersonaBar.UI
UI for pipeline settings in Site and Page Settings

#### DNNConnect.CKE
- HTML helper additions for CKEditor in MVC context

### Skins/Aperture
- Added mvc suport to the skin (Razor files)

[Skin conversion guide](https://github.com/dnnsoftware/Dnn.Platform/blob/0706d5901b8f16cdeb879a6af7ee7fea427e9f3f/DNN%20Platform/Skins/Aperture/SKIN_WEBFORMS_TO_MVC_GUIDE.md)

### Website
- MVC  Razor Layout
- MVC Default skin
- mvc.js minimal stuff needed for dnn to work
- Terms and privacy pages (razor files)
- DefaultController
- ModuleActions Controller

### AdminExperiance
- settings ui for new settings
- mvc controls for persona and edit bar

This PR is a first submission of the work of the MVC Pipeline team to bring the main project up to date with what has been developed in a forked repo.

### Documentation

[MVC Module Control Implementation](https://github.com/dnnsoftware/Dnn.Platform/pull/6700)
[MVC pipeline (routing / settings)](https://github.com/dnnsoftware/Dnn.Platform/pull/6721)
[Razor Module Development Guide](https://github.com/dnnsoftware/Dnn.Platform/blob/feature/mvc-pipeline-old/DNN%20Platform/DotNetNuke.Web.MvcPipeline/razor-module-development.md)
[How to compile Samples](https://github.com/dnnsoftware/Dnn.Platform/tree/d39d9e4b2eec73c2d529ebb6c24b55880f6bda1d/DNN%20Platform/Modules/Samples)

[DNN MVC Pipeline - Developer User Guide](https://github.com/dnnsoftware/Dnn.Platform/blob/feature/mvc-pipeline-old/DNN%20Platform/DotNetNuke.Web.MvcPipeline/DNN_MVC_PIPELINE_USER_GUIDE.md)

---
# Review Tasklist

- [ ] Walkthrough: Ask the author for a live, screen-share session for explaining the high-level intent, architectural changes, and where the high-risk logic lives.

## 1) Prerequirement, Build and packaging (9 folders, A: 2, M: 9)

- [ ] DNN Platform/Library/Properties
- [ ] DNN Platform/Library/UI/UserControls
- [ ] DNN Platform/Library/Services/Tokens
- [ ] DNN Platform/Library/Services/Tokens/PropertyAccess
- [ ] DNN Platform/Library/UI/Modules
- [ ] DNN Platform/DotNetNuke.ContentSecurityPolicy
- [ ] DNN Platform/DotNetNuke.ContentSecurityPolicy/Properties
- [ ] Build/Tasks
- [ ] DNN Platform/Components/Microsoft.AspNetMvc

## 2) Settings (portal, page, modules) (19 folders, A: 3, M: 29)

- [ ] DNN Platform/DotNetNuke.Abstractions/Portals
- [ ] DNN Platform/Library/Entities/Modules
- [ ] DNN Platform/Library/Entities/Portals
- [ ] DNN Platform/Library/Entities/Tabs
- [ ] DNN Platform/Library/Data
- [ ] DNN Platform/Library/Services/Upgrade
- [ ] DNN Platform/Library/Services/Installer/Writers
- [ ] Dnn.AdminExperience/ClientSide/Pages.Web/src/components/More
- [ ] Dnn.AdminExperience/ClientSide/Pages.Web/src/services
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Pages
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/DTO/SiteSettings
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Pages/App_LocalResources
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SiteSettings/App_LocalResources
- [ ] Dnn.AdminExperience/ClientSide/Extensions.Web/src/components/EditExtension/CustomSettings/Module/ModuleDefinitions/ModuleDefinitionRow/Controls
- [ ] Dnn.AdminExperience/ClientSide/SiteSettings.Web/src/components/moreSettings
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Components/Extensions/Dto/Editors
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/DTO
- [ ] Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Extensions/App_LocalResources
- [ ] DNN Platform/Website/Providers/DataProviders/SqlDataProvider

## 3) Rewriter (1 folders, A: 0, M: 1)

- [ ] DNN Platform/Library/Entities/Urls

## 4) MvcPipeline (19 folders, A: 108, M: 0)

- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Containers
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Exceptions
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Extensions
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Framework
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Framework/JavascriptLibraries
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/ModelFactories
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Models
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/ModuleControl
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/ModuleControl/Page
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/ModuleControl/Razor
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/ModuleControl/WebForms
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Modules
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Pages
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Routing
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Skins
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/UI/Utilities
- [ ] DNN Platform/DotNetNuke.Web.MvcPipeline/Utils

# 5) Website & Admin (17 folders, A: 23, M: 5)
- [ ] DNN Platform/Website
- [ ] DNN Platform/Website/Controllers
- [ ] DNN Platform/Website/Models
- [ ] DNN Platform/Website/Resources/Shared/scripts/jquery
- [ ] DNN Platform/Website/Views
- [ ] DNN Platform/Website/Views/Default
- [ ] DNN Platform/Website/Views/PersonaBarContainer
- [ ] DNN Platform/Website/admin/Portal/Controls
- [ ] DNN Platform/Website/admin/Portal/Views
- [ ] DNN Platform/Website/js
- [ ] Dnn.AdminExperience/EditBar/Dnn.EditBar.UI
- [ ] Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/Mvc
- [ ] Dnn.AdminExperience/Library/Dnn.PersonaBar.UI
- [ ] Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/Mvc

## 6) HtmlEditorProviders (3 folders, A: 2, M: 2)

- [ ] DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE
- [ ] DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Utilities
- [ ] DNN Platform/Providers/HtmlEditorProviders/DNNConnect.CKE/Web

## 7) DDRMenu (4 folders, A: 2, M: 12)

- [ ] DNN Platform/Modules/DDRMenu
- [ ] DNN Platform/Modules/DDRMenu/Common
- [ ] DNN Platform/Modules/DDRMenu/Properties
- [ ] DNN Platform/Modules/DDRMenu/TemplateEngine

## 8) HTML (8 folders, A: 19, M: 7)

- [ ] DNN Platform/Modules/HTML
- [ ] DNN Platform/Modules/HTML/Components
- [ ] DNN Platform/Modules/HTML/Controllers
- [ ] DNN Platform/Modules/HTML/Controls
- [ ] DNN Platform/Modules/HTML/Models
- [ ] DNN Platform/Modules/HTML/Mvc
- [ ] DNN Platform/Modules/HTML/Views
- [ ] DNN Platform/Modules/HTML/js

## 9) Skins (6 folders, A: 17, M: 1)

- [ ] DNN Platform/Website/Portals/_default/Containers/_default/Views
- [ ] DNN Platform/Website/Portals/_default/Skins/_default/Views
- [ ] DNN Platform/Skins/Aperture
- [ ] DNN Platform/Skins/Aperture/Views
- [ ] DNN Platform/Skins/Aperture/Views/partials
- [ ] DNN Platform/Skins/Aperture/containers/Views

# Not distributed (not critical)

## 10) Samples (13 folders, A: 25, M: 7)

- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Mvc
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Mvc/Controllers
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Mvc/Controls
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Mvc/Views/Contact
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/App_LocalResources
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/Controls
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/Models
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/Properties
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/Services
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/Views
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Razor/js
- [ ] DNN Platform/Modules/Samples/Dnn.ContactList.Spa

# Can be moved to other PR (we can start only with Razor+)

## 11) SPA Modules Support (1 folders, A: 0, M: 4)

- [ ] DNN Platform/Library/UI/Modules/Html5

## 12) MVC Modules Support (5 folders, A: 0, M: 5)

- [ ] DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionResults
- [ ] DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers
- [ ] DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules
- [ ] DNN Platform/DotNetNuke.Web.Mvc/Properties
- [ ] DNN Platform/DotNetNuke.Web.Mvc/Routing


mitchelsellers:
@bdukes @donker I believe we discussed that this is just a reset PR and initial review, as it is into their branch correct?  High level review only?

valadas:
Awesome, I'll try to get my review in this weekend...

mitchelsellers:
@donker In looking at this at this moment the solution isn't building, would you be able to review this and resolve the issue?

Build/Tasks/BuildPackages.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <remarks>This is the task run during CI.</remarks>
+    [IsDependentOn(typeof(CleanArtifacts))]
+    [IsDependentOn(typeof(UpdateDnnManifests))]
+    [IsDependentOn(typeof(GenerateSecurityAnalyzerChecksums))]
+    [IsDependentOn(typeof(SetPackageVersions))]
+    [IsDependentOn(typeof(CreateInstall))]
+    [IsDependentOn(typeof(CreateUpgrade))]
+    [IsDependentOn(typeof(CreateNugetPackages))]
+    [IsDependentOn(typeof(GeneratePackagesChecksums))]
+    public sealed class BuildPackages : FrostingTask<Context>
```
mitchelsellers: What is the purpose/goal of this task addition?  I'm concerned that this will be deleting a source controlled file.

DNN Platform/DotNetNuke.ContentSecurityPolicy/BaseCspContributor.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.ContentSecurityPolicy
```
mitchelsellers, line 5: Is this the same as the pending CSP PR into `develop`?

DNN Platform/DotNetNuke.Web.Client/Providers/DnnStandardRenderer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether checks if the composite files option is set for the current portal (DNN site settings).
+        /// If not enabled at the portal level it defers to the core CDF setting (web.config).
+        /// </summary>
+        public override bool EnableCompositeFiles
+        {
+            get
+            {
+                var settingsVersion = this.dnnSettingsHelper.AreCompositeFilesEnabled();
```
mitchelsellers: Given the changes to CDF (In pending state now) should this be removed?

DNN Platform/DotNetNuke.Web.Client/MvcClientResourceManager.cs
```diff
@@ -0,0 +1,588 @@
...
+
+        /// <summary>The default javascript provider.</summary>
+        internal const string DefaultJsProvider = "DnnBodyProvider";
+
+        private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(ClientResourceManager));
+        private static readonly Dictionary<string, bool> FileExistsCache = new Dictionary<string, bool>();
+        private static readonly ReaderWriterLockSlim LockFileExistsCache = new ReaderWriterLockSlim();
+
+        /// <summary>Adds the necessary configuration to website root <c>web.config</c> to use the Client Dependency component.</summary>
+        public static void AddConfiguration()
```
mitchelsellers: Why is this beind done as a Web.Client item, shouldn't this be handled by the existing functionality?  I'm concerned this is going to introduce a duplicate or race condition

DNN Platform/DotNetNuke.ContentSecurityPolicy/BaseCspContributor.cs
sachatrauwaen: Yes it's the same. It need to be removed from here after the pr in develop is merged, so we can remerg it in this branch.
Or some other sujestion how to manage this ?

DNN Platform/DotNetNuke.Web.Client/Providers/DnnStandardRenderer.cs
sachatrauwaen: Yes, the idee is to remove all CDF for mvc that is done here once the changes to CDF are ready to be integrated in this branch.

DNN Platform/DotNetNuke.Web.Client/MvcClientResourceManager.cs
sachatrauwaen: Yes, you are right, it's will beter to do all the web.config modification in one place.
In reality, this class will be removed once the new CDF is used.

Build/Tasks/BuildPackages.cs
sachatrauwaen: @donker Can you answer this ?

sachatrauwaen:
> @donker In looking at this at this moment the solution isn't building, would you be able to review this and resolve the issue?

@mitchelsellers  I just try to build a fresh checkout of feature/mvc-pipeline-old without issuues (.\build.ps1).

Can you tell witch build you use or what kind of errors you get ?

mitchelsellers:
@sachatrauwaen It is the AzureDevOps build that is failing, it looks like it is unit tests that are failing to execute.

The most recent run - https://github.com/dnnsoftware/Dnn.Platform/pull/6749/checks?check_run_id=51891737747

Build/Tasks/BuildPackages.cs
donker: Sorry ... that is my code bleed. I can delete it. It just builds the packages. But I believe the "default" task also does this.

donker:
> @bdukes @donker I believe we discussed that this is just a reset PR and initial review, as it is into their branch correct? High level review only?

Yes. This is what we discussed in an earlier meeting to have the first code submission a proper PR.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleSettingsController.cs
```diff
@@ -0,0 +1,386 @@
...
+        {
+            // Verify that the current user has access to edit this module
+            if (!ModulePermissionController.HasModuleAccess(SecurityAccessLevel.Edit, "MANAGE", this.Module))
+            {
+                if (!(this.IsSharedViewOnly() && TabPermissionController.CanAddContentToPage()))
+                {
+                    throw new UnauthorizedAccessException("Access denied");
+                }
+            }
+        }
```
valadas: This one hurts my brain a bit specially since it relates to permissions and should be very clear and intuitive.

```suggestion
        private void CheckPermissions()
        {
            // Verify that the current user has access to edit this module
            if (ModulePermissionController.HasModuleAccess(SecurityAccessLevel.Edit, "MANAGE", this.Module))
            {
                return;
            }
            
            if (this.IsSharedViewOnly() && TabPermissionController.CanAddContentToPage())
            {
                return;
            }
            
            throw new UnauthorizedAccessException("Access denied");
        }
```

Which then begs the question, should we allow the module settings control to be viewed for a shared module to content editors? It looks wrong to me unless I am missing something.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleSettingsViewController.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+            catch (Exception exc)
+            {
+                // Log the exception
+                DotNetNuke.Services.Exceptions.Exceptions.LogException(exc);
+                throw;
+            }
+        }
+
+        private void CheckPermissions()
```
valadas: Same concern here as my previous comment on the similar code in ModuleSettingsController.cs

Also what is the difference between ModuleSettingscontroller and MOduleSettingsViewController.cs?

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+
+        public int TabId
+        {
+            get
+            {
+                return this.ModuleContext.TabId;
+            }
+        }
```
valadas: If we expose ModuleContext, do we need to also expose those properties, could we make the API surface simper by limiting the redundant properties?

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        public PortalAliasInfo PortalAlias
+        {
+            get
+            {
+                return this.PortalSettings.PortalAlias;
+            }
+        }
```
valadas: Same for PortalSetings which is already exposed, do we need those properties also exposed?

DNN Platform/DotNetNuke.Web.MvcPipeline/Extensions/StartupExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+            var allTypes = TypeExtensions.SafeGetTypes();
+            allTypes.LogOtherExceptions(Logger);
+
+            var mvcControllerTypes = allTypes.Types
+                .Where(
+                    type => typeof(IMvcController).IsAssignableFrom(type) &&
+                            type is { IsClass: true, IsAbstract: false });
+            foreach (var controller in mvcControllerTypes)
+            {
+                services.TryAddTransient(controller);
```
valadas, line 33: Should this be scoped @bdukes ?

valadas (commented):
Ok, I found a bit of time and reviewed about the first 1/4 of this PR.

I think we should bring this PR to the minimal viable MVC working state (even if incomplete) just with the minimum to load up an mvc theme and module.

MVC Pipeline and CSP are 2 big ticket items and bringing both in the same PR makes putthing this in a reviewable state hard IMO. The goal of this PR was to have a baseline so that we can review smaller additions in an easier way.

Other than that, it does not have to be in this PR but I would like that we add our stylecop rules to the new projects to enforce our coding standards on this huge amount of new code.

I'll try to review some more the next coming days.

Thanks @sachatrauwaen for this huge contribution!

sachatrauwaen:
> MVC Pipeline and CSP are 2 big ticket items and bringing both in the same PR makes putthing this in a reviewable state hard IMO. The goal of this PR was to have a baseline so that we can review smaller additions in an easier way.
 
There an separate PR for CSP on the develop branch.

But removing it from this PR needs to remove a lot of code.
The idee is to replace the CSP project of this PR with the PR on develop when the PR is merged in develop.

Actually we heave a lot of sync issues between the differnt PR's.
I don'tknow the best way. So your advice is welcome.

DNN Platform/DotNetNuke.Web.MvcPipeline/Security/Controllers/ModulePermissionsGridController.cs
```diff
@@ -0,0 +1,375 @@
...
+                if (this.permissionsList == null && this.modulePermissions != null)
+                {
+                    this.permissionsList = this.modulePermissions.ToList();
+                }
+
+                return this.permissionsList;
+            }
+        }
+
+        public ActionResult Index(int tabId, int moduleId, bool inheritViewPermissionsFromTab)
```
valadas: What protects this endpoint?

DNN Platform/DotNetNuke.Web.MvcPipeline/dnn.json
valadas: Can you explain the dnn.json files?

DNN Platform/DotNetNuke.Web.MvcPipeline/Startup.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <inheritdoc/>
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.TryAddEnumerable(new ServiceDescriptor(typeof(IRoutingManager), typeof(MvcRoutingManager), ServiceLifetime.Singleton));
+
+            services.AddMvcControllers();
+            services.AddTransient<IPageModelFactory, PageModelFactory>();
+            services.AddTransient<ISkinModelFactory, SkinModelFactory>();
+            services.AddTransient<IPaneModelFactory, PaneModelFactory>();
+            services.AddTransient<IContainerModelFactory, ContainerModelFactory>();
```
valadas: Is transient the right lifecycle for all of these

DNN Platform/DotNetNuke.Web.MvcWebsite/Controllers/ModuleActionsController.cs
```diff
@@ -0,0 +1,53 @@
...
+    using DotNetNuke.Entities.Users;
+    using DotNetNuke.Services.Localization;
+    using DotNetNuke.Services.Log.EventLog;
+    using DotNetNuke.Web.MvcWebsite.Models;
+
+    public class ModuleActionsController : Controller
+    {
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public ActionResult Delete(ModuleActionsDeleteModel model)
```
valadas: What protects this endpoint?

valadas:
> > MVC Pipeline and CSP are 2 big ticket items and bringing both in the same PR makes putthing this in a reviewable state hard IMO. The goal of this PR was to have a baseline so that we can review smaller additions in an easier way.
> 
> There an separate PR for CSP on the develop branch.
> 
> But removing it from this PR needs to remove a lot of code. The idee is to replace the CSP project of this PR with the PR on develop when the PR is merged in develop.
> 
> Actually we heave a lot of sync issues between the differnt PR's. I don'tknow the best way. So your advice is welcome.

So if that is the case, we need to review/agree on the CSP one first before spending too much time reviewing this one...

valadas:
> > MVC Pipeline and CSP are 2 big ticket items and bringing both in the same PR makes putthing this in a reviewable state hard IMO. The goal of this PR was to have a baseline so that we can review smaller additions in an easier way.
> 
> There an separate PR for CSP on the develop branch.
> 
> But removing it from this PR needs to remove a lot of code. The idee is to replace the CSP project of this PR with the PR on develop when the PR is merged in develop.
> 
> Actually we heave a lot of sync issues between the differnt PR's. I don'tknow the best way. So your advice is welcome.

So if that is the case, we need to review/agree on the CSP one first before spending too much time reviewing this one...

DNN Platform/DotNetNuke.Web.MvcPipeline/Extensions/StartupExtensions.cs
bdukes: We're using scoped for web API controllers, but transient for MVC controllers:
https://github.com/dnnsoftware/Dnn.Platform/blob/3ce24744492a57f2802db7e7b5f338a77ecc0604/DNN%20Platform/DotNetNuke.Web/Api/Extensions/StartupExtensions.cs#L33
https://github.com/dnnsoftware/Dnn.Platform/blob/3ce24744492a57f2802db7e7b5f338a77ecc0604/DNN%20Platform/DotNetNuke.Web.Mvc/Extensions/StartupExtensions.cs#L33

Scoped it probably the right way

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
sachatrauwaen: It was inspirid from PortalModuleBase. But, yes i can make the API surface simper.
I can for exemple, inspire from [DnnController ](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web.Mvc/Framework/Controllers/DnnController.cs) from mvc modules witch expose : ModuleContext, PortalSettings, ActiveModule, ActivePage and User

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
sachatrauwaen: It was inspirid from PortalModuleBase. But, yes i can make the API surface simper.
I can for exemple, inspire from [DnnController ](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/DotNetNuke.Web.Mvc/Framework/Controllers/DnnController.cs) from mvc modules witch expose : ModuleContext, PortalSettings, ActiveModule, ActivePage and User

sachatrauwaen:
To simplify the review 
1) CSP is removed from this branch (i have just leaved some comments for later integration)
2) remove module settings stuff, because actually out of scope of mvc pipeline

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleSettingsViewController.cs
sachatrauwaen: to simplify, module settings stuff is removed from the pr, because actually out of scope of mvc pipeline

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleSettingsController.cs
sachatrauwaen: to simplify, module settings stuff is removed from the pr, because actually out of scope of mvc pipeline

sachatrauwaen:
My suggestion is

> before reviewing "First code submission for MVC Pipeline", merge this PRs without review
> PRs are inter dependend and replace a lot of code and are on this branch
> So after you will review the whole with a cleaner and more definitive view

* MVC pipeline - Module control — Type: Feature [#6700](https://github.com/dnnsoftware/Dnn.Platform/pull/6700)
* MVC pipeline popups [#6723](https://github.com/dnnsoftware/Dnn.Platform/pull/6723) 
* MVC pipeline (routing / settings) [#6721](https://github.com/dnnsoftware/Dnn.Platform/pull/6721)
* MVC pipeline - module actions [#6725](https://github.com/dnnsoftware/Dnn.Platform/pull/6725)

thienvc:
This is fantastic, I've been waiting for this for a long time. It feels like DNN is experiencing a strong resurgence. I hope the core version is finalized soon so I can experience it and gradually migrate the webform modules.

thienvc:
In the classic webform version, we could use a .html skin and then DNN auto convert it to .ascx. Does the newer version still support this, as it automatically converts to .cshtml? I'm interested because I want to allow the portal admin to edit the portal theme without affecting host security.

sachatrauwaen:
> In the classic webform version, we could use a .html skin and then DNN auto convert it to .ascx. Does the newer version still support this, as it automatically converts to .cshtml? I'm interested because I want to allow the portal admin to edit the portal theme without affecting host security.

In the first version the .cshtml skin need to be created manually.
But it is an option to generate the .ascx and .cshtml from the .html

thienvc:
> > In the classic webform version, we could use a .html skin and then DNN auto convert it to .ascx. Does the newer version still support this, as it automatically converts to .cshtml? I'm interested because I want to allow the portal admin to edit the portal theme without affecting host security.
> 
> In the first version the .cshtml skin need to be created manually. But it is an option to generate the .ascx and .cshtml from the .html

Great, when deploying multiple portals on the same host, I think allowing site administrators to customize the theme (skin) in .html format is standard practice, because if they were given permission to edit the skin in .ascc or .cshtml formats It's like giving them SuperUser privileges because it's too powerful.
In the future, I think we should only allow site administrators to install standard .html skins/themes to ensure overall security.

thienvc:
That's awesome, new documentation, I follow it daily, thanks @sachatrauwaen @donker  and everyone else!
Razor+ pattern!! It makes me want to code again!
It's not that I can't write modern modules without it, it's just that with it I feel it's more synchronized, more seamless with the core, and the system feels like it has a brighter future.
I apologize for not trying to rush you, but I'm just curious about what version it will officially be released.

sachatrauwaen:
I’d like to check the status of this PR. There hasn’t been any review activity for several months—are there any issues or blockers? How can I help move things forward?

It might also be worth bringing this up during a TAG meeting ?

I completely understand that reviewing this PR is a significant effort. At the same time, on the otherhand inoticing quite a few major changes being made elsewhere, which also require substantial merge work on my side.

I believe it’s generally more efficient to process PRs in order (except for bug fixes, of course). Otherwise, each new PR tends to create additional merge overhead for those still pending.

thienvc:
> I’d like to check the status of this PR. There hasn’t been any review activity for several months—are there any issues or blockers? How can I help move things forward?
> 
> It might also be worth bringing this up during a TAG meeting ?
> 
> I completely understand that reviewing this PR is a significant effort. At the same time, on the otherhand inoticing quite a few major changes being made elsewhere, which also require substantial merge work on my side.
> 
> I believe it’s generally more efficient to process PRs in order (except for bug fixes, of course). Otherwise, each new PR tends to create additional merge overhead for those still pending.

I wish this PR was a custom package that could be installed and uninstalled like a plugin. 
If that were the case, I think it wouldn't affect the official version.

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspExtensions.cs
```diff
@@ -16,17 +16,46 @@ public static class CspExtensions
...
             csp.FrameSource.AddSelf();
             csp.FormAction.AddSelf();
             csp.ConnectSource.AddSelf();
+            csp.FrameAncestors.AddSelf();
+            csp.ObjectSource.AddNone();
             csp.BaseUriSource.AddSelf();
+
+            if (isAuthenticated)
+            {
+                csp.FrameSource.AddHost("https://dnndocs.com").AddHost("https://docs.dnncommunity.org");
```
mitchelsellers: I don't think this should be a global approval, this will result in some really funny security responses most likely

DNN Platform/DotNetNuke.Web.Mvc/StandardTabAndModuleInfoProvider.cs
```diff
@@ -99,6 +99,11 @@ private static int FindInt(HttpRequestBase requestBase, string key)
                 value = requestBase.Headers[key];
             }
 
+            if (requestBase.Form[key] != null)
```
mitchelsellers: I would like to review/discuss the possible risks wit htis change, by accpeting the value from the form in any situation , as I believe this might expose a different method of manipulation that could have unintended consequences.

DNN Platform/DotNetNuke.Web.MvcPipeline/Commons/PropertyHelper.cs
```diff
@@ -0,0 +1,156 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Web.MvcPipeline.Commons
```
mitchelsellers: This entire implementation already exists within the solution at `DotNetNuke.Web.Mvc.Common`

Should this be a move?  Should the old file be marked as Obsolete?

DNN Platform/DotNetNuke.Web.MvcPipeline/Commons/TypeHelper.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.Web.MvcPipeline.Commons
```
mitchelsellers: To match other naming conventions if needed this namespace/folder should be `Common` not `Commons`

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    using System;
+    using System.Web.Mvc;
+
+    using DotNetNuke.Entities.Portals;
+
+    /// <summary>
+    /// Base controller for DNN MVC page controllers, exposing common services and portal context.
+    /// </summary>
+    public abstract class DnnPageController : Controller, IMvcController
```
mitchelsellers, line 18: Similar question here for the fate of the existing IDNNController and related items, as I can see a lot of confusion without some level of guidance being provided at a programming perspective with different behaviors and purposes.  For example this ONLY provides PortalSettings the legacy one provides much more rich context (User, etc)

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
```diff
@@ -0,0 +1,406 @@
...
+            {
+                return this.GetType().Name.Replace("ViewController", string.Empty);
+            }
+        }
+
+        /// <summary>
+        /// Builds an edit URL for the current module instance.
+        /// </summary>
+        /// <returns>The edit URL.</returns>
+        public string EditUrl()
```
mitchelsellers: I echo @valadas comments on this duplication of exploding items that are already there, the long-term maintenance of this, as we have seen with other changes, is not exactly "fun".  I would prefer exposing the ModuleContext and leaving it at that personally

DNN Platform/DotNetNuke.Web.MvcPipeline/Framework/JavascriptLibraries/MvcJavaScript.cs
```diff
@@ -0,0 +1,338 @@
...
+
+            var controller = GetClientResourcesController();
+            controller.CreateScript(GetScriptPath(jsl, HttpContextSource.Current?.Request))
+                  .SetPriority(GetFileOrder(jsl))
+                  .SetProvider(GetProvider(jsl))
+                  .SetNameAndVersion(jsl.LibraryName, jsl.Version.ToString(3), false)
+                  .Register();
+
+            /*
+            if (Host.CdnEnabled && !string.IsNullOrEmpty(jsl.ObjectName))
```
mitchelsellers, line 265: This commented out code should be removed

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspExtensions.cs
sachatrauwaen: This urls are used by some helpurl of dnn core modules with are loaded in a iframe.
In place of add them globaly to csp policy, maybe they can be added only on the help page.
This this solve your concerns ? Or is it more about the url it self ?

valadas: Do we actually load any resources from dnndocs? If so maybe we should rethink that as many implementations won't even be able to.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
sachatrauwaen: This one is indended to work in mvc pipeline, real mvc. IDNNController is for webforms , have a reference to a Page. Dous not use standard mvc. Use hacks in mvc to write the output to a string and generate module actions in a way that is not compatible with real mvc.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
sachatrauwaen: This a base controller is specifically for module controls implemented as a child controller. But is in reality not realy the recommended way because removed by MS in dotnet core. Maybe better to remove it ?

DNN Platform/DotNetNuke.Web.MvcPipeline/Framework/JavascriptLibraries/MvcJavaScript.cs
sachatrauwaen: this commended code, use DnnJsIncludeFallback with is actually not implemented in the mvc pipeline. Do you think this have some sense this days ? Is it needed to be implement in the mvc pipeline ?

DNN Platform/DotNetNuke.Web.Mvc/StandardTabAndModuleInfoProvider.cs
sachatrauwaen: because this, can be avoided and is based on convensions that are not common in mvc , i remove the change and avoid  unintended consequences

DNN Platform/DotNetNuke.Web.MvcPipeline/Commons/TypeHelper.cs
sachatrauwaen: i remove it and use mvc project version.

DNN Platform/DotNetNuke.Web.MvcPipeline/Commons/PropertyHelper.cs
sachatrauwaen: removed and use the mvc version

sachatrauwaen:
> > I’d like to check the status of this PR. There hasn’t been any review activity for several months—are there any issues or blockers? How can I help move things forward?
> > It might also be worth bringing this up during a TAG meeting ?
> > I completely understand that reviewing this PR is a significant effort. At the same time, on the otherhand inoticing quite a few major changes being made elsewhere, which also require substantial merge work on my side.
> > I believe it’s generally more efficient to process PRs in order (except for bug fixes, of course). Otherwise, each new PR tends to create additional merge overhead for those still pending.
> 
> I wish this PR was a custom package that could be installed and uninstalled like a plugin. If that were the case, I think it wouldn't affect the official version.

We look at it, but not possible do the the use of lot of code that is not exposed for external project and need some UI changes in the core

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspExtensions.cs
mitchelsellers: My larger concern is that its included anywhere, rather than where only needed, so yes, if we could get it added into that module that woudl help.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
mitchelsellers: @donker @valadas @bdukes @david-poindexter given the above comment from @sachatrauwaen do we need to think about the addition of an Obsolete or other attribute on the others?  I'm worried about multiple items that do similar things and those not being aware of what exactly this is, or this important difference

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
mitchelsellers: I think in many ways, yes, it would be a "bigger lift" for somone converting an existing module, but at the same time, very easy document using ModuleContext

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspExtensions.cs
sachatrauwaen: it's moved to the help control

bdukes: Should these be added dynamically, based on the help URL of the module? Will this still fail for 3rd party modules that have help content on other domains?

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets the dependency injection service provider for the current request.
+        /// </summary>
+        public IServiceProvider DependencyProvider { get; private set; }
+
+        /// <summary>
+        /// Gets the current portal settings.
+        /// </summary>
+        public PortalSettings PortalSettings
```
bdukes: I would prefer that we only expose `IPortalSettings` (i.e. via the interface)

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
bdukes: Are you suggesting that we mark the current MVC pattern as obsolete?

mitchelsellers: Maybe?  I'm worried that with multiple ways of doing something, and/or without doing so, it is going to create even more confusion? Maybe it's too early for that?

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspExtensions.cs
sachatrauwaen: > Should these be added dynamically, based on the help URL of the module? Will this still fail for 3rd party modules that have help content on other domains?

I changed it that way. Dynamic add domain of helpurl.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
bdukes: Yeah, I think it's too early for that. We're just going to have to live with two MVC patterns side-by-side for a while.

DNN Platform/DotNetNuke.Web.MvcPipeline/Framework/JavascriptLibraries/MvcJavaScript.cs
bdukes: This may have already been lost with @donker's CDF revamp.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
sachatrauwaen: In the mvc pipeline there is a way to run actual mvc modules as child controllers. But with some limitations. And when looking at the way actuel mvc modules are implemented is i think the only way that was possible in mvc 3. It's really bad hacking witch bring a lot of restructions to the real MS MVC. So opinion, is that depreciating the actual mvc modules is not a bad thing. Also because this patern can not work in dotnet core (because in use child controllers witch is not available in dotnet core). 

Depreciating the actual mvc modules, can also make this PR smalller.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
sachatrauwaen: I have tryed, but endup in a nogo, because ActiveTab was not exposed thue IPortalSettings and found no way to get ActiveTab an oter way. Or do you know a way to get ActiveTab without PortalSettings ?

bdukes: [`TabController.CurrentPage`](https://docs.dnncommunity.org/api/DotNetNuke.Entities.Tabs.TabController.html#DotNetNuke_Entities_Tabs_TabController_CurrentPage) is what I've been using. We want to have a better way to get that context, but it hasn't been fully designed yet.

sachatrauwaen: I try againt to go for IPortalSettings but it need a lot of changes in cascade :
- HtmlTextContoller.FormatHtmlText
- HtmlTokenReplace
- TokenContext

It not realy cause breaking changes because IPortalSetings can accept Portalsettings parameters.
But is it a good idea to to all this changes inside this PR ?

Other issue is that IPortalSettings dous not implement IPropertyAccess in TokenReplace
this.AddPropertySource("portal", this.PortalSettings);
and also missing ActiveTab
this.AddPropertySource("tab", this.PortalSettings.ActiveTab);
So it need a more in detail restructuration of TokenReplace

To minimize the size of this PR i can use IPortalSettings in the controller and fallback to PortalSettings.Current when there is an issue with a class outside of the mvc pipeline project.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/ModuleViewControllerBase.cs
sachatrauwaen: I remove the class, it is actually not used inside the mvc pipeline and the idea is to recommend the Razor+ pattern over Child Controller. So less code to maintain.

DNN Platform/DotNetNuke.Web.MvcPipeline/Controllers/DnnPageController.cs
bdukes: That sounds like a good compromise, thanks!

sachatrauwaen: I apply it but a lot of PortalSettings.Current was needed.

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/Mvc/PersonaBarContainerController.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    using Globals = DotNetNuke.Common.Globals;
+
+    /// <summary>
+    /// Controller for managing the Persona Bar container in the DNN platform.
+    /// </summary>
+    public class PersonaBarContainerController : Controller
+    {
+#pragma warning disable CS0618 // Type or member is obsolete
+        private readonly IPersonaBarContainer personaBarContainer = Dnn.PersonaBar.Library.Containers.PersonaBarContainer.Instance;
```
mitchelsellers: This line and all others that are referencing Obsolete should be updated and the `#pragma` items go away

DNN Platform/Modules/Samples/Dnn.ContactList.Mvc/RouteConfig.cs
```diff
@@ -10,7 +10,7 @@ public class RouteConfig : IMvcRouteMapper
     {
         public void RegisterRoutes(IMapRoute mapRouteManager)
         {
-            mapRouteManager.MapRoute("ContactList", "ContactList", "{controller}/{action}", new[]
+            mapRouteManager.MapRoute("Dnn/ContactList", "ContactList", "{controller}/{action}", new[]
```
mitchelsellers, line 13: Why was this changed?  Was it broken before or does this create a needed breaking change for MVC modules?

Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/Mvc/PersonaBarContainerController.cs
sachatrauwaen: done

DNN Platform/Modules/Samples/Dnn.ContactList.Mvc/RouteConfig.cs
sachatrauwaen: 1) This parameter is foldername witch is Dnn/ContactList (defined in the module manifest).
2) In the mvc pipeline old mvc modules run as child controllers and use mvc routing (witch is not realy the case in webforms). Child controllers can not use namespace routing but only area route value to differenciate 2 controllers with same name. So this first parameter is used as area name and it is required that it is equal to folder name of module definition.
3) As say before, the actual mvc modules contain big hack in the standard mvc. It can be an option, not to support actual mvc modules in mvc pipeline. I personnaly esitating.

bdukes: are you saying that the sample module is currently broken without these changes? or did you adjust it for the MVC pipeline in a way that required these changes?

sachatrauwaen: it is adjusted for the mvc pipeline and work also outside the mvc pipeline (in webforms)

DNN Platform/DotNetNuke.Web.MvcPipeline/Containers/ContainerHelpers.Content.cs
```diff
@@ -0,0 +1,97 @@
...
+            try
+            {
+                moduleControl = ModuleControlFactory.CreateModuleControl(model.ModuleConfiguration, model.ModuleConfiguration.ModuleControl.ControlSrc);
+                moduleDiv.InnerHtml += htmlHelper.Control(moduleControl);
+            }
+            catch (Exception ex)
+            {
+                if (TabPermissionController.CanAdminPage())
+                {
+                    moduleDiv.InnerHtml += htmlHelper.ModuleErrorMessage(ex.Message, "Error loading module");
```
mitchelsellers, line 54: This should be localized for replacement

mitchelsellers: Possibly look for others as well

DNN Platform/DotNetNuke.Web.MvcPipeline/ModuleControl/MvcModuleControl.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            get
+            {
+                var segments = this.GetSegments();
+                if (segments.Length < 2)
+                {
+                    return string.Empty;
+                }
+
+                return segments[1];
```
dimarobert, line 95: I think this needs to be
```suggestion
                if (segments.Length == 3)
                {
                    return segments[1];
                }
                
                return segments[0];
```

formats for the ControlSrc are:
- `{ControllerName}/{ActionName}.mvc` - segments[0] needed here
- `{Namespace}/{ControllerName}/{ActionName}.mvc` - segments[1] needed here

Also, probably `RouteNamespace` below is valid only when `segments.Length == 3`.

dimarobert:
A took another look at whether there could be a way to support async MVC Modules with the new pipeline. The issue that we come to in [this thread](https://github.com/dnnsoftware/Dnn.Platform/pull/7193#issuecomment-4215415592), that prevents us from doing it boils down to the fact that the entire logic of the MVC Module execution in the new pipeline happens during the Razor View render phase of the page through a call to `HtmlHelper.Action(controllerName, actionName, routeValues)` and that the entire rendering phase in ASP.NET MVC is synchronous.

So my next thought was, instead of running this pipeline as it is running now:
1. Run **page controller phase** (`DefaultController.Page()` method) which builds the PageModel and results in a ViewResult.
2. Run the **page view rendering phase**, which uses the PageModel to output the HTML content of the page.
3. **For each** MVC Module on the page, during this phase, start yet another complete MVC sub-pipeline run (controller and view phase) using `HtmlHelper.Action()` which captures the result into a StringWriter and then appends it to the page output. (also generating a bunch of intermediary strings in memory in the process)

What if we could run it in the proper order, all the Controller phases first, Page controller then each Module controller, then similarly the View rendering phases?

To sanity check and see if it could maybe work from the point of view of what the APIs in ASP.NET MVC provide us, I worked through a possible implementation architecture / description with a bit of help from AI. The following proposal was the result of that after a couple of iterations, mostly output from the AI, with some of my observations as "[My note here: ...]" sections:

https://gist.github.com/dimarobert/ec0212c6de66085ba890187d1ca32368

I think that something like that could really work and it would even improve the order of execution a bit. Allowing even Module results that can redirect the request, coming in handy when implementing a [PRG pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get).

sachatrauwaen:
> What if we could run it in the proper order, all the Controller phases first, Page controller then each Module controller, then similarly the View rendering phases?

Thanks for your comment. I have no easy response to your proposition. For sure there are some benefits. For now i give you only some general information from what i know.

- For sure async is the future but for async to be effetive it need to be used over whole line witch is difficult in .net framework and dnn. And not used the right way there can be locking problems in .net framework.
- One of the main goal of the mvcpipeline is to get shorter to .net core where ViewComponent sound like the best way to implement Module Controls.
- Actually the MVC module pattern of dnn is not the one best integrates in the MVC pipeline, because it trys to run a mvc controller as a child controller witch need routing with area witch was not intended to run like this in the webforms pipeline. And the new Razor pipeline fit most of the features of the actual dnn mvc module pattern.
- Initially a was thinking about a 3th alternative, where the controller phase will build the model of the page and also the models from all modules of the page and the View rendering then only render the layout and all modules models. Witch stay in a standard mvc concept of MVC (controller create the models, the views render the model).
- Actually we just hope this will be reviewed at some time by the core team to be integrated in a feature branch.  I don't think we will big restructuration before that.

dimarobert:
> Initially a was thinking about a 3th alternative, where the controller phase will build the model of the page and also the models from all modules of the page and the View rendering then only render the layout and all modules models. Witch stay in a standard mvc concept of MVC (controller create the models, the views render the model).

That is exactly what my proposed alternative above tries to do. It also gets rid of:
- the need for a child controller phase during the View phase.
- I do not think it needs routing with area (you just need the ControllerName and the ActionName from the ModuleControl definition to know what method to call)
- the need of rendering the modules to a bunch of StringWriters that are then appended to the page output in [MvcModuleControl.cs](https://github.com/dnnsoftware/Dnn.Platform/blob/feature/mvc-pipeline-old/DNN%20Platform/DotNetNuke.Web.MvcPipeline/ModuleControl/MvcModuleControl.cs#L223). (probably pretty costly as the output content of each module is manipulated at least 3 times before reaching the network).

Using the ViewComponent in net core or the current Razor pipeline for modules in DNN (basically module being able to run business logic only in the View phase) might work for modules that only render a shim HTML element and relies on a client-side js library (e.g. angular, react) to bootstrap the data, which would be retrieved through a secondary WebAPI call. But that can easily end up with a rendered page that is full of loading animations for each module, or maybe worse, a mostly empty page that starts filling up module by module. For modules that need to actually query and render data at page load (e.g. display a data table), having async would free up the request WorkerThread to handle other requests while the async I/O operation runs. It would be even more important for scenarios where MVC Views would be used for POSTing data back to the server, as that data would need to be processed, which is definitely slower.

> For sure async is the future but for async to be effetive it need to be used over whole line witch is difficult in .net framework and dnn. And not used the right way there can be locking problems in .net framework.

I am not sure this is true, async is safe in net framework and it is properly implemented in all three rendering technologies. MVC and WebApi are probably the safest, but WebForms is good as well. ASP.NET handles everything so that the execution can be async from top to bottom in the application code itself and you would never need to call `task.Result`, `task.Wait()`, or `task.GetAwaiter().GetResult()`, which are the APIs that on netfw could end up deadlocking on the SynchronizationContext. If you end up in a situation that you need to do that, then yes, it is very bad.
The same is equally true for net core if async methods are used from synchronous code and one of the three APIs above are used to wait their execution. There is indeed no SynchronizationContext, but you can as easily end up deadlocking if you do that.

PR #6752, opened 2025-10-03
Content security policy headers (CSP)

### LINKED ISSUE #6720: [Enhancement]: Content security policy headers

### Description of problem
Create a possibility to generate Content security policy (csp) headers .
Content Security Policy is a crucial security standard that helps protect your web applications from various types of attacks, including Cross-Site Scripting (XSS), clickjacking, and other code injection attacks. It works by allowing you to specify which resources (scripts, styles, images, etc.) your browser should be allowed to load.
[more info about Content security policy](https://centralcsp.com/docs/what-is-csp)

### Description of solution
3 http headers will be managed
- Content-Security-Policy
- Content-Security-Policy-Report-Only
- Reporting-Endpoints

Portal Settings
- activate csp : Enabled/Report Only/Disabled.
- default csp policy.
- activate reporting endpoint with 3 kind of reporting : email, log file, DNN Event logs.
- custom Reporting url

Image

Risk for DDOS attack on this endpoint, we to manage some  rate limits on this reporting endpoint.
There is also need to write some documentation for module developers and skin developers to tell how to contribute to the CSP policy.

### Suggested Implementation
A new CSP service where everyone (dnn core, skins, modules) can contribute to csp.
It is also responsible for generating a nonce for inline js.

Every inline javascript need to get a nonce attribute.

In the webforms pipeline we need to accept inline and eval js.
In the mvc pipeline we can be strict.

Analyze if all resources added by Client Resources Manager can be added automatically.

- [x] Yes

A first version of it exists already in the MVC Pipeline.
[DNN Platform/DotNetNuke.ContentSecurityPolicy](https://github.com/dnnsoftware/Dnn.Platform/tree/ae00a1508e4cae60c02a8694e3dfd512dd8c5500/DNN%20Platform/DotNetNuke.ContentSecurityPolicy)

- [ ] Yes

### Code of Conduct

sachatrauwaen:
### Summary
Create a possibility to generate Content security policy (csp) headers .

Content Security Policy is a crucial security standard that helps protect your web applications from various types of attacks, including Cross-Site Scripting (XSS), clickjacking, and other code injection attacks. It works by allowing you to specify which resources (scripts, styles, images, etc.) your browser should be allowed to load.
[more info about Content security policy](https://centralcsp.com/docs/what-is-csp)

### Description of solution

The intend is to manage CSP for WebForms and MVC pipeline.
- Webforms can not be very struct in the policy that can be used. Here  script-src 'unsafe-inline' and 'unsafe-eval' will be automatically added to csp. It is not added in the settings because it's specific for webforms.

- In the future MVC pipeline can be very strict. Here no js evaluation will be used and all inline javascript will be marked with a nonce.

This is the default csp for the setting 

```default-src 'self'; script-src 'self' 'report-sample';  style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self'; object-src 'none'; base-uri 'none'; form-action 'self'; frame-ancestors 'none'; frame-src 'self'; connect-src 'self';```

- It dous not accept external resources.
- It accept inline css : actually used by the default page of dnn and generated by ckeditor with default configuration. But it is not recomended. ckeditor can be configurated to not use inline css. 
- It accept image src starting with data: bacause it is used by default dnn page. But it is not recomended.

3 http headers will be managed : Content-Security-Policy, Content-Security-Policy-Report-Only and Reporting-Endpoints

### Persona bar - Security settings

image

### Implementation
It commes in a new project for csp management and a test project.

The IContentSecurityPolicy service that can be used with DI had all the stuff for skin and module developers to contribute to the policy.

For webforms skin developers actually the way to contribute is
```
<script runat="server">
    protected void Page_Init()
    {
        var defaultPage = (DotNetNuke.Framework.DefaultPage)this.Page;
        defaultPage.AddCsp("script-src https://ajax.aspnetcdn.com");
    }
</script>
```

## Details of DotNetNuke.ContentSecurityPolicy library

The `DotNetNuke.ContentSecurityPolicy` library provides a fluent API for building and emitting Content Security Policy (CSP) headers in DNN. The `IContentSecurityPolicy` interface is the main entry point to compose directives, manage sources, configure reporting, and generate final header strings.

### Interface: `IContentSecurityPolicy`
Namespace: `DotNetNuke.ContentSecurityPolicy`

### Properties
- **Nonce**: Cryptographically secure nonce value to use with inline script/style tags.
- **DefaultSource**: `SourceCspContributor` for `default-src`.
- **ScriptSource**: `SourceCspContributor` for `script-src`.
- **StyleSource**: `SourceCspContributor` for `style-src`.
- **ImgSource**: `SourceCspContributor` for `img-src`.
- **ConnectSource**: `SourceCspContributor` for `connect-src`.
- **FontSource**: `SourceCspContributor` for `font-src`.
- **ObjectSource**: `SourceCspContributor` for `object-src`.
- **MediaSource**: `SourceCspContributor` for `media-src`.
- **FrameSource**: `SourceCspContributor` for `frame-src`.
- **FrameAncestors**: `SourceCspContributor` for `frame-ancestors`.
- **FormAction**: `SourceCspContributor` for `form-action`.
- **BaseUriSource**: `SourceCspContributor` for `base-uri`.

### Methods
- **RemoveScriptSources(CspSourceType cspSourceType)**: Remove script sources of the specified type (e.g., `Inline`, `Self`, `Nonce`).
- **AddPluginTypes(string value)**: Add values for `plugin-types` (e.g., `application/pdf`).
- **AddSandboxDirective(string value)**: Add `sandbox` options (e.g., `allow-scripts allow-same-origin`).
- **AddFormAction(CspSourceType sourceType, string value)**: Add a `form-action` source.
- **AddFrameAncestors(CspSourceType sourceType, string value)**: Add a `frame-ancestors` source.
- **AddReportEndpoint(string name, string value)**: Add a named reporting endpoint.
- **AddReportTo(string value)**: Add a `report-to` group name to the policy.
- **AddHeaders(string cspHeader)**: Parse and merge a CSP header string; returns the same `IContentSecurityPolicy` for chaining.
- **GeneratePolicy()**: Build the `Content-Security-Policy` header value.
- **GenerateReportingEndpoints()**: Build the reporting header value(s).
- **UpgradeInsecureRequests()**: Add the `upgrade-insecure-requests` directive.

### Working with sources
Directive properties expose a `SourceCspContributor`, which supports adding/removing sources such as:
- `AddSelf()` → `'self'`
- `AddNone()` → `'none'`
- `AddInline()` → `'unsafe-inline'`
- `AddEval()` → `'unsafe-eval'`
- `AddStrictDynamic()` → `'strict-dynamic'`
- `AddNonce(string)` → `'nonce-<value>'`
- `AddHash(string)` → `'sha256-...'`, `'sha384-...'`, `'sha512-...'`
- `AddHost(string)` → `example.com`, `https://cdn.example.com`
- `AddScheme(string)` → `https:`, `data:`, `blob:`
- `RemoveSources(CspSourceType)` to remove by type

See: `CspSourceType.cs`, `CspSource.cs`, `SourceCspContributor.cs`.

### Configure a baseline policy with a nonce
```csharp
using DotNetNuke.ContentSecurityPolicy;

public class CspExample
{
    private readonly IContentSecurityPolicy _csp;

    public CspExample(IContentSecurityPolicy csp)
    {
        _csp = csp;
    }

    public void Configure()
    {
        // Default baseline
        _csp.DefaultSource.AddSelf();
        _csp.ScriptSource.AddSelf().AddNonce(_csp.Nonce);
        _csp.StyleSource.AddSelf();
        _csp.ImgSource.AddSelf().AddScheme("data:");

        // Lock down frames and forms
        _csp.FrameAncestors.AddNone();
        _csp.FormAction.AddSelf();

        // Reporting
        _csp.AddReportEndpoint("csp-endpoint", "/api/csp/report");
        _csp.AddReportTo("csp-endpoint");

        // Optionally upgrade insecure requests
        _csp.UpgradeInsecureRequests();

        // Generate header values
        var cspHeader = _csp.GeneratePolicy();
        var reportingHeader = _csp.GenerateReportingEndpoints();

    }
}
```

### Parse and merge an existing CSP header
```csharp
_csp.AddHeaders("default-src 'self'; img-src 'self' data:")
    .ScriptSource.AddNonce(_csp.Nonce);

var headerValue = _csp.GeneratePolicy();
```

### Remove an unsafe source
```csharp
_csp.RemoveScriptSources(CspSourceType.Inline);
```

## Notes
- Nonce: use `Nonce` in your inline tags: `<script nonce="{policy.Nonce}">`.
- Reporting: ensure your endpoint exists to accept violation reports.
- Parsing: `AddHeaders` is useful to import settings from configuration and extend them programmatically.
- This is not only about js and css but also about images, iframes and more (parts of the content).

## Update 22 october 2025

There are 2 point of views :
1) The site settings are there for the site admin to enforce some policies. In this case, the policy need to include all requirements (from skins, modules, skin object, content, ...) and all pages (not only dnn pages but also edit urls need to be included in this policies to get them work properly). Actually this can already done with a the web.config setting. Or you need different policies for different use cases/ user profils (Anonymous, loged in, page admin, module editor, host, ...). 

2) (The point of view of this PR) The site settings are there for defining a starting policy where skins, modules, ... can add dynamically there policy. So the policy will be adjusted automatically to make the site work.

What is sure is that modules and skins will paticipate in the ability to make the policies strict.
Knowing witch policies are required by a skin or modules is key even what kind of content is permited, if you want to be strict.

Know that browsers include also a way to report csp policy violations to the browser console and to a endpoint.

Fixes #6720


mitchelsellers:
With this, how exactly does the configured process in the persona bar work with the fact that anyone could manipulate this via other methods in the skin, module, etc?

Is the portal done first and then changes applied?  

How is the users input validated in the personarbar validated?  is it serialized to an object with detailed validation errors?

sachatrauwaen:
> With this, how exactly does the configured process in the persona bar work with the fact that anyone could manipulate this via other methods in the skin, module, etc?
> Is the portal done first and then changes applied?

Yes, the portal settings are applied first in the OnInit of the page (default.aspx).
Then skin and modules can add there policy. Normally skins and modules will only add addional policy.
And finally in the OnPreRender of the page the full policy string is generated and added to the http response header.

> How is the users input validated in the personarbar validated? is it serialized to an object with detailed validation errors?

The library containt a policy parser. But actully no validation error is showed to the user that enter the settings.
I will look to add this.

mitchelsellers:
Argh, sorry about that!, deleted and moved

donker:
@sachatrauwaen: can you add a switch for the admin that governs whether a module can enlarge the CSP scope or not? That way the admin decides what is allowed and knows what is going to happen.

sachatrauwaen:
> @sachatrauwaen: can you add a switch for the admin that governs whether a module can enlarge the CSP scope or not? That way the admin decides what is allowed and knows what is going to happen.

In reality there 2 ways to add policies to CSP : 1) site settings 2) api (used by the core, modules, skins, editor providers  and potentialy all other extensions).

The site settings where actually for define the default strict policy, policy for content (images, iframe, css) and policy for modules that not automatically add there policy.

When we want to add a "switch for the admin that governs whether a module can enlarge the CSP scope or not" it is for all policies added by api not only modules. So we add only the policy from the site settings.

sachatrauwaen:
New Sitting is added

image

DNN Platform/Website/Default.aspx.cs
```diff
@@ -177,6 +184,11 @@ protected string HtmlAttributeList
 
         private IPortalAliasInfo PrimaryPortalAlias => this.PortalSettings.PrimaryAlias;
 
+        public void AddCsp(string policy)
+        {
+            this.ContentSecurityPolicy.AddHeader(policy);
+        }
```
valadas: Does this method need to be public?

DNN Platform/DotNetNuke.ContentSecurityPolicy/SourceCspContributor.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <returns>The current instance for method chaining.</returns>
+        public SourceCspContributor AddInline()
+        {
+            return this.AddSource(new CspSource(CspSourceType.Inline));
+        }
+
+        /// <summary>
+        /// Ajoute une source 'self' qui autorise les ressources de la même origine.
+        /// </summary>
+        /// <returns>L'instance courante pour chaîner les méthodes.</returns>
```
valadas: This documentation appears to be in french, we usually do it in english for a wider audience of developers

valadas: Same for multiple others in this file

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicy.cs
```diff
@@ -0,0 +1,476 @@
...
+        /// Gets a cryptographically secure random nonce value for use in CSP policies.
+        /// </summary>
+        public string Nonce
+        {
+            get
+            {
+                if (this.nonce == null)
+                {
+                    var nonceBytes = new byte[32];
+                    var generator = System.Security.Cryptography.RandomNumberGenerator.Create();
```
valadas: Should we create a single instance of the RandomNumberGenerator at the class level so we dont do a new one each time a resource is added. This service is registered as scoped so that way we could reuse that existing instance if many resources get added in the same request.

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspDirectiveType.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace DotNetNuke.ContentSecurityPolicy
+{
+    /// <summary>
+    /// Represents different types of Content Security Policy directives.
+    /// </summary>
+    public enum CspDirectiveType
+    {
+        /// <summary>
+        /// Directive qui définit la politique par défaut pour les types de ressources non spécifiés.
```
valadas: This class also has most of its documentation in french

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspParsingExample.cs
```diff
@@ -0,0 +1,104 @@
...
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.ContentSecurityPolicy
+{
+    using System;
+
+    /// <summary>
+    /// Example class demonstrating how to parse Content Security Policy headers.
+    /// </summary>
+    public static class CspParsingExample
```
valadas: Can we move usage examples to the tests project maybe, or maybe tests are good enough examples?

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspPolicyExample.cs
```diff
@@ -0,0 +1,39 @@
...
+// See the LICENSE file in the project root for more information
+
+namespace DotNetNuke.ContentSecurityPolicy
+{
+    using System;
+
+    /// <summary>
+    /// Démontre l'utilisation de la Content Security Policy en configurant différentes directives.
+    /// </summary>
+    public class CspPolicyExample
```
valadas: Same here, maybe tests could replace examples here...

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicy.cs
mitchelsellers: Additionally, RandomNumberGenerator is a class that implements IDisposable, it should at minimum be within a `using` to prevent resource leaking

sachatrauwaen: > Should we create a single instance of the RandomNumberGenerator at the class level so we dont do a new one each time a resource is added. 

Because it is scoped, a new one created for each request only (not for each resource added) and if Nonce is used.

And because it  implements IDisposable (as Mich say) if it will be create at class level, it will be never disposed.

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspParsingExample.cs
sachatrauwaen: i remove the examples

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicy.cs
sachatrauwaen: i add using

DNN Platform/DotNetNuke.ContentSecurityPolicy/SourceCspContributor.cs
sachatrauwaen: fixed

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// </summary>
+        private string ValidateHostSource(string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                throw new ArgumentException("Host source cannot be empty");
+            }
+
+            // Basic domain validation
+            var domainRegex = new Regex(@"^(https?://)?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?(/.*)?$");
```
valadas: I am no expert in CSP but I think this regex might be wrong and maybe not optimal for performance too. I believe HostSource should support extensionless domains like `localhost` and wildcards like `*.somecdn.com` or simply `*`, domains with long TLDs like `something.technology`, IP and port, IPV6, etc. So it that should be a supported scenario I would love that we have some tests for a list of those that should be supported.

Maybe we could allow some of CSP specific scenarios like "*" and similar, then run Uri.TryCreate(value) to validate the more normal ones and if it still does not pass, we can do IPAddress.TryParse(value)?

DNN Platform/DotNetNuke.ContentSecurityPolicy/DocumentCspContributor.cs
```diff
@@ -0,0 +1,114 @@
...
+                    // Add more specific validations as needed
+            }
+        }
+
+        /// <summary>
+        /// Validates plugin types.
+        /// </summary>
+        private void ValidatePluginTypes(string value)
+        {
+            string[] validPluginTypes = { "application/pdf", "image/svg+xml" };
```
valadas: Is this correct to hardcode only those mime-types, I am not sure but should this allow other types like word/excel, etc.?

DNN Platform/DotNetNuke.ContentSecurityPolicy/IContentSecurityPolicy.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <summary>
+        /// Parses a CSP header string into a ContentSecurityPolicy object.
+        /// </summary>
+        /// <param name="cspHeader">The CSP header string to parse.</param>
+        /// <returns>A ContentSecurityPolicy object representing the parsed header.</returns>
+        /// <exception cref="System.ArgumentException">Thrown when the CSP header is invalid or cannot be parsed.</exception>
+        IContentSecurityPolicy AddHeader(string cspHeader);
+
+        /// <summary>
+        /// Ajoute une directive de rapport à la politique.
```
valadas: This file also has some french documentation

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingCspContributor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Gets collection of reporting endpoints.
+        /// </summary>
+        private List<string> ReportingEndpoints { get; } = new List<string>();
+
+        /// <summary>
+        /// Adds a reporting endpoint.
+        /// </summary>
+        /// <param name="endpoint">L'URL de l'endpoint où envoyer les rapports.</param>
```
valadas: This file also has some french documentation

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingCspContributor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        private void ValidateReportUri(string endpoint)
+        {
+            if (string.IsNullOrWhiteSpace(endpoint))
+            {
+                throw new ArgumentException("Reporting endpoint cannot be empty");
+            }
+
+            // URL validation regex
+            var urlRegex = new Regex(@"^(https?://)?([a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)*)?(:\d+)?(/.*)?$");
```
valadas: Maybe we could use Uri.TryParse here too to validate without a complex regex?

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingEndpointContributor.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        /// <summary>
+        /// Gets collection of reporting endpoints.
+        /// </summary>
+        private Dictionary<string, string> ReportingEndpoints { get; } = new Dictionary<string, string>();
+
+        /// <summary>
+        /// Adds a reporting endpoint.
+        /// </summary>
+        /// <param name="name">Le nom de l'endpoint où envoyer les rapports.</param>
```
valadas: This file also has some french documentation

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingEndpointContributor.cs
```diff
@@ -0,0 +1,116 @@
...
+            switch (this.DirectiveType)
+            {
+                case CspDirectiveType.ReportUri:
+                    this.ValidateReportUri(value);
+                    break;
+                case CspDirectiveType.ReportTo:
+                    this.ValidateReportTo(value);
+                    break;
+
+                    // Add more specific validations as needed
```
valadas: Was somthing missed after this comment?

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingEndpointContributor.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        private void ValidateReportUri(string endpoint)
+        {
+            if (string.IsNullOrWhiteSpace(endpoint))
+            {
+                throw new ArgumentException("Reporting endpoint cannot be empty");
+            }
+
+            // URL validation regex
+            var urlRegex = new Regex(@"^(https?://)([a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)*)?(:\d+)?(/.*)?$");
```
valadas: Maybe we could use URI.TryParse also?

DNN Platform/DotNetNuke.ContentSecurityPolicy/dnn.json
```diff
@@ -0,0 +1,16 @@
+{
```
valadas: What is the purpose of this file?

DNN Platform/Website/Default.aspx.cs
```diff
@@ -225,6 +232,31 @@ protected override void OnInit(EventArgs e)
             // set global page settings
             this.InitializePage();
 
+            if (!this.PortalSettings.CspHeaderFixed &&
```
valadas: The code for when CspHeaderFixed is true is pretty far from another block of code that handles when it is false. In am not sure if it needs to be that way because there is an important order or operations, but if not, can we possibly bring those together in a private method and improve the readability of this file?

valadas (changes requested):

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicy.cs
```diff
@@ -0,0 +1,479 @@
...
+
+    /// <summary>
+    /// Manages the entire Content Security Policy.
+    /// </summary>
+    public class ContentSecurityPolicy : IContentSecurityPolicy
+    {
+        private string nonce;
+
+        /// <summary>Initializes a new instance of the <see cref="ContentSecurityPolicy"/> class.</summary>
+        public ContentSecurityPolicy()
```
mitchelsellers, line 30: Given that we don't need to do anything special here with the constructor, this should be omitted

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicyParser.cs
```diff
@@ -0,0 +1,302 @@
...
+                contributor.AddHost(trimmedSource);
+            }
+        }
+
+        /// <summary>
+        /// Checks if a string represents a scheme.
+        /// </summary>
+        private static bool IsScheme(string source)
+        {
+            string[] knownSchemes = { "http:", "https:", "data:", "blob:", "filesystem:", "wss:", "ws:" };
```
mitchelsellers: For consistency in naming this should be `var knownSchemes`

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
mitchelsellers: Additionally, this is in a hot path for execution, compiling this regex at minimum should be done for performance, and limit processing time as well

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            return value.StartsWith("http") ? value : $"https://{value}";
+        }
+
+        /// <summary>
+        /// Validates scheme source (protocol).
+        /// </summary>
+        private string ValidateSchemeSource(string value)
+        {
+            string[] validSchemes = { "http:", "https:", "data:", "blob:", "filesystem:", "wss:", "ws:" };
```
mitchelsellers: For coding consistency this should be `var`

mitchelsellers: Similar comment for other items in this file

DNN Platform/DotNetNuke.ContentSecurityPolicy/DocumentCspContributor.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        /// Gets value of the document directive.
+        /// </summary>
+        public string DirectiveValue { get; private set; }
+
+        /// <summary>
+        /// Sets the directive value with validation.
+        /// </summary>
+        /// <param name="value">The value to set for the directive.</param>
+        public void SetDirectiveValue(string value)
```
mitchelsellers: Why was a separate method implement for this, rather than just doing this as a setter?

DNN Platform/DotNetNuke.ContentSecurityPolicy/ReportingCspContributor.cs
mitchelsellers: Additionally, this is a validation that is shared across multiple sections, it would be good to standardize this validaton if we can should it need to change.

DNN Platform/Library/Entities/Portals/PortalSettings.cs
```diff
@@ -132,6 +132,18 @@ public enum UserDeleteAction
             HardDelete = 3,
         }
 
+        public enum CspMode
+        {
+            /// <summary>Content Security Header is not added.</summary>
+            Off = 0,
+
+            /// <summary>Content Security Header is not added in Report Only.</summary>
```
mitchelsellers: I believe this summary is incorrect and it should be "Content Security Header is added in Report Only mode."

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
sachatrauwaen: I change it.
But I was thinking that all the regexp validation is maybe overkill. And because it is evaluated on each request , its maybe better to remove it.
What do you think ?

DNN Platform/DotNetNuke.ContentSecurityPolicy/DocumentCspContributor.cs
sachatrauwaen: modified

DNN Platform/DotNetNuke.ContentSecurityPolicy/dnn.json
sachatrauwaen: removed

DNN Platform/Library/Entities/Portals/PortalSettings.cs
sachatrauwaen: fixed

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
mitchelsellers: I think given it is executed on EVERY request, I'd be very inclined to NOT do this validation and test that it was correct, or find a way to validate at a different stage.

sachatrauwaen: i remove the syntax checks from all request. They only be done on site settings save.

valadas: I still see it in this file, do you mean that this only happens when saving settings for this specific place?

sachatrauwaen: Maybe you are looking at the outdated file?
Normally you see the constructor with the checkSyntax parameter.
 public CspSource(CspSourceType type, string value = null, bool checkSyntax = false)
By default no syntax check is done.

valadas: Well, I wee it has a boolean in the constructor to checkSyntax which defaults to false. 
So do you mean that with this change, it validates when you save settings but it does not on every page load? If so I am fine with this change.
image

sachatrauwaen: Yes, that's right

DNN Platform/Website/Default.aspx.cs
sachatrauwaen: done, for the common part

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicy.cs
sachatrauwaen: i now added a parameter to the constructor for avoid validation

DNN Platform/DotNetNuke.ContentSecurityPolicy/ContentSecurityPolicyParser.cs
sachatrauwaen: done

DNN Platform/DotNetNuke.ContentSecurityPolicy/CspSource.cs
sachatrauwaen: done

DNN Platform/DotNetNuke.ContentSecurityPolicy/DocumentCspContributor.cs
sachatrauwaen: changed to a property

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Security/App_LocalResources/Security.resx
```diff
@@ -1432,4 +1432,79 @@
...
+    <value>CSP settings updated</value>
+  </data>
+  <data name="CspSettingsError.Text" xml:space="preserve">
+    <value>An error has occurred. The CSP settings could not be updated. Please consult the admin log to determine the cause of this error.</value>
+  </data>
+  <data name="CspSettingsRestoreWarning.Text" xml:space="preserve">
+    <value>The changes to the CSP settings will be discarded. Do you wish to continue?</value>
+  </data>
+  <data name="plCspHeaderFixed.Text" xml:space="preserve">
+    <value>Lock CSP Header</value>
```
mitchelsellers: Just thowing out an idea here, I think Lock is better  than others, but still not sure it is fully understood by others, some thoughts after checking with my team

* Disallow Module/Theme/System Overrides
* Strict CSP Header

I think the first is the most "proper" as the name explains but its wordy.

sachatrauwaen: Strict : can sound as strict in a security point of view (that is not wath we want to tell, most of the time it will be less secure because it need to contain policy for all use cases)
Alternative can be : Disallow Extensions overrides (wich is also more correct, because all extesions are disallowed and now the framework will override to avoid crash of dnn)

donker: I like "Strict" but I'm probably biased due to how it's used in our technical world.

donker: I've gone ahead and changed the wording

valadas:
@donker it looks like turning mode to "Report Only" and attempting to save the default headers gives this error:

image

valadas:
When I replace the semi-colons with commas, it appears to save, but I don't know if that is right

sachatrauwaen:
i seen the issue. It already fixed locally. I will make pull request.

sachatrauwaen:
Pull request created : #7126

PR #6775, opened 2025-10-17
Change DNN version from 10.1.0 to 10.1.2 in template

david-poindexter:
## Summary
Updated the bug report template to reflect the latest DNN version.


valadas:

PR #6779, opened 2025-10-21
UI for management of upgrades

donker:
Since version 10.1 we have the ability to upgrade DNN by dropping an upgrade DNN package in the folder App_Data/Upgrade. The UI for this was the change of the upgrade button under the DNN logo top left. This PR creates a tab on the servers panel to manage these upgrades.

### Features

1. List of available packages in App_Data/Upgrade
2. Ability to delete packages
3. Ability to upload package

### DIFF
--- a/Dnn.AdminExperience/ClientSide/Servers.Web/package.json
+++ b/Dnn.AdminExperience/ClientSide/Servers.Web/package.json
@@ -42,6 +42,8 @@
     "react-hot-loader": "4.13.1",
     "style-loader": "^4.0.0",
     "typescript": "^5.9.3",
+    "url-loader": "4.1.1",
+    "webpack": "5.102.1",
     "webpack-bundle-size-analyzer": "3.1.0",
     "webpack-cli": "6.0.1",
     "webpack-dev-server": "5.2.2"
--- a/yarn.lock
+++ b/yarn.lock
@@ -105,6 +105,19 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/generator@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/generator@npm:7.28.5"
+  dependencies:
+    "@babel/parser": "npm:^7.28.5"
+    "@babel/types": "npm:^7.28.5"
+    "@jridgewell/gen-mapping": "npm:^0.3.12"
+    "@jridgewell/trace-mapping": "npm:^0.3.28"
+    jsesc: "npm:^3.0.2"
+  checksum: 10/ae618f0a17a6d76c3983e1fd5d9c2f5fdc07703a119efdb813a7d9b8ad4be0a07d4c6f0d718440d2de01a68e321f64e2d63c77fc5d43ae47ae143746ef28ac1f
+  languageName: node
+  linkType: hard
+
 "@babel/helper-annotate-as-pure@npm:^7.27.1, @babel/helper-annotate-as-pure@npm:^7.27.3":
   version: 7.27.3
   resolution: "@babel/helper-annotate-as-pure@npm:7.27.3"
@@ -141,6 +154,23 @@ __metadata:
   peerDependencies:
     "@babel/core": ^7.0.0
   checksum: 10/32d01bdd601b4d129b1d510058a19644abc764badcc543adaec9e71443e874ef252783cceb2809645bdf0e92b07f206fd439c75a2a48cf702c627aba7f3ee34a
+  languageName: node
+  linkType: hard
+
+"@babel/helper-create-class-features-plugin@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/helper-create-class-features-plugin@npm:7.28.5"
+  dependencies:
+    "@babel/helper-annotate-as-pure": "npm:^7.27.3"
+    "@babel/helper-member-expression-to-functions": "npm:^7.28.5"
+    "@babel/helper-optimise-call-expression": "npm:^7.27.1"
+    "@babel/helper-replace-supers": "npm:^7.27.1"
+    "@babel/helper-skip-transparent-expression-wrappers": "npm:^7.27.1"
+    "@babel/traverse": "npm:^7.28.5"
+    semver: "npm:^6.3.1"
+  peerDependencies:
+    "@babel/core": ^7.0.0
+  checksum: 10/0bbf3dfe91875f642fe7ef38f60647f0df8eb9994d4350b19a4d1a9bdc32629e49e56e9a80afb12eeb6f6bcc6666392b37f32231b7c054fc91a0d5251cd67d5b
   languageName: node
   linkType: hard
 
@@ -186,6 +216,16 @@ __metadata:
     "@babel/traverse": "npm:^7.27.1"
     "@babel/types": "npm:^7.27.1"
   checksum: 10/533a5a2cf1c9a8770d241b86d5f124c88e953c831a359faf1ac7ba1e632749c1748281b83295d227fe6035b202d81f3d3a1ea13891f150c6538e040668d6126a
+  languageName: node
+  linkType: hard
+
+"@babel/helper-member-expression-to-functions@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/helper-member-expression-to-functions@npm:7.28.5"
+  dependencies:
+    "@babel/traverse": "npm:^7.28.5"
+    "@babel/types": "npm:^7.28.5"
+  checksum: 10/05e0857cf7913f03d88ca62952d3888693c21a4f4d7cfc141c630983f71fc0a64393e05cecceb7701dfe98298f7cc38fcb735d892e3c8c6f56f112c85ee1b154
   languageName: node
   linkType: hard
 
@@ -278,6 +318,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/helper-validator-identifier@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/helper-validator-identifier@npm:7.28.5"
+  checksum: 10/8e5d9b0133702cfacc7f368bf792f0f8ac0483794877c6dca5fcb73810ee138e27527701826fb58a40a004f3a5ec0a2f3c3dd5e326d262530b119918f3132ba7
+  languageName: node
+  linkType: hard
+
 "@babel/helper-validator-option@npm:^7.27.1":
   version: 7.27.1
   resolution: "@babel/helper-validator-option@npm:7.27.1"
@@ -314,6 +361,17 @@ __metadata:
   bin:
     parser: ./bin/babel-parser.js
   checksum: 10/f54c46213ef180b149f6a17ea765bf40acc1aebe2009f594e2a283aec69a190c6dda1fdf24c61a258dbeb903abb8ffb7a28f1a378f8ab5d333846ce7b7e23bf1
+  languageName: node
+  linkType: hard
+
+"@babel/parser@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/parser@npm:7.28.5"
+  dependencies:
+    "@babel/types": "npm:^7.28.5"
+  bin:
+    parser: ./bin/babel-parser.js
+  checksum: 10/8d9bfb437af6c97a7f6351840b9ac06b4529ba79d6d3def24d6c2996ab38ff7f1f9d301e868ca84a93a3050fadb3d09dbc5105b24634cd281671ac11eebe8df7
   languageName: node
   linkType: hard
 
@@ -1224,6 +1282,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/plugin-transform-typescript@npm:^7.27.1":
+  version: 7.28.5
+  resolution: "@babel/plugin-transform-typescript@npm:7.28.5"
+  dependencies:
+    "@babel/helper-annotate-as-pure": "npm:^7.27.3"
+    "@babel/helper-create-class-features-plugin": "npm:^7.28.5"
+    "@babel/helper-plugin-utils": "npm:^7.27.1"
+    "@babel/helper-skip-transparent-expression-wrappers": "npm:^7.27.1"
+    "@babel/plugin-syntax-typescript": "npm:^7.27.1"
+  peerDependencies:
+    "@babel/core": ^7.0.0-0
+  checksum: 10/e4706379df70c2de9d3d3f573ff74a160e05221620a22dd0a64899ab45fddad9e4530fbba33014c75906f13aa78d8044fed80dba85068e11d84ed1f033dea445
+  languageName: node
+  linkType: hard
+
 "@babel/plugin-transform-unicode-escapes@npm:^7.27.1":
   version: 7.27.1
   resolution: "@babel/plugin-transform-unicode-escapes@npm:7.27.1"
@@ -1444,6 +1517,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/traverse@npm:^7.28.5":
+  version: 7.28.5
+  resolution: "@babel/traverse@npm:7.28.5"
+  dependencies:
+    "@babel/code-frame": "npm:^7.27.1"
+    "@babel/generator": "npm:^7.28.5"
+    "@babel/helper-globals": "npm:^7.28.0"
+    "@babel/parser": "npm:^7.28.5"
+    "@babel/template": "npm:^7.27.2"
+    "@babel/types": "npm:^7.28.5"
+    debug: "npm:^4.3.1"
+  checksum: 10/1fce426f5ea494913c40f33298ce219708e703f71cac7ac045ebde64b5a7b17b9275dfa4e05fb92c3f123136913dff62c8113172f4a5de66dab566123dbe7437
+  languageName: node
+  linkType: hard
+
 "@babel/types@npm:^7.0.0, @babel/types@npm:^7.18.9, @babel/types@npm:^7.20.7, @babel/types@npm:^7.27.1, @babel/types@npm:^7.27.3, @babel/types@npm:^7.28.2, @babel/types@npm:^7.28.4, @babel/types@npm:^7.4.4, @babel/types@npm:^7.7.0":
   version: 7.28.4
   resolution: "@babel/types@npm:7.28.4"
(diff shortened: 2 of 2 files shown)


Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Body/Body.jsx
```diff
@@ -47,30 +48,30 @@ class Body extends Component {
...
         let registeredTabs = application.getRegisteredServerTabs();
         if (!isHost) {
             registeredTabs = registeredTabs.filter(function (tab) {
                 return !tab.isHostOnlyVisible;
             });
-        }        
+        }
 
-        const systemInfoTabHeaders = isHost ? [Localization.get("tabApplicationTitle"), Localization.get("tabWebTitle"), Localization.get("tabDatabaseTitle"), Localization.get("tabServersTitle")] 
+        const systemInfoTabHeaders = isHost ? [Localization.get("tabApplicationTitle"), Localization.get("tabWebTitle"), Localization.get("tabDatabaseTitle"), Localization.get("tabServersTitle"), "Upgrades"]
```
bdukes: Is there a reason this tab is bypassing localization?

Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Upgrades/UpgradeRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </GridCell>
+      <GridCell columnSize={20}>
+        <TextOverflowWrapper text={version} />
+      </GridCell>
+      <GridCell columnSize={10}>
+        {props.upgrade.IsValid ? "Yes" : "No"}
+      </GridCell>
+      <GridCell columnSize={20}>
+        {props.upgrade.IsOutdated ? "Yes" : "No"}
+      </GridCell>
```
bdukes, line 42: These need to be localized, right?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Common/Extensions.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+
+namespace Dnn.PersonaBar.Extensions.Common
+{
+    using System;
+
+    internal static class Extensions
+    {
+        internal static bool Equals(this Version version, Version compareTo)
```
bdukes: I would think that this method wouldn't override the built-in [`Version.Equals`](https://learn.microsoft.com/en-us/dotnet/api/system.version.equals?view=netframework-4.8#system-version-equals(system-version)) when used as an extension method. Should this be given a different name to avoid confusion?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+        {
+            var upgrades = await this.localUpgradeService.GetLocalUpgrades(cancellationToken);
+            var upgrade = upgrades.FirstOrDefault(u => u.PackageName.Equals(data.PackageName, StringComparison.InvariantCultureIgnoreCase));
+            try
+            {
+                var packagePath = Path.Combine(this.applicationStatusInfo.ApplicationMapPath, "App_Data", "Upgrade", upgrade.PackageName + ".zip");
+                if (File.Exists(packagePath))
+                {
+                    File.Delete(packagePath);
+                }
```
bdukes: I would expect this logic to belong in LocalUpgradeService

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+            await request.Content.ReadAsMultipartAsync(provider).ConfigureAwait(false);
+
+            object result = null;
+            var fileName = string.Empty;
+            Stream stream = null;
+
+            foreach (var item in provider.Contents)
+            {
+                var name = item.Headers.ContentDisposition.Name;
+                switch (name.ToUpper())
```
bdukes:
```suggestion
                switch (name.ToUpperInvariant())
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                {
+                    case "\"POSTFILE\"":
+                        fileName = item.Headers.ContentDisposition.FileName.Replace("\"", string.Empty);
+                        if (fileName.IndexOf("\\", StringComparison.Ordinal) != -1)
+                        {
+                            fileName = Path.GetFileName(fileName);
+                        }
+
+                        if (Globals.FileEscapingRegex.Match(fileName).Success == false && Path.GetExtension(fileName).ToLowerInvariant() == ".zip")
```
bdukes:
```suggestion
                        if (Globals.FileEscapingRegex.IsMatch(fileName) == false && Path.GetExtension(fileName).Equals(".zip", StringComparison.OrdinalIgnoreCase))
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                        {
+                            stream = item.ReadAsStreamAsync().Result;
```
bdukes:
```suggestion
                            stream = await item.ReadAsStreamAsync();
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+            Stream stream = null;
+
+            foreach (var item in provider.Contents)
+            {
+                var name = item.Headers.ContentDisposition.Name;
+                switch (name.ToUpper())
+                {
+                    case "\"POSTFILE\"":
+                        fileName = item.Headers.ContentDisposition.FileName.Replace("\"", string.Empty);
+                        if (fileName.IndexOf("\\", StringComparison.Ordinal) != -1)
```
bdukes:
```suggestion
                        if (fileName.IndexOf('\\') != -1)
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                            stream = item.ReadAsStreamAsync().Result;
+                        }
+
+                        break;
+                }
+            }
+
+            if (!string.IsNullOrEmpty(fileName) && stream != null)
+            {
+                var info = this.localUpgradeService.GetLocalUpgradeInfo(Path.GetFileNameWithoutExtension(fileName), stream, CancellationToken.None).Result;
```
bdukes:
```suggestion
                var info = await this.localUpgradeService.GetLocalUpgradeInfo(Path.GetFileNameWithoutExtension(fileName), stream, CancellationToken.None);
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                return this.UploadFileAction();
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex);
+                return Task.FromResult(this.Request.CreateErrorResponse(HttpStatusCode.InternalServerError, ex.Message));
+            }
+        }
+
+        private async Task<HttpResponseMessage> UploadFileAction()
```
bdukes: It would be good to take in a `CancellationToken` from the web API method and pass it through here, so that if the request gets cancelled this method can respond immediately.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                var info = this.localUpgradeService.GetLocalUpgradeInfo(Path.GetFileNameWithoutExtension(fileName), stream, CancellationToken.None).Result;
+                if (!info.IsValid)
+                {
+                    result = new { message = $"The uploaded file {fileName} is not a valid DNN package.", };
+                }
+                else if (info.IsOutdated)
+                {
+                    result = new { message = $"The uploaded file {fileName} is an outdated DNN package.", };
+                }
```
bdukes: Do we need to localize these messages?

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                else
+                {
+                    var upgradePath = Path.Combine(this.applicationStatusInfo.ApplicationMapPath, "App_Data", "Upgrade", info.PackageName + ".zip");
+                    using (var fileStream = File.Create(upgradePath))
```
bdukes: I would like the path to be owned by the `LocalUpgradeService`, rather than copied around in a few different files.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,228 @@
...
+                    result = info;
+                }
+            }
+
+            var mediaTypeFormatter = new JsonMediaTypeFormatter();
+            mediaTypeFormatter.SupportedMediaTypes.Add(new MediaTypeHeaderValue("text/plain"));
+
+            // Response Content Type cannot be application/json
+            // because IE9 with iframe-transport manages the response
+            // as a file download
```
bdukes: I assume this is copied from somewhere, and you're not explicitly trying to support IE9, right?

bdukes:
This is looking great, thanks @donker!

david-poindexter:
I apologize as I am a bit late to the party on this one.  Would it be possible to add a `web.config` flag to enable/disable this feature (the UI)?

donker:
> I apologize as I am a bit late to the party on this one. Would it be possible to add a `web.config` flag to enable/disable this feature (the UI)?

We could. I've just added an issue describing my thought process around why we shouldn't in this issue: https://github.com/dnnsoftware/Dnn.Platform/issues/6782

I'd love to hear your thoughts on this.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: Done

Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Body/Body.jsx
donker: Done

Dnn.AdminExperience/ClientSide/Servers.Web/src/components/Tabs/Upgrades/UpgradeRow.tsx
donker: Done

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Common/Extensions.cs
donker: I didn't need it anymore ... it's gone

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: Moved to the service

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: Done

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: Done

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: OK

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
donker: Done

david-poindexter:
> > I apologize as I am a bit late to the party on this one. Would it be possible to add a `web.config` flag to enable/disable this feature (the UI)?
> 
> We could. I've just added an issue describing my thought process around why we shouldn't in this issue: #6782
> 
> I'd love to hear your thoughts on this.

I may have typed this up too quickly to make clear my hope.  I am more concerned about who is able to do this - not so much that it can/cannot be done.  Adding the UI makes it more obvious this is a feature.  I'd just like the ability to make it a "hidden" feature and not exposed in the UI.

donker:
From the DNN tech meeting 28 Oct. Todo:
- Master switch to block the upload function. Use case: some of our community members offer DNN with host access to customers but wish to remain in control of DNN upgrades because of potential breaking changes.
- Roll back the increased maxRequestLength and instead solve using chunked uploads.

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,280 @@
...
+                        stream = item.ReadAsStreamAsync().Result;
+                        break;
+                    case "\"START\"":
+                        long.TryParse(item.ReadAsStringAsync().Result, out startPosition);
+                        break;
+                    case "\"TOTALSIZE\"":
+                        long.TryParse(item.ReadAsStringAsync().Result, out totalSize);
+                        break;
+                    case "\"FILEID\"":
+                        fileId = item.ReadAsStringAsync().Result;
```
bdukes:
```suggestion
                        stream = await item.ReadAsStreamAsync();
                        break;
                    case "\"START\"":
                        long.TryParse(await item.ReadAsStringAsync(), out startPosition);
                        break;
                    case "\"TOTALSIZE\"":
                        long.TryParse(await item.ReadAsStringAsync(), out totalSize);
                        break;
                    case "\"FILEID\"":
                        fileId = await item.ReadAsStringAsync();
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/UpgradesController.cs
```diff
@@ -0,0 +1,280 @@
...
+            if (stream != null && totalSize != 0 && !string.IsNullOrEmpty(fileId))
+            {
+                var fileToWriteTo = Path.Combine(this.applicationStatusInfo.ApplicationMapPath, "App_Data", "Upgrade", fileId + ".resources");
+
+                // use the file stream and start position to write the chunk to the file
+                using (var fileStream = new FileStream(fileToWriteTo, FileMode.OpenOrCreate, FileAccess.Write))
+                {
+                    fileStream.Position = startPosition;
+                    await stream.CopyToAsync(fileStream).ConfigureAwait(false);
+                }
```
bdukes, line 269: Should this file writing be done in the `LocalUpgradeService`?

donker: I'd prefer to handle all file management in the upgrade folder in the extensions project and keep all upgrade logic in the upgrade service.

DNN Platform/Website/App_Data/Upgrade/upgrade.json
```diff
@@ -0,0 +1,15 @@
...
+    "/favicon.ico",
+    "/Robots.txt",
+    "/web.config",
+    "/App_Data/Database.mdf",
+    "/App_Data/Database_log.LDF",
+    "/bin/Providers/DotNetNuke.Providers.AspNetClientCapabilityProvider.dll",
+    "/bin/Newtonsoft.Json.dll",
+    "/bin/System.Runtime.CompilerServices.Unsafe.dll",
+    "/Config/DotNetNuke.config",
+    "/Install/InstallWizard.aspx"
```
mitchelsellers, line 11: I like this concept, howevewr, I guess my question here is if we are "excluding" explicitily named DLL's don't we still have the same issue that we are trying to fix?  I understand the other items, but 2 of the 3 DLL's listed in this list are common ones that third-parties upgrade and would result in a failure.

DNN Platform/Website/App_Data/Upgrade/upgrade.json
```diff
@@ -0,0 +1,15 @@
+﻿{
+  "minimumDnnVersion": "10.02.00",
```
mitchelsellers, line 2: Is the upgrade automatically performed via another process?  Or is this something we need to ensure is addressed in every release?

mitchelsellers (commented):
Just a few questions

DNN Platform/Website/App_Data/Upgrade/upgrade.json
donker: I think those dlls can go from this list because the upgrade process should take care of it.

DNN Platform/Website/App_Data/Upgrade/upgrade.json
donker: I think this is something we need to ensure before each release. It will require us to test the specific upgrade proposed. But it allows us to make changes that require an upgrade "path" and protect users from breaking their sites. Basically it's an extra protection mechanism.

DNN Platform/Website/App_Data/Upgrade/upgrade.json
bdukes: These DLLs are on the exclude list b/c they're already in packages. We'll need to think about whether it still makes sense to include DLLs in packages if the upgrade process takes care of the assembly version (is having a package linked to that assembly still helpful?)

donker:
This addresses issue #6782

PR #6812, opened 2025-11-21
Remove FromSrc in the new CDF solution in favor of CreateX method.

donker:
After feedback from developers adjusting old code, I propose to shorten

``` cs
 clientResourcesController.CreateScript().FromSrc("path/to/script.js").Register();
```

to

``` cs
 clientResourcesController.CreateScript("path/to/script.js").Register();
```

After all, the source is required in both all script elements as well as all link elements. So there is no use case where FromSrc would not be used.


valadas (approved):
I love it

PR #6814, opened 2025-11-25
Fixed Cookie Timeout Lookup Values

### LINKED ISSUE #6811: [Bug]: Cookie persistent timeout not enforced on DNN 10.1.2

### What happened?

Since DNN 10.1.2 fixed some security issues, web.config contains some <location> nodes.

```
  <location path="Providers/HtmlEditorProviders/DNNConnect.CKE/Browser/FileUploader.ashx">
    <system.web>
      <authorization>
        <deny users="?" />
        <!-- Deny anonymous users -->
      </authorization>
    </system.web>
  </location>
  <location path="Providers/HtmlEditorProviders/DNNConnect.CKE/Browser/Browser.aspx">
    <system.web>
      <authorization>
        <deny users="?" />
        <!-- Deny anonymous users -->
      </authorization>
    </system.web>
  </location>
```

The logic on the GetAuthCookieTimeout checks for this:

``` var locationNav = configNav.SelectSingleNode("configuration/location");```

Notice that this will succeed. 
The logic then will check for the cookie timeout under this node, which fill fail.

The should be changed to account for the new location nodes.

This is an issue if you have customized the cookie timeout of the forms section

<forms name=".DOTNETNUKE" protection="All" **timeout="480"** cookieless="UseCookies" />

### Steps to reproduce?

Customize cookie timeout on a DNN 10.1.2 to more than 30 minuts.
Login to the site.
Leave the browser alone for more that 30 mins.
Click anywhere on the website after 31 minutes and youll be logged out.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.1.2 (latest v10 release)

_No response_

### Code of Conduct

mitchelsellers:
- Fixes #6811 

## Summary

The existing behavior blindly assumed that if a location node exists that the system.web.authencation/forms element is in the location node.

This updates the behavior to start with the normal location, IF if funds forms authenticaton it continues, otherwise, it tries to find it under a location node IF location nodes exist.

This should restore functionality after the updates recently resulted in ALL installations having a location node.  This has been tested and appears to be working well

### DIFF
--- a/DNN Platform/Library/Common/Utilities/Config.cs
+++ b/DNN Platform/Library/Common/Utilities/Config.cs
@@ -468,21 +468,16 @@ namespace DotNetNuke.Common.Utilities
         {
             var configNav = Load(appStatus).CreateNavigator();
 
-            // Select the location node
-            var locationNav = configNav.SelectSingleNode("configuration/location");
-            XPathNavigator formsNav;
-
-            // Test for the existence of the location node if it exists then include that in the nodes of the XPath Query
-            if (locationNav == null)
+            // Try to get the forms authentication from the default location
+            var formsNav = configNav?.SelectSingleNode("configuration/system.web/authentication/forms");
+            if (formsNav is null)
             {
-                formsNav = configNav.SelectSingleNode("configuration/system.web/authentication/forms");
-            }
-            else
-            {
-                formsNav = configNav.SelectSingleNode("configuration/location/system.web/authentication/forms");
+                // If unable, look for a location node, if found try to get the settings from there
+                formsNav = configNav?.SelectSingleNode("configuration/location/system.web/authentication/forms");
             }
 
-            return formsNav != null ? XmlUtils.GetAttributeValueAsInteger(formsNav, "timeout", 30) : 30;
+            const int DefaultTimeout = 30;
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good, thanks!

DNN Platform/Library/Common/Utilities/Config.cs
```diff
@@ -468,18 +468,17 @@ public static int GetAuthCookieTimeout(IApplicationStatusInfo appStatus)
...
-            if (locationNav == null)
-            {
-                formsNav = configNav.SelectSingleNode("configuration/system.web/authentication/forms");
-            }
-            else
+            if (formsNav == null)
             {
-                formsNav = configNav.SelectSingleNode("configuration/location/system.web/authentication/forms");
+                //If unable, look for a location node, if found try to get the settings from there
+                var locationNav = configNav.SelectSingleNode("configuration/location");
```
vmasanas: I think this line should be looking for the node "configuration/location/system.web/authentication/forms"
Otherwise you have to do two checks.

mitchelsellers: I left this as a two step, more to limit the change of scope/impact, in 99.99% of installations this line will never execute.  I can clean this up further to be a sinigle additional check with the full node if desired

Issue #6817, opened 2025-11-28
[Enhancement]: Representative sample modules should be included in the main repo


donker:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of problem

We have sample projects such as https://github.com/dnnsoftware/Dnn.Platform.Samples.Mvc that are meant to show developers how to develop extensions for DNN. It has proved to be impossible to keep this in step with the developments in DNN. Given that we intend to bring some major improvements to DNN that will impact modules, it is not ideal that these demo projects are separate from the main project. I.e. if one changes the platform in a way that requires module to make adjustments, these modules should serve as validation for that.

### Description of solution

I propose to include a set of representative (i.e. covering the various module patterns) modules in the main repository that serve both to demonstrate good practice and secondly to test the latest version of DNN. As a bonus changes to the API that would risk breaking modules, would immediately provoke a build error of the entire project in this way.

### Do you plan to contribute code for this enhancement?

- [x] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mitchelsellers:
I don't necessarily disagree with this, but the solution structure and build complexity is something that I think we should really take into consideration.

For example I do believe

* It should NOT be something that impacts the normal build process for creating the DNN Platform project if we can avoid it, our build is already long enough
* It should NOT create/install the sample modules when developing/working locally, for similar reasons

But I do agree it would be easier for these to live in the main repoository.

Ao maybe the following.

* Create a new folder at the root "samples"
* Move the existing samples within there, such as `samples/mvc`, `samples/spa`, or `samples/webforms`

donker:
> * It should NOT create/install the sample modules when developing/working locally, for similar reasons

My idea was to build them as assets. I agree they shouldn't find their way into the debug/release build.

valadas:
That would be fine by me. However if for any reason we would want to have samples in its own repo but still in sync with latest DNN code, one way would be to use git submodules to this repo. Then to trigget a build on those repos we could use github api to dispatch a build event.

But I believe the idea here is that one would want to work on DNN code and module code at the same time right, so I'd be fine if we some samples right here in that context.

valadas:
Done in #6821

PR #6821, opened 2025-11-29
Include sample modules

donker:
This PR addresses issue #6817 

With this PR we include a sample MVC and SPA module in the solution. You can opt to build these to the output using your own local settings:

DNN_Platform.local.build
```
<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <WebsitePath Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">D:\path\to\my\local\dnn\site</WebsitePath>
    <CopySampleProjects>True</CopySampleProjects>
  </PropertyGroup>
</Project>
```

settings.local.json
```
{
 ...
  "CopySampleProjects": true,
  "Version": "10.99.0"
}
```

In debug mode the projects build as any other and the dll gets copied to a destination based on the master switch. In release mode the modules get built to Artifacts/SampleModules.

### DIFF
--- a/DNN Platform/Library/Common/Utilities/Config.cs
+++ b/DNN Platform/Library/Common/Utilities/Config.cs
@@ -468,21 +468,16 @@ namespace DotNetNuke.Common.Utilities
         {
             var configNav = Load(appStatus).CreateNavigator();
 
-            // Select the location node
-            var locationNav = configNav.SelectSingleNode("configuration/location");
-            XPathNavigator formsNav;
-
-            // Test for the existence of the location node if it exists then include that in the nodes of the XPath Query
-            if (locationNav == null)
+            // Try to get the forms authentication from the default location
+            var formsNav = configNav?.SelectSingleNode("configuration/system.web/authentication/forms");
+            if (formsNav is null)
             {
-                formsNav = configNav.SelectSingleNode("configuration/system.web/authentication/forms");
-            }
-            else
-            {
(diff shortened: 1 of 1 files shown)


valadas:
I only had time to take a very quick look, but that looks awesome!
For me on the SPA module I think I would add a couple of .md files just to explain that it is done one way here but could be done with anything one would like (react, angular, ionic, web-components, stencil, vanilla js, typescript, etc.), that jQuery is available upon request but not needed or provided unless requested, a couple of little things like that, that can potentially evolve. But love the samples so far. I'll give those a test hopefully by the end of this weekend

donker:
> I only had time to take a very quick look, but that looks awesome! For me on the SPA module ...

I am somewhat torn on the SPA module. I've spent a couple of hours making it look a bit better than before. This module has a reason to be here and a reason not to be. The reason it should be here is that it shows how far you can take this whole HTML templating stuff. The reason, IMHO, it shouldn't be here is that I don't wish to create the impression how you *should* use this module pattern. Instead I'd much rather promote a SPA module with a React frontend as that is the way most web UI developers see frontend development. But yeah ... framework wars incoming.

mitchelsellers (approved):
I believe this is ok as it is, however, my only question/comment/thought is to add some sort of a Readme.md or similar to the /Samples directory to show that these are samples and for illustration purposes only.

PR #6826, opened 2025-12-03
Restored /Tests folder to state in previous commit

valadas:
It looks like in https://github.com/dnnsoftware/Dnn.Platform/commit/4a4bcbcdf3cedbf702816f8168c4d51bf688f7f6 all the test projects were modified in a way were they would reference a duplicate app.config

I tried reverting only that change but it did not fix the broken tests. This reverts the whole Tests folder to the state it was in the very previous commit where the tests were fine.

I am not sure this is the proper solution to the failing tests @bdukes do you have opinions or a better fix?


valadas:
Well, that was apparently not to fix. I will leave this PR here however as a reference until we pinpoint the proper fix.

bdukes:
I was able to grab some of the details of that commit and revert it more directly, I've pushed that to your branch

valadas:
Perfect, thanks @bdukes

donker (approved):
Thanks

PR #6828, opened 2025-12-05
Update test projects to use SDK-style projects

bdukes:
## Summary
I've updated the test projects to use the simpler SDK-style project format. I'm hoping this makes it easier to maintain, and especially manage updating NuGet packages for tests.

This PR builds on top of https://github.com/dnnsoftware/Dnn.Platform/pull/6826, that PR should be merged first.


bdukes:
I tried to get the Urls and Integration tests projects running locally in order to verify that I didn't break them, but only got far enough to confirm that some of them can run. They appear to be pretty fragile and out of date (e.g. some tests seem to assume that the site they're running against is a single language, but other tests add a second language, and they expect the Digital Assets module to be there).

There are a handful of change in this PR that I think make the tests work better. One that affects more than tests is that the portal templates include a _My Messsages_ page, with three `s` characters (which breaks the URL tests that expect the page to be there with a correctly spelled name). It looks like that typo was introduced in 10.0.0, so not too widespread of an impact.

bdukes:

PR #6830, opened 2025-12-07
Fix CDF webforms lifecycle issues

donker:
The new CDF detaches itself from the Webforms lifecycle. The old CDF was very much tied into it as it created and deleted controls. The Css/JsExclude controls might miss their effect if this is not taken into account. We ran into an issue where a CSS file was not removed that should have been removed.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
+++ b/DNN Platform/DotNetNuke.Web.Client.ResourceManager/Models/ResourceBase.cs
@@ -14,6 +14,8 @@ namespace DotNetNuke.Web.Client.ResourceManager.Models
     /// <summary>Base class for all resource types.</summary>
     public abstract class ResourceBase : IResource
     {
+        private string name;
+
         /// <inheritdoc />
         public string FilePath { get; set; }
 


mitchelsellers (commented):
Just a general question here, do we anticipate this being considered a breaking change in any possibility by changing the processes from `OnInit` to `OnLoad`, which is later in the lifecycle?

Could this have a breaking change implication to anyone that may be using this in a custom module or otherwise?

Additionally: There appears to be test failures

donker:
I am fixing an issue that is causing the current Css/JSExclude to not work as advertised because of the Webforms lifecycle, BTW.

> Just a general question here, do we anticipate this being considered a breaking change in any possibility by changing the processes from `OnInit` to `OnLoad`, which is later in the lifecycle?
> 
> Could this have a breaking change implication to anyone that may be using this in a custom module or otherwise?
> 

I have asked myself this but have not succeeded in coming up with a scenario where that would happen. You're not supposed to inherit from these controls AFAIK. And given the entire page lifecycle in CDF is going away, it begs the question how it could break something. But if anyone has an idea, I'd be happy to test.

> Additionally: There appears to be test failures

That appeared to be from another commit than this.

DNN Platform/DotNetNuke.Web.Client/Controls/DnnCssInclude.cs
```diff
@@ -33,18 +33,17 @@ public DnnCssInclude(IClientResourceController clientResourceController)
...
                         .SetProvider(this.ForceProvider)
                         .SetPriority(this.Priority)
                         .SetMedia(this.CssMedia)
                         .Register();
-        }
-
-        /// <inheritdoc/>
-        protected override void OnLoad(System.EventArgs e)
-        {
-            this.PathNameAlias = string.IsNullOrEmpty(this.PathNameAlias) ? string.Empty : this.PathNameAlias.ToLowerInvariant();
```
bdukes, line 47: Do we not need to normalize `this.PathNameAlias` anymore?

DNN Platform/DotNetNuke.Web.Client/Controls/DnnCssInclude.cs
```diff
@@ -33,18 +33,17 @@ public DnnCssInclude(IClientResourceController clientResourceController)
         /// <inheritdoc/>
         protected override void OnInit(EventArgs e)
```
bdukes: Should this empty handler get removed?

PR #6831, opened 2025-12-07
Fix the default for the web.config variable for upgrade uploading

donker:
The default behavior upon upgrade should be the same as on a new install. This is by default True.


DNN Platform/Website/Install/Config/10.02.00.config
```diff
@@ -2,7 +2,7 @@
   <nodes configfile="web.config">
     <node path="/configuration/appSettings" action="add">
       <!-- Can be set to true to allow the image handler to display text passed in the querystring. -->
```
bdukes:
```suggestion
```

david-poindexter (approved):

PR #6832, opened 2025-12-08
Re-introduced "role" JWT claim to attempt to limit impact of breaking change

### LINKED ISSUE #6829: [Bug]: Breaking change on JWT response in DNN10

### What happened?

JWT response message has changed slightly which could break implementations that parse the JWT response message.

### Steps to reproduce?

1. install JWT support extension
2. Implement JWT login
3. Login using JWT

### Current Behavior

Image

### Expected Behavior

Image

### Relevant log output

```shell

```

### Anything else?

I don't know honestly if one is more standard than the other but it is an undocumented breaking change for JWT consumers targeting DNN9 vs DNN10.

### Affected Versions

10.1.2 (latest v10 release)

Chrome

### Code of Conduct

valadas:
In #6356 a dependency update made an unintentional breaking change. Instead of the classic "role" claim we now give a "http://schemas.microsoft.com/ws/2008/06/identity/claims/role" claim for roles.

JWT in itself has no specs for roles or claims which is left to whatever implementation is targeted like OIDC, OAuth, etc.

In DNN context it is a simple JWT but we do provide roles and the new package enum values uses Microsoft "standard" values that have been around in the Microsoft ecosystem since .Net Framework 4.5 and still lives all the way up to .Net 10.

I am not sure if we should support our old "role" claim forever or the Microsoft claim as there are 0 specs about "role". IETF does have a spec about "roles" so if we would change it to that spec, it would still be a breaking change.

What this PR does is provide both the old (<=9.13.9) behavior of "role" as well as the new Microsoft way which may be better known in the .Net ecosystem. It also adds a deprecation note on both the plain and encrypted token to try and bring attention to this breaking change with the old way being obsolete and removed in v12.

Closes #6829

image

### DIFF
--- a/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
+++ b/DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
@@ -304,7 +304,19 @@ namespace Dnn.AuthServices.Jwt.Components.Common.Controllers
 
             var subject = new ClaimsIdentity();
             subject.AddClaim(new Claim(SessionClaimType, persistedToken.TokenId));
-            subject.AddClaims(roles.Select(r => new Claim(ClaimTypes.Role, r)));
+
+            // Add roles using both the standard schema URI (ClaimTypes.Role) for standards compliance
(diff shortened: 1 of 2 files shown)


CBPSC:
Nice!

PR #6833, opened 2025-12-08
Update all projects to use `PackageReference` instead of `packages.config`

bdukes:
## Summary
This PR removes all usages of `packages.config`, moving every project to use `PackageReference`.

It also updates some references from assembly references to project references.
It removes some unused files, e.g. `DotNetNuke.WebUtility.dll` and the `SolPartMenu` folder.

This PR builds on top of #6828, that PR should be merged first.


bdukes:

PR #6845, opened 2025-12-10
Don't package globbing assembly twice

bdukes:
## Summary
The install and upgrade package currently include two copies of the Microsoft.Extensions.FileSystemGlobbing library. This doesn't cause any particular issue, but it would be good to only have one copy of the package.


valadas:
I see but where is it included with this removal?

bdukes:
It's packaged by the `PackageMicrosoftGlobbing` task, so it didn't need to be in `thirdparty.json`

valadas:
Oh, I see, great!

valadas (approved):

PR #6846, opened 2025-12-10
Use SDK-style projects where possible

bdukes:
## Summary
This PR builds on top of #6833 to convert all projects that don't have `ascx` or `aspx` files into SDK-style projects.

This also enables .NET analyzers for all of those projects with the "recommended" analysis mode. For now, it downgrades all of the warnings (since _treat warnings as errors_ is on) so that the build doesn't fail. Resolving many of these warnings is the next task.


bdukes:

PR #6869, opened 2025-12-19
Prevent the installBlocker lock file to be removed by subsequent requests during Upgrade or Install.

### LINKED ISSUE #6859: [Bug]: Subsequent requests during Upgrade remove the installBlocker lock file.

### What happened?

When doing a DNN Upgrade using `/Install/Install.aspx?mode=upgrade`, if a second request to any page of the site is made by anyone, the lock file gets deleted prematurely.

The root cause is that the code in Install.aspx [`DotNetNuke.Services.Install.Install.UpgradeApplication()`](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Website/Install/Install.aspx.cs#L375-L377) that renders the `WriteInstallationInProgress()` is part of the same try-finally that is supposed to safely create then delete the lock file during installation.

So, what happens is that the first request (that starts the installation) enters the try-finally and runs execution, but then any request to the site gets redirected to UpgradeWizard.aspx initially (from Initialize.cs [`CheckVersion()`](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Common/Initialize.cs#L312)) and then to Install.aspx by the [`IsInstallInProgress()`](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Website/Install/UpgradeWizard.aspx.cs#L314-L317) check on the first line of `UpgradeWizard.OnLoad()`. These requests then enter the same try-finally to render the `WriteInstallationInProgress()` and then immediately returns, but due to the fact that they are in the try-finally, the finally block gets executed for them as well, which immediately removes the lock file during installation.

### Steps to reproduce?

1. Unzip a DNN Upgrade package over the site files.
2. Access `/Install/Install.aspx?mode=upgrade` (I think it can in fact be anything other than `mode=none`)
3. The lock file is created and Installation begins.
4. During installation make a request to anything on the site (e.g. Home page)
5. That request will get redirected to UpgradeWizard.aspx and then to Install.aspx by the `IsInstallInProgress()` check on the first line of `UpgradeWizard.OnLoad()` which will then fall through to render the `WriteInstallationInProgress()` and short circuit return, but not before also executing the `finally` block that removes the file.
6. The Installation continues after that without a lock file.

### Current Behavior

_No response_

### Expected Behavior

Lock file should be present for the entire Upgrade process.

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.1.2 (latest v10 release)

_No response_

### Code of Conduct

dimarobert:
Fixes #6859

## Summary
Uses a `cleanupLocker` boolean that gets set just before calling `RegisterInstallBegining` to control whether the locker gets cleaned up by the finally block. This prevents early returning code that is meant for any arbitrary requests made during the installation/upgrade process to remove the lock prematurely.

### DIFF
--- a/DNN Platform/Website/Install/Install.aspx.cs
+++ b/DNN Platform/Website/Install/Install.aspx.cs
@@ -224,6 +224,7 @@ namespace DotNetNuke.Services.Install
             }
             else
             {
+                bool cleanupLocker = false;
                 try
                 {
                     var synchConnectionString = new SynchConnectionStringStep();
@@ -259,6 +260,7 @@ namespace DotNetNuke.Services.Install
                                 return;
                             }
 
+                            cleanupLocker = true;
                             RegisterInstallBegining();
                         }
 
@@ -314,7 +316,10 @@ namespace DotNetNuke.Services.Install
                 }
                 finally
                 {
-                    RegisterInstallEnd();
+                    if (cleanupLocker)
+                    {
+                        RegisterInstallEnd();
+                    }
                 }
             }
         }
@@ -352,6 +357,7 @@ namespace DotNetNuke.Services.Install
 
         private void UpgradeApplication()
         {
+            bool cleanupLocker = false;
             try
             {
                 if (Upgrade.Upgrade.RemoveInvalidAntiForgeryCookie())
@@ -377,6 +383,7 @@ namespace DotNetNuke.Services.Install
                         return;
                     }
 
+                    cleanupLocker = true;
                     RegisterInstallBegining();
                 }
 
@@ -498,7 +505,10 @@ namespace DotNetNuke.Services.Install
             }
             finally
             {
-                RegisterInstallEnd();
+                if (cleanupLocker)
+                {
+                    RegisterInstallEnd();
+                }
             }
         }
 


dimarobert:
@bdukes I think this could be another solution as well https://github.com/dnnsoftware/Dnn.Platform/compare/develop...dimarobert:install-locker2

It makes the finally block execute cleanup only if the code exits after the `RegisterInstallBegining` call.
Eliminates both the `slight chance that if an exception in RegisterInstallBegining() happens the locker file will not be cleaned up` and the `no additional code should be written in the future between the lock and the try code blocks` cases.

bdukes:
> @bdukes I think this could be another solution as well [develop...dimarobert:install-locker2](https://github.com/dnnsoftware/Dnn.Platform/compare/develop...dimarobert:install-locker2)
> 
> It makes the finally block execute cleanup only if the code exits after the `RegisterInstallBegining` call.

I like that approach a lot better. Can you open a new PR for that branch (or push that commit to this PR's branch)?

dimarobert:
@bdukes pushed here

PR #6877, opened 2025-12-24
Fix packaging sample module dlls into DNN release packages

donker:
DNN 10.2.1 RC1 packed the ContactList sample module dlls into the release zip. This shouldn't happen.


donker:
Looking through the build artifacts I can confirm the sample module dlls have gone.

valadas:
Do we want to retarget this to `release/10.2.1` ?

valadas (approved):

PR #6878, opened 2025-12-31
Fixed some upgrade issues due to dependencies

valadas:
This allows proper upgrades from 9.13.10, 10.0.0 and 10.2.0 to 10.2.1.


valadas:

PR #6888, opened 2026-01-06
Add back RTL resource handling

### LINKED ISSUE #6885: [Bug]: Resource files (e.g. *.rtl.css) are not loaded after DNN 10.2.* (CDF changes)
### What happened?

After releasing DNN 10.2.* and introducing recent changes in the Client Dependency Framework (CDF), resource files with specific naming patterns—such as *.rtl.css—are no longer being loaded.

This behavior breaks RTL support in projects that rely on these files (e.g. Persian and Arabic websites), while the same setup worked correctly in versions prior to 10.2.

The issue was discussed extensively in the following Pull Request on this repository:
https://github.com/dnnsoftware/Dnn.Platform/pull/6622

As a result of that discussion, the necessary fix was implemented in the ClientDependency repository here:
https://github.com/dnnsoftware/ClientDependency/pull/15

However, in the released DNN 10.2.* versions, the issue still exists, and resource files like *.rtl.css are not included in the output, suggesting that the fix may not be fully integrated or is not functioning as expected within Dnn.Platform.

Expected behavior

Resource files such as *.rtl.css should be loaded correctly, consistent with behavior prior to DNN 10.2.

Actual behavior

Files matching patterns like *.rtl.css are ignored and never loaded.

Environment

DNN Platform: 10.2.*

Browser: All

Affected feature: Client Dependency Framework (CDF)

### Steps to reproduce?

In the path Portals/_default/Skins/Aperture/css, create a file named:

skin.min.rtl.css

Set the site (portal) language to an RTL language, for example Persian (fa-IR).

Load the site using this RTL language.

Expected Behavior

When the site is loaded in an RTL language, skin.min.rtl.css should be loaded instead of skin.min.css.

Actual Behavior

The RTL-specific stylesheet (skin.min.rtl.css) is not loaded, and skin.min.css is used instead.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.2.0 (latest v10 release)

Chrome

### Code of Conduct

bdukes:
Fixes #6885

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
+++ b/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
@@ -6,6 +6,7 @@ namespace DotNetNuke.Web.Client.ResourceManager
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Linq;
 
     using DotNetNuke.Abstractions.Application;
@@ -15,12 +16,23 @@ namespace DotNetNuke.Web.Client.ResourceManager
     public class ClientResourceController : IClientResourceController
     {
         private readonly IHostSettings hostSettings;
+        private readonly IApplicationStatusInfo appStatus;
 
         /// <summary>Initializes a new instance of the <see cref="ClientResourceController"/> class.</summary>
         /// <param name="hostSettings">The host settings.</param>
+        [Obsolete("Deprecated in DotNetNuke 10.2.2. Use overload with IApplicationStatusInfo. Scheduled removal in v12.0.0.")]
         public ClientResourceController(IHostSettings hostSettings)
+            : this(hostSettings, null)
+        {
+        }
+
+        /// <summary>Initializes a new instance of the <see cref="ClientResourceController"/> class.</summary>
+        /// <param name="hostSettings">The host settings.</param>
+        /// <param name="appStatus">The application status.</param>
+        public ClientResourceController(IHostSettings hostSettings, IApplicationStatusInfo appStatus)
         {
             this.hostSettings = hostSettings;
+            this.appStatus = appStatus;
             this.RegisterPathNameAlias("SharedScripts", "~/Resources/Shared/Scripts/");
         }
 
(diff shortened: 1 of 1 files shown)


valadas (approved):
Awesome!

mnouraei:
### Problem

The current RTL resource detection logic incorrectly combines a virtual path  
(`resource.FilePath`, starting with `~`) with a physical path (`ApplicationMapPath`)
when checking for file existence.

Example values:

```csharp
resource.FilePath = "~/Resources/Shared/stylesheets/dnndefault/10.0.0/default.css";
this.appStatus.ApplicationMapPath = "C:\\DnnTest\\dnn-10.2.1-v1";
```

This produces an invalid physical path:
```
C:\DnnTest\dnn-10.2.1-v1~/Resources/Shared/stylesheets/dnndefault/10.0.0/default.rtl.css
```

As a result, File.Exists(...) always returns false, even when the RTL file
actually exists.

```csharp
var ext = Path.GetExtension(resource.FilePath);
var rtlFilePath = Path.ChangeExtension(resource.FilePath, ".rtl" + ext);

if (!File.Exists(this.appStatus.ApplicationMapPath + rtlFilePath))
{
    return resource;
}

```

Issue:
resource.FilePath is a virtual path, but File.Exists requires a physical path.

Proposed Fix

The fix uses resource.ResolvedPath to correctly check the physical existence
of the RTL file, and only then updates the resource paths.

```csharp
var ext = Path.GetExtension(resource.FilePath);

// resource.FilePath is a virtual path (~),
// so we must use ResolvedPath to check physical file existence
var rtlResolvedPath = Path.ChangeExtension(resource.ResolvedPath, ".rtl" + ext);
var relativePath = rtlResolvedPath
    .TrimStart('~')
    .TrimStart(Path.DirectorySeparatorChar)
    .Replace("/", "\\");

if (!File.Exists($"{this.appStatus.ApplicationMapPath}\\{relativePath}"))
{
    return resource;
}

// Update virtual and resolved paths once RTL file is confirmed
var rtlFilePath = Path.ChangeExtension(resource.FilePath, ".rtl" + ext);
resource.FilePath = rtlFilePath;
resource.ResolvedPath = this.ResolvePath(rtlFilePath, resource.PathNameAlias);


```

Why This Fix

ResolvedPath already represents the physical file location in DNN
Prevents invalid physical paths caused by mixing virtual and physical paths
Aligns with DNN’s internal resource resolution mechanism
Ensures RTL stylesheets are correctly detected and loaded

bdukes:
I've pushed an update which uses the `ResolvePath` instead of the `FilePath`, I believe that should resolve the issue you raised. Thanks for taking a look and giving feedback.

uzmannazari (approved):
Looks good

valadas (approved):

PR #6897, opened 2026-01-08
Enhance font-weight for bold elements in SCSS

### LINKED ISSUE #6755: [Enhancement]: Aperture > Add separator to header when content scrolled

### Description of problem

Currently when scrolled you see this:

Image

Which "hides the header in a way.

### Description of solution

IMO adding a bit of a separation would be good:

Animated gif with and then without border:

Image

Used as test:
`
border-bottom: solid 1px #25678a;`

_No response_

### Anything else?

_No response_

- [x] Yes

- [ ] Yes

### Code of Conduct

uzmannazari:
Fixes #6755 

## Summary

This PR fixes an issue where `<strong>` and `<b>` elements were not rendered as bold in the Aperture theme.

A global reset rule was inheriting the full `font` property for these elements, which unintentionally removed the browser default bold behavior and caused them to render with the same font weight as their parent.

An explicit override using:

`b, strong { font-weight: bolder; }`

restores the expected semantic behavior by matching standard user-agent styling, without hard-coding a specific font weight or impacting layout and typography elsewhere.

### DIFF
--- a/DNN Platform/Skins/Aperture/src/scss/_base.scss
+++ b/DNN Platform/Skins/Aperture/src/scss/_base.scss
@@ -105,6 +105,11 @@ code {
   line-height: 1;
 }
 
+b,
+strong {
+  font-weight: bolder;
+}
+
 sup,
 sub {
   vertical-align: baseline;


valadas (commented):
This looks like an OK solution, however @david-poindexter would you be open to instead refactor the _reset.css which I believe may be a bit old reference with things maybe too strong and lacking more modern elements.

Something like https://www.npmjs.com/package/modern-normalize

mitchelsellers (approved):

mitchelsellers:
@dnnsoftware/approvers What do we think about just getting this one approved?

valadas:
I think we should instead replace https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/base/_reset.scss with https://www.npmjs.com/package/modern-normalize

This issue may affect parts where the theme is not used as it is very global, so it looks like a better fix to me and reduces code instead of adding.

donker:
> I think we should instead replace https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/base/_reset.scss with https://www.npmjs.com/package/modern-normalize
> 
> This issue may affect parts where the theme is not used as it is very global, so it looks like a better fix to me and reduces code instead of adding.

But that would change the scope of this issue. How about incorporating this and then fixing the reset in another PR and you can roll this change back at that point?

mitchelsellers:
I would agree with @donker I think at this point, as the change to reset at the 10.0.0 level would have more of an impact than just the Aperture theme

PR #6898, opened 2026-01-08
BreadCrumb skin object to render semantic, accessible breadcrumb markup

### LINKED ISSUE #6609: [Enhancement]: Breadcrumb WCAG compliant

### Description of problem

Currently the DNN breadcrumb isn't WCAG compliant. 

### Description of solution

It should be a listing `<ul><li>`

We tried to convert the breadcrumb skinobject in a listing but didn't succeed.

Change core breadcrumb so it's WCAG complaint

### Anything else?

_No response_

- [ ] Yes

- [ ] Yes

### Code of Conduct

uzmannazari:
## Summary
This PR refactors the BreadCrumb skin object to render semantic, list-based markup using `<nav>`, `<ol>`, and `<li>` elements instead of inline spans and separators.

The updated output improves accessibility and aligns breadcrumb rendering with common WCAG and ARIA breadcrumb patterns, while preserving existing behavior and customization options.

Fixes #6609

## Details
- Updated the BreadCrumb ASCX to wrap the generated breadcrumb items inside a semantic navigation container (`<nav aria-label="Breadcrumb">`) and an ordered list (`<ol itemscope itemtype="https://schema.org/BreadcrumbList">`).
- Replaced the `Label` with a `Literal` to avoid injecting wrapper markup inside the list container (ensuring valid list structure).
- Refactored backend rendering to output one `<li>` per breadcrumb item with Schema.org `BreadcrumbList` / `ListItem` microdata and `position` meta.
- Applied `aria-current="page"` to the `<li>` representing the current page (instead of the `<a>`), matching common ARIA breadcrumb patterns.
- Rendered separators inside each breadcrumb `<li>` (only when another crumb follows) and marked them decorative via `aria-hidden="true"`, avoiding extra list items while preserving the existing `Separator` customization and path resolution.
- Added default styling in `type.scss` for `nav.dnnBreadcrumb ol` to prevent list items stacking vertically after switching to list markup, and to ensure the ordered list does not display numeric markers while keeping the visual breadcrumb layout consistent:
  - `list-style: none;` (prevents numbering/markers)
  - `display: flex; align-items: center; gap: 0.5rem;` (keeps crumbs inline and spaced)
  - `nav.dnnBreadcrumb ol li span { margin-inline-start: 0.5rem; display: inline-block; }` (consistent spacing for the separator/content)
image

## Impact
- Improves accessibility and semantic correctness of breadcrumb markup.
- Preserves existing breadcrumb features and configuration (RootLevel, UseTitle, HideWithNoBreadCrumb, DisableLink, Profile/Group URL parameters, Separator customization).
- Visual output remains consistent with previous behavior via the added default styling.

## Testing
- Verified breadcrumb output for multi-level navigation paths renders inline (not stacked) and without ordered-list numbering.
- Tested root breadcrumb rendering and `HideWithNoBreadCrumb` behavior.
- Tested pages with disabled breadcrumb links.
- Confirmed structured data output remains valid for `BreadcrumbList`.
image

### DIFF
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/base/_type.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/base/_type.scss
@@ -45,14 +45,3 @@ ul, ol{
     padding-inline-start: 1.5rem;
     list-style-position: outside;
 }
-nav.dnnBreadcrumb ol {
-    list-style: none;
-    display: flex;
-    align-items: center;
-    gap: 0.5rem;
-}
-
-nav.dnnBreadcrumb ol li span {
-    margin-inline-start: 0.5rem;
-    display: inline-block;
-}
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/default.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/default.scss
@@ -1,6 +1,7 @@
 @use "base/base";
 @use "ui/ui";
 @use "modules/modules";
+@use "skinobjects/skinobjects";
 
 /* Hard to categorize as it is used in many places */
 h2.dnnFormSectionHead {
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/_breadcrumb.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/_breadcrumb.scss
@@ -1,0 +1,11 @@
+nav.dnnBreadcrumb ol {
+    list-style: none;
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+}
+
+nav.dnnBreadcrumb ol li span {
+    margin-inline-start: 0.5rem;
+    display: inline-block;
+}
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/skinobjects.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/skinobjects.scss
@@ -1,0 +1,1 @@
+@use "breadcrumb";


DNN Platform/Website/admin/Skins/breadcrumb.ascx
```diff
@@ -1,2 +1,6 @@
 <%@ Control Language="C#" AutoEventWireup="false" Inherits="DotNetNuke.UI.Skins.Controls.BreadCrumb" Codebehind="BreadCrumb.ascx.cs" %>
-<asp:label id="lblBreadCrumb" runat="server" EnableViewState="False" itemprop="breadcrumb" itemscope itemtype="https://schema.org/breadcrumb"/>
\ No newline at end of file
+<nav class="dnnBreadcrumb" aria-label="Breadcrumb">
```
mitchelsellers: Given that this is a literal name used for human interaction (via screen readers) I believe this should be pulling from localization

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -221,15 +209,15 @@ private void ResolveSeparatorPaths()
                     var url = match.Groups[3].Value;
                     var changed = false;
 
-                    if (url.StartsWith("/", StringComparison.Ordinal))
+                    if (url.StartsWith("/"))
```
bdukes: We want to keep the `StringComparison`

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -86,41 +82,35 @@ public string RootLevel
...
 
-        // Do not show when there is no breadcrumb (only has current tab)
         public bool HideWithNoBreadCrumb { get; set; }
 
-        /// <summary>Gets or sets a value indicating whether to take advantage of the enhanced markup (remove extra wrapping elements).</summary>
         public bool CleanerMarkup { get; set; }
 
         private IPortalAliasInfo CurrentPortalAlias => this.PortalSettings.PortalAlias;
 
-        /// <inheritdoc/>
```
bdukes, line 100: Please leave the doc comments

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -129,44 +119,39 @@ protected override void OnLoad(EventArgs e)
...
-                this.breadcrumb.Append("<a href=\"" + this.homeUrl + "\" class=\"" + this.cssClass + "\" itemprop=\"item\" ><span itemprop=\"name\">" + this.homeTabName + "</span></a>");
-                this.breadcrumb.Append("<meta itemprop=\"position\" content=\"" + position++ + "\" />"); // Notice we post-increment the position variable
-                this.breadcrumb.Append("</span>");
+                var hasMoreAfterRoot = crumbCount > this.rootLevel;
 
-                // Add a separator
-                this.breadcrumb.Append(this.separator);
-            }
+                breadcrumb.Append("<li itemprop=\"itemListElement\" itemscope itemtype=\"https://schema.org/ListItem\">");
+                breadcrumb.Append("<a href=\"" + this.homeUrl + "\" class=\"" + this.cssClass + "\" itemprop=\"item\"><span itemprop=\"name\">" + this.homeTabName + "</span></a>");
```
bdukes: Either use a formattable string or `AppendFormat`.

Also, the tab name will need to be HTML encoded

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -179,31 +164,34 @@ protected override void OnLoad(EventArgs e)
...
                 {
-                    if (this.CleanerMarkup)
-                    {
-                        this.breadcrumb.Append("<span class=\"" + this.cssClass + "\">" + tabName + "</span>");
-                    }
-                    else
-                    {
-                        this.breadcrumb.Append("<span><span class=\"" + this.cssClass + "\">" + tabName + "</span></span>");
-                    }
+                    breadcrumb.Append("<span class=\"" + this.cssClass + "\" itemprop=\"name\">" + tabName + "</span>");
```
bdukes: tabName needs to be HTML encoded

DNN Platform/Website/admin/Skins/breadcrumb.ascx
uzmannazari: Good call — agreed. I’ve switched the aria-label value to come from the control’s resx (e.g. BreadCrumbAriaLabel.Text) with a sensible fallback to "Breadcrumb" if the resource is missing. This keeps screen-reader text properly localized.

uzmannazari:
> One note/request included on a change to localization.
> 
> Additionally however I do believe this would be a substantially breaking change to introduce, so the planning of how this would be released is something we have to take into consideration. @dnnsoftware/approvers anyone else with thoughts on this?

Totally agree on the breaking-change risk. I’ve updated the implementation so the new semantic/list markup is opt-in via a new property (UseListMarkup, default false).
When `UseListMarkup="false" `the output stays in the legacy <span> format (to preserve existing skins/CSS), and only when true it renders the `<nav>/<ol>/<li>` semantic structure.

uzmannazari:
> I like adding this markup as an option, however I don't think we can change the markup by default, I think it needs to be opt-in (e.g. add a new property for display mode or `DisplayAsList` or something).

Agreed — I changed this to opt-in. Added UseListMarkup (default false) so legacy markup remains the default behavior. Setting `UseListMarkup="true"` enables the semantic list-based breadcrumb output.

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
uzmannazari: Fixed — restored StartsWith(..., StringComparison.Ordinal) (and kept StringComparison usage throughout) to avoid culture-sensitive comparisons.

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
uzmannazari: Updated — replaced string concatenations with AppendFormat/formattable strings and HTML-encoded tab names (HttpUtility.HtmlEncode) before rendering to prevent malformed markup and ensure safe output.

uzmannazari:
@bdukes and @mitchelsellers Thanks for the feedback! 

I’ve updated the PR to minimize breaking changes and address review notes:

Added UseListMarkup (default false) so legacy output remains the default, and semantic markup is opt-in.

Kept legacy wrapper `<span itemprop="breadcrumb" ...>` when `UseListMarkup=false `to preserve existing skin/CSS behavior.

Switched the control output to `Literal `to avoid invalid nesting; list-mode now starts with` <nav> `directly.

Localized the aria-label via resx with fallback.

Restored` StringComparison.Ordinal`, kept doc comments, and ensured tab names are HTML-encoded with `AppendFormat` usage.

DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/base/_type.scss
```diff
@@ -44,4 +44,15 @@ p a, p a:visited {
...
+    list-style: none;
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+}
+
+nav.dnnBreadcrumb ol li span {
+    margin-inline-start: 0.5rem;
+    display: inline-block;
+}
```
valadas: I am not a big fan of adding this css here, could we create a "skinobjects" folder similar to the "modules" folder and put in there one file per skinobject maybe?

Image

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -95,115 +102,265 @@ public string RootLevel
...
+        /// <summary>
+        /// Gets or sets a value indicating whether opt-in: render list-based semantic markup (nav/ol/li).
+        /// Default is false to preserve legacy output and reduce breaking changes.
+        /// </summary>
+        public bool UseListMarkup { get; set; }
+
         private IPortalAliasInfo CurrentPortalAlias => this.PortalSettings.PortalAlias;
 
-        /// <inheritdoc/>
         protected override void OnLoad(EventArgs e)
```
bdukes, line 113:
```suggestion
        /// <inheritdoc />
        protected override void OnLoad(EventArgs e)
```

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -95,115 +102,266 @@ public string RootLevel
...
+        private string BuildLegacyMarkup(int crumbCount)
+        {
+            var position = 1;
+
+            // IMPORTANT:
+            // We intentionally keep the legacy outer span wrapper so existing skins/CSS don't break.
+            // This mimics the old asp:Label output wrapper and keeps schema breadcrumb attrs.
+            var breadcrumb = new StringBuilder(
+                "<span itemprop=\"breadcrumb\" itemscope itemtype=\"https://schema.org/breadcrumb\">" +
+                "<span itemscope itemtype=\"http://schema.org/BreadcrumbList\">");
```
mitchelsellers: This should really be broken out into either one initialization & one Append call or just one big string to avoid the concatenated string being allocated.

mitchelsellers (changes requested):
Sorry upon further review one small changes suggested.

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
uzmannazari: Good catch — updated to initialize the StringBuilder and append the wrapper elements explicitly to avoid intermediate string concatenation allocations.

valadas:
@mitchelsellers are you good with this?

PR #6920, opened 2026-01-18
Made JWT TTL configurable

valadas:
JWT implementation had hardcoded values for the session timeout and the renewal timeout. With this PR we can set those values in the web.config instead.

I decided to use minutes for both values because:
- It is simpler to have the same unit of measure
- I needed to do some testing for another project and needed to make those very short just to trigger the bug.

The code fallback to the original hardcoded values if the settings don't exist and the initial default settings are put in the web.config for discoverability.


david-poindexter (approved):
Great!

mitchelsellers (approved):

Issue #6864, opened 2025-12-18
[Bug]: Filemanager page get's created as default page, not visible in personabar


trouble2:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

I installed DNN 10.2.0 with a blank template and in dutch. 
And I saw this in the menu :

<img width="367" height="104" alt="Image" src="https://github.com/user-attachments/assets/9554b678-cdec-48b3-ae1a-a7e4c41773ec" />

So I guess this might be a missing translation which causes this issue.
But the effect is that the filemanger for the portal is created as a 'normal' page, and that it is not a page in the personabar.
When I go to page management I actually see there are more missing items:

<img width="264" height="271" alt="Image" src="https://github.com/user-attachments/assets/4912b299-dca1-47d1-8d5a-7ef3647e7038" />

### Steps to reproduce?

1. install dnn 10.2.0 with a blank template in dutch  
(I'm not sure if this is an issue in other languages as well...)
2. check the pages

### Current Behavior

-

### Expected Behavior

-

### Relevant log output

```shell
-
```

### Anything else?

-

### Affected Versions

10.2.0 (latest v10 release)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

trouble2:
Looks like this is fixed!

Issue #6866, opened 2025-12-18
[Bug]: Dnn Prompt Showing Tokens instead of RESX Replacements


jeremy-farrance:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

After upgrading to v10.02.00 I was using Dnn PB, Settings / **Prompt** and noticed that the recently added LIST-SERVICES was not showing the expected text for the Category and Description as you can see in this screenclip.

<img width="1027" height="604" alt="Image" src="https://github.com/user-attachments/assets/8a517383-c9a7-4b93-be6b-c6429ed2ee89" />
<br>
I checked and found this problem on both an upgraded site as well as a clean install.

### Steps to reproduce?

When logged into DNN as host/super, go to PB/Settings, Prompt. At the prompt type "help" and press enter. Scroll up until you see the help text for the LIST-SERVICES command.

### Current Behavior

Tokens are showing instead of the text (probably) from resources.

### Expected Behavior

We should be seeing a Category and Description.

### Anything else?

Based on the screenshots, it seems possible this is somehow related to this recently posted Issue #6864 

### Affected Versions

10.2.0 (latest v10 release)

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Tychodewaard:
I have a similar thing on the website. Vanilla install but Dutch instead of English

<img width="1773" height="887" alt="Image" src="https://github.com/user-attachments/assets/f8925fc0-f92c-4930-84b4-b4768ceabc6e" />

valadas:
@Tychodewaard  I would probably report a different issue for the theme part. This works differently as it is not technically "static localization" but  content. It will get those values only once at site creation, it should use whatever fallback language is in place but that is probably a different part of code for a fix than the issue for the Prompt one.

PR #6939, opened 2026-01-27
Fix #6874 #6864 install missed resourcetext

### LINKED ISSUE #6874: [Bug]: Site Assets missing after fresh deployment of 10.2.0

### What happened?

Hi, 
I'm missing the Site Assets link in the persona bar (only Global Assets available) when signed in as super user. When trying to manually access  /Admin/File-Management url, I see a 404 in the event logs saying page not available. 

### Steps to reproduce?

1. Sign in as super user
2. Go to persona bar (manage)
3. No Site Assets available (only Global Assets)

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.2.0 (latest v10 release)

Microsoft Edge

### Code of Conduct

thienvc:
Fix #6874 #6864
Fix the "Site Assets" link in the persona bar (only "Global Assets" were available) when signed in as a super user. When trying to manually access the /Admin/File-Management URL, a 404 error appears in the event logs stating the page is unavailable when using a non-en-US language setting.....

Solution: I have modified the code to populate the missing language resource data based on the original en-US source.

https://github.com/dnnsoftware/Dnn.Platform/blob/7d85ff4fe4913b0793d291df11707cdf4cf7a0d4/DNN%20Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs#L45-L80

```
internal PortalTemplateImporter(IPortalTemplateInfo templateToLoad)
{
    var buffer = new StringBuilder(File.ReadAllText(templateToLoad.TemplateFilePath));

    if (!string.IsNullOrEmpty(templateToLoad.LanguageFilePath))
    {
        XDocument languageDoc;
        using (var reader = PortalTemplateIO.Instance.OpenTextReader(templateToLoad.LanguageFilePath))
        {
            languageDoc = XDocument.Load(reader);
        }

        // var localizedData = languageDoc.Descendants("data");
        var localizedData = languageDoc.Descendants("data")
            .Where(x => x.Attribute("name") != null)
            .ToDictionary(x => x.Attribute("name").Value, x => x);

        // check missed localize resource text
        if (templateToLoad.CultureCode != "en-US")
        {
            // var rootDoc = languageDoc.Root;
            var defLangFilePath = PortalTemplateIO.Instance.GetLanguageFilePath(templateToLoad.TemplateFilePath, "en-US");
            if (!string.IsNullOrEmpty(defLangFilePath))
            {
                // var isModified = false;
                XDocument defLanguageDoc;
                using (var reader = PortalTemplateIO.Instance.OpenTextReader(defLangFilePath))
                {
                    defLanguageDoc = XDocument.Load(reader);
                }

                var defData = defLanguageDoc.Descendants("data");
                foreach (var defNode in defData)
                {
                    var key = defNode.Attribute("name")?.Value;
                    if (string.IsNullOrEmpty(key))
                    {
                        continue;
                    }

                    if (!localizedData.ContainsKey(key))
                    {
                        // isModified = true;
                        // rootDoc.Add(new XElement(defNode));
                        localizedData.Add(key, defNode);
                    }
                }

                // if(isModified)
                // {
                //    using (var writer = PortalTemplateIO.Instance.OpenTextWriter(templateToLoad.LanguageFilePath))
                //    {
                //        languageDoc.Save(writer);
                //    }
                // }
            }
        }

        foreach (var node in localizedData.Values)
        {
            var name = node.Attribute("name")?.Value;
            if (string.IsNullOrEmpty(name))
            {
                continue;
            }

            var valueElement = node.Element("value");
            if (valueElement != null)
            {
                var value = valueElement.Value;
                buffer.Replace($"[{name}]", value);
            }
        }
    }

    this.TemplatePath = Path.GetDirectoryName(templateToLoad.TemplateFilePath);
    this.Template = new XmlDocument { XmlResolver = null };
    using var templateReader = XmlReader.Create(new StringReader(buffer.ToString()), new XmlReaderSettings { XmlResolver = null, });
    this.Template.Load(templateReader);
}
```

### DIFF
--- a/DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
+++ b/DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
@@ -74,66 +74,7 @@ namespace DotNetNuke.Entities.Portals.Templates
 
             if (!string.IsNullOrEmpty(templateToLoad.LanguageFilePath))
             {
-                XDocument languageDoc;
-                using (var reader = PortalTemplateIO.Instance.OpenTextReader(templateToLoad.LanguageFilePath))
-                {
-                    languageDoc = XDocument.Load(reader);
-                }
-
-                // var localizedData = languageDoc.Descendants("data");
-                var localizedData = languageDoc.Descendants("data")
-                    .Where(x => x.Attribute("name") != null)
-                    .ToDictionary(x => x.Attribute("name").Value, x => x);
-                
-                // check missed localize resource text
-                if (templateToLoad.CultureCode != "en-US")
-                {
-                    // var rootDoc = languageDoc.Root;
-                    var defLangFilePath = PortalTemplateIO.Instance.GetLanguageFilePath(templateToLoad.TemplateFilePath, "en-US");
-                    if (!string.IsNullOrEmpty(defLangFilePath))
-                    {
-                        // var isModified = false;
-                        XDocument defLanguageDoc;
-                        using (var reader = PortalTemplateIO.Instance.OpenTextReader(defLangFilePath))
-                        {
-                            defLanguageDoc = XDocument.Load(reader);
-                        }
-                
-                        var defData = defLanguageDoc.Descendants("data");
-                        foreach (var defNode in defData)
-                        {
-                            var key = defNode.Attribute("name")?.Value;
-                            if (string.IsNullOrEmpty(key))
-                            {
-                                continue;
-                            }
-                
-                            if (!localizedData.ContainsKey(key))
-                            {
-                                // isModified = true;
-                                // rootDoc.Add(new XElement(defNode));
-                                localizedData.Add(key, defNode);
-                            }
-                        }
-                
-                    }
-                }
-                
-                foreach (var node in localizedData.Values)
-                {
-                    var name = node.Attribute("name")?.Value;
-                    if (string.IsNullOrEmpty(name))
-                    {
-                        continue;
-                    }
-                
-                    var valueElement = node.Element("value");
-                    if (valueElement != null)
-                    {
-                        var value = valueElement.Value;
-                        buffer.Replace($"[{name}]", value);
-                    }
-                }
+                ReplaceTemplateTokens(templateToLoad, buffer);
             }
 
             this.TemplatePath = Path.GetDirectoryName(templateToLoad.TemplateFilePath);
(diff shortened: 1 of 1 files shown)


thienvc:
im not sure this relate with #6866

mitchelsellers:
@thienvc It looks like another ite got pulled in with this rather than just the change noted?

thienvc:
> @thienvc It looks like another ite got pulled in with this rather than just the change noted?

I’m not sure why, but whenever I try to push from Visual Studio, it includes a lot of unintended previous changes. To avoid that, I've been manually editing on the web and copying the code here.
Sorry for my messy GitHub and limited English. Could you help me test and then open a new PR? I don't mind who gets the credit for the PR at all

thienvc:
@mitchelsellers 
I’m not quite sure which issue you’re referring to. Could you please clarify?

Initially, I planned to modify the XML file by adding the missing entries and saving it. However, I now think it's sufficient to just ensure the installation process is valid by falling back any missing items to 'en-US'. Actual localization should be the responsibility of those who understand the language.

Should we add a suffix like '... (en-US)' so users know those specific items haven't been translated yet?

mitchelsellers:
Sorry it looks like my typing was maybe a bit wrong as well.

Yes, it looks like we have some other changes that came along here and we need to try to split it out.

I can try to re-play this tomorrow to get a clean change.

mitchelsellers:
@dnnsoftware/approvers Do we hav eany feedback on this one

DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
```diff
@@ -80,21 +80,65 @@ internal PortalTemplateImporter(IPermissionDefinitionService permissionDefinitio
...
+                    .Where(x => x.Attribute("name") != null)
+                    .ToDictionary(x => x.Attribute("name").Value, x => x);
+                
+                // check missed localize resource text
+                if (templateToLoad.CultureCode != "en-US")
                 {
-                    var nameAttribute = item.Attribute("name");
-                    if (nameAttribute != null)
+                    // var rootDoc = languageDoc.Root;
+                    var defLangFilePath = PortalTemplateIO.Instance.GetLanguageFilePath(templateToLoad.TemplateFilePath, "en-US");
```
bdukes: Does `def` here mean `default`? I would prefer the full word for clarity.

DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
```diff
@@ -80,21 +80,65 @@ internal PortalTemplateImporter(IPermissionDefinitionService permissionDefinitio
...
+                            }
                         }
+                
+                        // if(isModified)
+                        // {
+                        //    using (var writer = PortalTemplateIO.Instance.OpenTextWriter(templateToLoad.LanguageFilePath))
+                        //    {
+                        //        languageDoc.Save(writer);
+                        //    }
+                        // }
```
bdukes: Let's remove these comments. Thanks!
```suggestion
```

bdukes:
I fixed the build issues (they were all whitespace-related). While I was in there, I extracted the logic into two methods to it's easier to follow what's going on. I also did some more variable renaming and removed other code comments.

PR #6950, opened 2026-01-29
Get ContentEditorManager dependencies from request scope

bdukes:
## Summary
In DNN 10.2.2 RC1, the edit bar does not display in edit mode. There is an error in the console about `dnn.ContentEditorManager` being undefined, which points to [`ContentEditor.js`](https://github.com/dnnsoftware/Dnn.Platform/blob/fd590cf8208d3c0e93de6a8ef0cc0e4c9afd11dc/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/editBar/resources/ContentEditorManager/Js/ContentEditor.js) not being registered by [`ContentEditorManager`](https://github.com/dnnsoftware/Dnn.Platform/blob/0115d33b6ba769916049308e87a462dae2829488/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/Controllers/ContentEditorManager.cs#L347).

This line was successfully running, but when the scripts were rendered on the page, the script was not in the list of requested scripts. This is because the `IClientResourceController` passed to the constructor of `ContentEditorManager` did not come from the current request scope, but was [created when `EditBarModule` was instantiated](https://github.com/dnnsoftware/Dnn.Platform/blob/03ee406ad582ecdd2a334d55b5a8d74f011fa901/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/HttpModules/EditBarModule.cs#L72-L109) was instantiated (due to timing issues with DI support in `IHttpModule` implementations).

The solution in this PR is to get the dependencies from the request scope when needed, rather than using constructor injection.

### DIFF
--- a/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/HttpModules/EditBarModule.cs
+++ b/Dnn.AdminExperience/EditBar/Dnn.EditBar.UI/HttpModules/EditBarModule.cs
@@ -18,66 +18,35 @@ namespace Dnn.EditBar.UI.HttpModules
     using DotNetNuke.Common.Extensions;
     using DotNetNuke.Entities.Controllers;
     using DotNetNuke.Entities.Host;
-    using DotNetNuke.Entities.Modules;
     using DotNetNuke.Entities.Portals;
     using DotNetNuke.Entities.Users;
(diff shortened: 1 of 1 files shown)


valadas (approved):
Excellent! Thanks again!

PR #6953, opened 2026-01-29
Fixed an issue where getting permissions from DB would fail at times

valadas:
While working on #6952 this would fail after each restart for some reason, with this change it no longer does. I don't know if this is a good fix performance-wise, any ideas @bdukes ?

### DIFF
--- a/DNN Platform/Library/DotNetNuke.Library.csproj
+++ b/DNN Platform/Library/DotNetNuke.Library.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="AngleSharp" />
     <PackageReference Include="AngleSharp.Css" />
     <PackageReference Include="BouncyCastle.Cryptography" />
+    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-j92c-7v7g-gj3f" />
     <PackageReference Include="HtmlSanitizer" />
     <PackageReference Include="MailKit" />
     <PackageReference Include="Microsoft.AspNet.WebPages" />


thienvc:
1. `private static readonly DataProvider Provider = DataProvider.Instance();`
2. `private static DataProvider Provider => DataProvider.Instance();`

I believe the first approach is better for performance. Given that Provider is a high-traffic object in our codebase, using static readonly ensures we only resolve the dependency once, whereas the current => syntax triggers a factory lookup on every single call.

valadas:
Yes but `Provider` is null at times. So far I can reproduce this way:
- Run a full build and install DNN from the produced install artifact
- Build the library in debug mode. This will trigger a site restart and then Provider is null and we get these errors on page for each an every module:

image

DNN Platform/Library/Security/Permissions/PermissionController.cs
```diff
@@ -23,27 +23,33 @@ namespace DotNetNuke.Security.Permissions
     using DotNetNuke.Security.Roles;
 
     using Microsoft.Extensions.DependencyInjection;
+    using Org.BouncyCastle.Tls.Crypto;
```
valadas:
```suggestion
```

valadas:
I pushed an additional commit that makes it better, however I am scared that this hides an underlying issue with DI or the ComponentModel pattern thing. So I put this in draft. It looks like I also cannot place a module on a page, $.dnnSf is not on the page and I'll investigate that too which may show an issue affecting both maybe...

thienvc:
```
private static DataProvider _provider;
private static DataProvider Provider => _provider ??= DataProvider.Instance();

```
I'm curious if doing it this way would be more feasible.

donker:
IMHO I think the way now proposed is fine with the call inside the callback. The DataProvider.Instance() is only called when the cache expires. I'd prefer not to put anything at class level as down the line someone may inadvertently use it which is not desirable. It's only there for the callback.

DNN Platform/Library/Security/Permissions/PermissionController.cs
```diff
@@ -27,23 +27,28 @@ namespace DotNetNuke.Security.Permissions
     /// <summary>The default <see cref="IPermissionDefinitionService"/> implementation.</summary>
     public partial class PermissionController : IPermissionDefinitionService
     {
-        private static readonly DataProvider Provider = DataProvider.Instance();
         private readonly IEventLogger eventLogger;
+        private readonly DataProvider dataProvider;
```
mitchelsellers, line 31: This change has me slightly confused.

Why are we keeping a `dataProvider` private that is readony only, and introducing a new static, but changing some implementations to use `dataProvider` and others to use `Provider`?

I would expect all to the be the same.  The static nature of that provider I guess also scares me, but maybe that is the proper intention context.

DNN Platform/Library/Security/Permissions/PermissionController.cs
```diff
@@ -299,7 +304,7 @@ private static IEnumerable<PermissionInfo> GetPermissions()
                 DataCache.PermissionsCacheKey,
                 DataCache.PermissionsCacheTimeout,
                 DataCache.PermissionsCachePriority),
-                c => CBO.FillCollection<PermissionInfo>(Provider.ExecuteReader("GetPermissions")));
+                c => CBO.FillCollection<PermissionInfo>(DataProvider.Instance().ExecuteReader("GetPermissions")));
```
bdukes: Is there a problem using `this.dataProvider` here?

valadas: I am not sure, I put this PR is draft because I am not in love with it and suspect we have a widespread CI container management issue. I ran out of time to troubleshoot deeper

valadas: Because that method is static...

I am not in love with this PR but it appears to resolve the issue, I'll put it back to "ready to review",
if you are fine with it, please merge @bdukes but if you have better ideas, I am all ears too.

BedtimeMouseGIF

bdukes:
I've pushed a commit which gets rid of the static property (by passing a value through to the static method).

valadas:
> I've pushed a commit which gets rid of the static property (by passing a value through to the static method).

Looks good, I'll test and merge once the build completes. Thanks @bdukes

valadas:
@bdukes it gave me this:
image

bdukes:
Yes, I'm seeing that too, now. ~~Maybe we need to use `DataProvider.Instance` instead of `Globals.GetCurrentServiceProvider().GetRequiredService<DataProvider>()`…~~ Nevermind, we're already doing that.

valadas:
This is the call stack that appears to cause thsi.dataProvider to be null, still scratching my head as to why though...
image

bdukes:
The issue was that before the `ComponentFactory` was setup, `Globals` was requesting `INavigationManager` from the container, which has a transient dependency on `DataProvider`. Since it wasn't in the `ComponentFactory`, it was coming through as `null`. I've rearranged the `ComponentFactory` initialization timing so that this all works are expected.

valadas:
@bdukes that broke some tests

bdukes:
tests are fixed

valadas:
When either #6968 or #6969 are merged, we can rebase this and re-test...

valadas:
I gave it a test and can no longer reproduce the issue. Merging.

PR #6955, opened 2026-01-31
Fix RTL resource detection by resolving virtual path to physical path

mnouraei:
### Summary
Fixes an issue where RTL-specific resources were not detected correctly because
`File.Exists` was being called with a virtual path.

### Details
- Converts RTL resolved paths from virtual (`~/...`) to physical paths
- Uses `Path.Combine` to safely build the physical file system path
- Ensures RTL resources are only swapped when the corresponding `.rtl` file exists

### Impact
- RTL styles/scripts are now correctly loaded when available
- No behavior change for LTR cultures or CDN/external resources

### Notes
This change only affects local resources and preserves existing CDN logic.


uzmannazari (approved):
Nice improvement, thanks for the contribution!

PR #6960, opened 2026-02-01
Fix Search SkinObject clear button semantics and styles

### LINKED ISSUE #6959: [Enhancement]: Search SkinObject uses anchor without href causing SEO & accessibility issues

### Description of problem

The Search SkinObject renders a clear-text control using an` <a>` element without an `href `attribute:
`<a class="dnnSearchBoxClearText" title="<%= ClearQueryText %>"></a>`
This element is not a navigational link and is only used as a UI control to clear the search input via JavaScript.

As a result:

- Google Lighthouse and PageSpeed report “Links are not crawlable”
- Google Search Console shows persistent crawlability warnings
- The markup violates HTML semantics and accessibility guidelines (WCAG), since an anchor without a destination is announced as a link with no target

This warning appears consistently on DNN-based sites and cannot be resolved at the site level without modifying the core Search SkinObject.

Image

### Description of solution

Replace the anchor element with a semantic button element:

```
<button
  type="button"
  class="dnnSearchBoxClearText"
  title="Clear search"
  aria-label="Clear search">
</button>
```
Why this works:

- A button is the correct semantic element for this behavior
- Existing JavaScript and CSS continue to work without changes
- SEO crawlability warnings are resolved
- Accessibility is improved (keyboard and screen-reader friendly)
- No breaking changes to behavior or UI

Adding `href="#"`
❌ Causes unwanted page jumps and is considered an anti-pattern

Keeping `<a> `and adding role="button"
❌ Still exposes an anchor without a navigable destination

Replacing with` <span>`
❌ Lacks native accessibility and keyboard support compared to <button>

The `<button>` element is the most correct and standards-compliant solution.

### Anything else?

This is a long-standing, low-risk issue that affects:

- SEO scores (Lighthouse / PageSpeed)
- Accessibility compliance
- HTML validity

Fixing it in core improves all DNN sites automatically without requiring custom skins or overrides.

- [x] Yes

- [ ] Yes

### Code of Conduct

uzmannazari:
Fixes #6959

## Summary

This PR addresses an SEO and accessibility issue in the Search SkinObject and includes a small structural cleanup aligned with recent core discussions.

### Changes included:

- Replaced the non-semantic `<a>` element used as the "clear search text" control with a semantic `<button type="button">`
  - Resolves Lighthouse and Search Console warnings: **“Links are not crawlable”**
  - Improves accessibility and HTML semantics
  - No behavioral or visual changes

- Removed `SearchSkinObjectPreview.css` and relocated its styles into the centralized SkinObjects styling structure
  - Follows the suggestion raised in a previous PR review by @valadas to organize styles under a dedicated `skinobjects` folder (similar to the existing `modules` folder)
  - Keeps one logical stylesheet per SkinObject

- Converted the Search SkinObject styles to use CSS logical properties
  - Improves RTL/LTR support without requiring direction-specific overrides
  - Aligns with modern, direction-agnostic CSS practices
  - Output behavior remains unchanged

### Notes / Follow-up

Currently, only CSS for the Search SkinObject has been moved into the SkinObjects structure.

If there is an established or preferred mechanism in `Dnn.ClientSide` for registering JavaScript files for SkinObjects, please advise.  
In that case, `SearchSkinObjectPreview.js` could be relocated accordingly in a follow-up PR to keep CSS and JS organization consistent.

No JavaScript logic or runtime behavior was modified in this PR.


DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/_search.scss
valadas, line 1: I would prefer to not bloat the default.css with search specific styles as many sites don't even use search. I would prefer to keep it in the skinobject css

DNN Platform/Website/admin/Skins/search.ascx
```diff
@@ -6,7 +6,12 @@
     <asp:RadioButton ID="SiteRadioButton" runat="server" CssClass="SkinObject" GroupName="Search" />
     <span class="searchInputContainer" data-moreresults="<%= SeeMoreText %>" data-noresult="<%= NoResultText %>">
         <asp:TextBox ID="txtSearch" runat="server" EnableViewState="False" CssClass="NormalTextBox" Columns="20" MaxLength="255" aria-label="Search"></asp:TextBox>
-        <a class="dnnSearchBoxClearText" title="<%= ClearQueryText %>"></a>
```
valadas, line 9: I like the idea, but this is breaking for many themes that style the search expecting this to be a link. I am not saying we should not do it, but we either have to make this breaking change in a major version or provide some opt-in way for the new markup.

uzmannazari: Thanks for the clarification — understood 

Before proceeding further, what would be your preferred approach here to avoid breaking existing themes?
I’m happy to adjust the PR accordingly.

DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/skinobjects/_search.scss
uzmannazari: I moved the styles into the SkinObject-specific location based on your earlier breadcrumb PR feedback.
If that still results in Search CSS being bundled into default.css, I’m happy to change course and keep it fully isolated and conditionally loaded.

valadas: That will still get bundled into default.css, the place it was first was OK since we don't need that CSS unless the skinobject is used.

uzmannazari: I’ll revert the CSS back to its original conditional-loading location so it’s only included when the Search SkinObject is used.

DNN Platform/Website/admin/Skins/search.ascx
valadas: Well, a way to not be breaking would certainly be clunky and the classname is not changed, just something to discuss I guess, and I am not the only one having to decide, so we need to just be aware of it and discuss it probably in next TAG meeting.

1. We could hold the PR until DNN 11 which will mean probably a lot of rebases.
2. We could hold the PR until 10.3.0 and call attention to it in the release notes, it is a breaking change but not a major thing either.
3. We could have a "legacy" property to the skinObject that is true by default and then people pass false in their own themes if they want to last markup and deal with it knowing what they are doing. But then we live with a property that people may set and it makes it hard to removed it because then it becomes an even worse breaking change causing an error. We could keep that property around even if it does nothing, that way it helps with future changes..

Just throwing ideas, we need to discuss this and decide as a group.

uzmannazari: Thanks for outlining the options, that helps a lot 👍

Just to add a bit more context from my side: the main goal of this PR was to improve DNN’s default SEO and accessibility out of the box.
An opt-in or legacy-flag approach would technically work, but it does add long-term complexity and doesn’t really improve the default experience, since theme authors can already override the markup today.

Out of the options you mentioned, option 2 (including this in 10.3.0 with clear release notes calling out the breaking change) feels like a reasonable middle ground to me:
- the breaking surface is small and well-scoped
- it avoids carrying a legacy flag long-term
- and it still moves the platform forward in a meaningful way

That said, I agree this should be discussed and decided as a group in the next TAG meeting.
I’m happy to hold, retarget, or reshape the PR based on the outcome of that discussion.

mitchelsellers: Just commenting to throw my vote for option 2 here as well

DNN Platform/Website/admin/Skins/search.ascx
```diff
@@ -18,7 +23,12 @@
         </div>
         <span class="searchInputContainer" data-moreresults="<%= SeeMoreText %>" data-noresult="<%= NoResultText %>">
             <asp:TextBox ID="txtSearchNew" runat="server" CssClass="SearchTextBox" MaxLength="255" EnableViewState="False" aria-label="Search"></asp:TextBox>
-            <a class="dnnSearchBoxClearText" title="<%= ClearQueryText %>"></a>
+            <button
+              type="button"
+              class="dnnSearchBoxClearText"
+              title="<%= Clear search %>"
+              aria-label="<%= Clear search %>">
```
bdukes, line 30:
```suggestion
              title="<%: ClearQueryText %>"
              aria-label="<%: ClearQueryText %>">
```

DNN Platform/Website/admin/Skins/search.ascx
```diff
@@ -6,7 +6,12 @@
     <asp:RadioButton ID="SiteRadioButton" runat="server" CssClass="SkinObject" GroupName="Search" />
     <span class="searchInputContainer" data-moreresults="<%= SeeMoreText %>" data-noresult="<%= NoResultText %>">
         <asp:TextBox ID="txtSearch" runat="server" EnableViewState="False" CssClass="NormalTextBox" Columns="20" MaxLength="255" aria-label="Search"></asp:TextBox>
-        <a class="dnnSearchBoxClearText" title="<%= ClearQueryText %>"></a>
+        <button
+          type="button"
+          class="dnnSearchBoxClearText"
+          title="<%= Clear search %>"
+          aria-label="<%= Clear search %>">
```
bdukes, line 13:
```suggestion
          title="<%: ClearQueryText %>"
          aria-label="<%: ClearQueryText %>">
```

mitchelsellers:
@uzmannazari Can you rebase this and update with the new RenderingVersion item

uzmannazari:
@mitchelsellers 
Yes, absolutely.

I’ll rebase the PR and update the Search SkinObject to align with the new `RenderingVersion` approach.

The plan is to keep the current markup as the legacy rendering (`v1`) and introduce the updated semantic/accessibility improvements under a higher rendering version, so existing skins remain fully backward compatible.

I’ll push an updated revision shortly. Thanks for the guidance on this.

uzmannazari:
> @uzmannazari Can you rebase this and update with the new RenderingVersion item

Linking this work to #6972.

I’ve left a review there regarding `RenderingVersion` being `protected`, since making it publicly settable is required for this PR to proceed cleanly.

Will rebase and update once that’s addressed.

valadas:
> > @uzmannazari Can you rebase this and update with the new RenderingVersion item
> 
> Linking this work to #6972.
> 
> I’ve left a review there regarding `RenderingVersion` being `protected`, since making it publicly settable is required for this PR to proceed cleanly.
> 
> Will rebase and update once that’s addressed.

I don't see a comment about that, protected should work in this context, did you try it out?

uzmannazari:
> > > @uzmannazari Can you rebase this and update with the new RenderingVersion item
> > 
> > 
> > Linking this work to #6972.
> > I’ve left a review there regarding `RenderingVersion` being `protected`, since making it publicly settable is required for this PR to proceed cleanly.
> > Will rebase and update once that’s addressed.
> 
> I don't see a comment about that, protected should work in this context, did you try it out?

It was my mistake. My comment was pending and not posted.

valadas:
@uzmannazari it has been made public if you want to rebase and adjust this PR to use the RenderingVersion

mitchelsellers:
it has been almost 2 months without updates for this PR to the suggestions above.  At this time I'm going to close the PR if someone would like to make the additional changes this can be reopened.

PR #6968, opened 2026-02-04
Bump HtmlSanitizer

bdukes:
## Summary
In order to avoid NU1902 error, this PR bumps HtmlSanitizer (which requires bumping System.Collections.Immutable).

Locally, this caused issues with the source generator, pushing this to CI to see if it's a local only issue.

### DIFF
--- a/DNN Platform/Library/DotNetNuke.Library.csproj
+++ b/DNN Platform/Library/DotNetNuke.Library.csproj
@@ -23,7 +23,6 @@
     <PackageReference Include="AngleSharp" />
     <PackageReference Include="AngleSharp.Css" />
     <PackageReference Include="BouncyCastle.Cryptography" />
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-j92c-7v7g-gj3f" />
     <PackageReference Include="HtmlSanitizer" />
     <PackageReference Include="Lucene.Net.Contrib" />
     <PackageReference Include="MailKit" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -20,7 +20,7 @@
     <PackageVersion Include="Dnn.CakeUtils" Version="2.1.4" />
     <PackageVersion Include="Dnn.ClientDependency" Version="1.10.0" />
     <PackageVersion Include="Google.Apis.Auth" Version="1.69.0" />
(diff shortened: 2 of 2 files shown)


bdukes:
hmmm… to upgrade that, it bumps System.Collections.Immutable to 10.0.0 (from 9.0.2), and that bump causes errors with the source generator for some reason.

>error : Generator 'DnnDeprecatedGenerator' failed to initialize. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'FileNotFoundException' with message 'Could not load file or assembly 'System.Collections.Immutable, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.'.

david-poindexter (dismissed):

mitchelsellers (dismissed):
This looks good with one question.

Does this upgrade result in a need for a binding redirect?

bdukes:
It will require an update to the binding redirect. It's already included in the _HtmlSanitizer_ package, but that assembly is included in the `bin` folder of the upgrade package, so it'll cause an error using the traditional upgrade process.

donker:
> It will require an update to the binding redirect. It's already included in the _HtmlSanitizer_ package, but that assembly is included in the `bin` folder of the upgrade package, so it'll cause an error using the traditional upgrade process.

So does this mean we already move forward to the new upgrade process? Meaning: not release the upgrade package and point users to the upgrade module instead?

valadas (dismissed):

valadas:
@bdukes my understanding is that this works fine with the new upgrade process and we have adjusted the documentation for that right?

bdukes:
10.2.3 resolves an issue introduced in 10.2.2 which required using the upgrade process to avoid binding redirect issues. I'm not sure where we landed on requiring the upgrade process for 10.3.0.

valadas (dismissed):

bdukes:
I've updated the branch to bump both dependencies to their latest version:
- System.Collections.Immutable to 10.0.10 (previously it bumped from 9.0.2 to 10.0.1)
- HtmlSanitizer to 9.1.968-beta (previously it bumped from 9.1.878-beta to 9.1.893-beta)

I also added a commit to revert #6970

valadas (approved):

mitchelsellers (approved):

PR #6969, opened 2026-02-04
Updated HtmlSanitizer

valadas:
Replaced usage of string.Contains with string.IndexOf for compatibility with older .NET versions in Upgrade.cs. Updated NuGet package versions for HtmlSanitizer, Microsoft.CodeAnalysis.CSharp, and System.Collections.Immutable. Added MSBuild properties and explicit package versions in DotNetNuke.Internal.SourceGenerators.csproj to improve build configuration and compatibility.


valadas:
#6968 resolves this and is better

PR #6972, opened 2026-02-05
Added RenderingVersion to skinobjects

### LINKED ISSUE #6967: [Enhancement]: SkinObject Versioning & Moving Forward with Breaking Changes

### Description of problem

As the DNN Platform continues to evolve we will need the ability to change SkinObjects to render differently to meet modern standards and improve markup for usability, semantics, and accessibility.  To this point we may have to implement breaking changes such as [this change for the Search SkinObject](https://github.com/dnnsoftware/Dnn.Platform/pull/6960), or [this change to the Breadcrumb Skin Object](https://github.com/dnnsoftware/Dnn.Platform/pull/6898).

### Description of solution

In the Breadcrumb PR a change was added to set "UseListMarkup" to let someone opt-in to the functionality, this works, however, it creates a changing method signature and introduces a property that is impossible to remove without breaking changes in the future.  

During the Technology Advisory Group meeting on 2/3/2026 it was proposed by @sleupold that we instead should consider introducing a "Version" attribute to skin objects, that allows us to set a version, and render properly.

For example the breadcrumb could be updated to support

* version="1" - Would render as the existing style
* version="2" - Would render as the neew list style
* Missing Version - Would render as the "most current"

This style of implementation would make it easier for changes in the future without breaking; however, the discoverability and documentation might be slightly harder to manage.

There was also discussion of using DNN Version or DNN Major Version rather than arbitrary versions, but the documentation question arises once again.

The question to the community is around which strategy would make the most sense

1. Version is unique to the SkinObject, and the current if no version has been made yet is 1, increment with breaking changes
2. Version is tied to DNN Version when introduced, if unknown default is 1. (This gets tricky for all existing skin objects, but could be done, and then the breadcrumb version would be 100200.   
3. Do not go this route and keep with the flags, knowing that we cannot ever remove a flag.

I will post my personal opinion below

### Code of Conduct

valadas:
Adds support for `RenderingVersion` to all skinobjects so that we can do progressive enhancements to them with less fear of breaking changes.

Closes #6967

@mitchelsellers @david-poindexter @sleupold @jeremy-farrance @bdukes @uzmannazari 
Let me know if that is what you had in mind or something different..

### DIFF
--- a/DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
+++ b/DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
@@ -18,6 +18,10 @@ namespace DotNetNuke.UI.Skins.Controls
     using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>A skin/theme object which displays the hierarchy of the current page.</summary>
+    /// <remarks>
+    /// The legacy renderings uses spans to mimic the old asp:Label output for backward compatibility.
+    /// v2 rendering uses nav/ol/li markup for better semantics and accessibility.
+    /// </remarks>
     public partial class BreadCrumb : SkinObjectBase
     {
         private const string MyFileName = "BreadCrumb.ascx";


DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -104,20 +103,30 @@ public string RootLevel
...
 
         /// <summary>
-        /// Gets or sets a value indicating whether opt-in: render list-based semantic markup (nav/ol/li).
-        /// Default is false to preserve legacy output and reduce breaking changes.
+        /// Gets or sets the version of the rendering to use.
         /// </summary>
-        public bool UseListMarkup { get; set; }
+        /// <remarks>
+        /// This allows opting-in into new more modern rendering behaviors.
+        /// v2 will render semantic markup (nav/ol/li) instead of the legacy span-based markup.
```
donker: Shouldn't we add the RenderVersion property to the base class (SkinObjectVersion) so that all skin objects can benefit from this in the future?

bdukes: That was my understanding as well

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -104,20 +103,30 @@ public string RootLevel
...
-        // Do not show when there is no breadcrumb (only has current tab)
+        /// <summary>
+        /// Gets or sets a value indicating whether the control is hidden when no breadcrumb is available and only the
+        /// current tab is present.
+        /// </summary>
         public bool HideWithNoBreadCrumb { get; set; }
 
-        /// <summary>Gets or sets a value indicating whether to take advantage of the enhanced markup (remove extra wrapping elements).</summary>
+        /// <summary>Gets or sets a value indicating whether to take advantage of the enhanced markup
+        /// (remove extra wrapping elements). No effect when RenderingVersion is 2 or higher.
```
bdukes:
```suggestion
        /// (remove extra wrapping elements). No effect when <see cref="RenderingVersion" /> is 2 or higher.
```

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
valadas: Done in https://github.com/dnnsoftware/Dnn.Platform/pull/6972/changes/32e447cdd1031c51a06a928c57f6d0e4b215e319

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
```diff
@@ -135,7 +144,7 @@ protected override void OnLoad(EventArgs e)
                 return;
             }
 
-            this.lblBreadCrumb.Text = this.UseListMarkup
+            this.lblBreadCrumb.Text = this.RenderingVersion >= 2
```
bdukes, line 142: I think date-based or version-based versioning will be easier to reason about, but I know others lean towards the simplicity of `1`, `2`, `3`. It's not a hill that I'm going to die on, but I don't see a major downside to it.

bdukes (approved):

DNN Platform/Website/admin/Skins/BreadCrumb.ascx.cs
valadas: If people would like an actual .Net Version object, I can give it a try. I think there is an implicit conversion from a version as a string, where "2" would become 2.0.0 and "2.1" would be come 2.1.0 and so on. However I don't know the performance implications and thought and int would be sufficient for this level of rare version bumps... I am not personally a fan of using dates though (just a personal taste)...

But I will also not die on that hill ;)

donker: KISS. I agree with the sentiment but I'm also afraid of anything extra on the critical path.

bdukes: FWIW I was not thinking of a `Version` object, just an `int` like `100202`

DNN Platform/Library/UI/Skins/SkinObjectBase.cs
```diff
@@ -29,5 +29,13 @@ public class SkinObjectBase : UserControl, ISkinControl
...
         /// <summary>Gets or sets the associated ModuleControl for this SkinControl.</summary>
         public IModuleControl ModuleControl { get; set; }
+
+        /// <summary>
+        /// Gets or sets the version of the rendering to use.
+        /// </summary>
+        /// <remarks>
+        /// This allows opting-in into new more modern rendering behaviors.
+        /// </remarks>
+        protected int RenderingVersion { get; set; }
```
uzmannazari: Thanks for adding `RenderingVersion` to `SkinObjectBase` and for bringing versioned rendering into core – this is going to make incremental improvements much easier.

One challenge I'm running into, especially now that we're looking to apply the same pattern to the Search SkinObject, is that `RenderingVersion` is currently declared as `protected`. That means we can't set it via markup, so it's not possible to opt‑in to the new rendering mode from a skin or theme.

Would it be possible to expose this property publicly or provide a public wrapper? For example:
`public int RenderingVersionPublic
{
    get => RenderingVersion;
    set => RenderingVersion = value;
}`

This would allow skins to specify` <dnn:SEARCH runat="server" RenderingVersionPublic="2" />` to opt in to v2 rendering, while leaving the default version unspecified to preserve legacy behavior. It also keeps the core versioning logic in one place without needing per‑SkinObject flags or breaking changes.

Happy to adjust my PR once this is in place – I think it will make applying RenderingVersion to Search and other SkinObjects much cleaner. Thanks again for all the work on this!

uzmannazari, line 39: Thanks for adding RenderingVersion to SkinObjectBase and for bringing versioned rendering into core – this is going to make incremental improvements much easier.

One challenge I'm running into, especially now that we're looking to apply the same pattern to the Search SkinObject, is that RenderingVersion is currently declared as protected. That means we can't set it via markup, so it's not possible to opt‑in to the new rendering mode from a skin or theme.

Would it be possible to expose this property publicly or provide a public wrapper? For example:
public int RenderingVersionPublic { get => RenderingVersion; set => RenderingVersion = value; }

This would allow skins to specify <dnn:SEARCH runat="server" RenderingVersionPublic="2" /> to opt in to v2 rendering, while leaving the default version unspecified to preserve legacy behavior. It also keeps the core versioning logic in one place without needing per‑SkinObject flags or breaking changes.

Happy to adjust my PR once this is in place – I think it will make applying RenderingVersion to Search and other SkinObjects much cleaner. Thanks again for all the work on this!

valadas: You are right! Thanks for spotting that, it does in fact need to be public. I created this PR to adjust it https://github.com/dnnsoftware/Dnn.Platform/pull/6977

PR #6974, opened 2026-02-05
Move Microsoft.SourceLink.GitHub to be a global package reference

bdukes:
## Summary
Now that we're using NuGet's central package management, we can specify packages that apply to every project within the solution using `GlobalPackageReference`. This simplifies these utility packages that apply at dev-time.

A next step would be to do the same thing with StyleCop.Analyzers, but we need to exclude DotNetNuke.Log4Net from getting that package, which is not straightforward.


valadas (approved):
Awesome!

PR #6977, opened 2026-02-06
Made RenderingVersion property public in SkinObjectBase

valadas:
Made RenderingVersion property public in SkinObjectBase

From feedback in https://github.com/dnnsoftware/Dnn.Platform/pull/6972#pullrequestreview-3760775458 and after testing, the property does need to be public so it can be set from an ascx file.


uzmannazari:
Thanks Mr @valadas

mitchelsellers (approved):
Thanks sir!

donker (approved):
I've tested this and you can set it in markup now.

PR #6982, opened 2026-02-07
Fix broken upgrades using install package

donker:
If you currently upload an install package to the upgrade module it will break your installation when you try to upgrade. This is because the exclude file list was containing a leading slash which meant nothing matched and all files got copied over including the web.config. This PR fixes that error.


mitchelsellers (approved):

david-poindexter:
Oh yikes - great catch @donker

PR #6987, opened 2026-02-11
Fix Portal-based CDV version for client resource management

### LINKED ISSUE #6946: [Bug]: CRM/CDV Cache Busting Stopped working after upgrade to DNN v10.02+

#### In brief

With the removal of enable composite and minify, the ability to control the CDV (CrmVersion) in page source output is no longer working. The UI (PersonaBar settings) is stuck in "site" mode. The actual page output is using "host" mode.
 
Workaround: manually incrementing the dbo.HostSettings `CrmVersion` does still work.

Important thing to test as part of a fix - right now, in the database, manually changing PortalSettings' `CrmUseApplicationSettings` to True or False makes no difference.
 
#### Long detailed version w screenclips

The Enable Composite Files was removed:

Image

Above shows DNN v9.13.10, below is what we see now in DNN v10.02.01:

Image

Notice the CRM Mode is also missing. It should not be, right?

**Here is the problem**: on two of my upgraded sites (started life on DNN v9.x and are now v10.02.01) where I've checked, the site is locked on delivering the ?cdv=826 while in PB Server settings above, you can see that its showing 400.

So, I switched into the database and in dbo.HostSettings, I changed `CrmVersion` from 826 to 827, recycled the app pool and now on page refresh, as expected I get `?cdv=827`. In dbo.PortalSettings you can see I changed `CrmVersion` to 400 and it matches what you see in PersonaBar, but is NOT what is output in the page source.

Image

Basically, what i need someone to confirm in DNN v10.02+ is that...

1. incrementing the SITE VERSION does nothing
2. DNN no longer respects True or False in the portal settings' `CrmUseApplicationSettings` and
3. always uses only the host settings' CrmVersion

Also, just confirmed this does _NOT affect sites deployed on 10.02.00_ (started life after enable/min features were removed), they properly use only the PostalSettings' `CrmVersion` and incrementing it works (number goes up and matches what you see in your page source). In the DNN v10.02.01 version, PostalSettings has

Image

False (to me) should mean that we are using Portal/Site settings, which is 1 as you can see above.

In HostSettings CrmVersion is 52

Image

Since I think 'UseApplication' should mean host, then False means use portal settings and in my page source output I should see ?cdv=1

However, its 52:

Image

### Steps to reproduce?

Covered above I think.

### Current Behavior

On upgraded sites your `?cdv=555` stays stuck on whatever the HostSettings' `CrmVersion` was prior to upgrade.

### Expected Behavior

You should be able to switch between using the Portal/Site and Application/Host version of the CrmVersion value in the PersonaBar.

### Relevant log output

```shell
n/a
```

### Anything else?

_No response_

### Affected Versions

10.2.0 and 10.2.1 (latest v10 release)

Chrome

### Code of Conduct

donker:
This PR aims to fix #6946 

The new Client Resource Framework that was introduced in 10.2.0 did not include code to allow portals to override the crm version (appended as ?cdv=123 to resource urls). This PR aims to fix that omission and improve the UI at the same time. It:

1. Adds a column in the UI so that it is now clear what is host-level and what is portal-level CDV management
2. Improves the explanatory text
3. Marks as deprecated a number of properties in clientResourceSettings in the Web.Client library
4. Harmonises the setting that determines if the portal should override the crm settings by (a) removing the double setting and (b) renaming the core setting to one that is in sync with the code.
5. Adds a new IClientResourcesSettings interface and implementation for DI so the resources controller can access these settings.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "10.0.102",
+    "version": "10.0.103",
     "allowPrerelease": false,
     "rollForward": "latestMinor"
   }


DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -19,26 +19,44 @@ public class ClientResourceController : IClientResourceController
...
         public ClientResourceController(IHostSettings hostSettings)
-            : this(hostSettings, null)
+            : this(hostSettings, null, null)
         {
         }
 
         /// <summary>Initializes a new instance of the <see cref="ClientResourceController"/> class.</summary>
         /// <param name="hostSettings">The host settings.</param>
         /// <param name="appStatus">The application status.</param>
+        [Obsolete("Deprecated in DotNetNuke 10.2.3. Use overload with IApplicationStatusInfo. Scheduled removal in v12.0.0.")]
```
bdukes:
```suggestion
        [Obsolete("Deprecated in DotNetNuke 10.2.3. Use overload with IClientResourceSettings. Scheduled removal in v12.0.0.")]
```

bdukes (approved):
Looks great, thanks!

jeremy-farrance:
Peter, the build worked and I am pretty sure I found and used the right artifacts. 

image

But within seconds of opening the site and starting the installer, I got this yellowness.

image

If you try and don't see this, would you please give me some steps/help in getting to the right artifacts for the build of this PR?

donker:
Ah ... the installer. Rrright. That one has no portal so this new code bombs when trying to determine whether to override the app settings. OK, fix incoming.

jeremy-farrance:
Okay, logged in as Host, things look like they work well. I was able to flip between Global override and not, the SITE version appeared, tested a bit, no issues.

The rewrite of the CRM paragraph is really great. It always bugs me when things are only written so the developer in the moment knows what it means. Your new version will hopefully make sense to anyone!
 
Two issues now, **first is minor**: The UI feels a little awkward. Would you consider stacking them in just one of the columns? I am sure the blank right column will get taken by something upcoming in v10.02.04 or higher. Or, just put only the big paragraph in the left column and stack the 3 items in the right column: Override toggle, Host Inc, Site Inc. Also, rather than have the Site Inc group disappear, I'd pretty strongly suggest just disabling it instead so its grey'd out but still visible. For new users that will feel more intuitive and self-explanatory-at-a-glance as to how the toggle will work (since the default is off).

**Second, this one is more serious**... to me. When Override Global CDV is True, I believe a Portal Administrator should be able to increment their SITE VERSION. 

_If you disagree with that, the rest can be ignored._
 
If it should be an option, here is what a portal admin sees if he goes to the same panel in the PB. Maybe a future contribution if it doesn't make sense right now?
image

donker:
Point taken about 1. You can tell I struggled to make things a bit more coherent.

Point 2 opens quite the box of Pandora. Currently none of the server settings are accessible by an admin. The entire controller is locked down. Opening up stuff selectively is possible, of course, but is tricky and can introduce security vulnerabilities. I.e. a high risk intervention. Which I hate.

One way we could address that is to move the portal version increment button to some other place. Both in the UI as well as backend code. But where should it go?

jeremy-farrance:
Agreed. Leave idea 2 for a future discussion or issue, if/when. I double-checked that it was not an option in v9.any

Thanks again for accidentally teaching me how to install a build of a PR! 🙂

donker:
I've pushed a change to the layout to address the first point. I'm still quite attached to the two column approach so I left that (if we ever need that space we'll cross that bridge when we get there IMHO). But now the right hand is always visible and is just disabled if the switch is off.

jeremy-farrance:
Works!! ✅

PR #6996, opened 2026-02-13
Don't include System.Memory in install or upgrade

### LINKED ISSUE #6993: [Bug]: Upgrade from 9.13.10 to 10.2.2 results in "Could not load file or assembly 'System.Memory' or one of its dependencies."

### What happened?

Attempting to upgrade 9.13.10 to 10.2.2 results in "Could not load file or assembly 'System.Memory' or one of its dependencies."  I am still testing things but it appears that an intermediate upgrade from 9.13.10 to 10.2.1 will allow a follow-up upgrade to 10.2.2 to work as expected.  The tested site is a basic site with no extensions.

### Steps to reproduce?

Standard upgrade method I use
1. Stop the existing site (9.13.10)
2. Recycle application pool
3. Copy over the upgrade files (10.2.2)
4. Start the site
5. Browse to site and error shows (attached)
[9.13.10 to 10.2.2 error.txt](https://github.com/user-attachments/files/25271578/9.13.10.to.10.2.2.error.txt)

### Current Behavior

Immediate error upon accessing the site.

### Expected Behavior

Site to show the standard DNN upgrade host login prompt.

### Relevant log output

```shell

```

### Anything else?

As a test, I also installed, from scratch, a new install of 10.2.2 and that worked as expected.  As mentioned previously, and intermediate step of 9.13.10 to 10.2.1 then again to 10.2.1 does appear to work.

### Affected Versions

10.2.0 (latest v10 release)

Firefox

### Code of Conduct

bdukes:
Fixes #6993

This PR has _not_ yet been tested, I'm not certain if the installer works without the assembly, but I believe it should work.

### DIFF
--- a/DNN Platform/Library/Common/Lists/ListInfoCollection.cs
+++ b/DNN Platform/Library/Common/Lists/ListInfoCollection.cs
@@ -8,6 +8,7 @@ namespace DotNetNuke.Common.Lists
     using System.Collections.Generic;
 
     using DotNetNuke.Collections;
+    using DotNetNuke.Common.Utilities;
     using DotNetNuke.Instrumentation;
 
     /// <summary>Represents a collection of <see cref="ListInfo"/>.</summary>
--- a/DNN Platform/Library/Common/Utilities/StringExtensions.cs
+++ b/DNN Platform/Library/Common/Utilities/StringExtensions.cs
@@ -3,13 +3,34 @@
 // See the LICENSE file in the project root for more information
 namespace DotNetNuke.Common.Utilities
 {
+    using System;
     using System.Text;
     using System.Text.RegularExpressions;
 
+    /// <summary>Extension methods for <see cref="string"/>.</summary>
     public static class StringExtensions
     {
         private static readonly Encoding Iso8859Encoding = Encoding.GetEncoding("iso-8859-8");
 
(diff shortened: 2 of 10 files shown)


donker:
Bummer. The install wizard needs it.

Screenshot 2026-02-13 at 21 16 57

donker:
I tried moving this to exclude only in upgrade but now I am getting a different dll issue when trying to upgrade a plain vanilla 9.13.10 to 10.2.3:

Screenshot 2026-02-13 at 22 02 04

Maybe we need to rethink this.

bdukes:
I've pushed a commit which avoids using [the `string.Contains` extension method from System.Memory](https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.contains), instead introducing our own extension method in `DotNetNuke.Common.Utilities.StringExtensions`. This allows the install wizard to load without an error when System.Memory is missing.

bdukes:
I've also confirmed now that this works as expected when upgrading from DNN 9.13.9

donker:
I have confirmed the following scenarios:

1. Install without wizard
2. Install with wizard
3. Upgrade from 10.1.2
4. Upgrade from 9.13.9

mitchelsellers (approved):

Issue #6999, opened 2026-02-14
Duplication of javascript files (with packages)


valadas:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

We currently have packages for many javascript libraries. Those js files get put in `Javascript Libraries/DnnPlugins/` with their package name and version.

When this was done, the original files in `Website/Resources/Shared/scripts` were left in place. To prevent confusion and duplication, we should try our best to warn developers to not directly reference those old files and instead use their named package for registration.

This can also cause fixes that are not applied in both copies as what happened before https://github.com/dnnsoftware/Dnn.Platform/pull/6927

The proposed solution is to show a browser console warning when the old files are used with instructions to not reference them directly and use DNN APIs to request those libraries instead.

### Steps to reproduce?

Here is a list of our current javascript packages.

- [ ] DnnPlugins
- [ ] HoverIntent
- [ ] jQuery
- [ ] jQuery.Fileupload
- [ ] jQueryMigrate
- [ ] jQueryUI
- [ ] Knockout
- [ ] KnockoutMapping
- [ ] Selectize

### Current Behavior

Developers may do fixes or consume those older duplicated files.

### Expected Behavior

Developers should be warned to not do so.

### Relevant log output

```shell

```

### Affected Versions

10.2.2 (latest v10 release)

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #7001, opened 2026-02-14
Added warning about duplicated dnn.jquery.js file

valadas:
Part of https://github.com/dnnsoftware/Dnn.Platform/issues/6999 effort

### DIFF
--- a/DNN Platform/Website/Resources/Shared/scripts/dnn.jquery.js
+++ b/DNN Platform/Website/Resources/Shared/scripts/dnn.jquery.js
@@ -1,7 +1,8 @@
 ﻿document.addEventListener("DOMContentLoaded", function () {
     console.warn(
         "Do not reference this file directly.\n" +
-        "Use DNN ClientResourcesController to register DnnPlugins instead.\n" +
+        "Use JavaScript.RequestRegistration(CommonJs.DnnPlugins) instead.\n" +
+        "Or other appropriate CommonJs registration methods as described in https://docs.dnncommunity.org/content/tutorials/client-resources/index.html#javascript-libraries.\n" +
         "This file is unmaintained, may include breaking changes and will be removed in DNN v12."
     );
 });


valadas:
@dnnsoftware/approvers let me know if this is how we want to handle it so that it gets done a similar way for the others.

donker (approved):

DNN Platform/Website/Install/UpgradeWizard.aspx
```diff
@@ -282,13 +281,12 @@
...
-        });
-
-        upgradeWizard.dnnProgressbar = $(".dnnProgressbar").dnnProgressbar();
-      });
+                $('.dnnFormMessage.dnnFormError').each(function () {
+                  if ($(this).html().length)
+                    $(this).css('display', 'block');
+                });
+              });
+            });
```
bdukes: Did you intend to add an extra `});`?

DNN Platform/Website/Resources/Shared/scripts/dnn.jquery.js
```diff
@@ -1,4 +1,12 @@
-﻿(function ($) {
+﻿document.addEventListener("DOMContentLoaded", function () {
+    console.warn(
+        "Do not reference this file directly.\n" +
+        "Use DNN ClientResourcesController to register DnnPlugins instead.\n" +
```
bdukes: It would be registered via the `JavaScript` class, not `ClientResourcesController`, right?

DNN Platform/Website/Install/UpgradeWizard.aspx
```diff
@@ -446,7 +443,7 @@
       $('.flag[value=' + $("#PageLocale")[0].value + ']').addClass("selectedFlag");
 
       //Progressbar and button initialization
-      $("#progressbar").progressbar({ value: 0 });
+      $("#progressbar")[0].value = 0;
```
bdukes: I would think we'd either want to use jQuery or not, so either 1️⃣ `$("#progressbar").val(0);` or 2️⃣ `document.getElementById('progressbar').value = 0;`

valadas: Agreed and I am all for removing jQuery dependencies where we can progressivelly, so fixed in https://github.com/dnnsoftware/Dnn.Platform/pull/7001/changes/57313d937d74fe89b65dc176e85462e1de68fd1d

DNN Platform/Website/Install/UpgradeWizard.aspx
valadas: You are right, indentation messed me up. I tried to harmonize it and it is a mess on the whole file which would make a crazy diff. So I just corrected that block in https://github.com/dnnsoftware/Dnn.Platform/pull/7001/changes/c060edcd61d052864fcb12462a5be36e75283b52

DNN Platform/Website/Resources/Shared/scripts/dnn.jquery.js
valadas: What do you think of https://github.com/dnnsoftware/Dnn.Platform/pull/7001/changes/f7f58c5493d10a99c9241378539b3c47634c9d3c

bdukes: looks great, thanks!

PR #7002, opened 2026-02-14
Fix CDF issues

donker:
This PR fixes these issues:

1. Multiple attributes being added while using different casing (mentioned in #6903 )
2. Remove double slashes from links to resources in skin path (mentioned in #6850)
3. Ensure no attribute can be added that has been added by the other resource methods

PR #7007, opened 2026-02-18
Remove StyleCop.Analyzers reference from DotNetNuke.WebControls

bdukes:
This was added globally in #7006


valadas (approved):

bdukes:
I forgot that #7002 included this commit and I merged it in already. So, this commit is already on `develop` now.

PR #6783, opened 2025-10-22
Bump npm dependencies

bdukes:
## Summary
This PR updates all of the npm dependencies that don't require upgrading React.

This PR also upgrades Yarn and sets the `minimumReleaseAge` configuration to 1440 (1 day), to limit installing very new packages. See https://pnpm.io/supply-chain-security for rationale.

PR #7008, opened 2026-02-18
Upgraded Storybook to v10

valadas:
Supersedes #6961 

- We had a mix of v9 and v10 packages causing a build issue, I first reverted the v10 packages to their latest v9 versions.
- Then I ran into an issue with svgr support within storybook, which I fixed.
- I then ran storybook upprade tool to bring us to latest v9.
- Then I reran the storybook upgrade tool to bring us to latest (v10)


bdukes:
You may need to bump the versions down or wait a bit and re-run. [We require](https://github.com/dnnsoftware/Dnn.Platform/pull/6783) npm packages to have been published for at least 24 hours before we upgrade for security reasons.

valadas:
> You may need to bump the versions down or wait a bit and re-run. [We require](https://github.com/dnnsoftware/Dnn.Platform/pull/6783) npm packages to have been published for at least 24 hours before we upgrade for security reasons.

Oh, I did not know that, I can re-run it tomorrow or such...

bdukes:
I can't tell what the issue is with this PR. The commit that I pushed fixes one issue, but [your first commit](04470fafe88dc73df98b82ed7725ae8125a93abc) now causes an error in the Admin Experience build script. I don't know if maybe it's causing some files not to get generated, so the resource zip doesn't get created? I failed to track down a real source cause.

valadas:
Yeah, we may need to investigate further. I ran storybook from its local place and did not do a full build of everything.

valadas:
@bdukes I rebased and resolved conflicts and it worked (on my machine). Crossing fingers for CI ;)

PR #7043, opened 2026-02-27
Bulk Install: Add filtering and paging to event log

bdukes:
This is part of the ongoing effort for https://github.com/dnnsoftware/Dnn.Platform/issues/5920.
Since this is a feature branch, I will self-approve this PR.

This PR adds the ability to filter the Bulk Install event log by Severity and Event Type, as well as paging links:
Screenshot of event log tab with filter dropdowns and paging

Theoretically, Bulk Install has four severity levels, but only two are used (Info and Warning). Also, since its custom API auth was replaced with API Tokens (see https://github.com/dnnsoftware/Dnn.Platform/pull/7033), there are not logs about authentication issues here (only IP safelist violations). This opens to question to me of whether Bulk Install should have its own Event Log at all, or if it should just use DNN's event log (and then this entire tab could be removed).

### DIFF
--- a/DNN Platform/Modules/BulkInstall/BulkInstall.Web/src/components/tabs/dnn-bi-logs/dnn-bi-log-pagination/dnn-bi-log-pagination.tsx
+++ b/DNN Platform/Modules/BulkInstall/BulkInstall.Web/src/components/tabs/dnn-bi-logs/dnn-bi-log-pagination/dnn-bi-log-pagination.tsx
@@ -1,0 +1,85 @@
+import { Component, Host, h, Prop, Event, EventEmitter, Fragment } from '@stencil/core';
+import { Pagination } from '../../../../clients/event-log-client';
+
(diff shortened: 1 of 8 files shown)


DNN Platform/Modules/BulkInstall/BulkInstall.Web/src/components/tabs/dnn-bi-logs/dnn-bi-log-pagination/dnn-bi-log-pagination.tsx
```diff
@@ -0,0 +1,85 @@
...
+  if (end !== lastIndex) {
+    if (end === lastIndex - 2) {
+      yield toPageItem(lastIndex - 1, current);
+    } else if (end !== lastIndex - 1) {
+      yield ellipsisItem;
+    }
+
+    yield toPageItem(lastIndex, current);
+  }
+}
```
valadas, line 40: Fancy!

PR #7044, opened 2026-02-28
Implement Serilog in DNN

donker:
This PR replaces our Log4net underpinning with Serilog. The advantages are:

1. Real-time changes in logging when editing the config file without a need for an app restart
2. Structured logging
3. The ability to add your own "sinks" (places the log is sent to)

## TODO

- [x] Add new dlls to distribution
- [x] Add writing out default config file to the root folder
- [ ] Add logic to repair a damaged/old config file
- [x] Repairing all deprecated code in DNN (big effort)


DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
```diff
@@ -94,7 +96,7 @@ public string ValidateToken(HttpRequestMessage request)
         {
             if (!JwtAuthMessageHandler.IsEnabled)
             {
-                Logger.Trace(this.SchemeType + " is not registered/enabled in web.config file");
+                Logger.LogTrace(this.SchemeType + " is not registered/enabled in web.config file");
```
mitchelsellers:
```suggestion
                Logger.LogTrace("{SchemeType} is not registered/enabled in web.config file", this.SchemeType);
```

DNN Platform/Dnn.AuthServices.Jwt/Components/Common/Controllers/JwtController.cs
```diff
@@ -456,7 +428,7 @@ private static void ValidateConfiguration()
             // Check if session token TTL exceeds renewal token TTL
             if (sessionTtl > renewalTtl)
             {
-                Logger.Warn(
+                Logger.LogWarning(
```
mitchelsellers, line 451: Similar to others it woudl be great to get this one to templated

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// Sets up Serilog using the config file ~/Serilog.config.
+    /// </summary>
+    /// <param name="hostMapPath">Path to the root of the DNN installation.</param>
+    internal static void AddSerilog(string hostMapPath)
+    {
+        var configFile = Path.Combine(hostMapPath, "Serilog.config");
+        var config = new LoggerConfiguration()
+            .WriteTo.File(Path.Combine(hostMapPath, "Portals\\_default\\Logs\\log.txt"), rollingInterval: RollingInterval.Day, formatProvider: CultureInfo.InvariantCulture)
+            .MinimumLevel.Debug();
```
mitchelsellers: Do we want the default to be Debug?  Or should this be Informational?

donker: Informational I'd say

DNN Platform/Dnn.AuthServices.Jwt/Auth/JwtAuthMessageHandler.cs
```diff
@@ -35,6 +36,7 @@ public JwtAuthMessageHandler(bool includeByDefault, bool forceSsl)
             // ServicesRoutingManager.RegisterAuthenticationHandlers()
             // this scheme gets marked as enabled.
             IsEnabled = true;
+            this.logger = DnnLoggingController.GetLogger<JwtAuthMessageHandler>();
```
valadas, line 39: Can we have IDnnLogingController in DI registered so that we can inject different controllers (let's say for unit tests).

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
valadas: The old one I believe defaulted to "Error", not saying it should but I think that was to avoid having huge files there out of the box.

donker: Done

DNN Platform/DotNetNuke.Instrumentation/LoggerSourceImpl.cs
```diff
@@ -94,7 +86,14 @@ public void Debug(object message)
 
             public void Debug(object message, Exception exception)
             {
-                this.Logger.Log(this.stackBoundary, levelDebug, message, exception);
+                if (message is string)
```
bdukes: Do we need to check is `message` is `null`?

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,51 @@
...
+/// <summary>
+/// Controller for Serilog functions.
+/// </summary>
+internal sealed class SerilogController
+{
+    /// <summary>
+    /// Sets up Serilog using the config file ~/Serilog.config.
+    /// </summary>
+    /// <param name="hostMapPath">Path to the root of the DNN installation.</param>
+    internal static void AddSerilog(string hostMapPath)
```
bdukes: I would call this `applicationMapPath`. `hostMapPath` makes me thing of the path to the `Portals/_default` folder. It also matches [`IApplicationStatusInfo.ApplicationMapPath`](https://github.com/dnnsoftware/Dnn.Platform/blob/11c7c27ed0ccd1baa32f57d7c025a3cfae9ed42f/DNN%20Platform/DotNetNuke.Abstractions/Application/IApplicationStatusInfo.cs#L21)

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,51 @@
...
+        if (File.Exists(configFile))
+        {
+            config = new LoggerConfiguration()
+                .ReadFrom.Configuration(new ConfigurationBuilder()
+                    .AddJsonFile(configFile, optional: false, reloadOnChange: true)
+                    .Build());
+        }
+
+        Log.Logger = config.CreateLogger();
+    }
```
bdukes: Should the default config be defined in an `else` block?
```suggestion
        var configFile = Path.Combine(hostMapPath, "Serilog.config");
        if (!File.Exists(configFile))
        {
            var defaultConfigFile = Path.Combine(hostMapPath, "Config", "Serilog.default.config");
            if (File.Exists(defaultConfigFile))
            {
                File.Copy(defaultConfigFile, configFile);
            }
        }

        LoggerConfiguration config;
        if (File.Exists(configFile))
        {
            config = new LoggerConfiguration()
                .ReadFrom.Configuration(new ConfigurationBuilder()
                    .AddJsonFile(configFile, optional: false, reloadOnChange: true)
                    .Build());
        }
        else
        {
            config = new LoggerConfiguration()
                .WriteTo.File(
                    Path.Combine(hostMapPath, "Portals\\_default\\Logs\\log.txt"),
                    outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] [{SourceContext}] {Message:lj}{NewLine}{Exception}",
                    rollingInterval: RollingInterval.Day,
                    formatProvider: CultureInfo.InvariantCulture)
                .MinimumLevel.Error();
        }

        Log.Logger = config.CreateLogger();
    }
```

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Sets up Serilog using the config file ~/Serilog.config.
+    /// </summary>
+    /// <param name="hostMapPath">Path to the root of the DNN installation.</param>
+    internal static void AddSerilog(string hostMapPath)
+    {
+        var configFile = Path.Combine(hostMapPath, "Serilog.config");
+        var config = new LoggerConfiguration()
+            .WriteTo.File(
+                Path.Combine(hostMapPath, "Portals\\_default\\Logs\\log.txt"),
```
bdukes: Using a `txt` file extension will allow unauthenticated users to download the log files. Let's stick to `.log.resources` like we're currently doing.

DNN Platform/Website/Config/Serilog.default.config
```diff
@@ -0,0 +1,22 @@
...
+    ],
+    "MinimumLevel": {
+      "Default": "Error"
+    },
+    "WriteTo": [
+      {
+        "Name": "File",
+        "Args": {
+          "outputTemplate": "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] [{SourceContext}] {Message:lj}{NewLine}{Exception}",
+          "path": "%BASEDIR%/Portals/_default/Logs/log.txt",
```
bdukes: Using a `txt` file extension will allow unauthenticated users to download the log files. Let's stick to `.log.resources` like we're currently doing.
```suggestion
          "path": "%BASEDIR%/Portals/_default/Logs/log.log.resources",
```

DNN Platform/DotNetNuke.Instrumentation/StartupExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+public static class StartupExtensions
+{
+    /// <summary>
+    /// Add Serilog to the project.
+    /// </summary>
+    /// <param name="services">The IServiceCollection.</param>
+    /// <param name="hostMapPath">The path to the root of the DotNetNuke website. This is needed to find the correct directory to write the log files to.</param>
+    public static void AddSerilog(this IServiceCollection services, string hostMapPath)
+    {
+        Environment.SetEnvironmentVariable("BASEDIR", hostMapPath);
```
bdukes: Using an environment variable here feels hacky to me. Can we call [`.SetBasePath()`](https://github.com/serilog/serilog-settings-configuration/blob/e0b845c708a95a8bc34772f473d28f37bd1b3754/sample/Sample/Program.cs#L14) when configuring instead?

bdukes (commented):
Thanks for all of the work on this one!

DNN Platform/DotNetNuke.Instrumentation/StartupExtensions.cs
donker: I have no idea whether that is possible. I'll give it a try.

donker: No. Now I remember. I've dug into this before and couldn't get this to work with loading a config from a file.

DNN Platform/Website/Config/Serilog.default.config
donker: Serilog forces a convention whereby it takes only the last extension and then sticks the day stamp in front of that. So it needs to be log.resources and the file today would be log20260303.resources.

DNN Platform/DotNetNuke.Instrumentation/LoggerSourceImpl.cs
donker: Done

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
donker: Done

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
donker: Done

bdukes:
I've [opened a PR to this branch](https://github.com/donker/Dnn.Platform/pull/130) which migrates all logging calls to use `LoggerMessage` generated methods.

DNN Platform/DotNetNuke.Instrumentation/StartupExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+public static class StartupExtensions
+{
+    /// <summary>
+    /// Add Serilog to the project.
+    /// </summary>
+    /// <param name="services">The IServiceCollection.</param>
+    /// <param name="applicationMapPath">The path to the root of the DotNetNuke website. This is needed to find the correct directory to write the log files to.</param>
+    public static void AddSerilog(this IServiceCollection services, string applicationMapPath)
+    {
+        Environment.SetEnvironmentVariable("BASEDIR", applicationMapPath);
```
dimarobert: `SerilogController.AddSerilog()` is used from 5 places throughout the code (3x DnnLoggingController, 1x LoggerSourceImpl, and this one), but the `BASEDIR` Environment Variable gets set only here. Could one of the other places ever get called first?

I think this should be part of `SerilogController.AddSerilog()` itself instead to be safe.

DNN Platform/DotNetNuke.Instrumentation/DnnLoggingController.cs
```diff
@@ -0,0 +1,86 @@
...
+        public static ILogger<T> GetLogger<T>()
+        {
+            if (Log.Logger == null)
+            {
+                // initialize Serilog
+                var applicationMapPath = System.Web.Hosting.HostingEnvironment.MapPath("~");
+                SerilogController.AddSerilog(applicationMapPath);
+            }
+
+            return new SerilogLoggerFactory(Log.Logger).CreateLogger<T>();
```
dimarobert, line 32: `SerilogLoggerFactory` should be a singleton. `SerilogLoggerProvider` should be a singleton.

`SerilogLoggerFactory` creates a new `SerilogLoggerProvider` in its constructor. The provider then creates its "root" logger `_logger = logger.ForContext(new[] { this });` (where `this` is the provider instance). That root logger is then used (in our case only one time as the factory and provider are never used after this call) to create the logger returned by calling the `SerilogLoggerFactory.CreateLogger<T>()` (through `SerilogLoggerFactory.CreateLogger(string categoryName)` -> `SerilogLoggerProvider.CreateLogger(string name)`).

This means that for each call to any of the `GetLogger` methods in this `DnnLoggingController` class a new set of `SerilogLoggerFactory`, `SerilogLoggerProvider`, and that "root" logger of the provider will get created and persisted in memory for the entire duration of the application lifetime. Currently there are more that 270 references to the DnnLoggingController.GetLogger overloads, which will create and persist 270 or more of those sets of instances (in addition to the `ILogger<T>` instances that gets returned to the caller to perform the logging function).

Note that an additional `SerilogLoggerProvider` instance is also created as singleton through DI by the `loggingBuilder.AddSerilog()` registration call in [`StartupExtensions.cs`](https://github.com/dnnsoftware/Dnn.Platform/pull/7044/changes#diff-bdfbd98c23822046fef60ce84a0a1015cd3fbbb0c39736a8b53ebb5e7526b85e) which gets shared by all loggers that get resolved/created through DI.

Also, in Serilog, the `SerilogLoggerProvider` is the logging scope holder. It has a CurrentScope that looks like this:
```cs
readonly AsyncLocal<SerilogLoggerScope?> _value = new();

internal SerilogLoggerScope? CurrentScope
{
    get => _value.Value;
    set => _value.Value = value;
}
```
(note that it is not kept as a static field, it is a per instance, AsyncLocal field)

that gets used to keep track of the calls to `logger.BeginScope()` that is used to enrich logs based on the execution context, which is important for good, detailed structured logging. Having multiple instances of `SerilogLoggerProvider` or each logger having its own `SerilogLoggerProvider` instance mean that they have split brain and scopes from one logger cannot be shared with the other loggers. Think along the lines of `ILogger<Installer>` and `ILogger<PackageInstaller>` cannot share state in something like this:
```cs
public class Installer {
  private ILogger logger = DnnLoggingController.GetLogger<Installer>();

  public PackageInstaller[] Packages { get; set; }

  private void InstallPackages() {
    foreach (var installer in Packages) {
      using (logger.BeginScope(state: new Dictionary<string, object> {
          ["ScopeName"] = "PackageInstall",
          ["PackageName"] = installer.Package.Name,
          ["PackageVersion"] = installer.Package.Version 
      }) {
          installer.Install();
      }
    }
  }
}

public class PackageInstaller {
  private ILogger logger = DnnLoggingController.GetLogger<PackageInstaller>();

  public void Install() {
    foreach (var componentInstaller in componentInstallers) {
      using (logger.BeginScope(state: new Dictionary<string, object> {
          ["ScopeName"] = "ComponentInstall",
          ["ComponentType"] = compInstaller.Type
      }) {
          logger.LogInformation("Installing component");
      }
    }
  }
}
```
in this setup, that call to `logger.LogInformation()` would gather all the values passed in the opened scopes along the callstack and should result in the following structured log (represented in JSON format):
```json
{
  "Message": "Installing component",
  "ScopeName": "ComponentInstall",
  "PackageName": "MyModule",
  "PackageVersion": "1.2.3.4",
  "ComponentType": "Assembly"
}
```
additionaly, the default implementation of Microsoft Logging is also composing an additional property `"Scopes": ["PackageInstall", "ComponentInstall"]`, so that you can understand your log code path. Serilog does not do that by default, but probably can be extended to with a different implementation of the provider or some other extension point.

With the "split brain" issue above, this nesting would not be properly tracked as each logger (with its own provider) would only track its scopes, but would be unable to track scopes from loggers of other components / classes.

_(and yes that `logger.BeginScope(Dictionary<string, object> state)` API from Microsoft is wild, they also have a [`logger.BeginScope(List<KeyValuePair<string, object>> state)`](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-10.0#log-scopes) which does not fare much better from an ease of use point of view, but they probably expect extension methods to be created by developers for each of their particular scopes and leveraging these lower APIs there. they also have `logger.BeginScope(string messageTemplate, params object[] args)` but that has its own problems, as any data you want to attach to the scope through the `args` MUST be a named token in the `messageTemplate`, otherwise it is discarded, which makes it cumbersome to construct an intelligible `messageTemplate`. first two [are supported by Serilog](https://github.com/serilog/serilog-extensions-logging/blob/v8.0.0/src/Serilog.Extensions.Logging/Extensions/Logging/SerilogLoggerScope.cs#L80-L104), probably the third one as well)_

DNN Platform/DotNetNuke.Instrumentation/DnnEnricher.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// Enriches the log event with DNN-specific properties including UserId and PortalId.
+    /// </summary>
+    /// <param name="logEvent">The log event to enrich.</param>
+    /// <param name="propertyFactory">The factory used to create log event properties.</param>
+    public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
+    {
+        var portalId = -1;
+        if (HttpContext.Current != null)
```
dimarobert, line 27: Another option to do this without having to use the `HttpContext.Current` is to leverage `logger.BeginScope()` instead.

For `PortalId`, looks like the `HttpContext.Current.Items["PortalSettings"]` is only ever set in `BasicUrlRewriter` or `AdvancedUrlRewriter`, so a call to `var scope = logger.BeginScope(new Dictionary<string, object>{ ["PortalId"] = portalSettings.PortalId })` would be made there, which will enrich all logging calls made during the (request) scope. `HttpContext.Current` is already used there to assign the `context.Items["PortalSettings"]`, so to correctly cleanup the scope at the right time (end of the request), the `scope` instance returned from the `logger.BeginScope()` would be added as a disposable to the HttpContext using `context.DisposeOnPipelineCompleted(scope)` after it is created. HttpContext would then take care of its disposal at the right time.

For the `UserId`, there are likely similar places where it gets assigned to the `HttpContext.Current.Items["UserInfo"]`, where an approach similar to the `PortalId` above could be used.

DNN Platform/DotNetNuke.Instrumentation/DnnLoggingController.cs
donker: Thank you for your comments, Dima, I'm going to revise the PR based on your feedback. Trying to shoehorn DI stuff into the older classes in DNN is something of a challenge. What puzzles me is how the ```SerilogLoggerProvider``` holds the scope values but it should be a Singleton, which would share it between requests. But I'll give it a try and see how it goes. Using the ```BeginScope``` is a much more elegant solution than the enricher. But I don't want any Serilog reference to bleed beyond the Instrumentation project.

dimarobert: The key for how the `SerilogLoggerProvider` holds the scope values even when it itself is a Singleton, is the `readonly AsyncLocal<SerilogLoggerScope?> _value = new();` field I described above. Similar to how `ThreadLocal<T>` values are specific to a thread, the value of that field (or any AsyncLocal object) is specific for that async execution and it is tracked by the [`ExecutionContext`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.executioncontext?view=netframework-4.8.1) across the entire execution of an async operation (even across multiple threads if the operation happen to jump threads when resuming).

As for the `BeginScope` API, that is part of the [Microsoft.Extensions.Logging.ILogger](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger.beginscope?view=net-10.0-pp&viewFallbackFrom=net-8.0) interface, so there should be no Serilog reference bleed beyond the Instrumentation project.

As for pulling services from DI in the older classes, there could be an alternative to create the singletons for the `SerilogLoggerFactory` and `SerilogLoggerProvider` outside of DI (in a static constructor of a class, maybe the `DnnLoggingController`) and then simply reimplement the [`SerilogLoggingBuilderExtensions.AddSerilog()`](https://github.com/serilog/serilog-extensions-logging/blob/v8.0.0/src/Serilog.Extensions.Logging/SerilogLoggingBuilderExtensions.cs#L35-L51) to provide them to the DI container as instances:
```cs
public static ILoggingBuilder AddDnnSerilog(this ILoggingBuilder builder, SerilogLoggerProvider provider)
{
    if (builder == null) throw new ArgumentNullException(nameof(builder));

    builder.AddProvider(provider);
    builder.AddFilter<SerilogLoggerProvider>(null, LogLevel.Trace);

    return builder;
}
```

then replace the registration in [StartupExtensions.cs](https://github.com/dnnsoftware/Dnn.Platform/pull/7044/changes#diff-bdfbd98c23822046fef60ce84a0a1015cd3fbbb0c39736a8b53ebb5e7526b85e):

```cs
services.AddLogging(loggingBuilder => loggingBuilder.AddDnnSerilog(DnnLoggingController.ProviderInstance));
```

or even omit the extension method and do it in-place in the `services.AddLogging()` lambda as there are only two lines of code `loggingBuilder => loggingBuilder.AddProvider(DnnLoggingController.ProviderInstance).AddFilter<SerilogLoggerProvider>(null, LogLevel.Trace)`.

DNN Platform/DotNetNuke.Instrumentation/DnnEnricher.cs
donker: I've substantially rewritten the approach. Can you recheck your comments?

DNN Platform/HttpModules/UrlRewrite/BasicUrlRewriter.cs
```diff
@@ -256,11 +256,16 @@ internal override void RewriteUrl(object sender, EventArgs e)
                 HttpContext.Current.Server.Transfer(strUrl);
             }
 
+            // Add PortalId to the log context for use in any logging that occurs during the request
+            // DnnLoggingController.AddToLogContext("PortalId", portalId);
```
dimarobert: Small cleanup here?

DNN Platform/HttpModules/UrlRewrite/BasicUrlRewriter.cs
```diff
@@ -256,11 +256,16 @@ internal override void RewriteUrl(object sender, EventArgs e)
...
+            // Add PortalId to the log context for use in any logging that occurs during the request
+            // DnnLoggingController.AddToLogContext("PortalId", portalId);
+            this.serviceProvider.GetRequiredService<LogRequestContext>().AddToLogContext("PortalId", portalId);
+
             if (portalId != -1)
             {
                 // load the PortalSettings into current context
                 var portalSettings = new PortalSettings(tabId, portalAliasInfo as PortalAliasInfo);
                 app.Context.Items.Add("PortalSettings", portalSettings);
+                DotNetNuke.Common.Globals.GetCurrentServiceProvider().GetRequiredService<LogRequestContext>()?.AddToLogContext("PortalId", portalSettings.PortalId);
```
dimarobert: Could `this.serviceProvider` be used as above?

DNN Platform/DotNetNuke.Instrumentation/StartupExtensions.cs
```diff
@@ -20,8 +18,8 @@ public static class StartupExtensions
     /// <param name="applicationMapPath">The path to the root of the DotNetNuke website. This is needed to find the correct directory to write the log files to.</param>
     public static void AddSerilog(this IServiceCollection services, string applicationMapPath)
     {
-        Environment.SetEnvironmentVariable("BASEDIR", applicationMapPath);
-        SerilogController.AddSerilog(applicationMapPath);
+        DnnLoggingController.InitializeLoggerFactory();
         services.AddLogging(loggingBuilder => loggingBuilder.AddSerilog(null, true));
```
dimarobert: Also, as noted in [my initial review](https://github.com/dnnsoftware/Dnn.Platform/pull/7044#:~:text=Note%20that%20an%20additional):
> Note that an additional SerilogLoggerProvider instance is also created as singleton through DI by the loggingBuilder.AddSerilog() registration call in [StartupExtensions.cs](https://github.com/dnnsoftware/Dnn.Platform/pull/7044/changes#diff-bdfbd98c23822046fef60ce84a0a1015cd3fbbb0c39736a8b53ebb5e7526b85e) which gets shared by all loggers that get resolved/created through DI.

Which means that now there are two SerilogLoggerProvider instances, one part of the factory in `DnnLoggingController` and one part of DI through the `loggingBuilder.AddSerilog()` call.

That is why I also [recommended manually registering Serilog](https://github.com/dnnsoftware/Dnn.Platform/pull/7044#:~:text=then%20simply%20reimplement) instead:
> As for pulling services from DI in the older classes, there could be an alternative to create the singletons for the SerilogLoggerFactory and SerilogLoggerProvider outside of DI (in a static constructor of a class, maybe the DnnLoggingController) and then simply reimplement the [SerilogLoggingBuilderExtensions.AddSerilog()](https://github.com/serilog/serilog-extensions-logging/blob/v8.0.0/src/Serilog.Extensions.Logging/SerilogLoggingBuilderExtensions.cs#L35-L51) to provide them to the DI container as instances:
> ```cs
> public static ILoggingBuilder AddDnnSerilog(this ILoggingBuilder builder, SerilogLoggerProvider provider)
> {
>     if (builder == null) throw new ArgumentNullException(nameof(builder));
> 
>     builder.AddProvider(provider);
>     builder.AddFilter<SerilogLoggerProvider>(null, LogLevel.Trace);
> 
>     return builder;
> }
> ```
> then replace the registration in [StartupExtensions.cs](https://github.com/dnnsoftware/Dnn.Platform/pull/7044/changes#diff-bdfbd98c23822046fef60ce84a0a1015cd3fbbb0c39736a8b53ebb5e7526b85e):
> ```cs
> services.AddLogging(loggingBuilder => loggingBuilder.AddDnnSerilog(DnnLoggingController.ProviderInstance));
> ```

dimarobert: Looking a bit closer to it, does not look like there is a way to create a SerilogLoggerFactory from a provider or to retrieve the inner provider that the SerilogLoggerFactory creates, but at the same time, I do not think that the SerilogLoggerFactory is even needed for this implementation. As this issue https://github.com/serilog/serilog-extensions-logging/issues/183 states, the factory was designed only for cases when the default Microsoft LoggerFactory needed to be replaced.
In our case I think that you can use the provider directly, [the SerilogLoggerFactory simply forwards the calls to the provider](https://github.com/serilog/serilog-extensions-logging/blob/dev/src/Serilog.Extensions.Logging/Extensions/Logging/SerilogLoggerFactory.cs) and there is no other logic from the SerilogLoggerFactory that we need.

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                File.Copy(defaultConfigFile, configFile);
+            }
+        }
+
+        LoggerConfiguration config;
+        if (File.Exists(configFile))
+        {
+            config = new LoggerConfiguration()
+                .Enrich.FromLogContext()
```
dimarobert: Another thing, I do not think that by default Serilog logs information like the current ThreadId, AppDomain, or HostName. Things that I think the logs in DNN had and were pretty useful. They should also be added to the default `outputTemplate` below for file logging.

dimarobert: Here is an enricher that I created for myself based on the [existing code in DNN](https://github.com/dnnsoftware/Dnn.Platform/blob/955c1ff200357427389bfe8f065ee36b10cb92e4/DNN%20Platform/DotNetNuke.Log4net/log4net/Util/SystemInfo.cs#L228-L286):
```cs
    class SystemInfoEnricher : Serilog.Core.ILogEventEnricher {
        private readonly string hostName;
        private readonly int appDomainId;

        private LogEventProperty hostNameProperty;
        private LogEventProperty appDomainProperty;

        public SystemInfoEnricher() {
            appDomainId = AppDomain.CurrentDomain.Id;
            try {
                hostName = System.Net.Dns.GetHostName();
            } catch {
                hostName = Environment.MachineName;
            }
        }

        public void Enrich(LogEvent logEvent, Serilog.Core.ILogEventPropertyFactory propertyFactory) {
            hostNameProperty = hostNameProperty ?? propertyFactory.CreateProperty("HostName", hostName);
            appDomainProperty = appDomainProperty ?? propertyFactory.CreateProperty("AppDomain", appDomainId);

            logEvent.AddPropertyIfAbsent(hostNameProperty);
            logEvent.AddPropertyIfAbsent(appDomainProperty);
            logEvent.AddPropertyIfAbsent(propertyFactory.CreateProperty("ThreadId", Thread.CurrentThread.ManagedThreadId));
        }
    }
```

DNN Platform/DotNetNuke.Instrumentation/StartupExtensions.cs
donker: Like this (new code uploaded)? I'm not sure I'm following what you're getting at as the code examples are not everything that would change.

dimarobert: I think something like this should work: https://github.com/donker/Dnn.Platform/compare/serilog-2026...dimarobert:Dnn.Platform:serilog-2026

I did not notice initially that the ILoggerFactory extension methods from Microsoft, `Microsoft.Extensions.Logging.LoggerFactoryExtensions.CreateLogger<T>(this ILoggerFactory factory)` and `Microsoft.Extensions.Logging.LoggerFactoryExtensions.CreateLogger(this ILoggerFactory factory, Type type)` were used by the `DnnLoggingController`, so an ILoggerFactory was needed to leverage them.

So, the implementation above first ties together the initialization of the `Log.Logger` and the singleton `Provider` and ensures that it happens only once by leveraging the static class initializer. It then creates a `SimpleLoggerFactory` that simply forwards the calls made to `ILoggerFactory.CreateLogger(string categoryName)` to the provider, to support the implementation of the `DnnLoggingController` methods. This forwarding was the only thing that we actually needed from the `SerilogLoggerFactory`.

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
donker: Do you prefer Enriching over Pushing properties? And if so, why?

dimarobert: For "HostName" and "AppDomain" they could be pushed to the LogContext at the start of the application (initialization of Serilog) as they would not change throughout the application lifetime.

For the "ThreadId", I do not see a good place where to push it. It is not related to a request, any code that executes (even background tasks), executes on a thread so it will have a ThreadId, but there would not be a single good place (or just a small amount of places) where to push it, it will probably be littered in a lot of places, becoming a burden for the logging. Also, in an async context, the ThreadId would change during the execution when the async continuations resume, making it even harder (maybe even impossible) to maintain accurate. I think the only viable place for it is an Enricher and if we already need one for it, makes sense to also put the other ones there to have them in a single place. There is no performance benefit of having them in the LogContext, as all items from the LogContext are added the same way to each logEvent through the `AddProperty()/AddPropertyIfAbsent()`. Additionally, in the enricher above, the `LogEventProperty` for HostName and AppDomain are cached the first time and not re-created for each log, similar to how would behave if pushed to the LogContext.

donker: I am not a big fan of the enricher (vs adding log properties) for the following reasons:

1. I prefer to see the "injection" of a context property happen in the same place where it is created (as now for userinfo and portalsettings). I first went the enricher route, but quickly realised that at the time that runs it would be unpredictable for me if these values were built or not.
2. It sets in stone what we do as a platform. The approach I've now chosen allows more flexibility to add new properties down the line.

I'm also concerned about performance. Having a ton of extra properties would weigh down the application as a whole.

In the current proposal I add a Guid for each request. This will allow us to trace individual requests. This also separates out threads I believe. Or: I don't know if thread IDs would add something meaningful. As to domain and host names: this is for quite specific use cases. Is this something we should burden everyone with? Why not create a custom enricher which is by default off but can be added in the config? Or add them as a log message so that the request ID is visible for it?

dimarobert: First things first, I only said about them because they existed in DNN previously in the Log4Net implementation and I thought this implementation would be feature compatible.

For me personally, the AppDomain came in handy while I was debugging some overlapping application startups during a batch update of extension packages. Gory details:
> If during the installation (which copies dlls and marks currently running AppDomains for recycle) there are requests made to the site, new AppDomains are started, which create a mess when they are also marked for recycle (as the installation continued to copy its dlls), allowing new AppDomains to be created, etc. All these AppDomains will go through initialization and will start running migrations in parallel, as there is no locking mechanism that would work across AppDomains.

As for the ThreadId, there is not always a 1-to-1 relation between a request and a thread. Most often this happens for async-await (which I think is also impossible to do the _"injection" of [the ThreadId] property happen in the same place where it is created_), but also for code that needs to spawn some background work in parallel. [Couple](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/TaskSchedulerController.cs#L156-L170) of [examples](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SecurityController.cs#L944-L946) [here](https://github.com/dnnsoftware/Dnn.Platform/blob/develop/DNN%20Platform/Library/Common/Initialize.cs#L268-L274). A LogContext push would have to always be made every time such a background task is spawned. _(I think a closer look at the PortalId and UserId is in order for these cases as well, as I think the AsyncLocal value that keeps track of the current LogContext will not flow to the spawned thread and their values will not be logged. AsyncLocal works only on the current thread and any async work that it `await`s for)_

Going back to the async-await case, here is an example of the ThreadId changing from one line to the other:
```cs
public async Task DoSomethingAsync() {
    Logger.Log(...); // Let's say ThreadId here would be `1`.

    var item = await DownloadSomethingAsync(); // If this is actually done async the current thread will be suspended and the code that follows will be scheduled as a continuation on another thread when the async work completes.

    Logger.Log(...); // ThreadId here would be different, let's say `2`. But could also be the same if the previous line completed synchronously (e.g. the value was already cached and simply returned).
}
```
I do not see how the ThreadId could be kept updated by injecting it _"in the same place where it is ~created~ [changed]"_ in a case like this.

However, I am not against it being an optional enricher that can be added in the config when needed.

---

As for your second point, if I understood it correctly,
> 2. It sets in stone what we do as a platform. The approach I've now chosen allows more flexibility to add new properties down the line.

Enrichers and LogContext are not mutually exclusive, both can be used at the same time for different purposes, so I do not see why adding an enricher would lose _flexibility to add new properties down the line_. In fact, Serilog implements the LogContext behavior through an enricher of its own that is added with `logConfig.Enrich.FromLogContext()`.

donker: OK, I like that idea. I've added it. And enrichers are now in the default config file. So if you don't want them you can just remove them.

donker: @dimarobert  What I do notice, however, is that when using the Enricher the code in the Enrich method runs for every log message. Whereas the adding of a property to the beginning of the request thread just runs once and then gets output with every log message. So in terms of performance it's prudent to carefully select what you put where. The enricher constructor runs just once per app cycle, by the way. So the app domain id and hostname are determined just once. Is this intended behavior?

dimarobert: Adding of a property to the LogContext, yes it happens once per request, but that only adds the property to a collection on the LogContext, not in the log events themselves. That happens in the `.Enrich.FromLogContext()` [enricher](https://github.com/serilog/serilog/blob/dev/src/Serilog/Context/LogContextEnricher.cs), which runs like any other enricher [for each](https://github.com/serilog/serilog/blob/dev/src/Serilog/Context/LogContext.cs#L201-L211) log event to copy the values from the LogContext to the event itself. So not really different from using any other enricher. Even the implementation of the LogContext itself is based on yet another collection of enrichers and [`LogContext.PushProperty()`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Context/LogContext.cs#L55-L58) is just adding a `PropertyEnricher` to it for every call. So, it is enrichers all the way down.

As for the "app domain id and hostname being determined just once", the AppDomainId cannot change during the lifetime of the application, while the HostName could, but I would not expect it to happen very often, maybe only in cases where for example the provisioning of the machine (VM) to run the application on is automated through a series of scripts and it happens to set the machine name after deploying and starting the IIS Application. I do not expect that to be a frequent case, but if it needs to be supported, the additional cost of querying the HostName and constructing the property would have to be incurred for each log event. From a logical / hierarchical point of view I would expect these properties to be from most stable to least in the following order: HostName (the machine that the application runs on), AppDomainId (the instance of the running app), RequestId, and ThreadId.

bdukes (dismissed):

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
```diff
@@ -0,0 +1,98 @@
...
+                .ReadFrom.Configuration(new ConfigurationBuilder()
+                    .AddJsonFile(configFile, optional: false, reloadOnChange: true)
+                    .Build());
+        }
+        else
+        {
+            config = new LoggerConfiguration()
+                .WriteTo.File(
+                    Path.Combine(applicationMapPath, "Portals\\_default\\Logs\\log.resources"),
+                    outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] [{SourceContext}] {Message:lj}{NewLine}{Exception}",
```
dimarobert: With this outputTemplate all of the extra (structured) log properties (added through enrichers, LogContext, or BeginScope) would not get into the log file. Only the properties that are part of the Message will be written (in string form as part of the message) to the file. To fix this, the [`{Properties:j}`](https://github.com/serilog/serilog/wiki/Formatting-Output#:~:text=All%20event%20property%20values) format token can be used. Note that an additional new line after it is needed, so the outputTemplate will have to have `{Properties:j}{NewLine}` appended to it.

(I see that the `Serilog.default.config` sets the same outputTemplate)

DNN Platform/Dnn.AuthServices.Jwt/Auth/JwtAuthMessageHandler.cs
bdukes: `ILogger<>` is added to the DI container, so that's the type you would start relying on.

DNN Platform/DotNetNuke.Instrumentation/SerilogController.cs
bdukes: Thanks! I've updated the output template (in both places) to add `{Properties:j}{NewLine}` (between the message and the exception)

valadas (dismissed):
I had time to give this some good testing.
- The deprecated way still works as before with a proper warning.
- The new way works good too

I have a lot of unit tests that mock and/or verify logging:
- The old way still worked in tests with the proper deprecated warning
- Migrating to the new way needed a bit of change in the testing patterns but using Microsoft's FakeLogger was a breeze and I could kill a lot of code and have it a cleaner way.

Love this overall.

My only nit picking item is if we could make the logged json in the text file indented to make it easier to read by humans. Not a big thing and it would make scroller more but I think it would make it more readable.

mitchelsellers (dismissed):
I too have tested this extensively, and love it

bdukes:
I've rebased this branch to fix the merge conflicts. It looks like GitHub is requiring another reviewer to approve (I can't count since I did the last push to the branch).

mtrutledge (approved):

valadas (approved):

PR #7047, opened 2026-03-01
fixed schdule drift (happy path)

### LINKED ISSUE #7045: [Bug]: Scheduler's reliance on "StartDate" to set "NextStart" guarantees time drift

### What happened?

"Messaging Dispatch" task runs every minute, and if you look at the history, each following start is shifted by a few seconds. 

This happens because the logic is relying on "StartDate" from the previous run to set the "NextStart", and since the actual start time is triggered by a 10-second timer, it will be shifted by 1-9 seconds on each run.

> **This does not really matter for a task that runs every minute, but I want my custom tasks to run every hour at hh:00:00. With 5 second drift on each run, the task will drift by over a minute per day, I think the issue is obvious.**

I'm working on PR that will rely on NextStart from previous run to set the NextStart of the future run. This way, there won't be any drift. Actual start time will still be a few second off, but that's not an issue.

### Steps to reproduce?

1. Install DNN, login as host
2. Go to "Settings > Scheduler"
3. Wait a few minutes to "Messaging Dispatch" task can run a few times
4. Click on "Task History" icon, and note task start time drift (see screenshot)

Image

### Current Behavior

Scheduler task StartDate drifts as it relies on StartDate to set NextStart

### Expected Behavior

Scheduler task StartDate should NOT drift. Should rely on  NextStart of current task to set NextStart for next task

### Relevant log output

```shell

```

### Anything else?

working on PR

### Affected Versions

Current development branch (unreleased)

Chrome

### Code of Conduct

r90727:
Fixes #7045

## Summary
This change fixes scheduler cadence drift on the happy path (successful completion and manual stop completion).

Previously, the next execution time could be calculated from `StartDate` in parts of the success/stop flow, which shifts future runs when a task starts late. This patch consistently advances `NextStart` from the previously scheduled `NextStart` value so recurring tasks stay anchored to their configured cadence (for example, hourly tasks remain on the hour).

Retry/error behavior is intentionally unchanged (`WorkErrored` still uses `StartDate`), so retry timing semantics remain the same.

If you want me to address Error/Retry logic, let me know.

image


r90727:
@mitchelsellers CatchUpEnabled flag logic was originally based on NextStart for s/m/h/d TimeLapseMeasurement, but later commits switched to StartDate logic.

I'm just trying to fix the happy path, the CatchUpEnabled branch can be reverted. lmk

mitchelsellers:
Yeah it looks really odd that it isn't all done the same way

r90727:
@mitchelsellers here's the commit, probably copy/paste error: https://github.com/dnnsoftware/Dnn.Platform/commit/6bc83a2786508fe0ef2de3f619214ce45b2c1ed0

Issue #7014, opened 2026-02-19
[Bug]: DDRMenu stopped work properly after Upgrade to DNN 9.02.02


ufoloko:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

Today I updated some sites to 10.02.02 and menus stopped work.

### Steps to reproduce?

1. Upgrade site with DDRMenu to 10.02.02
2. Menus stopped to show properly

### Current Behavior

After upgrade to DNN 10.02.02, Menus are rendering but without scripts or styles, it appears update is not loading js or css files listed in DDRMenu manifest menudef.xml

### Expected Behavior

Menu should load styles and scripts properly and be showed according to styles.

### Relevant log output

```shell

```

### Anything else?

Just like I said, it appears js and css resources are not being loaded from menudef.xml manifest

### Affected Versions

10.2.2 (latest v10 release)

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

valadas:
First are you talking about 9.2.2 or 10.2.2 (because you have 9.2.2 in the tile and 10.2.2 in the description).

Then can you share your menu template to see if something in it might be the root cause?

ufoloko:
Sorry, my bad, was a bad typo, im talking abount 10.02.02

[MainMenu.zip](https://github.com/user-attachments/files/25431397/MainMenu.zip)

Im attaching my menu template

valadas:
As a first look, this uses jquery and DNN no longer loads jquery by default. Do you see errors in the browser console related the javascript you have in that menu system? If you you may want to edit the theme to request jquery registration and maybe also jqueryUI since you use .slideToggle() if that is in jqueryUI (not sure).

At any rate I will move this into a discussion as it is not something actionable in the current state and more a help request than a bug at this stage.

PR #7056, opened 2026-03-03
Fix DDR Menu CDF issue

donker:
Fixes #7020 and #7014

### DIFF
--- a/DNN Platform/Modules/DDRMenu/TemplateEngine/TemplateDefinition.cs
+++ b/DNN Platform/Modules/DDRMenu/TemplateEngine/TemplateDefinition.cs
@@ -10,7 +10,6 @@ namespace DotNetNuke.Web.DDRMenu.TemplateEngine
     using System.IO;
     using System.Text;
     using System.Text.RegularExpressions;
-    using System.Web;
     using System.Web.Caching;
     using System.Web.UI;
     using System.Xml;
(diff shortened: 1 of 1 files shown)


DNN Platform/Modules/DDRMenu/TemplateEngine/TemplateDefinition.cs
```diff
@@ -324,12 +324,12 @@ internal void PreRender()
 
             foreach (var stylesheet in this.StyleSheets)
             {
-                this.clientResourceController.RegisterStylesheet(stylesheet);
+                GetClientResourcesController().RegisterStylesheet(stylesheet);
```
bdukes: Should we also remove the `IClientResourceController` parameter to the constructor?

donker: We could. But given I was trying to squeeze this into 10.2.3 and hoping to get 10.2.3 out the door I wanted to hold off on changing the constructor signature.

donker:
I think we need to prune out the yarn lock changes. I guess that got in because of working from develop.

bdukes:
I've rebased and removed those changes. I also went ahead and did the constructor deprecation, I don't think we want that hanging out long-term, and now's the best time to fix it.

PR #7058, opened 2026-03-03
Don't blow up when sending email without portal

### LINKED ISSUE #7052: [Bug]: Mailkit Provider fails to send mails from a scheduled JOB

### What happened?

I have code that sends emails from a scheduler client job. This Jobs, as far as I know do not have HTTP Context, so, no user or portal exists.

In the commit 93542f2 on the MailKitMailProvider.cs, on the ParseSmtpServer method, the line 143 expects a current context and this is throwing a Object Reference Null.
var currentPortalId = PortalSettings.Current.PortalId; 
https://github.com/dnnsoftware/Dnn.Platform/commit/93542f268e2fc9a52695fdd2a7e92aa90278e05b#diff-ad437d13d3c1a2d722f3b06c8f91d20188f99e10ff85515986b33673c43d2cc1 

### Steps to reproduce?

Configure the system to use MailKit
Create a scheduler job that sends emails.
When running this will breat on each SendEmail call

### Current Behavior

_No response_

### Expected Behavior

Could we assume the root portal (zero) is the one running the code when using a scheduller client?

var currentPortalId=0;
Try{
currentPortalId = PortalSettings.Current.PortalId;
}
catch{}

### Relevant log output

```shell
Message:There is a problem with the configuration of your SMTP Server. Mail was not sent. Object reference not set to an instance of an object. 

Message:Object reference not set to an instance of an object.

StackTrace:

   at DotNetNuke.Services.Mail.CoreMailProvider.CreateMailMessage(IMailSettings mailSettings, IHostSettings hostSettings, MailInfo mailInfo, SmtpInfo smtpInfo)
   at DotNetNuke.Services.Mail.CoreMailProvider.SendMail(MailInfo mailInfo, SmtpInfo smtpInfo)
   at DotNetNuke.Services.Mail.Mail.SendMail(String mailFrom, String mailSender, String mailTo, String cc, String bcc, String replyTo, MailPriority priority, String subject, MailFormat bodyFormat, Encoding bodyEncoding, String body, ICollection`1 attachments, String smtpServer, String smtpAuthentication, String smtpUsername, String smtpPassword, Boolean smtpEnableSSL)
```

### Anything else?

_No response_

### Affected Versions

10.2.2 (latest v10 release)

_No response_

### Code of Conduct

bdukes:
Fixes #7052

This PR causes the mail providers to fall back to a portal ID of `-1` if `PortalSettings.Current` is `null`. This prevents a `NullReferenceException`.

Ideally, there would be a way to specify a portal when sending an email from a scheduled task. We could add a property to `MailInfo` to support that scenario.

### DIFF
--- a/DNN Platform/Library/Services/Mail/CoreMailProvider.cs
+++ b/DNN Platform/Library/Services/Mail/CoreMailProvider.cs
@@ -46,6 +46,10 @@ namespace DotNetNuke.Services.Mail
             this.hostSettings = hostSettings ?? Globals.GetCurrentServiceProvider().GetRequiredService<IHostSettings>();
         }
 
+        private static int CurrentPortalId => PortalSettings.Current?.PortalId ?? Null.NullInteger;
+
+        private static string CurrentPortalName => PortalSettings.Current?.PortalName ?? string.Empty;
+
         /// <inheritdoc />
         public override string SendMail(MailInfo mailInfo, SmtpInfo smtpInfo = null)
         {
@@ -116,7 +120,7 @@ namespace DotNetNuke.Services.Mail
                 return null;
             }
 
-            if (!string.IsNullOrWhiteSpace(mailSettings.GetServer(PortalSettings.Current.PortalId)))
+            if (!string.IsNullOrWhiteSpace(mailSettings.GetServer(CurrentPortalId)))
             {
                 return null;
             }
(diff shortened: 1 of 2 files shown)


valadas:
@bdukes since this is a regression in 10.2.3, should we fix that branch?

bdukes:
Yup, I hadn't looked closely at where I'd introduced the issue. I've changed the base branch for the PR

PR #7064, opened 2026-03-04
Inline build task

bdukes:
## Summary
This PR removes `DotNetNuke.MSBuild.Tasks.dll` and its dependencies, replacing them with the one task that was used from it inline in the `Package.targets` file.


bdukes:
Agreed, long-term I think removing or replacing this compression makes sense.

donker (approved):
Thanks for the cleanup

PR #7071, opened 2026-03-06
Update Mimekit to latest version

donker:
This PR updates Mimekit to 4.15.1


donker:
This needs more work because a lot of dependencies are connected to this

PR #7074, opened 2026-03-06
Add CDV cache buster argument to Persona Bar loads

### LINKED ISSUE #7066: [Bug]: Severs tab styling appears broken in v10.02.03

### What happened?

We recently upgraded a DNN site from 10.2.2 to 10.2.3 that only has EasyDnnNews installed on it, and after upgrading the site through the new in-DNN interface, the server settings tab from the persona bar appear to have broken CSS.

Image

### Steps to reproduce?

1. Upgrade a site from DNN 10.2.3. using the in-DNN upgrade process.
2. Clear cache, restart the application and visit the settings tab.

This site was originally established as a 10.02.02 build. DNN versions table show a few versions but I think they were all installed as part of the initial setup, based on their timestamp.

Image

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

Affected Versions does not yet support the latest release of v10.2.3

### Affected Versions

10.2.2 (latest v10 release)

Chrome

### Code of Conduct

bdukes:
Fixes #7066

## Summary
In 10.2.3, some of the Persona Bar extensions were [migrated from webpack to Rsbuild](https://github.com/dnnsoftware/Dnn.Platform/pull/6837). One of the consequences of this change is that the CSS for those modules moved out of the JS bundle into a CSS file. Specifically, in 10.2.2, `/DesktopModules/Admin/Dnn.PersonaBar/Modules/Dnn.Servers/css/Servers.css` was empty, whereas in 10.2.3 it's 20kb (and `/DesktopModules/Admin/Dnn.PersonaBar/Modules/Dnn.Servers/scripts/bundles/servers-bundle.js` went from 592kb to 506kb).

Previously, the webpack runtime dynamically loaded that CSS, now it is loaded externally by DNN via its Persona Bar infrastructure. For each Persona Bar component it loads, DNN looks for a file sharing the name of the component in a `css` directory, so `css/Servers.css` gets loaded automatically in 10.2.3 and everything works as expected.

However, the way that the CSS file is requested causes it to be cached by the browser. So, in the upgrade scenario, the old empty `Servers.css` is served instead of the new version that has content. Because the Persona Bar is dynamically loaded in an `iframe`, even doing a hard refresh in the browser does not clear this cache (but using browser dev tools with caching disabled does). Therefore, this PR adds a cache buster parameter to those CSS requests.

### DIFF
--- a/Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/util.js
+++ b/Dnn.AdminExperience/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/util.js
@@ -1,29 +1,28 @@
 'use strict';
 define(['jquery', 'purify.min'], function ($, DOMPurify) {
-    var initializedModules = {};
+    const initializedModules = {};
     return {
         init: function (config) {
-            var loadTempl;
-            var setDialogClass;
-
-            loadTempl = function (folder, template, wrapper, params, self, cb) {
-                var callbackInit, moduleFolder, scriptFolder, templateSuffix, cssSuffix, initMethod, moduleJs, loadMethod;
-
+            const loadTempl = function (folder, template, wrapper, params, self, cb) {
                 if (!initializedModules[template]) {
-                    templateSuffix = '.html';
-                    cssSuffix = '.css';
-                    initMethod = 'init';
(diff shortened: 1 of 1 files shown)


donker (approved):
Thank you

PR #7075, opened 2026-03-06
Fix config scripts getting missed

### LINKED ISSUE #7069: [Bug]: Upload Package button missing in Settings > Servers > Upgrades

### What happened?

After having upgraded many sites from 9.10.13 to 10.02.01 using the "old way", the upgrades tab in Settings > Servers doesn't show the "upload packages" button.

### Steps to reproduce?

I just applied the upgrade package downloaded from github DNN releases. After the successful upgrade, none of my sites shows the mentioned "upload packages" button.

### Current Behavior

Missing "upload packages" button.

### Expected Behavior

The "upload pacckage" button should appear in the panel under the available packages table.

### Relevant log output

```shell

```

### Anything else?

After investigation, I found out it was due to the lack of the following line in web.config > AppSettings:     <add key="AllowDnnUpgradeUpload" value="true" />. I suggest to add this to the documentation (https://docs.dnncommunity.org/content/getting-started/setup/upgrades/post-10.2.0-steps.html) and I wonder if the upgrade from previous DNN versions just forgot to add the line to the web.confing during the upgrade.

### Affected Versions

10.2.3 (latest v10 release)

_No response_

### Code of Conduct

bdukes:
## Summary
There are config migration scripts for [10.2.0](https://github.com/dnnsoftware/Dnn.Platform/blob/b446f75a24f464eaf6a5a66af8a7caadbd14c9e8/DNN%20Platform/Website/Install/Config/10.02.00.config) and [10.2.1](https://github.com/dnnsoftware/Dnn.Platform/blob/b446f75a24f464eaf6a5a66af8a7caadbd14c9e8/DNN%20Platform/Website/Install/Config/10.02.01.config). However, there are [not `SqlDataProvider` scripts for those versions](https://github.com/dnnsoftware/Dnn.Platform/tree/b446f75a24f464eaf6a5a66af8a7caadbd14c9e8/DNN%20Platform/Website/Providers/DataProviders/SqlDataProvider). The [upgrade process](https://github.com/dnnsoftware/Dnn.Platform/blob/b446f75a24f464eaf6a5a66af8a7caadbd14c9e8/DNN%20Platform/Library/Services/Upgrade/Internals/Steps/InstallDatabaseStep.cs) uses the list of data provider scripts to determine which config files to look for. Therefore, when upgrading to a version that skips 10.2.0 or 10.2.1 (e.g. upgrading from 10.1.1 to 10.2.3), those config migration scripts do _not_ run.

This PR adds `SqlDataProvider` scripts for those versions, but also moves the config logic from 10.2.0 and 10.2.1 into a new `10.02.04.config`. In addition to managing [the `AllowDnnUpgradUpload` setting](https://github.com/dnnsoftware/Dnn.Platform/issues/7069), this includes the migration to reset the `requestLengthDiskThreshold` value, so this will need another call out in the release notes for any sites that had intentionally adjusted that value.

Fixes #7069

### DIFF
--- a/DNN Platform/Website/DotNetNuke.Website.csproj
+++ b/DNN Platform/Website/DotNetNuke.Website.csproj
@@ -767,6 +767,7 @@
     <Content Include="Providers\DataProviders\SqlDataProvider\10.02.00.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\10.02.01.SqlDataProvider" />
     <Content Include="Providers\DataProviders\SqlDataProvider\10.02.02.SqlDataProvider" />
+    <Content Include="Providers\DataProviders\SqlDataProvider\10.02.04.SqlDataProvider" />
     <Content Include="Resources\Shared\components\DatePicker\moment.min.js" />
     <Content Include="Resources\Shared\components\DatePicker\pikaday.css" />
     <Content Include="Resources\Shared\components\DatePicker\pikaday.jquery.js" />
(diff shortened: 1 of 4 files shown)


donker:
At some point we're going to have these silly requirements for one thing to run have another thing ...

PR #7091, opened 2026-03-08
Prevented not runing PR task to show as a failure

valadas:
Running the CreateGitHubPullRequest build task and then returning false for ShouldRun was considered an error in azure. This PR fixes that and also removes some stuff that was used only to troubleshoot earlier.


valadas:

PR #7096, opened 2026-03-08
Made package.json files to use only Major.Minor.Patch versioning

valadas:
Made package.json files to use only Major.Minor.Patch versioning. This should also reduce the PR noise to update github per builds.

### DIFF
--- a/Build/Tasks/SetPackageVersions.cs
+++ b/Build/Tasks/SetPackageVersions.cs
@@ -32,15 +32,15 @@ namespace DotNetNuke.Build.Tasks
             // Set all package.json in Admin Experience to the current version and to consume the current (local) version of dnn-react-common.
             foreach (var file in packages)
             {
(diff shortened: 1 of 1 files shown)


bdukes:
@valadas is [the CI failure from the latest push to `develop`](https://dev.azure.com/dotnet/DNN/_build/results?buildId=124307&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=72c3483f-759b-5fe1-97d0-7523409e045c) because it thought there was a change, but it was only line endings, so it ended up being empty?

valadas:
Kinda, I have an uncommitted fix here but on the road most of the day, I'll push that tonight

valadas:
@bdukes this should fix it but I can't wait for the build as I have to leave https://github.com/dnnsoftware/Dnn.Platform/pull/7099

PR #7098, opened 2026-03-09
Fix the addition of assemblies and sql scripts to module build

donker:
The recent changes to the module packaging script meant that some files were not being packaged. This PR aims to fix that.


bdukes (changes requested):
This added a `dll` to the root of the package, it didn't put it in a `bin` folder in the zip

bdukes:
I've pushed a fix which accounts for the path coming from `sourceFileName`, in addition to `path`. Thanks for tracking down the source of the issue. I can't find where the inline task is any different than the compiled task, but somehow that `RecursiveDir` was being correctly populated before.

bdukes:
One more fix coming, it's picking up files marked with `action="UnRegister"`, which obviously shouldn't be included in the package. (And, again, I feel like I'm missing something, since this wasn't an issue before).

PR #7102, opened 2026-03-11
Create assembly binding redirects at build time for web.config

donker:
This PR changes the approach to creating the web.config in the DNN install package. Currently we need to carefully manage the binding redirects to ensure the platform will install correctly. We do this by hand. But the amount of assemblies we distribute is likely to keep increasing in the future and we need to automate the process of ensuring the web.config is correct for an install distribution.

### DIFF
--- a/Build/Tasks/CreateWebConfig.cs
+++ b/Build/Tasks/CreateWebConfig.cs
@@ -7,6 +7,8 @@ namespace DotNetNuke.Build.Tasks
     using System.Linq;
     using System.Xml.Linq;
 
+    using Cake.Common.Diagnostics;
+    using Cake.Common.IO;
     using Cake.Frosting;
     using Cake.Json;
     using Dnn.CakeUtils;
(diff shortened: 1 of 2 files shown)


Build/Tasks/CreateWebConfig.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var configFile = context.WebsiteFolder + "release.config";
+            var doc = System.Xml.Linq.XDocument.Load(configFile);
+
+            context.PackagingPatterns = context.DeserializeJsonFromFile<PackagingPatterns>("./Build/Tasks/packaging.json");
+            var files = context.GetFilesByPatterns(context.WebsiteFolder, BinFolderInclude, context.PackagingPatterns.InstallExclude);
+            var redirects = files.ParseAssemblies()
+                .Where(x => x.PublicKeyToken != "null" && !string.IsNullOrEmpty(x.PublicKeyToken))
+                .Where(x => !x.Name.StartsWith("DotNetNuke.", StringComparison.OrdinalIgnoreCase))
+                .Where(x => !x.Name.StartsWith("Dnn.", StringComparison.OrdinalIgnoreCase))
```
bdukes: None of the DNN assemblies should have a `PublicKeyToken` value (and if they do, we would need a binding redirect for them).
```suggestion
```

Build/Tasks/CreateWebConfig.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>A cake task to copy the <c>release.config</c> to the <c>web.config</c>.</summary>
+    public sealed class CreateWebConfig : FrostingTask<Context>
+    {
+        private static readonly string[] BinFolderInclude = ["bin/**/*.dll",];
+
+        /// <inheritdoc />
+        public override void Run(Context context)
+        {
+            var configFile = context.WebsiteFolder + "release.config";
+            var doc = System.Xml.Linq.XDocument.Load(configFile);
```
bdukes:
```suggestion
            var doc = XDocument.Load(configFile);
```

Build/Tasks/CreateWebConfig.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// <summary>A cake task to copy the <c>release.config</c> to the <c>web.config</c>.</summary>
+    public sealed class CreateWebConfig : FrostingTask<Context>
+    {
+        private static readonly string[] BinFolderInclude = ["bin/**/*.dll",];
+
+        /// <inheritdoc />
+        public override void Run(Context context)
+        {
+            var configFile = context.WebsiteFolder + "release.config";
```
bdukes: Let's use Cake's built-in file path handling
```suggestion
            var configFile = context.WebsiteDir + context.File("release.config");
```

Build/Tasks/CreateWebConfig.cs
```diff
@@ -0,0 +1,74 @@
...
+                throw new InvalidOperationException("Could not find configuration/runtime/assemblyBinding in release.config.");
+            }
+
+            var assemblyBindingNamespace = assemblyBinding.Name.Namespace;
+            foreach (var redirect in redirects)
+            {
+                assemblyBinding.Add(WithNamespace(redirect, assemblyBindingNamespace));
+            }
+
+            var targetFile = context.WebsiteFolder + "web.config";
```
bdukes:
```suggestion
            var targetFile = context.WebsiteDir + context.File("web.config");
```

PR #7108, opened 2026-03-13
Add CI build via GitHub Actions

bdukes:
## Summary
This PR enables running CI builds via GitHub Actions.

I had it [run against that branch](https://github.com/dnnsoftware/Dnn.Platform/actions/runs/23058636180) until it worked, and then updated it to only run for `develop`, `main`, `release/*`, and pull requests (as well as manual dispatch).

### Differences from Azure DevOps
- I don't run the `BuildServerSetVersion` Cake task, it doesn't seem to have a corollary in GitHub Actions
- I don't start the SQL Local DB service (since we don't have an integration testing environment setup anyway)

I wasn't able to test the new `CreateGitHubPullRequest` Cake target. At the very least, we don't yet have the required secrets defined in GitHub. I assume there's an easier way to do that in GitHub Actions, but maybe it's easier to keep it working this way until we turn off Azure DevOps. /cc @valadas 

## Reasoning
We can add on steps or workflows which do trusted publishing of NuGet packages. In general, there should be better integration of annotations from CI build errors.


mitchelsellers:
This looks awesome, and I LOVE the build time!

PR #7117, opened 2026-03-16
Add GitHub Actions job to deploy to NuGet

bdukes:
## Summary
This PR adds a second job to the GitHub Actions workflow for CI. It only runs when the build is manually triggered, and it uses an [environment](https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments) to gate its running until approved. It's setup to use [NuGet Trusted Publishing](https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing), which still needs to be setup on the NuGet side (currently configured to use `MitchelSellers` as the NuGet user via an environment secret).

### Alternatives Considered
Instead of tacking this on to the end of each CI request, it could be a separate workflow that builds everything again (maybe triggered on creating a release), or a manual workflow that requires specifying an artifact ID.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,10 +6,6 @@ env:
   RELEASE_MODE: "Beta"
   RUN_TESTS: "true"
 
-permissions:
-  checks: write
-  pull-requests: write
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


valadas:
@bdukes I did setup Nuget for trusted publishing, we can discuss in todays meeting...

valadas (approved):
Looks good, is this ready to merge @bdukes or you needed to fix something regarding versioning?

bdukes:
It's ready, the version issue is only on the `feature/bulk-install` branch.

PR #7118, opened 2026-03-17
Update badge in README to use GitHub Actions CI

bdukes:


valadas:

PR #7119, opened 2026-03-17
Fix GitVersion not able to read history from repo

bdukes:
## Summary
The default git checkout does not include history, so GitVersion was not able to determine the correct version. This PR includes all history when performing `git checkout` so that it works as expected.


valadas:

PR #7120, opened 2026-03-17
Update env variables from `workflow_dispatch` inputs

bdukes:
GitHub Actions allows [defining inputs when manually triggering a build](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs). However, these values are available only via [the `inputs` context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#inputs-context), which does not exist when the workflow runs due to another trigger.

Specifically, this meant that our current workflow does not actually use the manual input, only the defaults, since it just looks in the `env` context, instead of the `inputs` context.

Therefore, this PR adds a new step to [update the `env` context](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable) with either the `inputs` value or the default.

Based on https://github.com/orgs/community/discussions/26322#discussioncomment-11553893

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,13 @@ jobs:
       run:
         shell: "pwsh"
     steps:
+      - name: "Setup Environment Variables"
+        run: |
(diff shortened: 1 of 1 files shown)


valadas:

PR #7121, opened 2026-03-17
Pin actions in workflows using commit SHA

bdukes:
## Summary
When pulling in actions in GitHub Actions workflows, use the full commit SHA to identify the action version, instead of a tag name which can be moved if the account is compromised.


bdukes:
Once this is setup, we can configure GitHub Actions to require a commit SHA:
image

PR #7124, opened 2026-03-17
Fix CSP tests not running

bdukes:
## Summary
The new ContentSecurityPolicy project has tests, but they aren't named in a way that the build automatically picks them up.


bdukes:
I've included this in #7126

PR #7129, opened 2026-03-18
Add GitHub Actions job to attest artifacts

bdukes:
See
https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations


valadas:

PR #7130, opened 2026-03-18
Fix tests with inconsistent values

bdukes:
Avoid notifications about tests being added and removed


valadas:

PR #7132, opened 2026-03-18
Limit permissions available to GitHub Actions workflows

bdukes:
## Summary
This PR implements the suggestions from the [Token Permissions](https://github.com/ossf/scorecard/blob/c22063e786c11f9dd714d777a687ff7c4599b600/docs/checks.md#token-permissions) check of the OSSF scorecard, by setting the top-level permissions of the workflows to `content: "read"` and only specifying higher permissions at the job level.

It also sets up a token to use by the OSSF scorecard action to read branch protection policies.


valadas:

PR #7133, opened 2026-03-18
Bump Dependencies

bdukes:
## Summary
This PR implements all minor version npm dependency bumps, as well as a handful of other dependency changes (e.g. using newer babel packages).


donker (approved):

donker:
As an added bonus: our build process now works on the latest versions of node

PR #7192, opened 2026-04-07
FIX: Site crawler regex improvement #7191

### LINKED ISSUE #7191: [Bug]: Site Search Crawler never completes and uses excessive CPU resources

### What happened?

Search crawler runs for hours / days and never completes.
Uses excessive CPU.

### Steps to reproduce?

Duplicated on site with content containing invalid HTML.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

On my site, I had previously used Google search integration and didn’t use the Lucene indexing until we recently updated the forums module to use DNN search/Lucene rather than SQL fulltext.
 
Despite my previous thought that this issue was related to user indexing, it isn’t. Rather, it is content that contains markup of some kind that is incorrect/incomplete, and during indexing, there is a regex that is run to remove tags from the content but leaves the attributes. Unfortunately, the regex isn’t executed with a timeout and doesn’t work correctly if it’s not valid HTML content.
 
This is where things get stuck in the call stack
 

 
Image
 
In that method, this regex never returns if the content is not valid HTML
 
Image

In my case, someone had posted some hacked-up wsdl (gack!) into a forum post; in other words, partially looks like HTML but it’s perfectly valid “text” content 
 
Image

 
This causes the regex to never complete due to catastrophic backtracking. 
When I look at it in regex buddy
 
 
Image

What should be done to fix it.
 
 
Image

 
1.	Add a timeout when declaring the regexes, 
2.	also use the cached regex API in DNN; 
3.	add a try/catch in the method to handle any timeout exceptions.
4.	Fix the regex (adding atomic groups – this will actually fix the regex itself, and it won’t go into an endless loop)
 
From
        private const string HtmlTagsWithAttrs = "<[a-z_:][\\w:.-]*(\\s+(?<attr>\\w+\\s*?=\\s*?[\"'].*?[\"']))+\\s*/?>";
to
        private const string HtmlTagsWithAttrs = “<[a-z_:][\\w:.-]*(?>(?:\\s+(?<attr>[\\w+\\s*?=\\s*?[\"'].*?[\"']))*)?\\s*/](file://w+/s*%3f=/s*%3f%5b/%22'%5d.*%3f%5b/%22'%5d))*)%3f/s*/)?>”;
 
this causes the regex to ignore invalid html
 
Image

 
But correctly matches valid HTML (the point of the regex is to remove tags but keep the attributes)
 
 
Image

 

### Affected Versions

10.3.0 (latest release)

_No response_

### Code of Conduct

johnhenley:
Fixes #7191

## Summary

Bug fix and performance improvement for HTML tag processing in the search controller.

This PR enhances the reliability and efficiency of HTML tag stripping and attribute extraction in InternalSearchControllerImpl by improving regex usage and error handling.
- InternalSearchControllerImpl.cs: Switched to RegexUtils.GetCachedRegex for regex instantiation to enable caching; added timeout parameters
- InternalSearchControllerImpl.cs: Updated the HTML tag regex for better accuracy and performance
- InternalSearchControllerImpl.cs: Wrapped all regex operations in StripTagsRetainAttributes with try-catch blocks to handle RegexMatchTimeoutException and log errors, preventing application hangs.

### DIFF
--- a/DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
+++ b/DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
@@ -41,7 +41,7 @@ namespace DotNetNuke.Services.Search.Internals
         private const string SearchableModuleDefsCacheKey = "SearchableModuleDefs";
         private const string LocalizedResxFile = "~/DesktopModules/Admin/SearchResults/App_LocalResources/SearchableModules.resx";
 
-        private const string HtmlTagsWithAttrs = "<[a-z_:][\\w:.-]*(\\s+(?<attr>\\w+\\s*?=\\s*?[\"'].*?[\"']))+\\s*/?>";
+        private const string HtmlTagsWithAttrs = "<[a-z_:][\\w:.-]*(?>(?:\\s+(?<attr>\\w+\\s*?=\\s*?[\"'].*?[\"']))*)?\\s*/?>";
 
         private const string AttrText = "[\"'](?<text>.*?)[\"']";
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(InternalSearchControllerImpl));
@@ -49,10 +49,10 @@ namespace DotNetNuke.Services.Search.Internals
         private static readonly string[] HtmlAttributesToRetain = { "alt", "title" };
         private static readonly DataProvider DataProvider = DataProvider.Instance();
 
-        private static readonly Regex StripOpeningTagsRegex = new Regex(@"<\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex StripClosingTagsRegex = new Regex(@"</\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex HtmlTagsRegex = new Regex(HtmlTagsWithAttrs, RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex AttrTextRegex = new Regex(AttrText, RegexOptions.Compiled);
+        private static readonly Regex StripOpeningTagsRegex = RegexUtils.GetCachedRegex(@"<\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex StripClosingTagsRegex = RegexUtils.GetCachedRegex(@"</\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex HtmlTagsRegex = RegexUtils.GetCachedRegex(HtmlTagsWithAttrs, RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex AttrTextRegex = RegexUtils.GetCachedRegex(AttrText, RegexOptions.Compiled, 2);
 
         private readonly IHostSettings hostSettings;
         private readonly IServiceProvider serviceProvider;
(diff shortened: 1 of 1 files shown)


mitchelsellers (dismissed):
I think t his is a good fix for right nwo with the lower level risk. 

@dnnsoftware/approvers I know there was a larger discussion about using HtmlSanitizer for this, but not sure that we want to go to the larger changge here at this time?

Thoughts?

DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
```diff
@@ -298,43 +298,94 @@ private static string StripTagsRetainAttributes(string html, IEnumerable<string>
...
+                                        var text = AttrTextRegex.Match(val).Groups["text"].Value.Trim();
+                                        if (text.Length > 0 && !list.Contains(text))
+                                        {
+                                            list.Add(text);
+                                        }
+                                    }
+                                    catch (RegexMatchTimeoutException ex)
+                                    {
+                                        DotNetNuke.Services.Exceptions.Exceptions.LogException(ex);
+                                        return strippedString;
```
valadas: I see a couple of places where if the regex expires, it still returns the string. Could this be a potential security issue from DOS to SXSS ?

valadas (dismissed):

valadas:
As for HtmlSanitizer, I think it would be best since we already have it and we don't have to think about all the edge cases that might arise by doing our own regexes IMO... But I won't die on that hill ;)

DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
johnhenley: @valadas the content is already in the portal at this point and should have been sanitized prior to being indexed by the crawler. And I would say that a never-completing regex that makes the system unresponsive is a worse DOS :) if you want to return an empty string, that's fine by me. I just wanted to fix what was broken.

DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
```diff
@@ -41,18 +41,18 @@ internal class InternalSearchControllerImpl : IInternalSearchController
...
         private static readonly DataProvider DataProvider = DataProvider.Instance();
 
-        private static readonly Regex StripOpeningTagsRegex = new Regex(@"<\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex StripClosingTagsRegex = new Regex(@"</\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex HtmlTagsRegex = new Regex(HtmlTagsWithAttrs, RegexOptions.IgnoreCase | RegexOptions.Compiled);
-        private static readonly Regex AttrTextRegex = new Regex(AttrText, RegexOptions.Compiled);
+        private static readonly Regex StripOpeningTagsRegex = RegexUtils.GetCachedRegex(@"<\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex StripClosingTagsRegex = RegexUtils.GetCachedRegex(@"</\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex HtmlTagsRegex = RegexUtils.GetCachedRegex(HtmlTagsWithAttrs, RegexOptions.IgnoreCase | RegexOptions.Compiled, 2);
+        private static readonly Regex AttrTextRegex = RegexUtils.GetCachedRegex(AttrText, RegexOptions.Compiled, 2);
```
bdukes: 2 seconds is the default timeout, let's just use the default unless there's a specific reason to use 2 explicitly.
```suggestion
        private static readonly Regex StripOpeningTagsRegex = RegexUtils.GetCachedRegex(@"<\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
        private static readonly Regex StripClosingTagsRegex = RegexUtils.GetCachedRegex(@"</\w*\s*>", RegexOptions.IgnoreCase | RegexOptions.Compiled);
        private static readonly Regex HtmlTagsRegex = RegexUtils.GetCachedRegex(HtmlTagsWithAttrs, RegexOptions.IgnoreCase | RegexOptions.Compiled);
        private static readonly Regex AttrTextRegex = RegexUtils.GetCachedRegex(AttrText, RegexOptions.Compiled);
```

DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
```diff
@@ -298,43 +298,94 @@ private static string StripTagsRetainAttributes(string html, IEnumerable<string>
...
+                                    catch (RegexMatchTimeoutException ex)
+                                    {
+                                        DotNetNuke.Services.Exceptions.Exceptions.LogException(ex);
+                                        return strippedString;
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        DotNetNuke.Services.Exceptions.Exceptions.LogException(ex);
+                                        throw;
+                                    }
```
bdukes: I assume somewhere else is going to log this exception after it's thrown, we probably don't need an extra log here.
```suggestion
```

DNN Platform/Library/Services/Search/Internals/InternalSearchControllerImpl.cs
bdukes: A security issue could some if the display of search results assumes that they've been sanitized, [which appears to be the case](https://github.com/dnnsoftware/Dnn.Platform/blob/7895e964c1e390d97e83f6deb01f645772626de7/DNN%20Platform/Website/DesktopModules/Admin/SearchResults/dnn.searchResult.js#L124-L135). So returning `string.Empty` is probably the right choice, at least until we adjust how we're displaying results.

bdukes (commented):
Thanks!

johnhenley:
I made the requested changes.

bdukes (approved):

mitchelsellers (approved):
Thanks for the updates!

PR #7193, opened 2026-04-08
Async support for MVC controllers.

### LINKED ISSUE #4647: Add async support to MVC controllers

dimarobert:
Closes #4647

## Summary
Creates a secondary variant of the MvcHostControl that is executed asynchronously during the page execution.
To avoid any breaking changes to the execution of the existing MVC controls, the control has to opt-in / be marked as requiring asynchronous execution. To do that the control needs to be declared with a ControlSrc in the form of `{ControllerNamespace}/{ControllerName}/async/{ActionName}.mvc` instead of the `{ControllerNamespace}/{ControllerName}/{ActionName}.mvc` that existing MVC module controls are defined as.

Apart from the action itself, the retrieval of the ModuleActions (`Get{0}Actions` method) supports `Task<ModuleActionCollection>` for asynchronous execution as well.

A few of the methods had to be made `protected` and as such code had to be re-ordered to comply with StyleCop rules. As a result, it may initially appear that more code than expected was modified.

Other small fixes:
+ Fix selection of ControllerName from the segments of the ControlSrc when it has a length greater than 2 when assigning the value for `moduleControl.LocalResourceFile`.
+ MVC: Improve error messages when view template is not found by writing out the SearchedLocations (something that I think ASP.NET does by default).

There is one TODO left in ModuleInstanceContext.cs, and as it states I am not sure if that should return or throw, Throwing, with a good message explaining the issue, would probably help more than the NullReferenceException that would end up being thrown by existing code that would most likely directly iterate over the list received from the method call.

This is also a place where a breaking change could be expected in 3rd party skins or other extensions that try to synchronously read the `ModuleInstanceContext.Actions` (list of admin ModuleActions) for asynchronous controls.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionResults/DnnPartialViewResult.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionResults/DnnPartialViewResult.cs
@@ -31,7 +31,7 @@ namespace DotNetNuke.Web.Mvc.Framework.ActionResults
 
             if (this.View == null)
             {
-                result = this.ViewEngineCollection.FindPartialView(context, this.ViewName);
+                result = this.FindView(context);
                 this.View = result.View;
             }
 
--- a/DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
@@ -5,7 +5,6 @@
 namespace DotNetNuke.Web.Mvc.Framework.Controllers
 {
     using System.Diagnostics.CodeAnalysis;
-    using System.Threading.Tasks;
     using System.Web.Mvc;
     using System.Web.UI;
 
@@ -24,10 +23,6 @@ namespace DotNetNuke.Web.Mvc.Framework.Controllers
         string LocalResourceFile { get; set; }
 
         ModuleActionCollection ModuleActions { get; set; }
-
-        Task<ModuleActionCollection> ModuleActionsAsync { get; set; }
-
-        bool IsAsync { get; set; }
 
         ModuleInstanceContext ModuleContext { get; set; }
 
--- a/DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
@@ -5,6 +5,7 @@
 namespace DotNetNuke.Web.Mvc.Framework.Controllers
 {
     using System.Diagnostics.CodeAnalysis;
+    using System.Threading.Tasks;
     using System.Web.Mvc;
     using System.Web.UI;
 
@@ -23,6 +24,10 @@ namespace DotNetNuke.Web.Mvc.Framework.Controllers
         string LocalResourceFile { get; set; }
 
         ModuleActionCollection ModuleActions { get; set; }
+
+        Task<ModuleActionCollection> ModuleActionsAsync { get; set; }
+
+        bool IsAsync { get; set; }
 
         ModuleInstanceContext ModuleContext { get; set; }
 
--- a/DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules/IModuleExecutionEngine.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules/IModuleExecutionEngine.cs
@@ -5,10 +5,14 @@
 namespace DotNetNuke.Web.Mvc.Framework.Modules
 {
     using System.IO;
+    using System.Threading;
+    using System.Threading.Tasks;
 
     public interface IModuleExecutionEngine
     {
         ModuleRequestResult ExecuteModule(ModuleRequestContext moduleRequestContext);
+
+        Task<ModuleRequestResult> ExecuteModuleAsync(ModuleRequestContext moduleRequestContext, CancellationToken cancellationToken);
 
         void ExecuteModuleResult(ModuleRequestResult moduleResult, TextWriter writer);
     }
--- a/DNN Platform/DotNetNuke.Web.Mvc/MvcModuleControlFactory.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/MvcModuleControlFactory.cs
@@ -23,12 +23,22 @@ namespace DotNetNuke.Web.Mvc
         /// <inheritdoc />
         public override Control CreateControl(TemplateControl containerControl, string controlKey, string controlSrc)
         {
+            if (IsAsyncControl(controlSrc))
+            {
+                return new AsyncMvcHostControl(controlKey);
+            }
+
             return new MvcHostControl(controlKey);
         }
 
         /// <inheritdoc />
         public override Control CreateModuleControl(TemplateControl containerControl, ModuleInfo moduleConfiguration)
         {
+            if (IsAsyncControl(moduleConfiguration.ModuleControl.ControlSrc))
+            {
+                return new AsyncMvcHostControl();
+            }
+
             return new MvcHostControl();
         }
 
@@ -37,9 +47,9 @@ namespace DotNetNuke.Web.Mvc
         {
             ModuleControlBase moduleControl = base.CreateModuleControl(moduleConfiguration);
 
-            var segments = moduleConfiguration.ModuleControl.ControlSrc.Replace(".mvc", string.Empty).Split('/');
+            var segments = moduleConfiguration.ModuleControl.ControlSrc.Split('/');
 
-            moduleControl.LocalResourceFile = $"~/DesktopModules/MVC/{moduleConfiguration.DesktopModule.FolderName}/{Localization.LocalResourceDirectory}/{segments[0]}.resx";
+            moduleControl.LocalResourceFile = $"~/DesktopModules/MVC/{moduleConfiguration.DesktopModule.FolderName}/{Localization.LocalResourceDirectory}/{(segments.Length == 2 ? segments[0] : segments[1])}.resx";
 
             return moduleControl;
         }
@@ -47,7 +57,18 @@ namespace DotNetNuke.Web.Mvc
         /// <inheritdoc />
         public override Control CreateSettingsControl(TemplateControl containerControl, ModuleInfo moduleConfiguration, string controlSrc)
         {
+            if (IsAsyncControl(controlSrc))
+            {
+                return new AsyncMvcSettingsControl();
+            }
+
             return new MvcSettingsControl();
+        }
+
+        private static bool IsAsyncControl(string controlSrc)
+        {
+            var segments = controlSrc.Split('/');
+            return segments.Length == 4 && segments[2].Equals("async", System.StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/DNN Platform/Library/UI/Containers/ActionBase.cs
+++ b/DNN Platform/Library/UI/Containers/ActionBase.cs
@@ -111,6 +111,24 @@ namespace DotNetNuke.UI.Containers
         /// <param name="actionID">The id of the action.</param>
         protected void ProcessAction(string actionID)
         {
+            if (this.ModuleControl is IAsyncModuleControl)
+            {
+                this.Page.RegisterAsyncTask(new PageAsyncTask(ct => this.ProcessActionInternalAsync(actionID, ct)));
+            }
+            else
+            {
+                this.ProcessActionInternal(actionID);
+            }
+        }
+
+        protected Task ProcessActionInternalAsync(string actionID, CancellationToken cancellationToken)
+        {
+            this.ProcessActionInternal(actionID);
+            return Task.CompletedTask;
+        }
+
+        protected void ProcessActionInternal(string actionID)
+        {
             if (int.TryParse(actionID, out var output))
             {
                 ModuleAction action = this.Actions.GetActionByID(output);
@@ -128,16 +146,26 @@ namespace DotNetNuke.UI.Containers
         /// <param name="e">The event arguments.</param>
         protected override void OnLoad(EventArgs e)
         {
-            if (this.ModuleControl != null)
+            if (this.ModuleControl is IAsyncModuleControl)
             {
                 // We need to defer accesing this.Actions as it could only be accesible after the WebForms async point.
                 this.Page.RegisterAsyncTask(new PageAsyncTask(this.LoadActionsAsync));
+            }
+            else
+            {
+                this.LoadActions();
             }
 
             base.OnLoad(e);
         }
 
         private Task LoadActionsAsync(CancellationToken cancellationToken)
+        {
+            this.LoadActions();
+            return Task.CompletedTask;
+        }
+
+        private void LoadActions()
         {
             try
             {
@@ -147,8 +175,6 @@ namespace DotNetNuke.UI.Containers
             {
                 Exceptions.ProcessModuleLoadException(this, exc);
             }
-
-            return Task.CompletedTask;
         }
     }
 }
--- a/DNN Platform/Library/UI/Containers/ActionBase.cs
+++ b/DNN Platform/Library/UI/Containers/ActionBase.cs
@@ -113,6 +113,7 @@ namespace DotNetNuke.UI.Containers
         {
             if (this.ModuleControl is IAsyncModuleControl)
             {
+                // We need to defer accesing this.Actions as it could only be accesible after the WebForms async point.
                 this.Page.RegisterAsyncTask(new PageAsyncTask(ct => this.ProcessActionInternalAsync(actionID, ct)));
             }
             else
--- a/DNN Platform/Library/UI/Containers/ActionBase.cs
+++ b/DNN Platform/Library/UI/Containers/ActionBase.cs
@@ -5,6 +5,8 @@ namespace DotNetNuke.UI.Containers
 {
     using System;
     using System.Diagnostics.CodeAnalysis;
+    using System.Threading;
+    using System.Threading.Tasks;
     using System.Web.UI;
 
     using DotNetNuke.Abstractions.Logging;
@@ -126,13 +128,19 @@ namespace DotNetNuke.UI.Containers
         /// <param name="e">The event arguments.</param>
         protected override void OnLoad(EventArgs e)
         {
+            if (this.ModuleControl != null)
+            {
+                // We need to defer accesing this.Actions as it could only be accesible after the WebForms async point.
+                this.Page.RegisterAsyncTask(new PageAsyncTask(this.LoadActionsAsync));
+            }
+
+            base.OnLoad(e);
+        }
+
+        private Task LoadActionsAsync(CancellationToken cancellationToken)
+        {
             try
             {
-                if (this.ModuleControl == null)
-                {
-                    return;
-                }
-
                 this.ActionRoot.Actions.AddRange(this.Actions);
             }
             catch (Exception exc)
@@ -140,7 +148,7 @@ namespace DotNetNuke.UI.Containers
                 Exceptions.ProcessModuleLoadException(this, exc);
             }
 
-            base.OnLoad(e);
+            return Task.CompletedTask;
         }
     }
 }
--- a/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
+++ b/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
@@ -511,13 +511,10 @@ namespace DotNetNuke.UI.Modules
             if (actionable != null)
             {
                 // Async module controls populate ModuleActions only after their async task executes.
-                // Leave this.actions as null so the Actions getter retries once results are available.
                 if (this.moduleControl is IAsyncModuleControl && actionable.ModuleActions == null)
                 {
-                    this.actions = null;
-
-                    // TODO: Should we throw instead to be able to identify code that tries to call it too early?
-                    return;
+                    throw new InvalidOperationException("Too early to access the ModuleActions. For async controls, ModuleActions collection is available only after the framework executes the `Page.ExecuteRegisteredAsyncTasks()`. " +
+                        "More specifically, you have to either register an async task using `Page.RegisterAsyncTask()` or use any of the sync events starting from PreRenderComplete to access them.");
                 }
 
                 this.moduleSpecificActions = new ModuleAction(this.GetNextActionID(), Localization.GetString("ModuleSpecificActions.Action", Localization.GlobalResourceFile), string.Empty, string.Empty, string.Empty);
--- a/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
+++ b/DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
@@ -510,6 +510,16 @@ namespace DotNetNuke.UI.Modules
             var actionable = this.moduleControl as IActionable;
             if (actionable != null)
             {
+                // Async module controls populate ModuleActions only after their async task executes.
+                // Leave this.actions as null so the Actions getter retries once results are available.
+                if (this.moduleControl is IAsyncModuleControl && actionable.ModuleActions == null)
+                {
+                    this.actions = null;
+
+                    // TODO: Should we throw instead to be able to identify code that tries to call it too early?
+                    return;
+                }
+
                 this.moduleSpecificActions = new ModuleAction(this.GetNextActionID(), Localization.GetString("ModuleSpecificActions.Action", Localization.GlobalResourceFile), string.Empty, string.Empty, string.Empty);
 
                 ModuleActionCollection moduleActions = actionable.ModuleActions;
--- a/DNN Platform/DotNetNuke.Web.Mvc/AsyncMvcHostControl.cs
+++ b/DNN Platform/DotNetNuke.Web.Mvc/AsyncMvcHostControl.cs
@@ -1,0 +1,85 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information
+namespace DotNetNuke.Web.Mvc
+{
+    using System;
+    using System.Globalization;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using System.Web;
+    using System.Web.UI;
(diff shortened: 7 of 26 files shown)


DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
```diff
@@ -24,6 +25,10 @@ public interface IDnnController : IController
 
         ModuleActionCollection ModuleActions { get; set; }
 
+        Task<ModuleActionCollection> ModuleActionsAsync { get; set; }
+
+        bool IsAsync { get; set; }
+
```
bdukes, line 31: @dnnsoftware/approvers do we have concerns about 3rd parties implementing this interface and being broken by this change?

DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules/IModuleExecutionEngine.cs
```diff
@@ -5,11 +5,15 @@
...
 {
     using System.IO;
+    using System.Threading;
+    using System.Threading.Tasks;
 
     public interface IModuleExecutionEngine
     {
         ModuleRequestResult ExecuteModule(ModuleRequestContext moduleRequestContext);
 
+        Task<ModuleRequestResult> ExecuteModuleAsync(ModuleRequestContext moduleRequestContext, CancellationToken cancellationToken);
```
bdukes, line 15: Same question about modifying interface

DNN Platform/DotNetNuke.Web.Mvc/MvcModuleControlFactory.cs
```diff
@@ -37,17 +47,28 @@ public override ModuleControlBase CreateModuleControl(ModuleInfo moduleConfigura
         {
             ModuleControlBase moduleControl = base.CreateModuleControl(moduleConfiguration);
 
-            var segments = moduleConfiguration.ModuleControl.ControlSrc.Replace(".mvc", string.Empty).Split('/');
+            var segments = moduleConfiguration.ModuleControl.ControlSrc.Split('/');
```
bdukes, line 50: Do we not need to trim `.mvc` from `ControlSrc`?

DNN Platform/Library/UI/Containers/ActionBase.cs
```diff
@@ -108,6 +110,24 @@ protected virtual void OnAction(ActionEventArgs e)
...
+            {
+                this.ProcessActionInternal(actionID);
+            }
+        }
+
+        protected Task ProcessActionInternalAsync(string actionID, CancellationToken cancellationToken)
+        {
+            this.ProcessActionInternal(actionID);
+            return Task.CompletedTask;
+        }
```
bdukes, line 129: Why do we need this `Task` wrapper that's not doing any async behavior?

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
```diff
@@ -510,6 +510,16 @@ private void LoadActions(HttpRequest request)
...
             {
+                // Async module controls populate ModuleActions only after their async task executes.
+                // Leave this.actions as null so the Actions getter retries once results are available.
+                if (this.moduleControl is IAsyncModuleControl && actionable.ModuleActions == null)
+                {
+                    this.actions = null;
+
+                    // TODO: Should we throw instead to be able to identify code that tries to call it too early?
+                    return;
+                }
```
bdukes, line 518: It makes sense to me that we would throw.

What are the scenarios in which we need to populate actions asynchronously? Is this change necessary to allow async MVC views?

bdukes:
Thanks @dimarobert for your efforts and digging into all of the ins-and-outs of the processing going on here!

DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
mitchelsellers: Maybe, as this will result in a compilation error for any upstream user of IDnnController correct since they don't implement the interface members

DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionResults/DnnPartialViewResult.cs
```diff
@@ -31,7 +31,7 @@ public void ExecuteResult(ControllerContext context, TextWriter writer)
 
             if (this.View == null)
             {
-                result = this.ViewEngineCollection.FindPartialView(context, this.ViewName);
+                result = this.FindView(context);
```
mitchelsellers, line 34: I believe this should be calling FindPartialView with the ViewName passed rather than the FindView

mitchelsellers:
@dimarobert thank you so much for this, a few things to review, but great amounts of research.

@donker and @sachatrauwaen I'm curious to see what impacts if any this has on the MVC Pipeline work that is going on as well

DNN Platform/DotNetNuke.Web.Mvc/MvcModuleControlFactory.cs
dimarobert: no, as it has no bearing on the number of segments when splitting by `'/'` and only the controllerName was ever used from the resulting `segments`, never the actionName that had the `.mvc` suffix, so it was just an unnecessary additional cleanup operation on the value.

reminder, formats (of interest here) for the ControlSrc are:
- `{ControllerName}/{ActionName}.mvc` - segments[0] needed here
- `{Namespace}/{ControllerName}/{ActionName}.mvc` - segments[1] needed here
- and now the new one, `{Namespace}/{ControllerName}/async/{ActionName}.mvc` - segments[1] needed here as well

DNN Platform/Library/UI/Containers/ActionBase.cs
dimarobert: The code in `this.ProcessActionInternal(actionID);` needs to access `this.Actions` (ModuleActions) that in an async execution are only available after the async point, so the code needs to run as part of a `this.Page.RegisterAsyncTask(new PageAsyncTask(ct => this.ProcessActionInternalAsync(actionID, ct)));` to delay that execution. The `PageAsyncTask` class has only two (convenient to use) constructors that both expect a `Task` return type: `PageAsyncTask(Func<Task>)` and `PageAsyncTask(Func<CancellationToken, Task>)`.

bdukes: Okay, can we at least add a comment so it's clear why it's necessary (I saw it's in one place). Alternatively, adding `await Task.Yield()` might be a way for the code to clarify that it needs to wait.

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
dimarobert: (see correction below)
yes, IIRC in the sync execution, the ModuleActions where populated (the `Get{0}Actions` method was called) after running the controller action itself, so to keep the same convention in the async execution I had to do the same and populate the ModuleActions list after the call to the async controller action which means that they are available only after the async point.

Also, this allows the `Get{0}Actions` method for an async controller action to be async as well, so for a `public Task<ActionResult> Index() {...}` you can have a `public Task<ModuleActionCollection> GetIndexActions() {...}`. (we probably should support the `public Task<ModuleActionCollection> GetIndexActions(CancellationToken) {...}` signature as well, but I do not expect a lot of non-CPU bound work to be needed by the computation for the list of ModuleActions that would benefit from cancellation)

And the reason for reseting `this.actions` back to `null`, as explained in the comment in code, is that if a call to `LoadActions` (the method that the code here is part of) is made too early in the pipeline (before the async execution point) `this.actions` would remain an empty list (assigned at the start of `LoadActions`) and would not be properly re-evaluated once it becomes available. but again, if we would throw here, probably makes sense to remove the reset to `null` as well.

DNN Platform/DotNetNuke.Web.Mvc/Framework/ActionResults/DnnPartialViewResult.cs
dimarobert: This is the code for that `this.FindView()` which is part of the base class `PartialViewResult` from ASP.NET. It correctly calls the exact same method as before, but in addition it also throws an exception with the `SearchedLocations` so that it is easier to debug when a view is not found due to a path issue. Otherwise a NRE or a generic "Could not find view {ViewName}" was thrown IIRC.
image

DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
dimarobert: A new `IAsyncDnnController` can be declared that has these properties and `DnnController` would implement both. Then a cast check would have to be made when they are accessed in the code, with an eventual fallback (probably throw?) if the controller does not have it but wants to run through the async code path.

One difficult situation with this might be the `DnnMvcHandler` route (rendering an MVC view through a direct request to it to do a partial render of the page), as there is no easy way there to figure out if the sync or async code path needs to be taken as that is not relying on a ModuleControl being present for the view to render, so that the ControlSrc format could tell us what to do, it simply matches on the ModuleInfo, controllerName, and actionName pair from the request path / query. So to figure out if `DnnMvcHandler` needs to end up calling `ModuleApplication.ExecuteRequest()` or `ModuleApplication.ExecuteRequestAsync()`, it would need to get a hold of the controller type first based on the pair above (which the inner ModuleApplication.ExecuteRequest{Async} already does) and then use reflection on the controller action to see if it returns `Task`. So we would be running the same logic twice.

DNN Platform/DotNetNuke.Web.Mvc/Framework/Modules/IModuleExecutionEngine.cs
dimarobert: Same here, we could have a separate `IAsyncModuleExecutionEngine` with the current `ModuleExecutionEngine` implementing both.

DNN Platform/Library/UI/Containers/ActionBase.cs
dimarobert: using `await Task.Yield()` forces the runtime to always suspend the execution and perform an async switch, which requires an allocation and incurs task scheduling costs. For methods that do not in fact have any asynchronous execution the runtime is smart enough to run the entire method synchronously when `return Task.CompletedTask;` is used, with 0 overhead (`Task.CompletedTask` is a static singleton that can be reused by the runtime without making allocations for each async boundry).

dimarobert:
> @dimarobert thank you so much for this, a few things to review, but great amounts of research.
> 
> @donker and @sachatrauwaen I'm curious to see what impacts if any this has on the MVC Pipeline work that is going on as well

I will try to give some context for this.

In the WebForms implementation here it was easy enough to hook the async stuff over / in between the existing sync stuff due to the fact that Microsoft provided clear boundaries between the sync and async sections of the page execution through the `this.Page.RegisterAsyncTask(new PageAsyncTask(this.MyAsyncMethod));` and also ensures that the async methods that we give to it are safely interleaved with the sync logic of the WebForms page and the transition from sync to async and back to sync does not fall into the async over sync anti-pattern. Execution is also safely done in the correct SynchronizationContext and ExecutionContext so all AsyncLocal variables (like Thread.CurrentCulture, HttpContext.Current, and so on) are available in all stages.

With the MVC implementation I see in #6749 there might be two show stoppers or hard to tackle problems:
1. HtmlHelper does not have support for async in .net framework, like `HtmlHelper.PartialAsync()` which would be needed to async render child controls. Only ASP.NET Core has `HtmlHelper.PartialAsync()`. So, if even possible to (did not look into it), this would have to be implemented by DNN. Probably `HtmlHelper.ActionAsync()` and other helper overloads would have to be provided as an Async overload as well.
2. Everything will have to be async starting at the root. Every place (extension point) where we expect we would need to support async (now or at some point in the future) has to be async, and all callers recursively up the stack have to be async as well. I would say that most of the new abstractions would have to be converted to async-first. Even more important if we add to the discussion that to truly benefit from the async execution we need it to terminate into the different calls to services so threads are freed up while waiting for database calls and (distributed) cache calls for models, file and network I/O calls, and any other operations that are not CPU bound. So as an exercise, to support async module execution we would have to have a `Task<IHtmlString> IMvcModuleControl.HtmlAsync()`, and everything in the call path has to be async as well:
   - `IHtmlString IMvcModuleControl.Html()` => `Task<IHtmlString> IMvcModuleControl.HtmlAsync()`
   - `IHtmlString ContainerHelpers.Content()` => `Task<IHtmlString> ContainerHelpers.ContentAsync()`
   - `htmlHelper.Partial(container.Value.ContainerRazorFile, container.Value)` would be replaced with the DNN implementation of `HtmlHelper.PartialAsync()`
   - `IHtmlString SkinExtensions.Pane()` => `Task<IHtmlString> SkinExtensions.PaneAsync()`

    In this implementation, the interfaces that the system uses should not have any methods for sync execution, only async. For modules that do not desire async, a base implementation of IMvcModuleControl could be provided by DNN that exposes a `protected abstract IHtmlString Html()` and it implements the async method as a redirect to it `Task<IHtmlString> IMvcModuleControl.HtmlAsync() => Task.FromResult(this.Html())`.

sachatrauwaen:
> > @dimarobert thank you so much for this, a few things to review, but great amounts of research.
> > @donker and @sachatrauwaen I'm curious to see what impacts if any this has on the MVC Pipeline work that is going on as well
> 
> I will try to give some context for this.
> 
> In the WebForms implementation here it was easy enough to hook the async stuff over / in between the existing sync stuff due to the fact that Microsoft provided clear boundaries between the sync and async sections of the page execution through the `this.Page.RegisterAsyncTask(new PageAsyncTask(this.MyAsyncMethod));` and also ensures that the async methods that we give to it are safely interleaved with the sync logic of the WebForms page and the transition from sync to async and back to sync does not fall into the async over sync anti-pattern. Execution is also safely done in the correct SynchronizationContext and ExecutionContext so all AsyncLocal variables (like Thread.CurrentCulture, HttpContext.Current, and so on) are available in all stages.
> 
> With the MVC implementation I see in #6749 there might be two show stoppers or hard to tackle problems:
> 
> 1. HtmlHelper does not have support for async in .net framework, like `HtmlHelper.PartialAsync()` which would be needed to async render child controls. Only ASP.NET Core has `HtmlHelper.PartialAsync()`. So, if even possible to (did not look into it), this would have to be implemented by DNN. Probably `HtmlHelper.ActionAsync()` and other helper overloads would have to be provided as an Async overload as well.
> 2. Everything will have to be async starting at the root. Every place (extension point) where we expect we would need to support async (now or at some point in the future) has to be async, and all callers recursively up the stack have to be async as well. I would say that most of the new abstractions would have to be converted to async-first. Even more important if we add to the discussion that to truly benefit from the async execution we need it to terminate into the different calls to services so threads are freed up while waiting for database calls and (distributed) cache calls for models, file and network I/O calls, and any other operations that are not CPU bound. So as an exercise, to support async module execution we would have to have a `Task<IHtmlString> IMvcModuleControl.HtmlAsync()`, and everything in the call path has to be async as well:
>    
>    * `IHtmlString IMvcModuleControl.Html()` => `Task<IHtmlString> IMvcModuleControl.HtmlAsync()`
>    * `IHtmlString ContainerHelpers.Content()` => `Task<IHtmlString> ContainerHelpers.ContentAsync()`
>    * `htmlHelper.Partial(container.Value.ContainerRazorFile, container.Value)` would be replaced with the DNN implementation of `HtmlHelper.PartialAsync()`
>    * `IHtmlString SkinExtensions.Pane()` => `Task<IHtmlString> SkinExtensions.PaneAsync()`
>    
>    In this implementation, the interfaces that the system uses should not have any methods for sync execution, only async. For modules that do not desire async, a base implementation of IMvcModuleControl could be provided by DNN that exposes a `protected abstract IHtmlString Html()` and it implements the async method as a redirect to it `Task<IHtmlString> IMvcModuleControl.HtmlAsync() => Task.FromResult(this.Html())`.

Great ! 
For the mvc pipeline, i think it is not a problem not to have HtmlHelper.PartialAsync as long as you do not put logic in your partials (and html helpers used in your partials).

But Html helpers on witch a lot is based in the mvc pipeline are synchronous by design in MVC 5.
Remark : In .net framework you need to be very carryfull to avoid the Sync Context Deadlock problem.
The solution is dotnet core ViewComponent !
The mvc 5 controllers can use async and then put all you async stuff here, and build the whole model here.
But actually in the mvc pipeline, it is not this way. Modules are rendererd more la ViewComponent and are independent of the controller.

dimarobert:
> But actually in the mvc pipeline, it is not this way. Modules are rendererd more la ViewComponent and are independent of the controller.

So, if Modules in the MvcPipeline do not have access to the controller and the only API surface that the module developer has access to is the `IHtmlString IMvcModuleControl.Html()`, doesn't that mean that MvcPipeline will be unable to support async execution for its modules? (unless the approach I gave above is implemented - with everything being async top to bottom in the MvcPipeline execution - which I see as a considerable API change to be made after the fact)

DNN Platform/Library/UI/Containers/ActionBase.cs
dimarobert: I've added some additional comments, let me know if that is good enough.

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
dimarobert: I've changed the code to throw instead. I hope the message is clear and detailed enough.

sachatrauwaen:
> > But actually in the mvc pipeline, it is not this way. Modules are rendererd more la ViewComponent and are independent of the controller.
> 
> So, if Modules in the MvcPipeline do not have access to the controller and the only API surface that the module developer has access to is the `IHtmlString IMvcModuleControl.Html()`, doesn't that mean that MvcPipeline will be unable to support async execution for its modules? (unless the approach I gave above is implemented - with everything being async top to bottom in the MvcPipeline execution - which I see as a considerable API change to be made after the fact)

The modules can have access to the controller, but it run all in a unic controller. 
I think you are right, MvcPipeline will be unable to support async execution for its modules (because existing reusable components in mvc 5, partial and child controllers, dous not support async).
If i understand it well your approach is based on PageAsyncTask witch is a pure webform stuff, so not appicable to the mvc pipeline. 
I read, "Razor views in MVC 5 are rendered by a synchronous rendering engine. The Razor view engine does not understand await". 

That's a raison why we want DNN.Core :)

dimarobert:
As I said in my initial comment on this "... if even possible to implement such a `HtmlHelper.PartialAsync()` (did not look into it)". Now I looked more into it and looks close to impossible to do. The entire stack is sync, event the runtime generated class of the cshtml writes its output through a sync `Execute()` method that gets called from `WebPageBase.ExecutePageHierarchy()`.

There is however the `HtmlHelper.Action()` helper that I would expect to be able to somehow run for any controller action, even if it is async, so they are hacking it in somehow and waiting for it - similar to WebForms async point. But even if we figure out how they do it for Actions, the big hurdle is that sync `Execute()` method that gets generated by the compilation of the cshtml view. I do not see how that could be overcome.

So, looks quite unlikely for the MvcPipeline Razor+ Modules (I hope that is the correct terminology) to be able to support async.

sachatrauwaen:
> As I said in my initial comment on this "... if even possible to implement such a `HtmlHelper.PartialAsync()` (did not look into it)". Now I looked more into it and looks close to impossible to do. The entire stack is sync, event the runtime generated class of the cshtml writes its output through a sync `Execute()` method that gets called from `WebPageBase.ExecutePageHierarchy()`.
> 
> There is however the `HtmlHelper.Action()` helper that I would expect to be able to somehow run for any controller action, even if it is async, so they are hacking it in somehow and waiting for it - similar to WebForms async point. But even if we figure out how they do it for Actions, the big hurdle is that sync `Execute()` method that gets generated by the compilation of the cshtml view. I do not see how that could be overcome.
> 
> So, looks quite unlikely for the MvcPipeline Razor+ Modules (I hope that is the correct terminology) to be able to support async.

HtmlHelper.Action is for executing Child Action Controllers, and from what i read, they can not run async. And because there called from a html helper from a razor view, it confirm that they cannot run async. And because Child Controllers are not supported in dotnet core, i prefer to stay on partials and/or ViewComponent patern.

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
dimarobert: I think a small correction is in order here, the `Get{0}Actions` method is and was called as part of `ModuleActionItemsAttribute.OnActionExecuting()` ActionFilter, so right before the execution of the action. But for an async action that would still be executed by ASP.NET in the async portion of the controller execution (`await Task.Factory.FromAsync(asyncController.BeginExecute, asyncController.EndExecute, this.RequestContext, null);`) and I do not think there is a way around it, we only have access to ModuleActions after the async point.

Going back to the comment about IDnnController interface above, if we want to get rid of the `ModuleActionsAsync` we have two options:
1. Do not support async execution for the `Get{0}Actions` method, that means that `ModuleActionItemsAttribute.OnActionExecuting()` is able to directly assign ModuleActions like in the sync version.
2. Find a way to run it async in-place. ActionFilters are not async in net framework, so I needed to store the Task somewhere to be able to await it after the action execution (in the `ModuleApplication.ExecuteModuleAsync()`). One option would be to leverage the `ResultCapturingActionInvoker.BeginInvokeActionMethodWithFilters()` to search for the `ModuleActionItemsAttribute` specifically and run it separately there using a special (not from ActionFilter API) method that is async, then remove it (or skip it or no-op it) from the normal filter execution.

Looking at the code, the `IDnnController.IsAsync` (which is in fact a flag that is closer in meaning to IsExecutingInAsyncContext, as in it is executing through `ModuleApplication.ExecuteModuleAsync()`) is used only by the `ModuleActionItemsAttribute.OnActionExecuting` to determine if a `Task<ModuleActionCollection>` return type makes sense. If we go with either option from the above we might be able to also get rid of the `IDnnController.IsAsync` from the interface if we think we will not need it anywhere else in the future. In option 2, we are able to because we would get a call through the new async method of  `ModuleActionItemsAttribute` only from an async execution context.

bdukes: IMO it's not important to support async module actions and would be fine with option 1. If you're happy to implement option 2 and it feels solid to you, that's fine with me, also.

dimarobert: I've implemented the support for async ModuleActions here: https://github.com/dimarobert/Dnn.Platform/compare/async-mvc-actions...dimarobert:Dnn.Platform:async-mvc-moduleactions

Initially I attempted to do it generically with an `IAsyncActionFilter` interface, but there were two caveats:
1. The filters would have been pulled from the original list and executed separately from the rest of sync `IActionFilter` changing the order of execution for these filters. Also the short-circuiting `ActionExecutingContext.Result` would have been ignored for the execution of the async filters and supporting it would have been challenging due to the re-ordering. These still applies with the implementation above, but is only limited to the `ModuleActionItemsAttribute` so I do not foresee any major impact.
2. Due to the fact that the async execution of the Controller in ASP.NET FW uses the old APM async model (with BeginX / EndX methods) that only supports async execution on the BeginX path and once the internal `callback` gets called (and the EndX return path starts executing) everything needs to be sync; and the fact that `OnActionExecuting()` gets executed in the `BeginX` while `OnActionExecuted()` is executed on the `EndX` (with the actual Controller Action in the middle), it means that there is no way to run `OnActionExecuted` async, so it will always be a half implementation.

I feel like these caveats are somewhat acceptable for a specific `ModuleActionItemsAttribute.OnActionExecutingAsync()` implementation, but would be confusing for a generic `IAsyncActionFilter`.
If however you think the implementation is too much of a hack even for this, then we can go with the first option instead _'Do not support async execution for the Get{0}Actions method'_.

DNN Platform/DotNetNuke.Web.Mvc/Framework/Controllers/IDnnController.cs
mitchelsellers: One thought would be to have IAsyncDnnController inherit IDnnController, and then we could add this, balancing the breaking change.

This I believe is the final consideration/question on this PR.

dimarobert: @mitchelsellers Well, I think that with the approach from [here](https://github.com/dnnsoftware/Dnn.Platform/pull/7193#discussion_r3105048208), the need for the additions in the IDnnController (or the split into an IAsyncDnnController) are no longer needed.

Looks like this branch was rebased, here are the related changes https://github.com/dimarobert/Dnn.Platform/commit/7eb3b87902e228eb31517318b5062b9fc45e566b

bdukes: So it doesn't look like that commit is part of this PR branch, do we need to pick those changes into this branch?

dimarobert: I did not add the commit to this PR because I wanted a decision on whether we want the Async support for ModuleActions, with the implementation provided there which could be considered a bit hacky, or if we do not need/want async support for the retrieval of ModuleActions at all.

DNN Platform/Library/UI/Modules/ModuleInstanceContext.cs
bdukes: Thank you for the work on this implementation. I agree that the caveats you list are acceptable, and I'm good with including this change.

dimarobert: updated with the changes for ModuleActions.

bdukes (approved):

mitchelsellers (approved):

PR #7201, opened 2026-04-10
Report issues on GitHub actions

bdukes:
## Summary
This PR updates the Cake build process to report issues when running on GitHub Actions (it currently reports issues using specific Azure DevOps APIs).

Additionally, I adjusted the usage of paths and IO to use more Cake helpers when possible.

### DIFF
--- a/Build/Build.csproj
+++ b/Build/Build.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="Cake.Frosting.Git" />
     <PackageReference Include="Cake.Issues" />
     <PackageReference Include="Cake.Frosting.Issues.MsBuild" />
+    <PackageReference Include="Cake.Frosting.Issues.PullRequests.GitHubActions" />
     <PackageReference Include="Cake.Json" />
     <PackageReference Include="Cake.XdtTransform" />
     <PackageReference Include="Dnn.CakeUtils" />
--- a/Build/Tasks/CopyWebConfigToDevSite.cs
+++ b/Build/Tasks/CopyWebConfigToDevSite.cs
@@ -34,7 +34,7 @@ namespace DotNetNuke.Build.Tasks
                 .Replace("{DbOwner}", context.Settings.DbOwner)
                 .Replace("{ObjectQualifier}", context.Settings.ObjectQualifier);
             var res = context.XmlTransform(trans, conf);
-            var webConfig = context.File(System.IO.Path.Combine(context.Settings.WebsitePath, "web.config"));
+            var webConfig = context.Settings.WebsitePath + context.File("web.config");
             context.FileWriteText(webConfig, res);
         }
     }
--- a/Build/Context.cs
+++ b/Build/Context.cs
@@ -5,7 +5,6 @@ namespace DotNetNuke.Build
 {
     using System;
     using System.Globalization;
-    using System.IO;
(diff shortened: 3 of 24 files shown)


Build/Build.csproj
```diff
@@ -23,15 +23,16 @@
...
     <PackageReference Include="Octokit" />
     <PackageReference Include="Microsoft.Data.SqlClient" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" VersionOverride="10.0.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" VersionOverride="10.0.0" />
-    <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" VersionOverride="8.15.0" />
-    <PackageReference Include="Microsoft.Identity.Client" VersionOverride="4.81.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" VersionOverride="10.0.5" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" VersionOverride="10.0.5" />
+    <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" VersionOverride="8.17.0" />
+    <PackageReference Include="Microsoft.Identity.Client" VersionOverride="4.83.3" />
```
valadas: Was this intentional?

bdukes: I was trying to fit a few things into one PR, but it looks like it's more complicated than I hoped, so that'll be a different PR.

valadas (dismissed):

Build/Tasks/CopyWebConfigToDevSite.cs
```diff
@@ -34,7 +34,7 @@ public override void Run(Context context)
                 .Replace("{DbOwner}", context.Settings.DbOwner)
                 .Replace("{ObjectQualifier}", context.Settings.ObjectQualifier);
             var res = context.XmlTransform(trans, conf);
-            var webConfig = context.File(System.IO.Path.Combine(context.Settings.WebsitePath, "web.config"));
+            var webConfig = context.Settings.WebsitePath + context.File("web.config");
```
mitchelsellers, line 37: Was there a reason to remove the Path.Combine here?  I know traditionally that helps with pathing issues and / type situations, not sure its really needed or a concern here though

mitchelsellers (approved):
Looks good, jsut one question I added.

Given that we are switching to GitHubActions do we eventually want to remove the AzurePipelines stuff, or just leave it to help others that may neeed it?

valadas (approved):

valadas:
> Looks good, jsut one question I added.
> 
> Given that we are switching to GitHubActions do we eventually want to remove the AzurePipelines stuff, or just leave it to help others that may neeed it?

I think we could kill it IMO.

Build/Tasks/CopyWebConfigToDevSite.cs
bdukes: This code is using Cake's path types, so the `+` operator is actually doing a `Path.Combine` kind of operation.

bdukes:
Once we totally shut down our usage of Azure Pipelines we could remove it, though I don't know that it really hurts anything at this point.

PR #7239, opened 2026-04-25
Fixed an issue where the themes list were no longer vertical in Pages module

valadas:
I can't for the life of me remember where this was discussed, but in 10.3.1 the list of themes/layouts/containers was no longer listed horizontally but showed up as a long vertical list.

This PR fixes that.


donker (approved):

bdukes (approved):
Do we know why this changed? Is it possible there are other similar changes that need to happen elsewhere?

valadas:
> Do we know why this changed? Is it possible there are other similar changes that need to happen elsewhere?

My suspicion at this stage is that some css imports stopped working many years ago and they were fixed by adding local css. Then something fixed those imports and it started producing those issues.

Issue #7184, opened 2026-04-03
[Enhancement]: Generic head injection


Tychodewaard:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of problem

**Issue**
There are many scenarios where a site admin or SEO specialist wants to add a code snippet to the site or a page. For instance to validate Google Search console, Tag manager, Rich snippets, Hotjar etc.

**Current way of working**
We have the option to inject snippets in the site settings and/or the page settings. But this gets very cluttered and mistakes are easily made. 

### Description of solution

As a Site admin or SEO specialist, 

- I have the ability to add/edit/delete snippets
- Provide a Title
- Provide the injected snippet

The UI for this could be like

<img width="1123" height="1115" alt="Image" src="https://github.com/user-attachments/assets/a9504abe-028d-4707-b82d-4d8eacaa30d6" />

The same for the Site settings

### Do you plan to contribute code for this enhancement?

- [x] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Tychodewaard:
The solution we are working on

1.The page will pop up a dialog box after the administrator clicks the “Add tag” link.
2.The administrator can directly enter the name and code of the tag in the pop-up window.
3.All added tags will be displayed in a list for easy management.
The data will be stored uniformly in the PageHeaderTags field of the TabSettings sheet, as shown below:

1.The page will pop up a dialog box after the administrator clicks the “Add tag” link.
2.The administrator can directly enter the name and code of the tag in the pop-up window.
3.All added tags will be displayed in a list for easy management.
The data will be stored uniformly in the PageHeaderTags field of the TabSettings sheet, as shown below:

<img width="699" height="312" alt="Image" src="https://github.com/user-attachments/assets/2b2b5e4b-3389-4380-b22c-b0ca65bceeee" />

The value in the SettingValue field will be stored in a structured JSON format, as shown in the example below:
{
"PageHeaderTags": [
{
"name": "Google verification",
"code": "<meta name='google-site-verification' content='...' />",
"createdAt": "2026-04-02",
"generatedBy": "System/Admin"
},
{
"name": "Google Analytics",
"code": "...",
"createdAt": "2026-04-02",
"generatedBy": "System/Admin"
}
]
}

For HTML Page Header Tags in Site Settings, we will use the same processing logic as the Page Settings described above.

[...]

PR #7248, opened 2026-04-30
Enhance Page Header Tag Management Functionality

zzmzaizai:
## Summary

Implements #7184 — Provides a structured, user-friendly way for site administrators and SEO specialists to manage header code snippets (e.g., Google Search Console verification, Tag Manager, Rich snippets, Hotjar, etc.) at both the **site level** and **page level**, replacing the previous single text-area approach that was cluttered and error-prone.

## Problem

Previously, header code injection was done through a single `PageHeadText` text field at both the site and page level. This approach:
- Gets cluttered when multiple snippets are needed
- Makes it easy to accidentally break existing snippets when editing
- Provides no way to organize, name, or individually manage snippets
- Has no validation or duplicate detection

## Solution

Introduces a structured **Page Header Tags** system where each snippet has:
- A **Name** (identifier, max 30 chars, must be unique)
- **Content** (the actual HTML/code to inject)

### Backend Changes

1. **New Core Model**: `PageHeaderTagInfo` class in `DotNetNuke.Entities.Tabs` namespace
   - Stores snippets as individual settings with `PageHeaderTag_` prefix
   - Provides CRUD operations for both Tab-level and Portal-level snippets
   - Handles normalization (trim, deduplicate)

2. **New API DTO**: `PageHeaderTagItem` for Web API serialization

3. **API Integration**:
   - Pages API: `PageSettings` now includes `PageHeaderTags` list; saved on page create/update/clone
   - Site Settings API: `GetDefaultPagesSettings` / `UpdateDefaultPagesSettings` now handle `PageHeaderTags`
   - XSS sanitization applied to tag names

4. **Page Rendering**: `Default.aspx.cs` updated to render structured header tags with fallback to legacy `PageHeadText` for backward compatibility

5. **Template Import/Export**: Portal templates now include `<pageheadertags>` XML elements

6. **Export/Import Module**: Detects new-style header tags and clears legacy `PageHeadText` to avoid conflicts

### Frontend Changes

7. **New React Component**: `PageHeaderTags` with:
   - List view showing all tags by name
   - Inline add/edit forms with name + content fields
   - Delete with confirmation
   - Validation: required fields, duplicate name detection, name length limit (30 chars)

8. **Integration**:
   - Page SEO panel: added `PageHeaderTags` component
   - Site Settings → Default Pages: added `PageHeaderTags` component (replacing old text area)

### Backward Compatibility

- Legacy `PageHeadText` is preserved as fallback — if no structured tags exist, the old text is still rendered
- When new-style tags are present, `PageHeadText` is set to `Null.NullString`/`"false"` to avoid duplication
- Existing templates and exported data continue to work

## Screenshots

*(Add screenshots of the new UI here — the PageHeaderTags component showing the list view and add/edit form)*

## How to Test

1. **Site-level**: Go to Site Settings → Default Pages → add/edit/delete header tags. Verify they render in the page `<head>` section.
2. **Page-level**: Go to Pages → edit a page → SEO tab → add/edit/delete header tags. Verify they render on that specific page.
3. **Template export/import**: Export a portal template with header tags, import into a new portal, verify tags are preserved.
4. **Backward compatibility**: Verify that existing pages with `PageHeadText` still render correctly.
5. **Validation**: Try adding duplicate names, empty names, names > 30 chars — verify error messages.

## Files Changed

**New Files:**
- `DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs`
- `Dnn.PersonaBar.Extensions/Services/DTO/PageHeaderTagItem.cs`
- `ClientSide/Pages.Web/src/components/PageHeaderTags/PageHeaderTags.jsx`
- `ClientSide/Pages.Web/src/components/PageHeaderTags/style.less`
- `ClientSide/SiteSettings.Web/src/components/PageHeaderTags/PageHeaderTags.jsx`
- `ClientSide/SiteSettings.Web/src/components/PageHeaderTags/style.less`

**Modified Files:**
- `Dnn.PersonaBar.Extensions/Services/DTO/PageSettings.cs`
- `Dnn.PersonaBar.Extensions/Services/DTO/SiteSettings/UpdateDefaultPagesSettingsRequest.cs`
- `Dnn.PersonaBar.Extensions/Components/Pages/Converters.cs`
- `Dnn.PersonaBar.Extensions/Components/Pages/PagesControllerImpl.cs`
- `Dnn.PersonaBar.Extensions/Components/Pages/XssCleaner.cs`
- `Dnn.PersonaBar.Extensions/Components/Pages/BulkPagesController.cs`
- `Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs`
- `DNN Platform/Website/Default.aspx.cs`
- `DNN Platform/Library/Entities/Portals/Templates/PortalTemplateExporter.cs`
- `DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs`
- `DNN Platform/Modules/DnnExportImport/Components/Services/PagesExportService.cs`
- `ClientSide/Pages.Web/src/components/Seo/Seo.jsx`
- `ClientSide/SiteSettings.Web/src/components/defaultPagesSettings/index.jsx`


jeremy-farrance:
I downloaded and reviewed the results and this is a really useful enhancement, thank you!!
 
I have one quick visual - and hopefully easy and/or quick - improvement to the UI to suggest. Given the Content is usually going to be longer, how about stacking the fields vertically with widths at 60% and 100% like this:

image

jeremy-farrance:
Another issue that should be considered is where these load in the `<head>` - meaning the order or priority. When I added a new one in Site Settings / Behavior, my addition was added first, even before `<title... > ...`.
 
Maybe we should consider adding an additional field for Priority (similar to CDF)?
 
 Easier - and given the fact that these are probably/usually user-level after-thoughts, hacks, etc. - might be to just add them with a priority of 200 or 500, so they add late/last.

Tychodewaard:
@jeremy-farrance thank you for your feedback.

As I am the one funding this enhancement: would you be ok if I put extras on the backlog for further enhancements? This would go for any other input that are outside the initial scope. 
In my experience, we'll have multiple new wishes when we will actually use it. Your input are good and valuable examples of this. With this process I can remain in control of budgets :-)

Tychodewaard:
> Another issue that should be considered is where these load in the `<head>` - meaning the order or priority. When I added a new one in Site Settings / Behavior, my addition was added first, even before `<title... > ...`.
> 
> Maybe we should consider adding an additional field for Priority (similar to CDF)?
> 
> Easier - and given the fact that these are probably/usually user-level after-thoughts, hacks, etc. - might be to just add them with a priority of 200 or 500, so they add late/last.

I noticed this too. But this is already the case. Never realized that but Site settings - page head injections are above the title and Page settings - page head injections just below the title. So, you have multiple places for this. TBH, I don't know if this is correct and desirable or could be hurtful. After all these years leveragin both, I have never experience any issues, though.

jeremy-farrance:
> As I am the one funding this enhancement: would you be ok if I put extras on the backlog for further enhancements? This would go for any other input that are outside the initial scope...
 
All good, I am just sharing ideas, I did not intend for these to be requirements or anything. If they end up on a backlog or wish list, that is great!

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var normalizedItems = Normalize(items);
+            var targetSettingNames = new HashSet<string>(normalizedItems.Select(item => item.SettingName), StringComparer.OrdinalIgnoreCase);
+            var existingSettingNames = TabController.Instance.GetTabSettings(tabId)
+                .Cast<DictionaryEntry>()
+                .Select(entry => Convert.ToString(entry.Key, CultureInfo.InvariantCulture))
+                .Where(key => !string.IsNullOrEmpty(key))
+                .ToList();
+
+            foreach (var key in existingSettingNames)
```
mitchelsellers, line 93: I'm curious if we might be able to find a way to avoid the purge & reload, and update items if possible.  By deleting & reloading you will always have an entry that looks new, rther than having a created/updated history.  Not huge

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
```diff
@@ -0,0 +1,152 @@
...
+            var normalizedItems = Normalize(items);
+            var targetSettingNames = new HashSet<string>(normalizedItems.Select(item => item.SettingName), StringComparer.OrdinalIgnoreCase);
+            var existingSettingNames = (string.IsNullOrEmpty(cultureCode)
+                ? PortalController.Instance.GetPortalSettings(portalId)
+                : PortalController.Instance.GetPortalSettings(portalId, cultureCode))
+                .Keys
+                .Where(key => !string.IsNullOrEmpty(key))
+                .ToList();
+
+            foreach (var key in existingSettingNames)
```
mitchelsellers, line 118: Similar comment here, and I also agree with the note from Copilot in that we want to be sure to have the culture code here to avoid any cross-over in settings from one culture to another.

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
```diff
@@ -0,0 +1,152 @@
...
+                }
+            }
+
+            foreach (var item in normalizedItems)
+            {
+                PortalController.Instance.UpdatePortalSetting(portalId, item.SettingName, item.Content, true, cultureCode ?? Null.NullString);
+            }
+        }
+
+        private static List<PageHeaderTagInfo> Normalize(IEnumerable<PageHeaderTagInfo> items)
```
mitchelsellers: I'm concerned that we are going to be arbitrarily picking a winner here grouping on a normalized key, what specific case of user input would allow this, as it might be a very hard to diagnose user behavior if we do this.

DNN Platform/Website/Components/Portals/portal.template.xsd
```diff
@@ -41,6 +41,21 @@
               <xs:element name="controlpanelsecurity" type="xs:string" minOccurs="0" maxOccurs="1" />
               <xs:element name="controlpanelvisibility" type="xs:string" minOccurs="0" maxOccurs="1" />
               <xs:element name="pageheadtext" type="xs:string" minOccurs="0" maxOccurs="1" />
+              <xs:element name="pageheadertags" minOccurs="0" maxOccurs="1">
```
mitchelsellers, line 44: @dnnsoftware/approvers I believe this is safe with the minOccurs of 0 but just want to confirm that we all agree that this should be ok

DNN Platform/Website/Portals/_default/Blank Website.template
```diff
@@ -24,7 +24,10 @@
     <hostspace>0</hostspace>
     <userquota>0</userquota>
     <pagequota>0</pagequota>
-    <pageheadtext>&lt;meta content="text/html; charset=UTF-8" http-equiv="Content-Type" /&gt;</pageheadtext>
+    <pageheadtext></pageheadtext>
```
mitchelsellers: Given the new structure, should we just remove the older `<pageheadtext>` element entirely from the default templates?

DNN Platform/Website/Providers/DataProviders/SqlDataProvider/10.03.02.SqlDataProvider
```diff
@@ -7,3 +7,45 @@
...
+  AND ISNULL(ps.SettingValue, '') NOT IN ('', 'false')
+  AND NOT EXISTS (
+      SELECT 1
+      FROM {databaseOwner}[{objectQualifier}PortalSettings]  psExisting
+      WHERE psExisting.PortalID = ps.PortalID
+        AND ISNULL(psExisting.CultureCode, '') = ISNULL(ps.CultureCode, '')
+        AND psExisting.SettingName = 'PageHeaderTag_Default');
+GO
+
+-- Reset legacy PageHeadText to 'false' after migration
```
mitchelsellers: Is there a way to avoid this value of false?  Maybe not, and maybe more of a question for @dnnsoftware/approvers  but I worry that down the road this will become even more confusing.

DNN Platform/Website/Default.aspx.cs
```diff
@@ -729,13 +751,15 @@ private void InitializePage()
...
-                !(HeaderTextRegex.IsMatch(this.PortalSettings.ActiveTab.PageHeadText) ||
-                  HeaderTextRegex.IsMatch(this.PortalSettings.PageHeadText)))
+            // META generator
+            this.Generator = string.Empty;
+
+            var tabLegacyPageHeadText = this.PortalSettings.ActiveTab.PageHeadText ?? string.Empty;
+
+            // META Robots - hide it inside popups and if PageHeadText of current tab already contains a robots meta tag
+            if (!UrlUtils.InPopUp() &&
+                !(HeaderTextRegex.IsMatch(PageHeaderTagInfo.Render(PageHeaderTagInfo.GetTabItems(this.PortalSettings.ActiveTab.TabID)) + tabLegacyPageHeadText) ||
```
mitchelsellers: This is a second iteration that builds/renders the items within the same process, ideally, it would be best to avoid this if we can given the every page load nature of this

mitchelsellers (commented):
Thank you so much for this contribution, just a few notes and questions upon a quick initial review

mitchelsellers:
I would like to also note that I think to be usable, given how users do things today, we need the text area to be full width for individual items otherwise it will be almost impossible to see/manage safely.

zzmzaizai:
> I downloaded and reviewed the results and this is a really useful enhancement, thank you!!
> 
> I have one quick visual - and hopefully easy and/or quick - improvement to the UI to suggest. Given the Content is usually going to be longer, how about stacking the fields vertically with widths at 60% and 100% like this:
> 
> image

Great suggestion, thank you! The fields are now stacked vertically with 60% and 100% widths — that should handle longer content much better. The update has been pushed to this branch.

mitchelsellers:
@zzmzaizai Just making sure you saw the few other comments/questions inline with this PR.

zzmzaizai:
> I would like to also note that I think to be usable, given how users do things today, we need the text area to be full width for individual items otherwise it will be almost impossible to see/manage safely.

Thank you for your feedback. I've addressed this in commit [7208846](https://github.com/dnnsoftware/Dnn.Platform/pull/7248/changes/7208846a4bc2166163ce110e2c0ef2a8f8a0f88d) by setting the text area for individual items to full width in the edit form. The CSS change adds width: 100%; to the .dnn-multi-line-input-with-error class within the .editTag section. Please let me know if this resolves the usability issue.

DNN Platform/Website/Default.aspx.cs
zzmzaizai: I've addressed this performance concern in commit [2b62ccf](https://github.com/dnnsoftware/Dnn.Platform/pull/7248/changes/2b62ccf984c0a06cf13686754ce35e0218f34f02). The rendering logic has been refactored so that items are now built and rendered only during the initial render, avoiding unnecessary rebuilds on every page load.

DNN Platform/Website/Portals/_default/Blank Website.template
mitchelsellers: @dnnsoftware/approvers Thoughts?

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
zzmzaizai: > The current implementation already uses a delta approach rather than a full purge-and-reload:
> - **Deleted items**: Only settings that no longer exist in the new collection are removed via `DeleteTabSetting`/`DeletePortalSetting`.
> - **Added/changed items**: New and updated entries go through `UpdateTabSetting`/`UpdatePortalSetting`, which performs an upsert (INSERT or UPDATE).

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
zzmzaizai: I've already fixed this. The culture code was missing when deleting items, which could cause cross-culture issues. It's now properly passed in both save and delete operations.

donker:
Does this change make the PageHeadText column in Tabs obsolete?

If so we should probably clean things up by: 1. remove the column (SQL), 2. rewire the property to point to the new data location (TabSettings) and 3. mark it as obsolete.

DNN Platform/Website/Portals/_default/Blank Website.template
bdukes: yes, it should be removed entirely

zzmzaizai:
> Does this change make the PageHeadText column in Tabs obsolete?
> 
> If so we should probably clean things up by: 1. remove the column (SQL), 2. rewire the property to point to the new data location (TabSettings) and 3. mark it as obsolete.


The PageHeadText column is no longer used and the data has already been cleaned up.
 - Removing the column via SQL might break external modules that directly query it. I'm not sure if Dnn has a standard practice for this. Please advise.
 - Marking the property as [Obsolete] is fine and I can do that.

mitchelsellers (changes requested):
A detailed team review of this PR was completed today to aggregate the feedback and provide an overall update to the system.  I am going to dismiss most of the other comments to keep things clean here.

During the initial review, there are concerns regarding the processing that has to occur for each individual page load, including key enumeration, string concatenation, and collection copying.  

A suggestion/thought would be to store a single additional "PageHeaderTagsGrouped" or some similar name that had all of the processing done at save (Tab or Portal) so that the Default.aspx processing would only be grabbing keys that were already ready for render and cached. 

Additionally, regarding the existing field, there was a consensus that the following could be done for the existing "Tabs" PageHeadText.

* Remove the column after successful migration
* Update the TabInfo to mark that column as obsolete, to be removed in DNN v12
* Update the process within the DNN interface that saves the TabInfo, if a value was set in that field, update the "PageHeaderTag_Default" value
* Remove the field from PortalTemplates

zzmzaizai:
> A detailed team review of this PR was completed today to aggregate the feedback and provide an overall update to the system. I am going to dismiss most of the other comments to keep things clean here.
> 
> During the initial review, there are concerns regarding the processing that has to occur for each individual page load, including key enumeration, string concatenation, and collection copying.
> 
> A suggestion/thought would be to store a single additional "PageHeaderTagsGrouped" or some similar name that had all of the processing done at save (Tab or Portal) so that the Default.aspx processing would only be grabbing keys that were already ready for render and cached.
> 
> Additionally, regarding the existing field, there was a consensus that the following could be done for the existing "Tabs" PageHeadText.
> 
> * Remove the column after successful migration
> * Update the TabInfo to mark that column as obsolete, to be removed in DNN v12
> * Update the process within the DNN interface that saves the TabInfo, if a value was set in that field, update the "PageHeaderTag_Default" value
> * Remove the field from PortalTemplates

Here is a summary of the actions taken in response to each point:
﻿
**1. Remove the column after successful migration**
- The `PageHeadText` entry in the `PortalSettings` table has been removed.
- Regarding the `Tabs` table column: this change would be too invasive and has not been included in this PR. The field remains but is now effectively superseded by the new `PageHeaderTags` data structure.
﻿
**2. Mark TabInfo column as obsolete (to be removed in DNN v12)**
- Done. The `TabInfo.PageHeadText` property has been marked as `[Obsolete]` with a note for removal in v12.
﻿
**3. Migration of existing `TabInfo.PageHeadText` values to `PageHeaderTag_Default`**
- Handled during the SQL migration. Any existing `PageHeadText` values on tabs are migrated to the new `PageHeaderTags` collection with the `Default` tag type.
﻿
**4. Remove the field from PortalTemplates**
- Done.

mitchelsellers:
> During the initial review, there are concerns regarding the processing that has to occur for each individual page load, including key enumeration, string concatenation, and collection copying.
> 
> A suggestion/thought would be to store a single additional "PageHeaderTagsGrouped" or some similar name that had all of the processing done at save (Tab or Portal) so that the Default.aspx processing would only be grabbing keys that were already ready for render and cached.

@zzmzaizai any thoughts on this final item?

zzmzaizai:
> > During the initial review, there are concerns regarding the processing that has to occur for each individual page load, including key enumeration, string concatenation, and collection copying.
> > A suggestion/thought would be to store a single additional "PageHeaderTagsGrouped" or some similar name that had all of the processing done at save (Tab or Portal) so that the Default.aspx processing would only be grabbing keys that were already ready for render and cached.
> 
> @zzmzaizai any thoughts on this final item?

Our approach: we'll use DNN's built-in cache for both portal-level and page-level Header Tags. On first load, the concatenated Header Tags content is stored in cache. On subsequent loads, it's served directly from cache — no repeated processing.

When tags are saved/added/deleted (portal or page), we simply clear the corresponding cache entry. The next page load will re-concatenate and re-cache the Header Tags content automatically.

mitchelsellers:
> Our approach: we'll use DNN's built-in cache for both portal-level and page-level Header Tags. On first load, the concatenated Header Tags content is stored in cache. On subsequent loads, it's served directly from cache — no repeated processing.

I think this may work reasonably well, my only concern with cache entries, rather than utilizing already cached items, is that in large portal installations (note there are many installations with hundreds of portals and/or hundreds/thousands of pages, this may introduce a very large number of additional cache keys and entries that might be able to be absorbed in the other cache entries.

However, being that it only impacts those pages/portals with specific values, it MAY be ok to be its own additional cache.

@bdukes @donker @valadas or @david-poindexter any thoughts/comments to the above?

donker:
> I think this may work reasonably well, my only concern with cache entries, rather than utilizing already cached items, is that in large portal installations (note there are many installations with hundreds of portals and/or hundreds/thousands of pages, this may introduce a very large number of additional cache keys and entries that might be able to be absorbed in the other cache entries.
> 
> However, being that it only impacts those pages/portals with specific values, it MAY be ok to be its own additional cache.
> 
> @bdukes @donker @valadas or @david-poindexter any thoughts/comments to the above?

I think at this point we may as well let that be a problem when it really becomes a problem. We'll need good testing of the performance of larger sites when we go RC.

valadas:
Nothing strikes me other than already mentioned things here. I'll do an additional review/testing when the above are resolved.

donker:
@zzmzaizai There are build errors due to the fact that the DnnExportImport module was not adjusted for this change. This module still expects the PageHeadText of a tab to be present. Since we are changing this behavior that code will need to be adjusted as well.

zzmzaizai:
> @zzmzaizai There are build errors due to the fact that the DnnExportImport module was not adjusted for this change. This module still expects the PageHeadText of a tab to be present. Since we are changing this behavior that code will need to be adjusted as well.

I've updated the DnnExportImport/Export Class

DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
```diff
@@ -962,27 +962,22 @@ private void ParsePortalSettings(XmlNode nodeSettings, int portalId)
                 PortalController.UpdatePortalSetting(this.portalController, portalId, "ControlPanelVisibility", XmlUtils.GetNodeValue(nodeSettings, "controlpanelvisibility"));
             }
 
-            if (!string.IsNullOrEmpty(XmlUtils.GetNodeValue(nodeSettings, "pageheadtext", string.Empty)))
```
bdukes: Do we need to support importing from an older version that included `pageheadtext`?

zzmzaizai: For template import/export, the legacy pageheadtext field has been replaced with the new pageheadertags/pageheadertag structure.

bdukes: Yes, the question is whether an export from a previous version should be able to be imported into this version (without losing the header tags).

donker: IMHO we could do the same as for deprecation. This will work until v12 and then we remove support. Seems fair we leave a bit of time for others to adjust.

valadas: I agree

bdukes: @zzmzaizai are you able to update the import to continue supporting templates that include the old `pageheadtext`?

bdukes:
I've merged this with latest `develop` and bumped the version of the SQL file to 10.4.0

mitchelsellers:
@zzmzaizai any chance you can look at this PR and resolve the build issues that are there now?

DNN Platform/Library/Entities/Portals/Templates/PortalTemplateImporter.cs
zzmzaizai: @bdukes    Okay, I will get this done as soon as possible.

mitchelsellers:
@zzmzaizai We are working on getting the 10.4.0 release ready, and we wanted to check with you to see if you had an ETA on when we may have this ready for review again?  

We would love to see if we can get this merged/approved as part of the 10.4.0 RC.  If you can let us know your intentions ASAP. As this is something that we do need to include in a minor version release.

Ideally we are going to cut the RC next Tuesday (9/1)

zzmzaizai:
@mitchelsellers Thanks for the heads-up — the branch is updated: merged the latest develop (the only conflict was 10.04.00.SqlDataProvider, where upstream's copy was the empty placeholder, so our full migration script was kept), and fixed the CS0618 obsolete-API warnings in PagesExportService.cs

mitchelsellers:
Awesome, we will get this reviewed!

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+                {
+                    continue;
+                }
+
+                items.Add(new PageHeaderTagInfo
+                {
+                    Name = name,
+                    Content = Convert.ToString(entry.Value, CultureInfo.InvariantCulture),
+                });
+            }
```
valadas, line 54: Unless I am reading this wrong, this would get all the settings longer than the SettingPrefix.Length without checking if it is that prefix

valadas (changes requested):

DNN Platform/Library/Entities/Tabs/PageHeaderTagInfo.cs
zzmzaizai: Good question — the prefix is verified a few lines above the snippet: FromSettings skips any key that does not StartsWith(SettingPrefix, StringComparison.Ordinal) before calling Substring(SettingPrefix.Length), so only real PageHeaderTag_* keys reach the items.

valadas: Oh right, I missed that, looks good.

valadas (approved):
Nothing sticks out to me, thanks!

Tychodewaard:
Shout out to anyone involved. As I am the initiator of this feature and this is the first collab of @zzmzaizai with the usual suspects, I checkd "did everything go well for you?".

"Thank you for asking; the work itself was not complex, the workload was within our expectations, and the community's validation process was both smooth and well-structured."

So, compliments to the community. A warm welcome is a great foundation for collaboration!
Have great day!

mitchelsellers (approved):

PR #7258, opened 2026-05-04
Fix bug report version using RC

bdukes:
## Summary
#7236 introduced an issue where the bug-report version was set to the latest release, rather than the latest stable release.


valadas (approved):
Awesome

mitchelsellers (approved):

PR #7278, opened 2026-05-18
Add merge_group triggers for PR workflows

bdukes:
Required checks don't run once they're added to the merge queue


bdukes:

PR #7289, opened 2026-05-22
Create settings necessary for an MVC Pipeline - retrying in 1 - 2 - 3 ...

donker:
This PR creates the settings necessary for the MVC Pipeline. These are the following:

1. A site setting which determines the entire site's pipeline behavior: WebForms (default), MVC or Auto.The value becomes part of the PortalSettings. In the UI this can be found on the PB under Site Settings > Site Behavior > More at the bottom of the panel.
2. A page setting which tells DNN how to render: Inherited (default), WebForms, MVC. The setting is stored in the TabSettings. It is visible in the UI on the PB under Pages > Advanced > More towards the bottom.
3. A new field "MvcControlClass" on the ModuleControls table which informs DNN whether the module control supports the new pipeline. The value will be a class reference or empty in case there is no explicit support for MVC. There is support for this field in the PB extensions module (on the module control editor) and in the module's manifest (.dnn file)

This is the first PR in a series of PRs that aim to build out the MVC Pipeline functionality in DNN. It targets a feature branch as all PRs will need to be in before this can be merged into the main branch of the project. The reason for a more granular approach to the PRs is to make them easier to verify for the reviewers.


mitchelsellers:
@donker This seems to have grabbed a number of "rebase" type changes with it (version #'s, and other stuff).

Any chance you have a list of the specifically changed files?

donker:
> @donker This seems to have grabbed a number of "rebase" type changes with it (version #'s, and other stuff).
> 
> Any chance you have a list of the specifically changed files?

Crap. Redoing this https://github.com/dnnsoftware/Dnn.Platform/pull/7290/changes

PR #7290, opened 2026-05-22
Create settings necessary for an MVC Pipeline

donker:
This PR creates the settings necessary for the MVC Pipeline. These are the following:

1. A site setting which determines the entire site's pipeline behavior: WebForms (default), MVC or Auto.The value becomes part of the PortalSettings. In the UI this can be found on the PB under Site Settings > Site Behavior > More at the bottom of the panel.
2. A page setting which tells DNN how to render: Inherited (default), WebForms, MVC. The setting is stored in the TabSettings. It is visible in the UI on the PB under Pages > Advanced > More towards the bottom.
3. A new field "MvcControlClass" on the ModuleControls table which informs DNN whether the module control supports the new pipeline. The value will be a class reference or empty in case there is no explicit support for MVC. There is support for this field in the PB extensions module (on the module control editor) and in the module's manifest (.dnn file)

This is the first PR in a series of PRs that aim to build out the MVC Pipeline functionality in DNN. It targets a feature branch as all PRs will need to be in before this can be merged into the main branch of the project. The reason for a more granular approach to the PRs is to make them easier to verify for the reviewers.


DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
```diff
@@ -359,5 +359,8 @@ public interface IPortalSettings
 
         /// <summary>Gets a value indicating whether to display the dropdowns to quickly add a moduel to the page in the edit bar.</summary>
         bool ShowQuickModuleAddMenu { get; }
+
+        /// <summary>Gets the pipeline type for the portal.</summary>
+        string PagePipeline { get; }
```
bdukes: We need a strategy for adding fields to interfaces without making a breaking change.

DNN Platform/DotNetNuke.Abstractions/Portals/PagePipelineConstants.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static class PagePipelineConstants
+    {
+        /// <summary>WebForms.</summary>
+        public const string WebForms = "webforms";
+
+        /// <summary>MVC.</summary>
+        public const string Mvc = "mvc";
+
+        /// <summary>Auto.</summary>
+        public const string Auto = "auto";
```
bdukes: At the least, these comments need more details about how these are used, if not changing the field names to be more descriptive.

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -1329,14 +1329,21 @@ public virtual void UpdateModuleDefinition(int moduleDefId, string friendlyName,
                 lastModifiedByUserID);
         }
 
+        [Obsolete("Deprecated in DotNetNuke 10.99.0. Scheduled removal in v12.0.0.")]
         public virtual int AddModuleControl(int moduleDefId, string controlKey, string controlTitle, string controlSrc, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int createdByUserID)
```
bdukes:
```suggestion
        [DnnDeprecated(10, 99, 0, "Use overload with mvcControlClass")]
        public virtual partial int AddModuleControl(int moduleDefId, string controlKey, string controlTitle, string controlSrc, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int createdByUserID)
```

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -1356,7 +1363,7 @@ public virtual IDataReader GetModuleControls()
             return this.ExecuteReader("GetModuleControls");
         }
 
-        public virtual void UpdateModuleControl(int moduleControlId, int moduleDefId, string controlKey, string controlTitle, string controlSrc, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int lastModifiedByUserID)
+        public virtual void UpdateModuleControl(int moduleControlId, int moduleDefId, string controlKey, string controlTitle, string controlSrc, string mvcControlClass, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int lastModifiedByUserID)
```
bdukes, line 1385:
```suggestion
        [DnnDeprecated(10, 99, 0, "Use overload with mvcControlClass")]
        public virtual partial void UpdateModuleControl(int moduleControlId, int moduleDefId, string controlKey, string controlTitle, string controlSrc, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int lastModifiedByUserID)
            => this.UpdateModuleControl(moduleControlId, moduleDefId, controlKey, controlTitle, controlSrc, null, iconFile, controlType, viewOrder, helpUrl, supportsPartialRendering, supportsPopUps, lastModifiedByUserID);

        public virtual void UpdateModuleControl(int moduleControlId, int moduleDefId, string controlKey, string controlTitle, string controlSrc, string mvcControlClass, string iconFile, int controlType, int viewOrder, string helpUrl, bool supportsPartialRendering, bool supportsPopUps, int lastModifiedByUserID)
```

DNN Platform/Library/Entities/Portals/IPortalSettingsController.cs
```diff
@@ -14,6 +14,8 @@ public interface IPortalSettingsController
 
         PortalSettings.PortalAliasMapping GetPortalAliasMappingMode(int portalId);
 
+        string GetPortalPagePipeline(int portalId);
```
bdukes: Adding a member to an interface here as well, needs discussion

DNN Platform/DotNetNuke.Abstractions/Portals/PagePipelineConstants.cs
bdukes: Should some of these values be an enum, rather than string values? e.g. `PortalSettings.PagePipeline` could be a `PagePipelineType` value instead of `string`

DNN Platform/Library/Entities/Portals/PortalSettingsController.cs
```diff
@@ -280,7 +292,7 @@ public virtual void LoadPortalSettings(PortalSettings portalSettings)
             portalSettings.DataConsentDelayMeasurement = setting;
             setting = settings.GetValueOrDefault("AllowedExtensionsWhitelist", this.hostSettingsService.GetString("DefaultEndUserExtensionWhitelist"));
             portalSettings.AllowedExtensionsWhitelist = new FileExtensionWhitelist(setting);
-
+            portalSettings.PagePipeline = settings.GetValueOrDefault("PagePipeline", "webforms");
```
bdukes:
```suggestion
            portalSettings.PagePipeline = settings.GetValueOrDefault("PagePipeline", PagePipelineConstants.WebForms);
```

DNN Platform/Library/Entities/Tabs/TabInfo.cs
```diff
@@ -699,6 +699,27 @@ public string SkinDoctype
...
 
+        /// <summary>Gets a value indicating the pipeline type.</summary>
+        [XmlIgnore]
+        [JsonIgnore]
+        public string PagePipeline
+        {
+            get
+            {
+                string pagePipeline;
+                if (this.TabSettings.ContainsKey("PagePipeline") && !string.IsNullOrEmpty(this.TabSettings["PagePipeline"].ToString()))
```
bdukes: Should this be a new column on `Tabs` instead of using tab settings?

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -238,13 +238,14 @@ public static TabInfo AddHostPage(string tabName, string description, string tab
         /// <param name="controlKey">The key for this control in the Definition.</param>
         /// <param name="controlTitle">The title of this control.</param>
         /// <param name="controlSrc">The source of ths control.</param>
+        /// <param name="mvcControlClass">The mvc control class of ths control.</param>
```
bdukes:
```suggestion
        /// <param name="mvcControlClass">The MVC control class of this control.</param>
```

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -238,13 +238,14 @@ public static TabInfo AddHostPage(string tabName, string description, string tab
         /// <param name="controlKey">The key for this control in the Definition.</param>
         /// <param name="controlTitle">The title of this control.</param>
         /// <param name="controlSrc">The source of ths control.</param>
+        /// <param name="mvcControlClass">The mvc control class of ths control.</param>
         /// <param name="iconFile">The icon file.</param>
         /// <param name="controlType">The type of control.</param>
         /// <param name="viewOrder">The vieworder for this module.</param>
-        public static void AddModuleControl(int moduleDefId, string controlKey, string controlTitle, string controlSrc, string iconFile, SecurityAccessLevel controlType, int viewOrder)
+        public static void AddModuleControl(int moduleDefId, string controlKey, string controlTitle, string controlSrc, string mvcControlClass, string iconFile, SecurityAccessLevel controlType, int viewOrder)
```
bdukes, line 258: Deprecate this overload and add a new overload with `mvcControlClass`

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Extensions/App_LocalResources/Extensions.resx
```diff
@@ -1220,4 +1220,10 @@ Common issues with bad installation files:
   <data name="Loading.Tooltip" xml:space="preserve">
     <value>Your extensions will be here in a just a moment</value>
   </data>
+  <data name="AddModuleControl_MvcControl.HelpText" xml:space="preserve">
+    <value>The fully qualified class name of this mvc control required to suporting the Mvc Pipeline.</value>
```
bdukes:
```suggestion
    <value>The fully qualified class name of the MVC control required for supporting the MVC Pipeline.</value>
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.Pages/App_LocalResources/Pages.resx
```diff
@@ -1849,4 +1849,22 @@
...
+    <value>Pipeline Settings</value>
+  </data>
+  <data name="PagePipeline.Text" xml:space="preserve">
+    <value>Page rendering pipeline</value>
+  </data>
+  <data name="WebForms.Text" xml:space="preserve">
+    <value>WebForms</value>
+  </data>
+  <data name="Mvc.Text" xml:space="preserve">
+    <value>Mvc</value>
```
bdukes:
```suggestion
    <value>MVC</value>
```

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SiteSettings/App_LocalResources/SiteSettings.resx
```diff
@@ -1548,4 +1548,22 @@
...
+    <value>Page rendering pipeline</value>
+  </data>
+  <data name="PagePipeline_tooltip.Text" xml:space="preserve">
+    <value>The type of pipeline to use for the site. Can be overridden for each page.</value>
+  </data>
+  <data name="WebForms.Text" xml:space="preserve">
+    <value>WebForms</value>
+  </data>
+  <data name="Mvc.Text" xml:space="preserve">
+    <value>Mvc</value>
```
bdukes:
```suggestion
    <value>MVC</value>
```

DNN Platform/DotNetNuke.Abstractions/Portals/PagePipelineConstants.cs
donker: Agreed. Should be an enum.

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
donker: Any ideas? IPortalSettingsPlus will get very silly very soon. Then I'd prefer IPortalSettings2.

DNN Platform/Library/Services/Upgrade/Upgrade.cs
donker: Done

DNN Platform/DotNetNuke.Abstractions/Portals/PagePipelineConstants.cs
donker: Done

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
bdukes: In this case, there is not a need to expose it from the interface (the places where the property is currently used, the variable's type is `PortalSettings`, not `IPortalSettings`). But it would obviously be helpful to switch references to the interface, or have it available for 3rd parties.

I see four options for adding members to an interface (in rough order of my preference).
1. We create smaller named interfaces with only a small handful of properties, e.g. `public interface IRenderingSettings { PagePipeline.PortalRenderingPipeline PagePipeline { get; }  }`
2. We add a version number, e.g. `public interface IPortalSettingsV2 : IPortalSettings { PagePipeline.PortalRenderingPipeline PagePipeline { get; }  }`
3. We switch to abstract base classes instead of interfaces, e.g. `public abstract class PortalSettingsBase { PagePipeline.PortalRenderingPipeline PagePipeline { get; }  }`
4. We break interfaces when necessary on major versions

donker: But as long as you are not actively creating objects in your module to implement the interface, it wouldn't break, no? I mean: if I use an IPortalSettings in my module, my module wouldn't break because a new property gets added. Only when I create an object that implements the IPortalSettings would things go South, no? If this is true, then should we not have a strategy where certain interfaces are marked solely for consumption?

valadas: There are basically 2 strategies to add to an interface I have seen.

1. Extensions methods, if the new members are related, then they could be grouped into some extension to that basic interface. I have done so for the portal styles with .GetPortalStyles() which extends IPortalSettings or some such.
2. I have seen in many libraries including some from Microsoft when interfaces are simply versioned and intherit an older interface within the new one. One example is Visual Studio SDK itself that has IVsSolution, IVsSolution2, IVsSolution3, IVsSolution4, IVsSolution5, IVsSolution6. It is ugly but it works for both adding new stuff, changing stuff and removing deprecated stuff. And if we have an inherithance chain, consumers only have to migrate to the new type and only handle what has changed. This would be one example https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.VisualStudio/Extensibility/IVsPackageInstaller2.cs

With todays tooling (intellisense and AI), I personally don't have anything against option 2 which could allow us to move quite faster in some areas.

sachatrauwaen: It sound to me as this interface never will be implemented from outsite the core of DNN.
There is a possibility that interfaces cannot be implemented through an external assembly.

public interface IService
{
    void Execute();

    // Prevent external implementations
    internal void InternalOnly();
}

(may be to late to consider for this case)

We can also give a default implementation for new property in a interface (with some limitations).

In this specific case, I think we could remove string PagePipeline { get; } from the interface because it is only used internally and not thru the interface itself.

DNN Platform/Library/Entities/Portals/IPortalSettingsController.cs
bdukes: This would be a scenario where an extension methods might suffice

bdukes: But it would still be good to add it to the interface at some point in the future

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
bdukes: Are you saying that if we define the interface as `internal` then we don't need to worry about backward compatibility?

Default interface implementations [require a least .NET Core 3.0](https://andrewlock.net/understanding-default-interface-methods/#default-interface-methods-require-runtime-support)

Using extension methods (or extension properties) feels like a short-term compromise. It will make testing more complicated (you can't just override an interface for testing).

I'm happy to add numbered versions of interfaces. Do we want to match major DNN version (e.g. this field would be on `IPortalSettings11`), or just start at `2` and make changes any time we want?

sachatrauwaen: An alternative is to split internal and external stuff like this :

public interface IService
{
    void Execute();
}

internal interface IServiceInternal : IService
{
    void InternalMethod();
}

internal class Service : IServiceInternal
{
    public void Execute()
    {
        Console.WriteLine("Execute");
    }

    public void InternalMethod()
    {
        Console.WriteLine("Internal");
    }
}
services.AddScoped<IService, Service>();

you can then inject IServiceInternal internaly and IService externaly

valadas: Yes, if the thing is not meant for public usage, in this case I think people would want to be able to access those settings.

valadas: @bdukes regarding:
> I'm happy to add numbered versions of interfaces. Do we want to match major DNN version (e.g. this field would be on IPortalSettings11), or just start at 2 and make changes any time we want?

I am of the opinion we got go 2, 3, 4 that way we don't even need to wait for next major to move forward on "user non-breaking changes" as long as we keep the previous interface alive for 2 majors.

valadas: The only thing I did not test and this may be pretty simple but for DI I think we would do something like:

``` cs
services.AddScoped<IService, Service>();
services.AddScoped<Iservice2, Service>();
```

Now that brings another interesting thought though, `Service` if it was public now has the breaking changes for anyone that inherited it, unless it was `sealed`. So we may need a `Service` and a `Service2` as well... Maybe we should start sealing new classes we are adding unless there is a good reason to inherit from them. That might simplify future maintenance...

bdukes: It shouldn't be breaking to inherit from `PortalSettings` if we add `IPortalSettings2`, because the base class will already implement that interface.

sachatrauwaen: > Are you saying that if we define the interface as `internal` then we don't need to worry about backward compatibility?

I was thinking, Yes.

If the interface is internal, we only need to preserve compatibility for components we control and ship together.

DNN Platform/Library/Entities/Tabs/TabInfo.cs
sachatrauwaen: Most of the times this setting will be the default value (inherited). So nothing will be saved in the database.
Create a new column, make the ugrade hevier when a lot of pages.
Tab settings sound like as the aproach used by a lot of tab settings.
Why do you think a new column is better ?

DNN Platform/DotNetNuke.Abstractions/Portals/IPortalSettings.cs
donker: OK, this one I've done.

DNN Platform/Library/Entities/Portals/IPortalSettingsController.cs
donker: I'm moving it to an extension method for now.

DNN Platform/Library/Entities/Tabs/TabInfo.cs
donker: I've added a column to the Tabs table

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -56,18 +56,22 @@ public static string DecodeParameter(string value)
             return Encoding.UTF8.GetString(arrBytes);
         }
 
-        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(string)"/>. Decrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(string)"/>. Decrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
```
bdukes:
```suggestion
        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(string)"/>. Decrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -56,18 +56,22 @@ public static string DecodeParameter(string value)
             return Encoding.UTF8.GetString(arrBytes);
         }
 
-        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(string)"/>. Decrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(string)"/>. Decrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
         /// <param name="value">The encrypted value.</param>
         /// <returns>The decrypted value.</returns>
         [DnnDeprecated(10, 2, 2, "Use overload taking ICryptographyProvider")]
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
```
bdukes:
```suggestion
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -56,18 +56,22 @@ public static string DecodeParameter(string value)
...
         /// <param name="value">The encrypted value.</param>
         /// <returns>The decrypted value.</returns>
         [DnnDeprecated(10, 2, 2, "Use overload taking ICryptographyProvider")]
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
         public static partial string DecryptParameter(string value)
             => DecryptParameter(Globals.GetCurrentServiceProvider().GetRequiredService<ICryptographyProvider>(), value);
 
-        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(ICryptographyProvider,string)"/>. Decrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(ICryptographyProvider,string)"/>. Decrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
```
bdukes:
```suggestion
        /// <summary>Decrypts an encrypted value generated via <see cref="EncryptParameter(ICryptographyProvider,string)"/>. Decrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -56,18 +56,22 @@ public static string DecodeParameter(string value)
...
         /// <param name="cryptographyProvider">The cryptography provider.</param>
         /// <param name="value">The encrypted value.</param>
         /// <returns>The decrypted value.</returns>
         public static string DecryptParameter(ICryptographyProvider cryptographyProvider, string value)
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
```
bdukes:
```suggestion
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -108,18 +112,22 @@ public static string EncodeParameter(string value)
             return toEncode.ToString();
         }
 
-        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
```
bdukes:
```suggestion
        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -108,18 +112,22 @@ public static string EncodeParameter(string value)
             return toEncode.ToString();
         }
 
-        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
         /// <param name="value">The value to encrypt.</param>
         /// <returns>The encrypted value.</returns>
         [DnnDeprecated(10, 2, 2, "Use overload taking ICryptographyProvider")]
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
```
bdukes:
```suggestion
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -108,18 +112,22 @@ public static string EncodeParameter(string value)
...
             => EncryptParameter(Globals.GetCurrentServiceProvider().GetRequiredService<ICryptographyProvider>(), value);
 
-        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
         /// <param name="cryptographyProvider">The cryptography provider.</param>
         /// <param name="value">The value to encrypt.</param>
         /// <returns>The encrypted value.</returns>
         public static string EncryptParameter(ICryptographyProvider cryptographyProvider, string value)
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
```
bdukes:
```suggestion
```

DNN Platform/Library/Common/Utilities/UrlUtils.cs
```diff
@@ -108,18 +112,22 @@ public static string EncodeParameter(string value)
...
         /// <param name="value">The value to encrypt.</param>
         /// <returns>The encrypted value.</returns>
         [DnnDeprecated(10, 2, 2, "Use overload taking ICryptographyProvider")]
+#pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
         public static partial string EncryptParameter(string value)
             => EncryptParameter(Globals.GetCurrentServiceProvider().GetRequiredService<ICryptographyProvider>(), value);
 
-        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
+#pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
+        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettingsV2.GUID"/>.</summary>
```
bdukes:
```suggestion
        /// <summary>Encrypt a parameter for placing in a URL. Encrypted using the current portal's <see cref="IPortalSettings.GUID"/>.</summary>
```

DNN Platform/DotNetNuke.Abstractions/Logging/IEventLogger.cs
```diff
@@ -23,45 +23,45 @@ public interface IEventLogger
         /// <param name="portalSettings">The portal settings.</param>
         /// <param name="userID">The user id.</param>
         /// <param name="logType">The log type.</param>
-        void AddLog(string name, string value, IPortalSettings portalSettings, int userID, EventLogType logType);
+        void AddLog(string name, string value, IPortalSettingsV2 portalSettings, int userID, EventLogType logType);
```
bdukes: This is a breaking change. I don't think we need to change any of the usages of the interface, unless they're using the new `PagePipeline` property

DNN Platform/Library/Data/DataProvider.cs
```diff
@@ -889,7 +889,7 @@ public virtual int SaveTabVersionDetail(int tabVersionDetailId, int tabVersionId
...
         }
 
-        public virtual void UpdateTab(int tabId, int contentItemId, int portalId, Guid versionGuid, Guid defaultLanguageGuid, Guid localizedVersionGuid, string tabName, bool isVisible, bool disableLink, int parentId, string iconFile, string iconFileLarge, string title, string description, string keyWords, bool isDeleted, string url, string skinSrc, string containerSrc, DateTime startDate, DateTime endDate, int refreshInterval, string pageHeadText, bool isSecure, bool permanentRedirect, float siteMapPriority, int lastModifiedByuserID, string cultureCode, bool isSystem)
+        public virtual void UpdateTab(int tabId, int contentItemId, int portalId, Guid versionGuid, Guid defaultLanguageGuid, Guid localizedVersionGuid, string tabName, bool isVisible, bool disableLink, int parentId, string iconFile, string iconFileLarge, string title, string description, string keyWords, bool isDeleted, string url, string skinSrc, string containerSrc, DateTime startDate, DateTime endDate, int refreshInterval, string pageHeadText, bool isSecure, bool permanentRedirect, float siteMapPriority, int lastModifiedByuserID, string cultureCode, bool isSystem, int pagePipeline)
```
bdukes, line 902: This need to be deprecated and a new overload added

Dnn.AdminExperience/Dnn.PersonaBar.Extensions/Services/SiteSettingsController.cs
```diff
@@ -3189,6 +3193,7 @@ public HttpResponseMessage UpdateOtherSettings(UpdateOtherSettingsRequest reques
                 PortalController.Instance.UpdatePortalSetting(pid, "AllowJsInModuleHeaders", request.AllowJsInModuleHeaders.ToString(), false, null, false);
                 PortalController.Instance.UpdatePortalSetting(pid, "AllowJsInModuleFooters", request.AllowJsInModuleFooters.ToString(), false, null, false);
                 PortalController.Instance.UpdatePortalSetting(pid, "ShowQuickModuleAddMenu", request.ShowQuickModuleAddMenu.ToString(), false, null, false);
+                PortalController.Instance.UpdatePortalSetting(pid, PagePipeline.SettingName, request.PagePipeline, false, null, false);
```
bdukes: Should this use the `MvcPipelineSettingsRepository`?

bdukes (approved):

bdukes:
@donker is the current status of this PR that you're looking into unit test failures?

thienvc:
Excuse me, but I'd like to ask if the MVC Pipeline concept is still being developed?

sachatrauwaen:
> Excuse me, but I'd like to ask if the MVC Pipeline concept is still being developed?

Yes it is under review

mitchelsellers:
> Excuse me, but I'd like to ask if the MVC Pipeline concept is still being developed?

As Sacha mentioned yes, the current focus is on trying to get this into a state that we can review it for inclusion without introducing breaking changes to the larger DNN Platform install base.  As you can see some of the feedback on this particular PR addresses, for example the proper location to introduce new features, or settings to do so in a manner that will not break prior installations, and also not add too much to the API Surface for long-term support.

mitchelsellers (approved):

PR #7298, opened 2026-05-26
Use always-increasing CDV when debug mode is on

### LINKED ISSUE #7296: [Enhancement]: Add a setting to enable or disable cdv.

### Description of problem

Sometimes when you are still working on some design stuff, while in in development, sometimes you need css and js files being updated in real time to test changes, without need to go to server > performance and change host version, avoiding css and js cache.

Last dnn versions had minify and composite options to avoid that but now that options dissapear.

### Description of solution

Would be great to have some option or setting to avoid cdv cache, while in development. or maybe while site is in debug ( Security > More > More Security Settings > Enable debug mode).

_No response_

### Anything else?

_No response_

- [ ] Yes

- [ ] Yes

### Code of Conduct

bdukes:
Fixes #7296

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
+++ b/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
@@ -6,6 +6,7 @@ namespace DotNetNuke.Web.Client.ResourceManager
 {
     using System;
     using System.Collections.Generic;
+    using System.Globalization;
     using System.IO;
     using System.Linq;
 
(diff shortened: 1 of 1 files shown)


DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
```diff
@@ -52,7 +53,11 @@ public ClientResourceController(IHostSettings hostSettings, IApplicationStatusIn
             this.clientResourceSettings = clientResourceSettings;
             this.RegisterPathNameAlias("SharedScripts", "~/Resources/Shared/Scripts/");
             this.controllerId = Guid.NewGuid();
-            if (clientResourceSettings != null)
+            if (hostSettings.DebugMode)
+            {
+                this.crmVersion = int.Parse($"{DateTime.UtcNow:HHmmssfff}", NumberStyles.None, CultureInfo.InvariantCulture);
```
mitchelsellers, line 58:
```suggestion
                this.crmVersion = int.Parse($"{DateTime.UtcNow:MMddyyyyHHmmssfff}", NumberStyles.None, CultureInfo.InvariantCulture);
```
Maybe a bit of overkill, but this guarantees uniqueness across multiple days.

I do perfer the change here with the CRM version attribute/value

bdukes: I was trying to keep it to nine digits so that it was inside the bounds of an `Int32` value. We could take off `ff` from the end and add `dd` to the front…

valadas:
I think #7299 might be fine by me, it will wrap in 2028 or something and start increasing from int.MinValue for the next 2038 years ;)

bdukes:
Closing in favor of #7299

PR #7299, opened 2026-05-26
Use Unix timestamp as cdv in debug mode

donker:
Addresses #7296

When in debug mode (Security > More > More Security Settings) the CDV nr will be replaced with the unix timestamp in seconds.

### DIFF
--- a/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
+++ b/DNN Platform/DotNetNuke.Web.Client.ResourceManager/ClientResourceController.cs
@@ -283,7 +283,9 @@ namespace DotNetNuke.Web.Client.ResourceManager
                 });
             }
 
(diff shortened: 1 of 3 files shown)


bdukes:
Duplicate of https://github.com/dnnsoftware/Dnn.Platform/pull/7298

bdukes (dismissed):

valadas (dismissed):
Very nice, good catch on the "unchecked" part.

Not a blocker, but would it be hard to also do it for the PersonaBar resources? Right now it appears to only affect the portal resources.

Image

bdukes:
Maybe the logic needs to be moved into `IHostSettings.CrmVersion` so that it applies everywhere without hunting down each usage.

bdukes (approved):

mitchelsellers (approved):

PR #7302, opened 2026-05-29
Better support for empty elements in manifest.

### LINKED ISSUE #7301: [Bug]: Manifest has partial support for empty XML elements.

### What happened?

The manifest reading logic supports having a self-closing `<element/>` (e.g. `<supportedFeatures/>`) but not an empty `<supportedFeatures></supportedFeatures>` element. With the former, the installation throws the following error:
```
2026-05-29 14:19:00.512+03:00 [workstation][D:3][T:23][ERROR] DotNetNuke.Services.Install.InstallWizard - WIZARD ERROR:System.Xml.XmlException: There are multiple root elements. Line 25, position 2.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at DotNetNuke.Entities.Modules.DesktopModuleInfo.ReadXml(XmlReader reader)
   at DotNetNuke.Common.Utilities.CBO.DeserializeObject[TObject](XmlReader reader)
   at DotNetNuke.Services.Installer.Installers.ModuleInstaller.ReadManifest(XPathNavigator manifestNav)
   at DotNetNuke.Services.Installer.Installers.InstallerFactory.GetInstaller(XPathNavigator manifestNav, PackageInfo package)
   at DotNetNuke.Services.Installer.Installers.PackageInstaller.ReadComponents(XPathNavigator manifestNav)
   at DotNetNuke.Services.Installer.Installers.PackageInstaller.ReadManifest(XPathNavigator manifestNav)
   at DotNetNuke.Services.Installer.Installers.PackageInstaller..ctor(String packageManifest, InstallerInfo info)
   at DotNetNuke.Services.Installer.Installer.ProcessPackages(XPathNavigator rootNav)
   at DotNetNuke.Services.Installer.Installer.ReadManifest(Boolean deleteTemp)
   at DotNetNuke.Services.Upgrade.Upgrade.InstallPackage(String file, String packageType, Boolean writeFeedback)
   at DotNetNuke.Services.Upgrade.InternalController.Steps.InstallExtensionsStep.Execute()
   at DotNetNuke.Services.Install.InstallWizard.Install(IApplicationStatusInfo appStatus)
```

### Steps to reproduce?

1. Create a package with a manifest that has an empty `<supportedFeatures></supportedFeatures>`
2. Try to install it.

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Affected Versions

10.3.2 (latest release), develop build (unreleased)

_No response_

### Code of Conduct

dimarobert:
Fixes #7301

## Summary
Adds support for both forms of empty elements in manifest.

As the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreader.isemptyelement?view=netframework-4.8.1) states, `System.Xml.XmlReader.IsEmptyElement` supports only self-closing elements:
> gets a value indicating whether the current node is an empty element (for example, `<MyElement/>`).
> 
> Returns
> &nbsp;&nbsp; true if the current node is an element (NodeType equals XmlNodeType.Element) that ends with `/>`; otherwise, false.

### DIFF
--- a/DNN Platform/Library/Entities/Modules/DesktopModuleInfo.cs
+++ b/DNN Platform/Library/Entities/Modules/DesktopModuleInfo.cs
@@ -266,82 +266,81 @@ namespace DotNetNuke.Entities.Modules
                     continue;
                 }
 
-                if (reader.NodeType == XmlNodeType.Element && reader.Name == "moduleDefinitions" && !reader.IsEmptyElement)
+                switch (reader)
                 {
-                    this.ReadModuleDefinitions(reader);
-                }
-                else if (reader.NodeType == XmlNodeType.Element && reader.Name == "supportedFeatures" && !reader.IsEmptyElement)
-                {
-                    this.ReadSupportedFeatures(reader);
-                }
-                else if (reader.NodeType == XmlNodeType.Element && reader.Name == "shareable" && !reader.IsEmptyElement)
-                {
-                    this.ReadModuleSharing(reader);
-                }
-                else if (reader.NodeType == XmlNodeType.Element && reader.Name == "page" && !reader.IsEmptyElement)
-                {
-                    this.ReadPageInfo(reader);
-                }
-                else
-                {
-                    switch (reader.Name)
-                    {
-                        case "desktopModule":
-                            break;
-                        case "moduleName":
-                            this.ModuleName = reader.ReadElementContentAsString();
-                            break;
-                        case "friendlyName":
-                            this.FriendlyName = reader.ReadElementContentAsString();
-                            break;
-                        case "description":
-                            this.Description = reader.ReadElementContentAsString();
-                            break;
-                        case "foldername":
-                            this.FolderName = reader.ReadElementContentAsString();
-                            break;
-                        case "businessControllerClass":
-                            this.BusinessControllerClass = reader.ReadElementContentAsString();
-                            break;
-                        case "codeSubDirectory":
-                            this.CodeSubDirectory = reader.ReadElementContentAsString();
-                            break;
-                        case "page":
-                            this.ReadPageInfo(reader);
+                    case { NodeType: XmlNodeType.Element, Name: "moduleDefinitions", IsEmptyElement: false, }:
+                        this.ReadModuleDefinitions(reader);
+                        break;
+                    case { NodeType: XmlNodeType.Element, Name: "supportedFeatures", IsEmptyElement: false, }:
+                        this.ReadSupportedFeatures(reader);
+                        break;
+                    case { NodeType: XmlNodeType.Element, Name: "shareable", IsEmptyElement: false, }:
+                        this.ReadModuleSharing(reader);
+                        break;
+                    case { NodeType: XmlNodeType.Element, Name: "page", IsEmptyElement: false, }:
+                        this.ReadPageInfo(reader);
+                        break;
+                    default:
+                        switch (reader.Name)
+                        {
+                            case "desktopModule":
+                                break;
+                            case "moduleName":
+                                this.ModuleName = reader.ReadElementContentAsString();
+                                break;
+                            case "friendlyName":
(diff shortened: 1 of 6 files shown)


bdukes (dismissed):

bdukes:
Thanks! I pushed a couple of commits to add on to your changes.

1. I added some tests to verify what should be happening here.
2. I added a helper `XmlUtils.TryReadEndElement`, which also handles an element which has only whitespace.

DNN Platform/Library/Entities/Modules/DesktopModuleInfo.cs
```diff
@@ -266,82 +266,81 @@ public void ReadXml(XmlReader reader)
...
-                            if (reader.NodeType == XmlNodeType.Element && !string.IsNullOrEmpty(reader.Name))
-                            {
-                                reader.ReadElementContentAsString();
-                            }
+                    case { NodeType: XmlNodeType.Element, Name: "moduleDefinitions", }:
+                        this.ReadModuleDefinitions(reader);
+                        break;
+                    case { NodeType: XmlNodeType.Element, Name: "supportedFeatures", }:
+                        this.ReadSupportedFeatures(reader);
+                        break;
```
dimarobert: @bdukes Looks like the `IsEmptyElement: false` is missing / got removed from these two.

What actually happened in the initial issue was that because the `EndElement` was not properly handled for `SupportedFeatures` (and others) it kept searching for `supportedFeature` (singular) items from the reader until it reached the end of the component XML, then the remaining code that parses the `<component type="Module">` got confused of where the reader was and threw the unexpected `There are multiple root elements. Line 25, position 2.` error with the reader on the last line of the `<component type="Module">`, even if the XML itself was valid.

bdukes: The new helper `XmlUtils.TryReadEndElement` checks for whether it's an empty element. I suppose we can add the checks back (I originally removed them because of inconsistencies around whether `SupportedFeatures` was initialized to `-1` or `0`, but ultimately decided that I needed to just keep that inconsistency in there).

bdukes: I've pushed an updated commit which adds that check back in.

dimarobert: @bdukes Just a note, I think adding the check back is the correct approach, as `XmlUtils.TryReadEndElement` only partially checks for an empty element, it only checks for whitespace and `XmlNodeType.EndElement`, so it will handle:
1. ```
   <MyElement>
   </MyElement>
   ```
2. `<MyElement></MyElement>`

but it will not handle a self-closing element, e.g. `<MyElement />`, The reader would only ever be in a `reader.NodeType == XmlNodeType.Element` with `reader.IsEmptyElement == true` state for this type of element, it would never go though a `XmlNodeType.EndElement` state for it. I think that basically `reader.IsEmptyElement == true` is the only way to detect a self-closing element.
I think that it is ok that `XmlUtils.TryReadEndElement` doesn't handle this third case itself, as it's name clearly states that it should only read (as in consume) the next node if it is an `EndElement` node (with a little added bonus of consuming any whitespace before it) and nothing more.

bdukes: That makes sense. Looks like I need to add a test cases which demonstrates an issue without the `IsEmptyElement` check…

bdukes: I've added tests that data after a self-closing element can still be read. Thanks again!

bdukes (approved):

mitchelsellers (approved):

Issue #7306, opened 2026-06-03
[Enhancement]: Check for file change during install


donker:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of problem

On Azure installations various issues have been reported because Azure doesn't react well when you massively update (overwrite) files. E.g. issue #5377. This is in part because when we install the platform or extensions we copy files regardless of whether they have changed.

### Description of solution

I suggest we first check the file to be copied with the existing file to see if it has changed. IMO a simple MD5 hash compare would suffice.

### Do you plan to contribute code for this enhancement?

- [x] Yes

### Would you be interested in sponsoring this enhancement?

- [ ] Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mitchelsellers:
This is a great idea, the other side of this is the copying/backup of files that we don't fully use, that adds a second file operation which limits the file IO etc.

PR #7315, opened 2026-06-07
Add IsSameFile check when unpacking resources/zip files in DNN

donker:
As mentioned in issue #7306 it is possible one encounters problems when updating components (extensions or DNN itself) due to the amount of disk IO. This PR explores a way to reduce this by comparing a file on disk with a file it would be overwritten with using their Hash.

There are two places this PR intervenes: in the FileSystemUtils and the ResourceFileInstaller. The FileSystemUtils also include methods for "resource file installation" which is why this was included. The only way we can do this avoiding unnecessary disk IO is to load the file to be written into a memory stream. This, because zip streams are typically not seekable. To void having very large files loaded into memory the methods in FileSystemUtils cut off the file checking at 10 Mb. This would be acceptable as the main goal is to drive down the number of unnecessary file writing, not eliminate it all together.

I've not implemented the 10 Mb check on the ResourceFileInstaller yet, as I wish to first get feedback on this proposal. Is the 10 Mb cutoff even necessary? Should it be higher/lower?

### DIFF
--- a/DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
+++ b/DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
@@ -9,6 +9,7 @@ namespace DotNetNuke.Common.Utilities
     using System.IO;
     using System.IO.Compression;
     using System.Linq;
+    using System.Security.Cryptography;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -25,6 +26,9 @@ namespace DotNetNuke.Common.Utilities
     /// <summary>File System utilities.</summary>
     public partial class FileSystemUtils
     {
+        /// <summary>Maximum file size (in bytes) for performing content comparison during extraction. Files larger than this will always be extracted without comparison. Default is 10MB.</summary>
+        private const long MaxFileSizeForComparison = 10 * 1024 * 1024; // 10 MB
+
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(FileSystemUtils));
 
         /// <summary>Adds a File to a Zip File.</summary>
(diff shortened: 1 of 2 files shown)


DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
```diff
@@ -296,14 +435,65 @@ public static void UnzipResources(IEnumerable<ZipArchiveEntry> zipArchiveEntries
...
-                        File.Delete(fileNamePath);
-                    }
+                        if (File.Exists(fileNamePath))
+                        {
+                            // Check file size - skip comparison for large files
+                            var existingFileInfo = new SchwabenCode.QuickIO.QuickIOFileInfo(fileNamePath);
+
+                            if (existingFileInfo.Length > MaxFileSizeForComparison)
+                            {
+                                Logger.Info($"Skipping comparison for large file '{localFileName}' ({existingFileInfo.Length:N0} bytes) - exceeds {MaxFileSizeForComparison:N0} byte limit");
```
bdukes:
```suggestion
                                Logger.Debug($"Skipping comparison for large file '{localFileName}' ({existingFileInfo.Length:N0} bytes) - exceeds {MaxFileSizeForComparison:N0} byte limit");
```

DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
```diff
@@ -254,6 +258,141 @@ public static async Task<string> ReadFileAsync(string filePath, CancellationToke
...
+                byte[] fileHash;
+                byte[] streamHash;
+
+                using (var fileStream = File.OpenRead(filePath))
+                {
+                    fileHash = fileHasher.ComputeHash(fileStream);
+                }
+
+                stream.Position = 0;
+                streamHash = streamHasher.ComputeHash(stream);
```
bdukes, line 366: Do we need multiple hashers?
```suggestion
                using var hasher = CryptographyUtils.CreateSHA256();

                byte[] fileHash;
                using (var fileStream = File.OpenRead(filePath))
                {
                    fileHash = hasher.ComputeHash(fileStream);
                }

                stream.Position = 0;
                var streamHash = hasher.ComputeHash(stream);
```

DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
```diff
@@ -254,6 +258,141 @@ public static async Task<string> ReadFileAsync(string filePath, CancellationToke
...
```
bdukes, line 298:
```suggestion
                using var hasher = CryptographyUtils.CreateSHA256();

                byte[] fileHash;
                using (var fileStream = File.OpenRead(filePath))
                {
                    fileHash = hasher.ComputeHash(fileStream);
                }

                stream.Position = 0;
                var streamHash = hasher.ComputeHash(stream);
```

DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
```diff
@@ -360,14 +550,65 @@ public static async Task UnzipResourcesAsync(IEnumerable<ZipArchiveEntry> zipArc
...
-                        await File.DeleteAsync(fileNamePath);
-                    }
+                        if (await File.ExistsAsync(fileNamePath))
+                        {
+                            // Check file size - skip comparison for large files
+                            var existingFileInfo = new SchwabenCode.QuickIO.QuickIOFileInfo(fileNamePath);
+
+                            if (existingFileInfo.Length > MaxFileSizeForComparison)
+                            {
+                                Logger.Info($"Skipping comparison for large file '{localFileName}' ({existingFileInfo.Length:N0} bytes) - exceeds {MaxFileSizeForComparison:N0} byte limit");
```
bdukes:
```suggestion
                                Logger.Debug($"Skipping comparison for large file '{localFileName}' ({existingFileInfo.Length:N0} bytes) - exceeds {MaxFileSizeForComparison:N0} byte limit");
```

DNN Platform/Library/Common/Utilities/FileSystemUtils.cs
```diff
@@ -360,14 +550,65 @@ public static async Task UnzipResourcesAsync(IEnumerable<ZipArchiveEntry> zipArc
...
 
-                    using var fileStream = File.Open(fileNamePath, FileMode.CreateNew);
-                    await zipEntry.Open().CopyToAsync(fileStream, 25000, cancellationToken);
+                        if (shouldWriteFile)
+                        {
+                            if (memoryStream != null)
+                            {
+                                // Write from MemoryStream (already loaded for comparison)
+                                using var fileStream = File.Open(fileNamePath, FileMode.CreateNew);
+                                await memoryStream.CopyToAsync(fileStream, 81920, cancellationToken);
```
bdukes, line 598: Can you pull `81920` into a constant, so there's a name indicating what it is and why?

bdukes (dismissed):
Have you run any testing to see if this has an effect on an Azure app?

donker:
> Have you run any testing to see if this has an effect on an Azure app?

I am looking for some help for this part.

mitchelsellers:
I did an initial test on this tonight, and not seeing much of a difference in 2 tests, but they were quick

donker:
After testing we discovered this does not solve the issue in IO resource constricted environments.

PR #7327, opened 2026-06-16
Actions cleanup

bdukes:
## Summary
This PR resolves some issues raised by the zizmor action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,10 +67,15 @@ jobs:
     steps:
       - name: "Setup Environment Variables"
         run: |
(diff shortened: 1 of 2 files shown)


valadas:
@bdukes any clue why the build stays yellow and shows no progress?

bdukes:
Since I added a name to the _Dependency Review_ job, our configuration was looking for the default name instead of the new name. And there was a typo in the _Build and Validate_ workflow (using `with` instead of `env`) which kept it from working.

valadas (approved):
Awesome!

david-poindexter (approved):

PR #7344, opened 2026-06-30
Fix page workflow state when workflow disabled

### LINKED ISSUE #7305: [Bug]: On upgraded site (10.2.3 -> 10.3.2), new HTML modules remain Draft/Unpublished and do not render, while existing HTML modules still work

### What happened?

On one legacy DNN site that was recently upgraded from **10.2.3 to 10.3.2**, newly created **HTML modules** are not rendering in normal/view mode or for anonymous users.

The key finding is that **Version History shows the content as Draft / Unpublished only** after saving, and no Published version is created.

At the same time:

- **existing HTML modules on the site still render correctly**
- **OpenContent modules render correctly**
- the problem reproduces on a **new page** as well as on existing pages
- workflow-related settings appear to be **off**

This makes it look like a site-specific issue affecting the HTML module publish/versioning state after upgrade.

### Steps to reproduce?

Unfortunately, I have only reproduced this so far on one upgraded legacy site (other upgraded sites don't have the same issue), but the behavior is:

1. Upgrade an older site from **10.2.3 to 10.3.2**
2. Create a new page (or use an existing page)
3. Add a new **HTML module**
4. Enter simple test content and click **Save**
5. Exit edit mode / view page normally

### Current Behavior

- The newly created HTML module content is visible in **edit mode**
- The module does **not** render in normal/view mode
- The module does **not** render for anonymous/public users
- In **Version History**, the content shows only as **Draft / Unpublished**
- No Published version appears to be created

### Expected Behavior

Saving content in a new HTML module should result in a **Published** version (when workflow is not enabled), and the content should render in normal/view mode and for anonymous users according to page/module permissions.

### Relevant log output

```shell

```

### Anything else?

Here is what we tested:

- Existing HTML modules on the site still display correctly
- New HTML modules do not display
- If edit mode is turned off, the new HTML modules disappear even while still logged in
- OpenContent modules on the same test page do display correctly for anonymous users
- Changing page containers/layouts did not change the behavior
- Copying an existing HTML module to another page did not make it render there
- Adding a new HTML module to the homepage also did not make it render
- On another recently upgraded DNN **10.3.2** site, creating a new HTML module works normally and Version History shows it as **Published**
- The Workflow setting "Enable Versioning" is turned off

### Affected Versions

10.3.2 (latest release)

Microsoft Edge, Chrome

### Code of Conduct

tvatavuk:
Fixes #7305

## Summary

This PR prevents page-level workflow settings from being applied when workflow/versioning is disabled at the site level.

Root cause: the Pages settings UI could submit workflow/versioning values even when those features were disabled globally. The saved tab settings then poisoned the page workflow state, so newly added Text/HTML modules could be created as draft content and remain hidden in view mode.

The fix keeps page workflow/versioning disabled unless the portal-level feature is enabled, so Text/HTML content is published normally on sites where workflow is not active.

Verified locally with:
- clean DNN 10.3.2 install
- site workflow/versioning disabled
- page workflow/versioning toggled in page settings
- newly added Text/HTML modules remain visible in view mode

### DIFF
--- a/DNN Platform/Modules/HTML/Components/HtmlTextController.cs
+++ b/DNN Platform/Modules/HTML/Components/HtmlTextController.cs
@@ -26,6 +26,7 @@ namespace DotNetNuke.Modules.Html
     using DotNetNuke.Entities.Modules;
     using DotNetNuke.Entities.Portals;
     using DotNetNuke.Entities.Tabs;
+    using DotNetNuke.Entities.Tabs.TabVersions;
     using DotNetNuke.Entities.Users;
     using DotNetNuke.Internal.SourceGenerators;
     using DotNetNuke.Modules.Html.Components;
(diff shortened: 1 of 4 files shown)


bdukes (dismissed):
Thanks!

valadas (commented):
GetWorkflow_UsesDirectPublish_WhenTabHasStateButWorkflowIsDisabled currently fails but other than that nothing sticks out to me

mitchelsellers (approved):

bdukes (approved):

PR #7355, opened 2026-07-03
FIX: default.css > Accidental dnnTableDisplay nesting > tables.scss

Timo-Breumelhof:
No issue as minor typo

## Summary

Broken selector caused by accidental nesting
modules/_tables.scss:56 (inside .dnnTableDisplay { ... })
    .dnnTableDisplay tr:hover td { ... }
This rule is nested inside the .dnnTableDisplay block, so SCSS compiles it
to ".dnnTableDisplay .dnnTableDisplay tr:hover td" - a selector that
requires a .dnnTableDisplay element nested inside another .dnnTableDisplay
element. That never occurs in the real markup, so the row hover background
this rule defines never applies.


mitchelsellers (approved):

donker:
Looks to be duplicate of #7356

PR #7356, opened 2026-07-03
FIX: default.css > Default css/rgb var not dynamic

Timo-Breumelhof:
Minor issue, not issue created

## Summary

Two of three color channels hardcoded in an otherwise-variable rgba()
modules/_tables.scss:56-61 (same rule as 1.3)
    background: rgba(
        var(--dnn-color-info-r, 2),
        139,
        255,
        0.15);
The red channel comes from --dnn-color-info-r, but green/blue are hardcoded
literals (139, 255) instead of --dnn-color-info-g/-b. Even once 1.3 is
fixed, a theme changing --dnn-color-info can only shift one of the three
channels of this hover tint.

Solution: see 1.3 above - use var(--dnn-color-info-g, 139) and
var(--dnn-color-info-b, 255) so all three channels track the theme's
--dnn-color-info the same way -r already does.

### DIFF
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/modules/_tables.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/modules/_tables.scss
@@ -53,7 +53,7 @@
             var(--dnn-color-foreground-b, 0),
             0.04);
     }
-    .dnnTableDisplay tr:hover td {
+    tr:hover td {
         background: rgba(
             var(--dnn-color-info-r, 2),
             139,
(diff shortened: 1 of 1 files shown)


mitchelsellers:
This appears to be a better and more complete fix for #7355 am I correct?

Timo-Breumelhof:
@mitchelsellers hmm I think I accidentally have one commit in two PRs, sorry about that

Timo-Breumelhof:
@mitchelsellers I guess as this is all minor the easiest is to just drop the other one?

donker (approved):

valadas (approved):

PR #7371, opened 2026-07-10
Improved css reset for deafault.css

### LINKED ISSUE #7352: DNN 10 default css reset is too strong
The css reset in default.css for DNN 10 marks a lot of elements as `font: inherit` this breaks normal browser behavior for italic or bold, etc. 
We should revisit to modernize that reset as we do not then override some important things after they are reset. There are more lightweight resets available that only patch currently known browser inconsistencies without being so invasive.

### Discussed in https://github.com/dnnsoftware/Dnn.Platform/discussions/7350

<div type='discussions-op-text'>

<sup>Originally posted by **datafastlr** July  2, 2026</sup>
I have a number of DNN websites that were upgraded to version 10.3.2 and now text inside style tags like <em> or <strong> for example, look fine in the HTML module's editor, but look like plain text on the page.
Not sure what is causing that.

Regards,
Larry
html-tag-bg
</div>

valadas:
"modern-normalize" was already in the dev-dependencies for the clientside project but was unused. Instead a manual css reset was in place which was too strong and made the default `em` `<i>` `<strong>` `<b>` have wrong defaults that made them useless.

I believe this was temporary and the intention was to use modern-normalize in the first place and it was gorgotten.

This PR fixes that. The above elements now look fine and I went over the default theme and the DNN UI parts and it appers to not have affected them.

Closes #7352

### DIFF
--- a/DNN Platform/Dnn.ClientSide/build.ts
+++ b/DNN Platform/Dnn.ClientSide/build.ts
@@ -57,6 +57,7 @@ async function buildScss(input: string, output: string): Promise<void> {
         {
           sourceMap: true,
           sourceMapIncludeSources: true,
+          importers: [new sass.NodePackageImporter()],
         }
     );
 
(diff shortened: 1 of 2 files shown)


Timo-Breumelhof:
Thanks :-)

bdukes (approved):

mitchelsellers (approved):

bdukes:
I accidentally merged this, but we're intending to include this in 10.4.0 instead. I've pushed a revert commit

bdukes:
Moved to https://github.com/dnnsoftware/Dnn.Platform/pull/7372

PR #7372, opened 2026-07-14
Reapply "Improved css reset for default.css"

bdukes:
This reverts commit c5f81ddacfdcf75a2f6c8c06ec9a33e9a2b4f38a.

Second try for #7371


valadas (dismissed):

valadas (approved):

Timo-Breumelhof:
It's this one, I guess. https://cdn.jsdelivr.net/npm/modern-normalize/modern-normalize.css
Yes that will fix the issue.
Might lead to other ones though, we'll see.
As I said before, I would remove it as IMO it's the theme's responsibility to load a reset.css

Timo-Breumelhof:
@bdukes @valadas 
I tested some more with the same theme (BS5)
Without the current reset (custom default.css) my aligment issues in tabels are fixed but some elements (in FF) get borders they did not have before. This is also the case with this new reset.

Looked like this:

image


Without reset or with the new one included:

image

I think my initial issue are caused by the DNN 7 default.css not having a border reset for td and the current version does.
This new one however does not reset borders to 0 and thus unexpected borders show up in some places.

Not sure I would add this new reset to DNN 10.4 THB and I think the "no risk" fix would be to not remove baseline like I did here:

https://github.com/dnnsoftware/Dnn.Platform/issues/7406

But remove the lines for tables as they are a change.
image


And then postpone a more modern version to DNN 11?

valadas:
Discussing this now and both me and @bdukes agree that less is more here. If you want to send a PR with an alternative solution you have done more testing here than we did and default.css should try to be non-obtrusive as much as we can.

Timo-Breumelhof:
@valadas OK' I'll do my suggested PR

valadas:
Replaced by #7418 to be less opiniated

PR #7373, opened 2026-07-15
Update reset styles in _reset.scss

valadas:
Removed font-size and font properties from the reset styles.
This prevented normal elements like `small`, `em`, `strong` from having their normal behavior.


mitchelsellers (approved):
This looks correct per our discussion.

bdukes (approved):

PR #7379, opened 2026-07-18
Fix large legacy Site Export import failing with LiteDB FindAll loop

### LINKED ISSUE #7378: [Bug]: Large legacy Site Export import fails with LiteDB "Detected loop in FindAll({0})"

### What happened?

Importing a large legacy Site Export package (produced by DNN 9.10.2-era Site Export/Import, which uses LiteDB 3.1.0) fails validation with:

```json
{"Message": "Package is not valid. Technical Details:Detected loop in FindAll({0})"}
```

`VerifyImportPackage` returns HTTP 400 and the import never starts. I expected the legacy export database to be upgraded/read successfully (or, if genuinely incompatible, to return a specific diagnostic rather than the low-level LiteDB message).

### Steps to reproduce?

1. Using DNN Platform 9.10.2 (or any product built on it), create a Full Site Export whose `ExportFolder` collection contains **more than ~2,550 records** (e.g. create 3,000+ folders in the source site's file system before exporting).
2. Copy the resulting export package folder (containing `export.json` and `export_db.zip`) into `App_Data/ExportImport/<packageName>/` on a fresh DNN Platform 10.3.x instance.
3. Go to Settings > Import/Export > Import.
4. Select the package and click Continue (this triggers `VerifyImportPackage`).

### Current Behavior

After ~10 seconds, the request returns HTTP 400 with `Package is not valid. Technical Details:Detected loop in FindAll({0})`. The import wizard shows a generic error and does not proceed. Packages with fewer than ~2,550 records in their largest collection import successfully.

### Expected Behavior

The legacy export database should be upgraded and read successfully by `VerifyImportPackage`, allowing the import to proceed — regardless of collection size.

### Relevant log output

```
System.Exception: Detected loop in FindAll({0})
   at LiteDB.Engine.IndexService.FindAll(...)
   at LiteDB.Engine.RebuildService.Rebuild(...)
   ...
```

(Thrown from inside the `LiteDatabase` constructor when `ConnectionString.Upgrade = true` triggers an in-place rebuild of a legacy-format file.)

### Anything else?

**Root cause:** `ExportImportRepository` (`DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs`) opens every export database with:

```csharp
this.liteDb = new LiteDatabase(new ConnectionString(dbFileName) { Upgrade = true });
```

A DNN 9.10.2-era export database is written in the legacy LiteDB v3.x on-disk format (internally "v7"). Opening it with `Upgrade = true` forces LiteDB to run a full in-place `Rebuild()`. LiteDB 5.0.21 — the version pinned in `Directory.Packages.props` — has a bug in that rebuild path: `IndexService.FindAll` uses a `MAX_ITEMS_COUNT` loop-detection guard that is initialized as if the target collection were empty, so any single collection above roughly 2,550 records falsely throws `Detected loop in FindAll({0})`.

This is a known upstream LiteDB defect:
- [litedb-org/LiteDB#2525](https://github.com/litedb-org/LiteDB/issues/2525) — same exception thrown from `Rebuild()` on 5.0.21
- [litedb-org/LiteDB#2582](https://github.com/litedb-org/LiteDB/issues/2582) — independent report of the same failure
- [litedb-org/LiteDB#2603](https://github.com/litedb-org/LiteDB/pull/2603) (unmerged) — proposes computing `maxItemsCount` from source data instead of assuming an empty database

Since LiteDB is a binary dependency and there's no fixed stable release available, bumping the version isn't currently an option. I have a fix ready (detects the legacy file format by its header and migrates it into a fresh LiteDB 5.x database via normal inserts — which don't hit the broken rebuild guard — instead of the in-place upgrade, with the original fast path preserved for native 5.x and small legacy files) and will follow up with a pull request referencing this issue.

matt468779:
Fixes #7378

## Summary

Importing a legacy Site Export package (LiteDB 3.x on-disk format, produced by DNN 9.10.2-era Site Export/Import) whose largest collection exceeds ~2,550 records fails `VerifyImportPackage` with HTTP 400 `Package is not valid. Technical Details:Detected loop in FindAll({0})`, so the import never starts.

**Root cause:** `ExportImportRepository` (`DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs`) opens the export database with `Upgrade = true`, which forces LiteDB to run a full in-place `Rebuild()` when the file is in the legacy v3 on-disk format. LiteDB 5.0.21 (the version pinned in `Directory.Packages.props`) has a bug in that rebuild path: `IndexService.FindAll` uses a `MAX_ITEMS_COUNT` loop-detection guard initialized as if the target collection were empty, so any single collection above roughly 2,550 records falsely throws `Detected loop in FindAll({0})`. This is documented upstream in [litedb-org/LiteDB#2525](https://github.com/litedb-org/LiteDB/issues/2525) and [litedb-org/LiteDB#2582](https://github.com/litedb-org/LiteDB/issues/2582), with an unmerged fix at [litedb-org/LiteDB#2603](https://github.com/litedb-org/LiteDB/pull/2603). Since LiteDB is a binary dependency here and there's no fixed stable release available, bumping the version isn't currently an option.

**Fix:** detect a legacy-format (LiteDB v7) export database by its file header, and instead of the broken in-place rebuild, migrate it into a fresh LiteDB 5.x database by streaming every collection's documents through normal inserts — which don't go through the broken rebuild guard — then operate on the migrated copy. Native 5.x databases and small legacy databases that fail to migrate fall back to the original `Upgrade = true` fast path, so unaffected imports keep their exact previous behavior. The public `ExportImportRepository` API is unchanged; the migrated copy is a temporary file cleaned up on `Dispose`.

## Release Note

Fixed an issue where importing a large legacy (DNN 9.10.2-era) Site Export package failed with a LiteDB "Detected loop in FindAll" error, blocking the import before it could start.

## Testing Steps

1. On a DNN 9.10.2-era instance (or any product built on it), create a Full Site Export whose `ExportFolder` collection contains more than ~2,550 records (e.g. create 3,000+ folders in the source site before exporting).
2. Copy the resulting export package folder into `App_Data/ExportImport/<packageName>/` on a DNN Platform 10.3.x instance built from this branch.
3. Go to Settings > Import/Export > Import, select the package, and click Continue.
4. Confirm the package validates and the import proceeds successfully (previously: HTTP 400 with `Detected loop in FindAll({0})`).

Also added `ExportImportRepositoryTests` with genuine LiteDB v3.1.0 fixtures:
- `legacy_v3_large.dnndb` — 3,000 `ExportFolder` rows (above the threshold; reproduces the original failure on the old `Upgrade = true` path — the repository must now open it and read all rows back correctly)
- `legacy_v3_small.dnndb` — 50 rows (below the threshold; regression check that small legacy packages keep working)
- A native LiteDB 5.x regression case (the common 10.x → 10.x import path, confirming no behavior change there)

image


matt468779:
@microsoft-github-policy-service agree company="Trilogy"

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
```diff
@@ -267,12 +292,158 @@ public void CleanUpLocal(string collectionName)
             collection.Update(documentsToUpdate);
         }
 
-        protected virtual void Dispose(bool disposing)
```
bdukes, line 270: Changing `Dispose` from `protected` to `private` is a breaking change

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
```diff
@@ -267,12 +292,158 @@ public void CleanUpLocal(string collectionName)
...
+                var fileReaderType = liteDbAssembly.GetType("LiteDB.Engine.FileReaderV7");
+                if (engineSettingsType == null || fileReaderType == null)
+                {
+                    return null;
+                }
+
+                var engineSettings = Activator.CreateInstance(engineSettingsType);
+                engineSettingsType.GetProperty("Filename").SetValue(engineSettings, sourceDbFileName);
+
+                var reader = (IDisposable)Activator.CreateInstance(
```
bdukes, line 415: Could we use the `IFileReader` type here, so we can avoid some of the reflection?

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
```diff
@@ -267,12 +292,158 @@ public void CleanUpLocal(string collectionName)
...
+            var targetDbFileName = sourceDbFileName + ".migrated";
+            try
+            {
+                if (File.Exists(targetDbFileName))
+                {
+                    File.Delete(targetDbFileName);
+                }
+
+                var liteDbAssembly = typeof(LiteDatabase).Assembly;
+                var engineSettingsType = liteDbAssembly.GetType("LiteDB.Engine.EngineSettings");
```
bdukes: `EngineSettings` is public, do we need to create it via reflection?

bdukes:
Thanks, this seems like a reasonable resolution to the issue. However, I'm wanted to get a better understanding of the scenario in which this happens. Why would you be exporting a 9.10.2 site and importing it into a 10.3.x site? Wouldn't it be better to upgrade the site first? I'm wondering if this is a scenario we want to support. If we don't want to support imports across versions, you make a good point that we should give a better error message.

Any thoughts @dnnsoftware/approvers?

mitchelsellers:
My $0.02 here is that we shouldn't really support export/import in this manner, as there are so many things that could be a problem, especially 9.x -> 10.x or any other major version crosses given breaking change possibilities.

valadas:
I do agree that it would be very difficult to support. We should have a message saying something along the lines of "this export cames from a different version and importing it may succeed, completely fail or partially work, we do not recommend import/export between different versions"

valadas (commented):
Other than @bdukes comments, I would be OK with this and we could add the better messaging in another PR.

matt468779:
Looked into the actual scope of the LiteDB rebuild bug to help inform this decision, since the affected code path can be pinned down precisely.

**Confirmed: this only affects cross-major-version (9.x export → 10.x import).** Same-version imports in either direction are unaffected:

- **9.x → 9.x**: DNN 9.10.2 uses LiteDB 3.1.0, and its `ExportImportRepository` opens the database with `new LiteDatabase(dbFileName)` — no `Upgrade` flag at all. LiteDB 3.1.0 doesn't have the buggy `IndexService.FindAll` rebuild-guard code that was introduced later in the 5.x line, so this bug can't occur regardless of collection size.
- **10.x → 10.x**: A 10.x export is already written in native LiteDB 5.x on-disk format. LiteDB's own `TryUpgrade()` (`LiteDB/Engine/Engine/Upgrade.cs`) checks `FileReaderV7.IsVersion(buffer)` before doing anything, and bails out immediately if the file isn't legacy v7 format. So opening a 10.x-native export with `Upgrade = true` never reaches the broken rebuild path either.
- **9.x → 10.x**: This is the only case that trips it — a genuinely legacy (LiteDB ≤4.x, on-disk "v7") file gets opened by a 10.x (LiteDB 5.0.21) engine with `Upgrade = true`, which runs `Recovery()`/rebuild on the legacy data, and that's where `IndexService.FindAll`'s loop-detection guard misfires on any collection over ~2,550 records.

Given @mitchelsellers and @valadas's points about not wanting to support cross-major-version import generally, happy to go either way on this PR:

1. Keep this fix as-is (it only ever activates for the legacy-format case, so it's inert for same-version imports either way), or
2. Pivot it (or split into a separate PR) to detect a legacy v7-format file up front and fail fast with a clear message instead of migrating and proceeding — something along the lines of @valadas's suggested wording ("this export comes from a different version and importing it may succeed, completely fail, or partially work; we do not recommend import/export between different versions").

Let me know which direction the group prefers and I'll adjust accordingly.

bdukes:
Can we do both (either in this PR or in two separate PRs)? Give a warning message about cross-version imports, but fix the upgrade if they still continue despite the warning?

mitchelsellers:
I agree with @bdukes as the best option if we can

bdukes:
@matt468779 will you be able to make any of the requested adjustments to this PR?

matt468779:
@bdukes Yes I will, I got lost in other work. Is this going to be included in the next release? What is the timeline of the next release?

bdukes:
We would love to get it included in the next release. The RC period for that release will probably start next week.

matt468779:
@bdukes I have added the warning in this PR
image

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
bdukes: @matt468779 was there a need to change the visibility of `Dispose`?

mitchelsellers (approved):

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
bdukes: No, it's `internal`

DNN Platform/Modules/DnnExportImportLibrary/Repository/ExportImportRepository.cs
bdukes: This has been done, thanks!

bdukes (approved):

PR #7380, opened 2026-07-20
FIX: Remove font-size and line-height value for .Normal

### LINKED ISSUE #7341: [Enhancement]: default.css: .Normal line-height uses rem instead of em, breaks themes with non-root font-size

### Description of problem

I upgraded a site to DNN 10 today and the main text looked weird. I then found that `.Normal` has styling that it shouldn't IMO.

If you want `default.css` to be less intrusive, there are some things that could be improved.

https://github.com/dnnsoftware/Dnn.Platform/blob/163aad777361c9fc7262cd9c7666848ad38590c0/DNN%20Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/modules/modules.scss#L16

I don't see why `.Normal` needs any styling at all TBH, apart from maybe `inherit`?

Both `font-size` and `line-height` are issues here.

The theme in use sets `font-size: 10px` on the `html` element as a `rem` base trick (so `1rem = 10px`), while actual readable font sizes are set on `body` or container elements. In DNN 9, `.Normal` inherited correctly from the cascade.

With DNN 10:
- `font-size: 1rem` resolves to **10px** instead of inheriting the intended body font size
- `line-height: 1.125rem` resolves to **11.25px** instead of scaling with the element's actual font size

Both values will be wrong in any theme that sets its `rem` base differently from its readable font size.

```css
/* theme sets 1rem = 10px as a base trick, readable size on body */
html { font-size: 10px; }
body { font-size: 16px; }

/* .Normal now renders at 10px with 11.25px line-height — both wrong */
```

Example:

Image

### Description of solution

The fix for both is the same: either **remove these declarations** (preferred — let `.Normal` inherit like it did in DNN 9), or if a value must be set, use `em` so it scales relative to the element's own font size:

```scss
font-size: 1em;       // instead of 1rem
line-height: 1.125em; // instead of 1.125rem
```

_No response_

### Anything else?

_No response_

- [x] Yes

- [ ] Yes

### Code of Conduct

Timo-Breumelhof:
## Summary

Here's the why demo: https://youtu.be/MOcLN9YfXTQ

This is what these pages look like with the fix:

Home:

image

Body 22px:

image

fixes #7341

### DIFF
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/modules/modules.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/modules/modules.scss
@@ -12,13 +12,10 @@ td.SubHead{
     margin-bottom: 0;
 }
 
-/* Set text size in modules */
-.Normal,
+/* Color of Disabled and Deleted */
 .NormalDisabled,
 .NormalDeleted {
     color: var(--dnn-color-foreground, #444);
-    font-size: 1rem;
-    line-height: 1.125rem;
     word-wrap: break-word;
 }
 
--- a/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/ui/_forms.scss
+++ b/DNN Platform/Dnn.ClientSide/src/styles/default-css/10.0.0/ui/_forms.scss
@@ -35,13 +35,13 @@
     input[type="email"],
     input[type="search"],
     input[type="password"] {
-        margin-bottom: 1.125rem;
-        padding: 0.5rem;
+        margin-bottom: 1.125em;
+        padding: 0.5em;
         background: var(--dnn-color-background, #fff);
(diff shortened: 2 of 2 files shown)


valadas (dismissed):
I do agree with this. The build error is unrelated to these changes (we need to bump AngleSharp) so we may have to rebase this.

However I think it sold go in next minor so that we can call attention to default.css fixes as it is a little bit of a potential breaking change if people have worked around those issues.

Timo-Breumelhof:
@valadas agreed 10.4 would be best

Timo-Breumelhof:
@valadas :-( I accidentally added another commit, I remove that

Timo-Breumelhof:
incorrect commit reverted

mitchelsellers (approved):

valadas (approved):

PR #7453, opened 2026-08-13
Apply sitemap eligibility rules to localized alternates

### LINKED ISSUE #7449: [Bug]: Non-indexable localized pages can be emitted as sitemap alternates

### What happened?

After #7447 restored serialization of hreflang links in localized XML
sitemaps, I reviewed how CoreSitemapProvider constructs the alternate URL
collection.

DNN applies its sitemap eligibility rules, including AllowIndex, when
selecting the primary sitemap page. However, the same rules are not applied
when selecting localized alternates or the separately appended
default-language page.

Consequently, a localized page configured with Allow Indexing disabled can
still be advertised as an xhtml:link alternate of an otherwise eligible
sitemap entry.

This is a remaining gap from #3896 / #3897, which added the AllowIndex check
to primary sitemap entries but did not apply it to localized alternates.

### Steps to reproduce?

1. Use the current develop branch containing the merged #7447 change.
2. Enable Content Localization with a default language such as en-US and
   another language such as fr-FR.
3. Publish equivalent versions of a page in both languages.
4. Edit the fr-FR page and, under Page Settings > SEO, disable
   "Allow Indexing."
5. Expire or regenerate the cached sitemap and request Sitemap.aspx for the
   default culture.
6. Find the en-US page entry in the generated XML.
7. Observe that its hreflang collection can still contain an xhtml:link for
   the noindex fr-FR page, even though that page is excluded as a primary
   sitemap entry.

### Current Behavior

Primary sitemap pages are filtered using sitemap eligibility rules,
including AllowIndex.

Localized and default-language alternate URLs are selected separately
without applying the same complete eligibility rules. This can expose a
noindex or otherwise ineligible page inside an hreflang group. It can also
leave a meaningless self-only alternate group after ineligible pages are
removed.

### Expected Behavior

Primary pages, localized alternates, and the default-language alternate
should all use the same sitemap eligibility rules.

Pages that are deleted, disabled, unpublished, outside their publication
window, non-public, excluded by visibility settings, or configured with
AllowIndex=false should not appear anywhere in a sitemap hreflang group.

If fewer than two eligible language versions remain, no alternate group
should be emitted.

### Relevant log output

```shell

```

### Anything else?

Related history:

- #3896 / #3897 excluded non-indexable primary sitemap entries.
- #7446 / #7447 restored serialization of DNN's existing AlternateUrls
  collection.
- #3194 concerns localized sitemaps becoming empty and is a different issue.
- #2819 concerns hreflang tags in HTML page headers rather than XML sitemap
  alternate eligibility.

Culture-specific portal-alias handling was also traced through DNN's existing
URL-generation path. CoreSitemapProvider passes each culture to NavigateURL,
and the friendly URL provider selects the culture-matched portal alias.
No production alias-selection change appears necessary; regression coverage
has been prepared to preserve that behavior.

Google requires each hreflang group to list the eligible language versions,
including itself:
https://developers.google.com/search/docs/specialty/international/localized-versions

A focused implementation with automated tests is ready. The optional
hreflang="x-default" enhancement will be proposed separately to keep this
bug fix narrowly scoped.

Existing cached sitemap files may need to expire or be regenerated before
the corrected output becomes visible.

### Affected Versions

develop build (unreleased)

_No response_

### Code of Conduct

alex-pommier:
## Summary

Fixes #7449

Applies the same sitemap eligibility rules to localized and default-language
alternates that DNN already applies to primary sitemap entries.

This prevents pages that are deleted, disabled, unpublished, outside their
publication window, non-public, excluded by sitemap visibility settings, or
configured with `AllowIndex=false` from appearing inside hreflang groups.

The change also:

- avoids emitting a self-only alternate group when fewer than two eligible
  language versions remain;
- preserves DNN's existing culture-specific URL and portal-alias selection path;
- adds focused regression coverage without changing portal-alias production logic.

`x-default` is intentionally excluded and preserved for a separate enhancement
proposal.

## Tests

Focused sitemap tests:

`dotnet test "DNN Platform/Tests/DotNetNuke.Tests.Core/DotNetNuke.Tests.Core.csproj" --no-restore --filter "FullyQualifiedName~Services.Sitemap" --verbosity quiet`

Result: 8 passed, 0 failed, 0 skipped.

### DIFF
--- a/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
+++ b/DNN Platform/Library/Services/Sitemap/CoreSitemapProvider.cs
@@ -54,21 +54,16 @@ namespace DotNetNuke.Services.Sitemap
             {
                 try
                 {
-                    if (!tab.IsDeleted && !tab.DisableLink && tab.TabType == TabType.Normal &&
-                        (Null.IsNull(tab.StartDate) || tab.StartDate < DateTime.Now) &&
-                        (Null.IsNull(tab.EndDate) || tab.EndDate > DateTime.Now) && this.IsTabPublic(tab.TabPermissions))
+                    if (this.IsTabEligibleForSitemap(tab, DateTime.Now))
                     {
-                        if ((this.includeHiddenPages || tab.IsVisible) && tab.HasBeenPublished && tab.AllowIndex)
+                        try
                         {
-                            try
-                            {
-                                pageUrl = this.GetPageUrl(tab, currentLanguage, ps);
-                                urls.Add(pageUrl);
-                            }
-                            catch (Exception exception)
-                            {
-                                Logger.CoreSitemapProviderErrorGettingPageUrl(exception, tab.TabName);
-                            }
+                            pageUrl = this.GetPageUrl(tab, currentLanguage, ps);
+                            urls.Add(pageUrl);
+                        }
+                        catch (Exception exception)
+                        {
+                            Logger.CoreSitemapProviderErrorGettingPageUrl(exception, tab.TabName);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):
Upon initial code review this looks good, I tested with a single dual language site and it appears to be correct, would appreciate others testing with more langauges/situations enabled.

valadas (approved):
Nothing sticks out to me. I'll merge this and add the `noteworty` tag to call out attention to RC stage testing of this.

PR #7482, opened 2026-08-27
Deprecate UserController.PortalId

### LINKED ISSUE #7481: [Bug]: Username Minimum Length does not work outside portal 0

### What happened?

I've set the minimum username length to 4, but cannot create a user with a short username.

### Steps to reproduce?

1. Create a new portal
2. Update minimum username length to 4
3. Attempt to create a user with username 1234

### Current Behavior

Says the username is invalid

### Expected Behavior

Accepts the username

### Relevant log output

```shell

```

### Anything else?

Introduced in 9.9.0 via https://github.com/dnnsoftware/Dnn.Platform/pull/4444

### Affected Versions

10.3.3 (latest release)

_No response_

### Code of Conduct

bdukes:
Fixes #7481

`UserController.IsValidUserName` and `UserController.UpdateDisplayNames` both use `UserController.PortalId`, but nothing sets this property, so it always returns `0`.

This PR deprecates `UserController.PortalId`. (I considered making it an error deprecation, since the use of that property is almost certainly an error, thoughts @dnnsoftware/approvers? In that case, compiling against that property would emit a compiler error rather than a warning.)

I updated those two methods to use `PortalSettings.Current.PortalId` instead, and introduced an overload for `UpdateDisplayNames` that takes a `portalId`.

### DIFF
--- a/DNN Platform/Library/Entities/Users/UserController.cs
+++ b/DNN Platform/Library/Entities/Users/UserController.cs
@@ -74,6 +74,7 @@ namespace DotNetNuke.Entities.Users
         public string DisplayFormat { get; set; }
 
         /// <summary>Gets or sets the site (portal) id.</summary>
+        [Obsolete("Deprecated in DotNetNuke 10.4.0. This property does nothing, do not use. Scheduled removal in v12.0.0.")]
         public int PortalId { get; set; }
 
         /// <summary>Gets the number count for all duplicate e-mail addresses in the database.</summary>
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me.

I think we need to keep deprecations as warnings. If we would be sure that all 3rd party would compile before deploying, we could use errors but since there are 3rd party solutions that use razor and dynamically compiled code, it could create unexpected runtime errors.

bdukes:
Good point, I hadn't thought about dynamically compiled usages.

mitchelsellers (approved):

PR #7512, opened 2026-09-15
Register non-registered assemblies

### LINKED ISSUE #7510: [Bug]: DNN 10.04.00 RC1 - After Installing 2sxc Module, DNN Throws an Error

### What happened?

Installing 2sxc v21.7.0 LTS or v22.0.0 (latest as of 9/10/2026) makes DNN unable to startup. The install appears to be successful and complete, but on the next reload/startup DNN is no longer in a usable state and presents this error (full screenshot below):

```
Could not load type 'Microsoft.Extensions.Logging.ILoggingBuilder' from assembly 
'Microsoft.Extensions.Logging.Abstractions, Version=2.2.0.0, Culture=neutral, 
PublicKeyToken=adb9793829ddae60'. 
Description: An unhandled exception occurred during the execution of the current 
web request. Please review the stack trace for more information about the error and 
where it originated in the code.

Exception Details: System.TypeLoadException: Could not load type 
'Microsoft.Extensions.Logging.ILoggingBuilder' from assembly 
'Microsoft.Extensions.Logging.Abstractions, Version=2.2.0.0, Culture=neutral, 
PublicKeyToken=adb9793829ddae60'.
```

### Steps to reproduce?

1. Clean new install of release candidate DNN 10.04.00 RC1
2. Install 2sxc (v21 or v22), this completes with no errors logged (though while still in PersonaBar Extensions, when the list tries to refresh, you get a red top-status-bar that says, "There was an error retrieving your content. Please check your internet connection.")
3. When you reload again, you you quick get a Server Error (YSOD, see below) about Serilog and "System.EntryPointNotFoundException: Entry point was not found."

### Current Behavior

After 2sxc v21 or v22 is installed (no errors logged), DNN startup fails showing this:

Image

### Expected Behavior

DNN should be working after a module install.

### Anything else?

I tried to confirm with the vendor that it was a DNN problem, but I did not hear back after two days. So, posting here.

I did not try a DNN Upgrade scenario, clean install only.
 
Also, the dropdown below in the Bug Template only has 10.03.03 (10.3.3), no option for "10.04.00 RC1".

### Affected Versions

10.3.3 (latest release)

Chrome

### Code of Conduct

donker:
This PR attempts to fix #7510 by registering all assemblies in the bin folder after install or upgrade that have not yet been registered.


DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -2785,5 +2791,45 @@ private static bool NewtonsoftNeedUpdate()
 
             return true;
         }
+
+        /// <summary>Scans the bin folder and registers every assembly in the database, ignoring whether it is already registered.</summary>
+        private static void CheckAssembliesForRegistration()
```
bdukes: What about a name like `EnsureAssembliesRegistered`?

bdukes (dismissed):

DNN Platform/Library/Services/Upgrade/Upgrade.cs
```diff
@@ -2785,5 +2791,45 @@ private static bool NewtonsoftNeedUpdate()
...
+                DnnInstallLogger.InstallLogInfo(Localization.GetString("LogEnd", Localization.GlobalResourceFile) + "CheckAssembliesForRegistration");
+                return;
+            }
+
+            foreach (var strAssemblyPath in Directory.GetFiles(binFolder, "*.dll"))
+            {
+                Version version;
+                try
+                {
+                    version = System.Reflection.AssemblyName.GetAssemblyName(strAssemblyPath).Version;
```
bdukes: I think we need to borrow the logic from https://github.com/dnnsoftware/Dnn.Platform/blob/09a495ef8d19dbcaca467ef084d347ed0cd58536/Build/ContextExtensions.cs#L20-L25

donker: I copied the logic from the LocalUpgradeService. So should we change it there, too?

bdukes: Yes, I'm fairly certain that what the build process does is the right approach for as many assemblies as we can.

donker: OK. I've proposed a new approach for this.

bdukes (approved):

valadas (approved):

PR #7498, opened 2026-09-02
Bump NuGet dependencies

bdukes:
## Summary
This PR bumps NuGet dependencies. There are more to bump, but my local environment is not happy with me and I can't see what's working and what's not, so I'll need to do some experimentation via this branch.

PR #7517, opened 2026-09-17
Fix error copying DLLs from multi-target projects

bdukes:
## Summary
Since #7498 the build has failed intermittently. This is due to adding multi-targeting to a handful of projects. Those projects call `xcopy` after building to move their output assemblies into the website `bin` folder. But it's now possible that the `net48` build ends and then tries to copy the `netstandard2.0` DLLs before the `netstandard2.0` build is complete. This changes the copy step to only run for the target framework that produces the desired DLLs.

See https://github.com/dnnsoftware/Dnn.Platform/actions/runs/35225627899/job/105216359902?pr=7503 as an example (and https://github.com/dnnsoftware/Dnn.Platform/actions/runs/35225627899?pr=7503 as an example of the build being fixed just by retrying)


donker (approved):
Thanks for that

valadas (approved):

